PR #112, opened 2014-05-12
Fix create conference form

ChrisBr:
- redirects to new path if conference is not valid
- adds HTML 5 validations


app/views/layouts/application.html.haml
```diff
@@ -1,4 +1,4 @@
-!!!
+%html{:xmlns => "http://www.w3.org/1999/html"}
```
AndrewKvalheim, line 1: 😢

hennevogel: ?

Issue #545, opened 2014-11-18
Schedule Highlights


hennevogel:
For the splash I would like to have the ability to 'highlight' events. This could be a simple boolean 'is_highlight' on the Event model.

differentreality:
Closed in #562

Issue #544, opened 2014-11-18
Keynotes


hennevogel:
Currently I can assign the role keynote_speaker to a User. That however doesn't give me the keynote which is the important part about it...

I would suggest to get rid of this role and introduce an EventType keynote (by default) which would give me the event and through it also the speaker.

differentreality:
More to do for this?

hennevogel:
Nope

PR #562, opened 2014-12-08
add highlights and fix cfp in splashpage

differentreality:
https://github.com/openSUSE/osem/issues/544
https://github.com/openSUSE/osem/issues/545

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,6 +121,9 @@ gem 'money-rails'
 # for lists
 gem 'acts_as_list'
 
+# for switch checkboxes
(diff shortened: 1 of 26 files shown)


hennevogel:
Is the switch on events_controller#index supposed to work? It doesn't

differentreality:
Indeed it is not working after adding %td{'data-sort' => "#{event.is_highlight}"} for fixing the sorting of the table. Any thoughts?

hennevogel:
Better :) :+1: please :ship:

PR #568, opened 2014-12-13
fix users list actions

differentreality:


app/assets/stylesheets/osem.css.scss
```diff
@@ -53,4 +53,8 @@ body {
 
 fieldset {
   margin: 20px 0 20px 0;
-}
\ No newline at end of file
+}
+
+.col-actions {
+  width: 170px;
```
hennevogel: This is pixel pushing. What when then UI is wider or smaller? If you want the Action <td> to be a specific width use percentages or the bootstrap classes like in [ticket.html.haml](https://github.com/openSUSE/osem/blob/master/app/views/tickets/_ticket.html.haml)

Issue #573, opened 2014-12-17
Transifex Translation Broken


hennevogel:
The translation feature on the live instance is broken.

PR #575, opened 2014-12-26
put translation javascript code in separate file

differentreality:
That should work and hide our key.
@hennevogel ?

https://github.com/openSUSE/osem/issues/573

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -34,6 +34,7 @@
 //= require osem-datepickers
 //= require osem-datatables
 //= require osem-tickets
+//= require osem-translation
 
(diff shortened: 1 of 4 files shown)


hennevogel:
This would mean you have to change code (*osem-translation.js*) to configure transifex. Bad idea

differentreality:
@hennevogel do you have any other proposal at all to make the translation work?

hennevogel:
How it is implemented now is the right way. We just have to make it work...

hennevogel:
I don't get it. Have you tried this?

hennevogel:
See #585 for a correct fix

PR #582, opened 2015-01-15
Events open now in new tab when clicking in schedule

### LINKED ISSUE #71: Cannot open talks in separate tabs
I am on an unreliable network and I want to preload data. When opening, i.e. https://conference.opensuse.org/osem/conference/osc14/schedule, I expected to be able to middle click the talks to load them in several tabs. That, however, does not work and I cannot preload the abstracts of the talks.

ChrisBr:
close #71


hennevogel:
:+1:

PR #583, opened 2015-01-18
Rework Roles and their interface

differentreality:
Move into their own controller
Add description to roles
Make it possible to add new roles (as labels)
Add abilities
Clean code

### DIFF
--- a/app/views/admin/registrations/index.xlsx.axlsx
+++ b/app/views/admin/registrations/index.xlsx.axlsx
@@ -27,6 +27,5 @@ wb.add_worksheet(name: 'registrations') do |sheet|
 
     sheet.add_row row
   end
-  sheet.column_info[8].width = [sheet.column_info[8].width, 30].min
 end
 


hennevogel:
Whats the difference between a Role and a Label and what's the use of Labels?

hennevogel:
The interface is also not very user friendly. If I'm on an empty Role (which is kind of the default) and I click on another Role that has no Users yet only very little on the page changes, it's almost indistinguishable. Why don't you just use a simple CRUD interface?

differentreality:
Some of the 'default' roles we are using have special access (eg. organizer), some others don't (eg 'Attendee'). That was the original setup (the second category I call label-kind roles, or simply labels).

hennevogel:
Okay so what is the use of the labels? I really don't understand sorry :confused:

differentreality:
Labels are purely informative. For example 'Sponsor', we can apply the role 'sponsor' to the representatives of our sponsors.
When they come to the registration desk we know they are affiliated with one of our sponsors.
I belive the reason it does not make all the sense it should, yet, is that the registration list does not currently show the roles of the user. (https://github.com/openSUSE/osem/issues/595)

hennevogel:
Users of OSEM already have to know what implications roles have. This is, so far, a very obscure part of OSEM that badly needs clarification. Adding labels to the concept of roles just complicates this further. I would strongly suggest to get rid of labels in this context.

Sorry if I appear so picky about this :worried: :bouquet:

hennevogel:
How about roles are roles and labels are labels? And how about if labels are free text, tags so to speak?

differentreality:
Labels *are* free text, you can create your own new role. If that's what you meant.

What about dropping the whole differentiation from our frontend?
We have roles. Several roles. 
Each role means something different. 
(**Organizer** role means you have full access,
**Sponsor** role means you are a sponsor of the conference, etc)

And I have also updated the wiki about roles: https://github.com/openSUSE/osem/wiki/Roles

Be (a bit) picky!
I introduced roles, everything makes total sense to me! But it should really make sense to everybody else too, well at least after they read the wiki info at least. :)

hennevogel:
That's the problem I'm having with label roles. **Roles** mean different behaviour of OSEM. Assigning someone a 'label' **role** doesn't change anything in OSEM. That's why I really don't like the mixture of labels with roles in this PR.

differentreality:
But it does not really mean different behaviour for OSEM actually. 

Since we setup roles, only some of the roles defined a different behaviour, while other roles simply defined a characteristic for the user. 
The differentiation between 'actionable roles' and 'labels' was mine, just for the frontend. Since it is confusing, we can drop the differentiation and just keep Roles which may or may not give you special access, but will definitely give you a label box next to your name in our 'registrations' list.

In each case, the number of roles is pre-defined and it is small. Do you think it really makes sense to implement the same thing twice just to have a different table and view for roles and a different one called label doing the exact same thing? 

Or do you feel we should just split the views?
(One page for roles one page for labels, still save them in the same table 'roles'.)

For what it's worth, I still think that roles don't necessarily mean different behaviour and 'Organizer' is equally a role as 'Sponsor'.

hennevogel:
Yes what I'm saying is: Don't intermix roles with labels. Implement meaningful roles _AND_ some general tagging mechanism for people and events...

differentreality:
That's the updated index so far (some meaningful roles and some general tags without different names). 

What else do you have in mind?

screenshot-10

differentreality:
@hennevogel you proposed using a different gem for tags, but I am not sure we would like to introduce the new logic. What about we use rolify for tags as well and making tags a new thingie on its own?

screenshot-osem - chromium-3
screenshot-osem - chromium
screenshot-osem - chromium-2
screenshot-osem - chromium-1

hennevogel:
What if we want to tag events too?

differentreality:
In both cases (rolify and act_as_taggable_on) we have an issue for the list of tags we will make available. Some tags are only for users, some are only for events and in each case we have different tags available for each conference.

With rolify we need to have different models, eg EventTag and UserTag. That would work simple and fast and allows for different tags for the events of every conference. 
But if we need to add tagging possibility for another model, we would again need to create a new tag model, and its corresponding join table (events_tags, users_tags, etc)

With act_as_taggable_on, in order to use 1 single table for tags, we need to make changes. Eg:

Tags table:

Name       conference_id   context
MyVolunteer        1            User              (tag for users for conference with ID 1)
MyKeynote          1            Event            (tag for events of conference with ID 1)
Volunteer                           User              (tag for users. Available for all conferences)              
Keynote                             Event            (tag for events. Avaiable for all conferences)

Obviously some data will be duplicated between tags and taggings tables.
The join table (taggings) contains the following fields:
tag_id, taggable_id, taggable_type, tagger_id, tagger_type, context

I think it might be simpler to have 2 different models (EventTag and UserTag) using rolify.
That would give us 4 new tables: event_tags, user_tags and their corresponding joing tables: events_tags and users_tags respectively.

Other options/ideas?

hennevogel:
I can only *again* strongly advise you against mixing up roles and tags. No matter what this means for the amount of objects we need....

differentreality:
What do you think of this layout?

screenshot-2
screenshot-osem - chromium
screenshot-osem - chromium-1

Issue #589, opened 2015-01-27
Registrations PDF broken


ChrisBr:
NoMethodError: undefined method `attending_social_events' for #<Registration:0x0000000597c9f8>

We have to remove the attending_social_events attribute from the prawn file

shubhankardesh:
What is the cause of NoMethodError?

shubhankardesh:
Suppose the attribute is removed, then what could be a substitute for 'attending_social_events'?

differentreality:
We have introduced questions.
Organizers can create their own custom questions with stuff they want to ask to attendees.
These questions (if they exist) they appear in the registration page.

shayonj:
I am planning on taking this up. I see that both ```attending_social_events``` and ```attending_with_partner``` are no longer the columns. 

So we would need to remove this from both - https://github.com/openSUSE/osem/blob/master/app/views/admin/registrations/index.pdf.prawn  and https://github.com/openSUSE/osem/blob/master/app/views/admin/registrations/index.xlsx.axlsx  , right?

differentreality:
Yes that is correct.  We need to remove those attributes from everywhere, as they are no longer used.
Some work is already being done in https://github.com/openSUSE/osem/pull/601

ChrisBr:
@differentreality I'll close this one because it's fixed in #645 right?

PR #601, opened 2015-02-13
Fixes Issue #589 (Update prawn file)

shubhankardesh:
The attribute attending_social_events has been removed. The question "Are you attending any social events?" could now be asked by creating a custom question instead.

### DIFF
--- a/app/views/admin/registrations/index.pdf.prawn
+++ b/app/views/admin/registrations/index.pdf.prawn
@@ -3,7 +3,6 @@ prawn_document(:force_download=>true, :filename => @pdf_filename) do |pdf|
   header_array = ["   ",
                   "Name",
                   "Email",
-                  "Attending Social Events",
                   "Attending With Partner",
                   "Arrival Date",
                   "Departure Date"]
@@ -13,10 +12,6 @@ prawn_document(:force_download=>true, :filename => @pdf_filename) do |pdf|
     row << ""
     row << registration.name
     row << registration.email
-    if registration.attending_social_events
-      row << "X"
-    else
-      row << " "
     end
 
     if registration.attending_with_partner.to_s
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -11,13 +11,13 @@
     = csrf_meta_tags
     :javascript
       window.liveSettings = {
-        api_key: "<%= CONFIG['transifex_live_api_key'] %>",
+        api_key: "#{CONFIG['transifex_live_api_key']}",
         picker: "bottom-right",
         detectlang: true,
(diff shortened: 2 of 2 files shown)


app/views/admin/registrations/index.pdf.prawn
```diff
@@ -13,10 +12,6 @@ prawn_document(:force_download=>true, :filename => @pdf_filename) do |pdf|
     row << ""
     row << registration.name
     row << registration.email
-    if registration.attending_social_events
-      row << "X"
-    else
-      row << " "
     end
 
     if registration.attending_with_partner.to_s
```
ChrisBr: I suppose we also moved attending_with_partner!
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L310

ChrisBr:
Hey @shubhankardesh,
thank you for your PR! :)

Have you tested your changes? Because I suppose we also moved the attribute attending_with_partner.
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L310

Otherwise LGTM :+1:

shubhankardesh:
Great! Well, it looks like the build has passed. We could move the attribute attending_with_partner as well. Maybe, it could be created as a custom question.

ChrisBr:
Hey @shubhankardesh,
thanks for your response!

Yes, the build passed because there is no test for generating the .pdf file! Maybe you can investigate if there is a possibility to test it?

Attending with partner is already a dynamic question.
https://github.com/openSUSE/osem/blob/master/db/seeds.rb#L26

shubhankardesh:
Yes of course!

shubhankardesh:
Any pointers as to how to test the correct generation of the .pdf?

ChrisBr:
@shubhankardesh testing the generation should be enough! Maybe a feature test!

shayonj:
Great. While you are at it, I noticed few things while playing around with this bug.

You may want to take care of
```
    row << reg.attending_social_events
    row << reg.attending_with_partner
    row << reg.handicapped_access_required
```
from https://github.com/openSUSE/osem/blob/master/app/views/admin/registrations/index.xlsx.axlsx . As all of them are throwing errors. And finally setting the ```sheet.column_info``` reference to a not nil value, accordingly.

My two cents :)

Cheers

differentreality:
@shubhankardesh Your changes do not work.
Please try the app with the changes you make before you send a PR!!
And as Chris said, add a relevant test. Where do you stand with that? Did you find how to test something about the feature of 'PDF generation' ?

ChrisBr:
@shubhankardesh Any progress?

ChrisBr:
No response since 4 weeks, so I will close this one!

Issue #599, opened 2015-02-10
Rewrite ApplicationHelper.target_process_color to use case rather than elsif


kalabiyau:
https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb#L2

Ensure testing

PR #602, opened 2015-02-26
replaced if-else with case-when

v0dro:
Solves #599

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,13 +1,13 @@
 module ApplicationHelper
   def target_progress_color(progress)
     progress = progress.to_i
-    if progress > 90
-      result = 'green'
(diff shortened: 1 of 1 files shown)


hennevogel:
You don't catch the case where progress is 90 or 80. 
```ruby
irb(main):021:0> target_progress_color(90)
=> "red"
irb(main):022:0> target_progress_color(80)
=> "red"
```

v0dro:
I just took the `if-else` above as a reference.

Should the code be something like this? :

``` ruby

case
  when progress >= 90 then 'green'
  when progress < 90 && progress >= 80 then 'orange'
  else 'red'
end
```

v0dro:
@hennevogel updated :)

hennevogel:
Looks good to me! :bouquet:

Issue #519, opened 2014-10-07
dashboard status is minutes, not hours


ghost:
@ChrisBr @bear454 

Chris wrote of the categories at the top of the conference dashboard...
The Hours category sums up the length of each submitted talk. This is useful if you know that you have room / time for X hours program then you can compare it with this number to check if you have already enough submissions.

This is actually the number of _minutes_ (not hours). Nice statistic for measuring various conference areas!!

differentreality:
When in hours, we run into the problem of accuracy, as it doesn't make much sense to show 3.2 hours but we should be able to see that it's more than 3 hours.

Any ideas?

![minutes](https://cloud.githubusercontent.com/assets/4322737/5592723/8b83a14a-91e7-11e4-9c3e-a216da3c78f3.png)

kallecarlkde:
On 01/01/2015 08:54 AM, Stella Rouzi wrote:
> When in hours, we run into the problem of accuracy, as it doesn't make
> much sense to show 3.2 hours but we should be able to see that it's more
> than 3 hours.
>
> Any ideas?
>
> minutes
> <https://cloud.githubusercontent.com/assets/4322737/5592723/8b83a14a-91e7-11e4-9c3e-a216da3c78f3.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openSUSE/osem/issues/519#issuecomment-68491437>.
>

showing "days, hours, minutes" helps convey the information; minutes 
gives it a sense of urgency despite the remaining duration. This format 
can stay in place from start to finish.

Each category needs to accommodate the range of expected numbers...
Days could be as much as triple digits as some CfP deadlines are 4 or 
more months in the future, and "hours" doesn't work for long lead times.
Hours - 24 to 0; minutes 60 to 0.
I think that people would understand ###d, ##h, ##m

There's also an issue of making the information fit in the dashboard 
appropriately...in a line or stacked?

As a conference organizer, I like the countdown to close information 
being available to potential participants.

Carl

shayonj:
I can get going with this. From the above, I feel ```Day, Hour, Min ``` makes sense and can convey accurate information right away. So yes, a countdown in place.

Please correct me if wrong, or there is some other way(s) we can handle this.

hennevogel:
Just hours, rounded please. This is about getting an overview of the state of your submissions not about total accuracy.

ChrisBr:
@shayonj Please note that it's also necessary to change the unit of targets / goals to hours!
[...]

PR #611, opened 2015-03-02
Fixing functions for minutes to hours conversion for dashboard

shayonj:
This solves https://github.com/openSUSE/osem/issues/519

* Rounding the minute conversion to the nearest hour. So ```1.5 hours``` becomes ```2 hours```. Where as ```1.3 hours``` becomes ```1 hour```.
* Changed some naming conventions in ```conference.rb```. So that the method names sound more accurate.
 - Changed previously ```current_program_hours``` to ```current_program_minutes```. Added ```current_program_hours``` which is deriving rounded hours from ```current_program_minutes``` now.
 - Changed previous ```new_program_hours``` to ```new_program_minutes```. Added ```new_program_hours``` which is deriving rounded hours from ```new_program_minutes``` now.
 - Changed ```calculate_program_hours``` to ```calculate_program_minutes``` because it was returning minutes in the first place.
* Changed ```current_program_hours``` to ```current_program_minutes``` in ```target.rb```. Naming changed, rest stays same.
* Updated test (```conference_spec.rb```) to test against the closes hour conversion and not minute.

Maybe we would need tweak more things around or undo previous one. Looking forward to comments/feedback.

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -387,25 +387,43 @@ class Conference < ActiveRecord::Base
   end
 
   ##
-  # Calculates the overall programm hours
+  # Calculates the overall program minutes
   #
   # ====Returns
-  # * +hash+ -> Fixnum hours
-  def current_program_hours
+  # * +hash+ -> Fixnum minutes
+  def current_program_minutes
     events_grouped = events.group(:event_type_id)
     events_counted = events_grouped.count
-    calculate_program_hours(events_grouped, events_counted)
+    calculate_program_minutes(events_grouped, events_counted)
   end
 
   ##
-  # Calculates the overall programm hours since date
+  # Calculates the overall program hours
   #
   # ====Returns
-  # * +hash+ -> Fixnum hours
-  def new_program_hours(date)
+  # * +Fixnum+ -> Fixnum hours. Example: 1.5 gets rounded to 2. 1.3 gets rounded 1.
+  def current_program_hours
+    (current_program_minutes / 60.to_f).round
+  end
+
+  ##
+  # Calculates the overall program minutes since date
+  #
+  # ====Returns
+  # * +hash+ -> Fixnum minutes
+  def new_program_minutes(date)
     events_grouped = events.where('created_at > ?', date).group(:event_type_id)
     events_counted = events_grouped.count
-    calculate_program_hours(events_grouped, events_counted)
+    calculate_program_minutes(events_grouped, events_counted)
+  end
+
+  ##
+  # Calculates the overall program hours since date
+  #
+  # ====Returns
+  # * +Fixnum+ -> Fixnum hours
+  def new_program_hours(date)
+    (new_program_minutes(date) / 60.to_f).round
   end
 
   ##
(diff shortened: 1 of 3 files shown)


app/models/conference.rb
```diff
@@ -387,25 +387,45 @@ def self.user_distribution
...
   ##
-  # Calculates the overall programm hours since date
+  # Calculates the overall program hours
   #
   # ====Returns
-  # * +hash+ -> Fixnum hours
-  def new_program_hours(date)
+  # * +Fixnum+ -> Fixnum hours. Example: 1.5 gets rounded to 2. 1.3 gets rounded 1.
+  def current_program_hours
+    minutes_standard = 60.to_f
```
hennevogel: Whats the use of this assignment?

app/models/conference.rb
```diff
@@ -387,25 +387,45 @@ def self.user_distribution
...
+    hours = (current_program_minutes / minutes_standard).round
```
hennevogel: Whats wrong with 
```ruby
def current_program_hours
  (current_program_minutes / 60).round
end
```

shayonj: I assumed using a float value, we can round of to the nearest hour. Because right if the minutes is ```110``` then without the float value the rounding off will result in ```1 hour``` instead of ```2 hours```. 

But yes I am removing the variable assignment. Not really needed, i realized.

hennevogel:
Now all that is missing is testing both functions...

ChrisBr:
@shayonj You didn't changed the function call in the controller?
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/conference_controller.rb#L107

Also I think it's necessary to write a migration for the targets because we saved hours and not minutes!
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L417

@hennevogel

hennevogel:
@ChrisBr Conference.current_program_hours is now really hours, not jut in the name of the method.
And how can Target have saved hours if there was no code to convert the old Conference.current_program_hours which always returned minutes?

shayonj:
@ChrisBr So yes, I changed the contents of the function that actually calculates the hours now. Because, before it was just minutes. So at - https://github.com/openSUSE/osem/blob/master/app/controllers/admin/conference_controller.rb#L107 it is actually calculating hours in instead of minutes. Which is based on the below code changes

```
def current_program_minutes
    events_grouped = events.group(:event_type_id)
    events_counted = events_grouped.count
    calculate_program_minutes(events_grouped, events_counted)
  end

  def current_program_hours
    (current_program_minutes / 60.to_f).round
  end
```

Similar for new program hours and minutes.

ChrisBr:
@shayonj I will have a look at it on the weekend!

shayonj:
Thanks @ChrisBr. No rush :).

ChrisBr:
@shayonj LGTM :+1: 

@hennevogel But what we have to bear in mind: There is now a mismatch between tagets / goals and the dashboard! A target has the unit minutes and in the dashboard we now have hours. So we have to migrate the targets / goals also to hours, otherwise it's hard to compare the values ... What do you think, do we want to fix that in an additional PR?
@shayonj Can you fix that also?

shayonj:
@ChrisBr sure, I can do.

I will wait for @hennevogel to get back on the new PR reply and then accordingly proceed :)

hennevogel:
@ChrisBr moving target to hours sounds like a plan

ChrisBr:
@shayonj I will merge this PR! As described, we want to switch the targets unit also from minutes to hours, if you could do that, it would be awesome!

PR #612, opened 2015-03-03
Implements program section for unscheduled events

ChrisBr:
As long as the schedule is not public an overview of the confirmed events is shown instead.
Links to the program are on the index page and the conference splash page.

Program


ChrisBr:
@hennevogel @differentreality please review

@hcderaad

app/models/event.rb
```diff
@@ -209,7 +209,7 @@ def speaker_names
     speakers.each do |speaker|
       result.push(speaker.name)
     end
-    result.to_sentence
+    result.uniq.to_sentence
```
hennevogel: Please use a [Ruby set](http://ruby-doc.org//stdlib-2.1.1//libdoc/set/rdoc/Set.html) here.

app/views/conference/_program_splashpage.html.haml
```diff
@@ -23,7 +23,12 @@
         %p.cta-button.text-center
           = link_to(schedule_conference_path(@conference.short_title), class: 'btn btn-default btn-lg') do
             Full Schedule
-
+  - elsif @conference.events.confirmed.any?
+    .row
+      .col-md-12
+        %p.cta-button.text-center
+          = link_to(schedule_conference_path(@conference.short_title), class: 'btn btn-default btn-lg') do
+            View confirmed events
```
hennevogel: No visitor knows what that means.

hennevogel:
How do I not show the confirmed events?

app/models/event.rb
ChrisBr: Okay

app/views/conference/_program_splashpage.html.haml
ChrisBr: Do you have a better name? Maybe just "Program"?

ChrisBr:
@hennevogel That's not possible! 
- If there are confirmed events the program page is displayed
- If the schedule is ready and public the schedule is displayed
- Otherwise the schedule page is not visible at all

Any concerns with this workflow?

hennevogel:
Yeah sure. What if I don't want to reveal my program before I have a schedule ready?

ChrisBr:
@hennevogel Okay, what do you suggest? A boolean flag in cfp like we did it for schedule public?

hennevogel:
I'm still a fan of integrating schedule and listing and control it via the setting that is already there.

ChrisBr:
But then you have an empty schedule displayed which is
- ugly 
- and maybe misleading

app/views/conference/_program_splashpage.html.haml
hennevogel: Schedule? ;-)

hennevogel:
What's the problem with catching this case and displaying nothing?

ChrisBr:
@hcderaad sorry for the delay ...

No scheduled events:
Program


No scheduled events:
Program

With scheduled events:
Program

@hennevogel @differentreality please review

hcderaad:
@ChrisBr  No problem, thanks a lot for your efforts, this looks very usable! Looking forward to see this added when done!

differentreality:
When we have a schedule, we should only show extra content (confirmed, not scheduled) in the other tab, right? So perhaps Program tab should be named *More content* ?

On a sidenote, we should stick to the usage of a single word in general. Splashpage configuration says **include program**', when it should really say **'include schedule**.

hennevogel:
Don't make it too complicated. It's not "extra content" it's another view of our events.

ChrisBr:
@hennevogel @differentreality So we will stay with Program?

@differentreality For the second point I agree with you!

hennevogel:
LGTM :+1:

hcderaad:
@ChrisBr @hennevogel @differentreality  Thank you all for your effort in implementing this feature!

Issue #215, opened 2014-06-06
Better User Onboarding


danimo:
People seem to be confused about the two-step way of registering for a conference. They think they are registered for the conference after creating an OSEM account.

Proposal: Registering for your first conference should transparently create an OSEM account (unless you have one, which could be detected through the email address).

jospoortvliet:
I was wondering if there has been any progress on this - seeing one part got merged but I dunno how much that deals with this issue... Before we can use it for 2015, this is kind'a a must-have feature... as the painful onboarding it turns off a lot of people and we want to *attract* people to our event :D

ChrisBr:
@jospoortvliet Thanks for your response!
For events.o.o we now use the opensuse account, so it's not necessary to create an OSEM account explicit! But an explicit registration for the conference is still necessary!
But I don't have a good idea at a first glance how we can improve the onboarding with the detour to the opensuse registration proxy!
@hennevogel @differentreality maybe we should discuss / think about this issue again

hennevogel:
We definately need to think about onbording. Not only limited to the sign up process but in general.

jospoortvliet:
We're considering using OSEM again for the ownCloud conference - but forcing people to sign up twice as it is today is not going to work. If it isn't possible to present a single page where you at once can register for account AND conference, I'll have to use another tool :(

It can't be that hard to add user registration to the registration for a conference, is it? I don't care for a second about email verification - that is an entirely unnecessary step in the process.

I'll be announcing the conference next week so I guess I can forget about this - but I don't understand why this isn't a bigger priority.

@ChrisBr you are saying that you don't allow people to register for the openSUSE conference without having created an openSUSE account? You really don't want people to register for the openSUSE Conference, do you? That is a HUGE barrier for many people, not much better than asking an entree fee!

Heck, I PREFER giving my credit card details and paying 10 bucks over creating another account on another random website... And I know that goes for many, if not most internet users. Making people give their mail account and lots of other details, making them go through a verification step and then ANOTHER registration step just to say "yes, I am coming" - it's no surprise we always had so many people show up who had not registered!

Sorry for the rant, but registration is the first thing users see from a conference system - and in case of OSEM, it is the last many see, too. That just plain sucks.

hennevogel:
Just for you :rabbit: I'll spend an evening on this on Thursday.

:kiss: your Henne

jospoortvliet:
You're too nice. And that's serious :D

You will get beer from me in Den Hague.

[...]

PR #615, opened 2015-03-12
Implements sign up during registration. Solves most of #215

hennevogel:


spec/features/ability_spec.rb
```diff
@@ -226,77 +229,5 @@
 
     visit admin_conference_commercials_path(conference3.short_title)
     expect(current_path).to eq(admin_conference_commercials_path(conference3.short_title))
-
-  end
-
-  scenario 'when user is volunteer coordinator' do
```
ChrisBr: Why do we remove this test?

config/initializers/devise.rb
```diff
@@ -105,7 +105,7 @@
   # able to access the website for two days without confirming his account,
   # access will be blocked just in the third day. Default is 0.days, meaning
   # the user cannot access the website without confirming his account.
-  # config.allow_unconfirmed_access_for = 2.days
+  config.allow_unconfirmed_access_for = 2.days
```
ChrisBr: So we don't want to get of the confirmation as Jos suggested?
Anyway, I think this is a good compromise...

ChrisBr:
Otherwise LTGM :+1: 
Long desired feature implemented :)

jospoortvliet:
/me is awesomely happy ;-)

I can't judge the code very well (I did look over it but what do I know about Ruby? PHP still scares me) but we'll test it before we send people to it ;-)

config/initializers/devise.rb
hennevogel: I think it's fair to require confirmation after 2 days...

spec/features/ability_spec.rb
differentreality: @hennevogel ?

hennevogel: Because we don't have an ability based on volunteer coordinator. Just for dead objects (Vposition, Vday) which we don't expose anymore

PR #620, opened 2015-03-16
Changed minutes to hours for target/program

shayonj:
Showing hours for targets, just like it is for conference now.

Something to note:  The same rounding up logic applies here. That is, 1.5 hours rounds up to 2. Like it is being done for the conference dashboard. Maybe we need a different behavior?

### DIFF
--- a/app/controllers/admin/conference_controller.rb
+++ b/app/controllers/admin/conference_controller.rb
@@ -149,7 +149,6 @@ module Admin
       # get targets
       @registration_targets = @conference.get_targets(Target.units[:registrations])
       @submission_targets = @conference.get_targets(Target.units[:submissions])
-      @program_minutes_targets = @conference.get_targets(Target.units[:program_minutes])
 
       # get campaigns
       @campaigns = @conference.get_campaigns
--- a/db/migrate/20150317210343_convert_target_minutes_to_hours.rb
+++ b/db/migrate/20150317210343_convert_target_minutes_to_hours.rb
@@ -1,0 +1,22 @@
+class ConvertTargetMinutesToHours < ActiveRecord::Migration
+  class TempTarget < ActiveRecord::Base
+    self.table_name = 'targets'
+    attr_accessible :target_count, :unit
+  end
+
+  def up
+    targets = TempTarget.where(unit: 'Program minute').where.not(target_count: nil)
+    targets.each do |target|
+      new_target_count = (target.target_count / 60.to_f).round
+      target.update_attributes(target_count: new_target_count, unit: 'Program hours')
+    end
+  end
+
+  def down
+    targets = TempTarget.where(unit: 'Program hours').where.not(target_count: nil)
+    targets.each do |target|
+      new_target_count = (target.target_count * 60.to_f).round
+      target.update_attributes(target_count: new_target_count, unit: 'Program minute')
+    end
+  end
+end
--- a/app/models/target.rb
+++ b/app/models/target.rb
@@ -9,7 +9,7 @@ class Target < ActiveRecord::Base
     {
       registrations: 'Registration',
       submissions: 'Submission',
-      program_minutes: 'Program minute'
+      program_hours: 'Program hours'
     }
   end
 
@@ -33,8 +33,8 @@ class Target < ActiveRecord::Base
       numerator = conference.events.where('created_at < ?', due_date).count
     elsif unit == Target.units[:registrations]
       numerator = conference.registrations.where('created_at < ?', due_date).count
-    elsif unit == Target.units[:program_minutes]
-      numerator = conference.current_program_minutes
+    elsif unit == Target.units[:program_hours]
+      numerator = conference.current_program_hours
     end
     (numerator / target_count.to_f  * 100).round(0).to_s
   end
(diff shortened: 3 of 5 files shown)


app/controllers/admin/conference_controller.rb
```diff
@@ -149,7 +149,6 @@ def show
       # get targets
       @registration_targets = @conference.get_targets(Target.units[:registrations])
       @submission_targets = @conference.get_targets(Target.units[:submissions])
-      @program_minutes_targets = @conference.get_targets(Target.units[:program_minutes])
```
shayonj: this instance variable isnt being used anywhere.

ChrisBr:
@shayonj I just skimmed it but I think here you need a migration for the existing targets!

db/migrate/20150317210343_convert_target_minutes_to_hours.rb
```diff
@@ -0,0 +1,9 @@
+class ConvertTargetMinutesToHours < ActiveRecord::Migration
+  def change
```
ChrisBr: Please add a temporary target class to the migration to be more robust against renaming/changing the model.
(like we did it in previous migrations
https://github.com/openSUSE/osem/blob/master/db/migrate/20141128073306_migrate_data_remove_column_include_cfp_in_splash_add_column_include_cfp.rb#L6)
See http://railsguides.net/change-data-in-migrations-like-a-boss/ for details!

ChrisBr:
@shayonj I will review your PR in detail tomorrow!

db/migrate/20150317210343_convert_target_minutes_to_hours.rb
shayonj: @ChrisBr thanks for the note. That was a new thing to learn.

I am now wondering if we would need a ```down``` method for rollback changes, where would need to shift the hours back to minutes? Pretty much reverse whatever and exactly how it is happening in the ```up``` method?

db/migrate/20150317210343_convert_target_minutes_to_hours.rb
```diff
@@ -0,0 +1,13 @@
+class ConvertTargetMinutesToHours < ActiveRecord::Migration
+  class TempTarget < ActiveRecord::Base
+    self.table_name = 'targets'
+  end
```
shayonj: @ChrisBr thanks for the note. That was a new thing to learn.

I am now wondering if we would need a ```down``` method for rollback changes, where would need to shift the hours back to minutes? Pretty much reverse whatever and exactly how it is happening in the ```up``` method?

differentreality: Yes, an up and a down.

Otherwise, as far as I know, except for add/remove columns and such, the same code of the change fuction will be run and this is not exactly what we ideally need (we would need to get the previous value back). Makes sense?

shayonj: @differentreality yes having an up and down definitely makes more sense. I will push the new changes in the morning. Thanks

shayonj:
@ChrisBr , @differentreality I added the ```up``` and ```down``` methods accordingly to the migration file. But the issue is with the rounding up that is being done on the conference (dashboard) and target methods. That is, at the time of running the migration if the minutes is ```112``` then the hours will be resulted in ```2``` (taking the nearest value). Now if you rollback, then ```60*2``` will get us ```120```, so we already lose the actual value which was ```112```. 

Thoughts?

differentreality:
Right, should we be storing minutes (our base unit) in DB and present the value formatted the way we want in the view instead?

ChrisBr:
Good point @differentreality, that would be the best solution. @shayonj I would discard the idea with the migration and add instead a conversion method to target. Sound that reasonable?

shayonj:
@ChrisBr , @differentreality so reset everything back to how it was, in this PR or just the migration?

ChrisBr:
@shayonj Not reset back to how it was. I would drop the migration and still save the minutes. In the target model I would implement a conversion method from minutes to hours!
@differentreality

differentreality:
The previous PR https://github.com/openSUSE/osem/commit/ff124632ef4337c6c03dc82960f20b989ea34614 mainly changed the name of the functions from hours to minutes. We still need those functions to depict what we really have (which is minutes, not hours).

Since minutes is our base unit, I think we should store minutes into our DB, calculate in minutes the value we will storte to the DB, fetch the value from DB in minutes, and while we have that value in minutes we can have 1 function to convert minutes to hours (helper function perhaps to be used from other models if/when needed?).

shayonj:
I see. Ok. Yes I was thinking of a helper too. Where you pass the minutes to get the hours. 

Now, there are few functions from the previous PR as @differentreality mentioned, that calculate the hour in conference like ```current_program_hours``` and ```new_program_hours```. We touch them now, or let them be as is ?

ChrisBr:
@shayonj Is it possible to simply move this methods to a helper class and then use it in both scenarious (targets and dashboard)?

shayonj:
@ChrisBr I dont think so it is 100% viable for dashboard/conference. A good of it is being used in the controllers.

Plus for target, to use a helper we would have to write a condition to check and make sure the unit is "Program minutes". I think it will be easier/neater to have a method in the model that does the conversion like it is right now in conference and then let the model method for conversion pass the values accordingly to controllers.

ChrisBr:
@shayonj Okay, sounds reasonable :)

shayonj:
@ChrisBr , cool I will try to get it updated asap :)

hennevogel:
Please reopen once you find the time. Thanks!

PR #621, opened 2015-03-28
no authorization for Conference in subscriptions controller

differentreality:
[Authorization in nested resources] (https://github.com/CanCanCommunity/cancancan/wiki/Nested-Resources) Conference is being authorized with read action which includes index and show actions, however action show is authorized only for Conference with public splashpage. 

Therefore Conference is not authorized and raises an exception when trying to subscribe to a conference without a public splashpage.


ChrisBr:
@differentreality LGTM

PR #624, opened 2015-04-06
Show registration help link even when openID is disabled

differentreality:
Also use .container for sign_in page so that the input fieds look properly:

screenshot-osem - chromium-1
screenshot-osem - chromium


hennevogel:
:+1:

differentreality:
Also fixed in #639, do we prefer changes as in #639 (with embedded forms moved into their own partials)?

hennevogel:
Yes we do

PR #627, opened 2015-04-09
Redirect to registration edit page when user is already registered

ChrisBr:
Fixes 

```
I tried to update the resgistration, but this error appears: An error
prohibited the registration for openSUSE Conference: User already
Registered!. Once is fixed I will retry.
```
reported by @aplanas


hennevogel:
:+1:

PR #630, opened 2015-04-14
Last stage of refactoring of the ability model and spec

hennevogel:


spec/models/conference_spec.rb
```diff
@@ -510,13 +510,13 @@
...
-      @workshop = create(:event_type, title: 'Workshop', color: '#000000')
-      @lecture = create(:event_type, title: 'Lecture', color: '#ffffff')
+      @workshop = create(:event_type, title: 'Workshop', color: '#000000', conference: subject)
+      @lecture = create(:event_type, title: 'Lecture', color: '#ffffff', conference: subject)
     end
 
     describe '#event_type_distribution' do
       it 'calculates correct for different event types' do
-        create(:event, conference: subject, event_type: @workshop)
+        blah = create(:event, conference: subject, event_type: @workshop)
```
ChrisBr: This was only for debugging, we should remove it.
(Rubocop also reports about it!)

spec/features/proposal_spec.rb
```diff
@@ -1,19 +1,15 @@
 require 'spec_helper'
 
 feature Event do
-  let!(:conference) { create(:conference) }
+  let!(:conference) { create(:conference, call_for_paper: create(:call_for_paper) ) }
```
ChrisBr: Reported by Rubocop: Space inside parantheses!

PR #634, opened 2015-04-16
Refactoring of proposal workflow. Fixes #215

### LINKED ISSUE #215: Better User Onboarding

differentreality:


app/views/proposal/_encouragement_text.html.haml
```diff
@@ -0,0 +1,23 @@
+%p.lead
+  - if @conference.event_types.any?
+    You can submit proposals for
+    = "#{event_types(@conference)}."
+  - if @conference.tracks.any?
+    Proposals should fit in one of the
+    = "#{pluralize(@conference.tracks.count, 'track')}:"
+    = "#{tracks(@conference)}."
+  - if @conference.call_for_paper
```
ChrisBr: This was not a bug, normally this should NEVER be happen. If there is no CfP it shouldn't be possible to enter this page at all. We should catch this with an ability, something like this: 

```
can :show, :index, :new Event do |event|
   event.conference.cfp && event.conference.cfp.start_date <= Date.today etc
end
```

differentreality: I think the ability should only refer to :new, :create actions. 
Why not access your proposals index page whenever? Before cfp, after cfp... 

We should also make the 'New Proposal' button only appear accordingly.
This view could still stay as it is I guess.

hennevogel:
then :+1: let's go!

Issue #633, opened 2015-04-15
Commercial Submission


hennevogel:
Instead of explaining what to copy from what link, parse whatever people input to the ID field. And once this field looses focus, display the commercial.

ChrisBr:
@sysrich requested a owncloud support! I will have a lookt at it!

hennevogel:
Please do this in a separate issue :-)

hennevogel:
This feature is done since some time

PR #637, opened 2015-04-17
Implement simplified commercial integration (#633)

ChrisBr:
Fixes (#633)
It's now possible to simply enter an URL instead of copying the id / whatever.
http://paste.opensuse.org/view/raw/14705047
http://paste.opensuse.org/view/raw/64575612

TODO:
- tests / testing
- migration (don't know how to generate the old url out of the id for speakerdeck)

@differentreality @hennevogel please review!


db/migrate/20150417050953_add_url_to_commercial.rb
```diff
@@ -0,0 +1,40 @@
+class AddUrlToCommercial < ActiveRecord::Migration
+  class TempCommercial< ActiveRecord::Base
+    self.table_name = 'commercials'
+  end
+
+  def change
+    add_column :commercials, :url, :string
+
+    # Generate the url from the commercial type and id
+    TempCommercial.each do |commercial|
```
differentreality: TempCommercial.all.each

ChrisBr: True!

ChrisBr:
I dropped the idea with the migration. The problem is that it isn't possible to automatically generate the url from the id and type. Therefore I simply check in the view if there is an url otherwise I render the old partial!

@hennevogel @differentreality

ChrisBr:
Still a few tests are failing but hopefully I can fix it this week!

spec/features/ability_spec.rb
```diff
@@ -167,7 +167,7 @@
     expect(current_path).to eq(root_path)
 
     visit admin_conference_commercials_path(conference2.short_title)
-    expect(current_path).to eq(admin_conference_commercials_path(conference2.short_title))
+    expect(current_path).to eq(root_path)
```
ChrisBr: @differentreality @hennevogel any idea why this test passed before?
I looked in the ability.rb and this behavior should be correct!

spec/features/ability_spec.rb
```diff
@@ -238,6 +238,6 @@
     expect(current_path).to eq(admin_conference_questions_path(conference3.short_title))
 
     visit admin_conference_commercials_path(conference3.short_title)
-    expect(current_path).to eq(admin_conference_commercials_path(conference3.short_title))
+    expect(current_path).to eq(root_path)
```
ChrisBr: same here!

ChrisBr:
@hennevogel @differentreality Fixed tests, please review!

spec/features/ability_spec.rb
differentreality: https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L97

ChrisBr: Any idea why the test fails now with 
```
expect(current_path).to eq(admin_conference_commercials_path(conference2.short_title))
```
No idea what I have changed what the ability altered?

hennevogel:
There is no indication on the page that you fetch the oembed. Either you add some [spinner](http://fortawesome.github.io/Font-Awesome/examples/#animated) or as soon as focusout occurs you render a "blank" commercial and move the input field down (that's how people usually do this)

hennevogel:
Oh and fetching the embed also should trigger on [paste](http://www.quirksmode.org/dom/events/cutcopypaste.html)

app/models/commercial.rb
```diff
@@ -1,10 +1,43 @@
...
+        OEmbed::Providers::Instagram,
+        speakerdeck
+    )
 
-  validates :commercial_type, inclusion: { in: CONFIG['commercial_types'].values }
+    begin
+      resource = OEmbed::Providers.get(url, maxwidth: 560, maxheight: 315)
+      resource.html
+    rescue StandardError
+      ''
```
hennevogel: huh?

ChrisBr: OEmbed::Providers.get throws an Error when you enter an invalid url (not supported provider eg) and then I return an empty string to display nothing in the view! 
But yes, we can discuss about a better approach :) Any suggestion?

hennevogel: Tell people the truth! :-) Especially about the cases like OEmbed::NotFound, OEmbed::FormatNotSupported and maybe [timeout](https://github.com/judofyr/ruby-oembed/pull/35)

ChrisBr: Okay, good idea!

app/controllers/admin/commercials_controller.rb
```diff
@@ -42,6 +38,10 @@ def destroy
       redirect_to admin_conference_commercials_path, notice: 'Commercial was successfully destroyed.'
     end
 
+    def get_html
```
hennevogel: The function does not what it's name says it will do...

app/controllers/commercials_controller.rb
```diff
@@ -35,10 +28,14 @@ def update
 
   def destroy
     @commercial.destroy
-    redirect_to edit_conference_proposal_path(conference_id: @conference.short_title, id: @event.id),
+    redirect_to edit_conference_proposal_path(conference_id: @conference.short_title, id: @event.id, anchor: 'commercials-content'),
                 notice: 'Commercial was successfully destroyed.'
   end
 
+  def get_html
```
hennevogel: Same here. The function does not what it's name says it will do...

app/models/commercial.rb
```diff
@@ -1,10 +1,43 @@
...
+
   belongs_to :commercialable, polymorphic: true
 
-  attr_accessible :commercial_id, :commercial_type
+  attr_accessible :url
+
+  validates :url, presence: true
+  validates :url, format: URI::regexp(%w(http https))
+
+  def self.provider
```
hennevogel: This and OEmbed::Providers.register seems to be duplication. Can't you ask OEmbed::Providers in this method?

app/views/shared/_media_item.html.haml
```diff
@@ -1,12 +1,12 @@
--if commercial_type == CONFIG['commercial_types']['slideshare']
+-if commercial_type == Commercial.provider['slideshare']
   %iframe{:width=>"560", :height=>"315", :frameborder=>"0", :allowfullscreen=>"true", :src=> (request.ssl? ? "https://" : "http://") + "www.slideshare.net/slideshow/embed_code/#{commercial_id}"}
```
hennevogel: :width=>"100%"
:height=>"100%"

should work well with a [responsive embed](http://getbootstrap.com/components/#responsive-embed)

spec/features/ability_spec.rb
hennevogel: As the tests run through, is this still an issue?

ChrisBr: I guess yes, as @differentreality stated out, we allow indexing of conference Commercials! @hennevogel any idea why this fails now?

Anyway, I will investigate it!

app/models/commercial.rb
ChrisBr: I think we can drop this anyway! This was necessary for the drop down menu. I guess we could also hard code the provider in the partial for backward compatibility!

hennevogel: Isn't there a way to see which providers are registered in OEmbed?

spec/features/ability_spec.rb
hennevogel: We also allow [show](https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L50)

app/models/commercial.rb
ChrisBr: This is only for backward compatibility because I don't drop the old fields in the DB! If there is a way to see the registered providers, I don't think that they correspond with the value we stored in the DB! Anyway, this is only for backward compatibility, so I think dropping this method and hardcode in the view should be OK. We needed the function for the drop down menu of the old functionality...

hennevogel: I don't get it. backward compatibility?

ChrisBr:
JFYI: I'm still working on this :)
It still needs some styling ;)

hennevogel:
What's the status of this now? :-)

ChrisBr:
Haha :) It's summer... I don't think that I'll work the next few weeks on it ... But I'll definitely finish it :)

ChrisBr:
I plan to finish this this week! I close this PR and open a new one as soon as I finished it, too much noise in this one ...

Issue #625, opened 2015-04-08
medium usability issue: no "forgot password" at signin option on conference registration


jospoortvliet:
Hi,

When registering for a conf, there's the 'sign in' option for those who already have an account. But unlike the sign in option on top of the screen, this one has no 'forgot password' or 'reset password' option. Would be nice to have ;-)

Probably a relatively simple thingy...

jancborchardt:
Seconding that, this confused me a lot.

On a similar note: In the top right sign in dropdown, why is a click on »Need help?« needed to reveal the »Forgot your password« option? If it only reveals that one thing, might as well just show that already.

hennevogel:
It reveals whatever is configured which can be one or a gazillion options...

hennevogel:
Oh and as #639 is merged this is fixed. You fine ownCloud people should update your instance :-)

jancborchardt:
Great stuff, thanks @differentreality for fixing! @jospoortvliet get to updating! :D

jospoortvliet:
Unless there's a proper release which has had testing and can be upgraded (taking care of migrating database etc) too I'd rather keep our instance running as it is ;-)

PR #639, opened 2015-04-18
Move common views to partial

differentreality:
https://github.com/openSUSE/osem/issues/625
We ended up with a lot of common code in multiple sections of the app, such as 
* sign_in form (embedded in conference_registration and proposal)
* sign_up form (embedded in conference_registration and proposal)
* help link for devise links
* openid sign_in

Moreover in some places the code was slightly different for no actual reason.
I gathered them in partials so that we have 1 single point of reference for doing the exact same thing.

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -21,6 +21,8 @@ class ConferenceRegistrationsController < ApplicationController
     end
 
     @registration = Registration.new
+
+    # @user variable needs to be set so that _sign_up_form_embedded works properly
     @user = @registration.build_user
   end
 
@@ -34,7 +36,15 @@ class ConferenceRegistrationsController < ApplicationController
 
   def create
     @registration = @conference.registrations.new(registration_params)
-    @registration.user = current_user || ( @user = @registration.build_user(user_params))
+
+    if current_user == nil
+      # @user variable needs to be set so that _sign_up_form_embedded works properly
+      @user = @registration.build_user(user_params)
+    else
+      @user = current_user
+    end
+
+    @registration.user = @user
 
     if @registration.save
       # Trigger ahoy event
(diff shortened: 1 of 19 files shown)


app/controllers/conference_registrations_controller.rb
```diff
@@ -27,9 +33,8 @@ def show
   def edit; end
 
   def create
-    @registration = Registration.new(registration_params)
-    @registration.conference = @conference
-    @registration.user = current_user if current_user
+    @registration = @conference.registrations.new(registration_params)
+    @registration.user = current_user || ( @user = @registration.build_user(user_params))
```
hennevogel: Are you sure this is doing what you think it does? :-)

app/views/devise/shared/_help.html.haml
```diff
@@ -0,0 +1,5 @@
+%p.text-right
+  %a.small.btn.btn-default{"data-toggle" => "collapse", "data-target" =>  "#devise-help"}
```
hennevogel: I guess you meant btn-xs?

hennevogel:
password#new and confirmation#new are still unstyled

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,12 @@
 module ApplicationHelper
+  def resource_name
+    :user
+  end
+
+  def devise_mapping
+    @devise_mapping ||= Devise.mappings[:user]
+  end
+
```
hennevogel: Please document what they do and why they are there

app/views/devise/shared/_help.html.haml
differentreality: No, but could be :)

app/controllers/conference_registrations_controller.rb
differentreality: Could you elaborate on that?

hennevogel: To me it looks not so clear what happens. Maybe more brackets would help...

hennevogel:
Anyway, let's merge this :+1:

differentreality:
@hennevogel are we good to go with this PR?

hennevogel:
Sure :fireworks:

Issue #640, opened 2015-04-20
Scheduling problems with drag and drop


hcderaad:
There are some serious problems with the scheduling in OSEM as we're in the process of organising openSUSE Conference 2015.
At the scheduling page I get all kinds of sessions that have nothing to do with oSC15.
https://events.opensuse.org/admin/conference/osc15/schedule
 
Also the list on the left is not scrollable so it's also impossible to ignore the irrelevant (at least for oSC15) talks.
Can this interface be adapted so that it only shows talks that have actually been accepted and confirmed? And also show the same titles as in the list overview?

ChrisBr:
Hi @hcderaad,

first of all, before you can schedule your events you need to create some rooms
(https://events.opensuse.org/admin/conference/osc15/rooms), otherwise scheduling isn't possible!

What do you mean with ```all kinds of sessions that have nothing to do with oSC15.``` exactly? Which events aren't from the OSC15? Selecting only the accepted and confirmed events shouldn't be a big deal, I will have a look at it!

hcderaad:
Hi @ChrisBr Thanks for responding, I'm seeing all kinds of talk in the schedule list that are not registered for oSC15. Regarding the rooms, valid point thanks, will do that!

differentreality:
The 'Unscheduled Events' list shows 4 confirmed but unscheduled events from summit asia and 1 confirmed/scheduled talk from osc14.

Except for the first 5, the others seem to be valid proposals of oSC15.

@hcderaad 
Step 1: Add rooms
Step 2: Ignore the first 5 entries and start scheduling the rest
(You should not be able to put in the schedule talks that are not of your conference anyway)
(As you schedule talks, more on the list will - magically - appear.)

Maybe not the most optimized way to schedule, but it has worked before :)

As soon as the rooms are into place, let us know if you need further assitance with anything.

In the meantime, we are obviously investigating the bug :)

hcderaad:
@differentreality Thanks for your feedback, Step 1 rooms have now been created. Step 2 is a problem, especially since the tracks for oSC15 have quite some overlap with earlier years (and other events like the Asia Summit). And since the titles also tend to overlap how do we know that the tracks we're scheduling are actually oSC15 tracks? 
Looking forward to your feedback on this issue, we'd really like to start scheduling today.

differentreality:
Fixed in #643

PR #643, opened 2015-04-20
fix scrolling of left column in schedule, show only @conference.events

differentreality:
https://github.com/openSUSE/osem/issues/640


hennevogel:
LGTM

PR #647, opened 2015-04-23
Some more sign in/up & ichain stuff

hennevogel:


app/models/ability.rb
```diff
@@ -49,12 +49,13 @@ def not_signed_in
...
-      registration.new_record?
-    end
-
-    can [:show, :create], Event do |event|
-      event.new_record?
+    unless CONFIG['authentication']['ichain']['enabled']
+      can [:show, :create], Registration do |registration|
+        registration.new_record?
+      end
+      can [:show, :create], Event do |event|
```
ChrisBr: It should be possible to show the event as unregistered user, shouldn't it?

config/environments/development.rb
```diff
@@ -85,4 +85,12 @@
                               }
                             )
 
+
+  config.after_initialize do
```
ChrisBr: Why do we need this?

ChrisBr:
and don't we need a conditional statement also here:
https://github.com/openSUSE/osem/blob/master/app/views/proposal/new.html.haml#L25

app/models/ability.rb
hennevogel: [It is](https://github.com/hennevogel/osem/blob/master/app/models/ability.rb#L45) see above

config/environments/development.rb
hennevogel: To make it easier to develop in ichain mode.

hennevogel:
No as we never reach the case where there isn't a current user

app/models/ability.rb
ChrisBr: :+1:

config/environments/development.rb
ChrisBr: Okay, is it now enabled by default for development or how can I enable it?

ChrisBr:
@hennevogel Ah yes true, because of the ability! Got it!

:ship: it!

config/environments/development.rb
hennevogel: You still have to enable it in your [config](https://github.com/openSUSE/osem/blob/master/config/config.yml.example#L22) but when you do it is straight in test mode. No need to edit the [initializer](https://github.com/openSUSE/osem/blob/master/config/initializers/devise.rb#L270) anymore.

PR #654, opened 2015-05-02
Update edit.html.haml

skriesch:
I want to update the page.


hennevogel:
In the end we took Tommys change but we could easily have taken yours, so thanks anyway Sarah! What are you going to fix next? :-)

Issue #603, opened 2015-02-27
EmailSetting incosistency


hennevogel:
In EmailSettings attributes should be called

:send_on_$ACTION
:$ACTION_body
:$ACTION_subject

For instance:

```ruby
:send_on_registration
:registration_body
:registration_subject
```
Refactor all the EmailSettings attributes to follow this schema.

shubhankardesh:
So for example the attribute accepted_email_template should be called accepted_body and likewise for the remaining attributes, right?

hennevogel:
Right.

skriesch:
registration_body doesn't exist at the moment. But I can add it.

We have got the problem, that all the EmailSettings attributes are global and will be used in different classes. You can't change it in /app/models/email_settings.rb without changing it in the other classes or having another solution.

My solutions:
1) using sed and changing the attributes of email_settings.rb everywhere
2) adding all the new names for attributes and assigning the old attributes to the new attributes, so that you can use both

skriesch:
my workflow steps:
- adding new attributes
- changing all in other classes 
- adding a if for testing, whether something was forgotten
- removing old attributes

The other way would cost O(n^x) for x attributes.

skriesch:
Before editing all I want to know, how do we get all the parameters for $ACTION like accepted, rejected and registration after this change?

ChrisBr:
@skriesch It's not as complicated as you think!

I suggest the following workflow:
1. Write a migration to alter the attributes to follow the suggested naming schema
  - See @hennevogel comment above for the suggested naming schema
  - http://edgeguides.rubyonrails.org/active_record_migrations.html
2. Rename the attributes in model, controller, views, tests etc.

That should be all!

hennevogel:
@skriesch does that answer your question? :-)

skriesch:
Sorry. No. The tutorial is explaining the migration of records in a database or renaming them there. They exist here. We want to use the old names and $ACTION should give the right parameters to the special global attribute. I haven't seen something like that in attributes anywhere before... Ok.

My question is: How and where does $ACTION get all the parameters like registration, accepted and rejected? I have looked after something like that in the code, but didn't find it.

ChrisBr:
@skriesch iirc this issue is about renaming the attributes (and then you also have to rename the attributes in the database). See
https://github.com/openSUSE/osem/blob/master/app/models/email_settings.rb#L2
The attributes don't follow @hennevogel naming schema!

[...]

PR #655, opened 2015-05-14
Update email_settings.rb

skriesch:
add 'send_on_$ACTION false', registration_body, $ACTION_subject and $ACTION_body for #603


ChrisBr:
See my comment in #603

PR #656, opened 2015-05-16
Make talks on schedules table real links

bgeuken:
Previously the link functionality (open in window or tab) of those talks was
done via javascript. Because of that they couldn't be selected via right click
to be opened in a tab, copied or bookmarked.

This PR fixes this.


ChrisBr:
Awesome :clap: :clap: :clap: 
That was really ugly before...

LGTM :+1:

PR #635, opened 2015-04-17
Refactor calculate progress

bgeuken:

PR #657, opened 2015-05-16
Calculate progress

bgeuken:
Basically that's a c&p from #635. Might be worth to merge the code in a helper.


hennevogel:
LGTM :+1: thanks

PR #661, opened 2015-05-31
Create 201505312045_rename_email_settings.rb

skriesch:
Refactoring of email_settings for #603 

I have spoken with @differentreality on IRC. I have misunderstood the issue. I created this file for renaming all email_template columns to body.

### DIFF
--- a/db/migrate/201505312045_rename_email_settings.rb
+++ b/db/migrate/201505312045_rename_email_settings.rb
@@ -1,4 +1,4 @@
-class rename_email_settings < ActiveRecord::Migration
+class RenameEmailSettings < ActiveRecord::Migration
   def up
     change_table :email_settings do |t|
       t.rename :registration_email_template, :registration_body
--- a/db/migrate/201505312045_rename_email_settings.rb
+++ b/db/migrate/201505312045_rename_email_settings.rb
@@ -1,10 +1,10 @@
-class FixColumnNames < ActiveRecord::Migration
+class rename_email_settings < ActiveRecord::Migration
   def up
     change_table :email_settings do |t|
       t.rename :registration_email_template, :registration_body
       t.rename :accepted_email_template, :accepted_body
-	t.rename :rejected_email_template, :rejected_body
-	t.rename :confirmed_email_template, :confirmed_body
+      t.rename :rejected_email_template, :rejected_body
+      t.rename :confirmed_email_template, :confirmed_body
     end
   end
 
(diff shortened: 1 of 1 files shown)


differentreality:
I think the build failure is due to your filename (rename_email_settings) being different from the class name inside the migration file (class FixColumnNames < ActiveRecord::Migration)

differentreality:
I think we are now on the right track :) I hope it is no longer as frustrating as it seemed to be in the beginning!

Next step would be to migrate the existing data.

We already have a few conferences in our DB, the conferences already have data in the dropped columns (:registration_email_template, :accepted_email_template, :rejected_email_template, :confirmed_email_template), that data needs to be copied in the new columns now, before the old columns are dropped.

We have done the same thing in several occasions, such as: 
https://github.com/openSUSE/osem/blob/master/db/migrate/20141128073306_migrate_data_remove_column_include_cfp_in_splash_add_column_include_cfp.rb

* We use temp class names
https://github.com/openSUSE/osem/blob/master/db/migrate/20141128073306_migrate_data_remove_column_include_cfp_in_splash_add_column_include_cfp.rb#L2

* To which we assign the actual DB table:
https://github.com/openSUSE/osem/blob/master/db/migrate/20141128073306_migrate_data_remove_column_include_cfp_in_splash_add_column_include_cfp.rb#L3

* We add the new column
* We copy the data from old column to new column
* We remove the old column
* This needs to be done for all the conferences
https://github.com/openSUSE/osem/blob/master/db/migrate/20141128073306_migrate_data_remove_column_include_cfp_in_splash_add_column_include_cfp.rb#L18

Again feel free to ping us on IRC for more discussion about this!

skriesch:
I have got a question for the file names. Which numbers should I use for migration files? The first ones are for the date and what are the others?

skriesch:
I have not dropped the columns. I have renamed them...
All data will exist all the time with another column name.

differentreality:
You should automatically generate the migration files from the command line:
rails g migration RenameAttributesInEmailSettings

Then keep the filename and class name given.

differentreality:
@skriesch yes, right, my bad :) We just need to change the column reference in our code then.
By the way I think we have a couple of more columns that need changing?

ChrisBr:
@skriesch Regarding the failing build

1. You should run the migration locally and then include the changes in db/schema.rb also in your commit
2. You should also change the factories accordingly (I guess this causes the failing build)
(e.g. https://github.com/openSUSE/osem/blob/master/spec/factories/email_settings.rb)

Hope that helps :)

skriesch:
Another question about the db migration: I have written a script for renaming the columns. Will it be executed, so that the columns will be changed with it after a new installation? Or do I have to edit the other db scripts with email_template, too?

ChrisBr:
@skriesch What script do you mean? Please elaborate!

skriesch:
@ChrisBr : I meant this one: https://github.com/skriesch/osem/commit/a31d0bf9f5df8fb4edd224ce53b5280a7554c740

differentreality:
@skriesch no you don't have to edit other migration files. Once your change is merged, we run the command *rake db:migrate*, your migration file will run and your changes will be in effect immediately in any existing instance of OSEM (not just new installation). I hope this is what you were asking?

differentreality:
Closing due to inactivity

Issue #663, opened 2015-06-02
Error when trying to export as XLS files registration


Doctor-Who:
On menu Registration, I can export to PDF but not as XLS file.

![capture du 2015-06-02 15 32 59](https://cloud.githubusercontent.com/assets/5890416/7936888/d1ec2b80-093c-11e5-8cfe-abc5f55b00f9.png)

PR #671, opened 2015-06-11
Remove redundant code.

differentreality:
Fixes https://github.com/openSUSE/osem/issues/663

(The error is the wrong count of columns after we removed the columns we no longer have for *registrations* table. If we set the width of the columns, we should probably provision for all the columns.)


hennevogel:
:+1:

Issue #613, opened 2015-03-05
RecordNotFound in Admin::QuestionsController#edit


sonalkr132:
I get this error when I try to edit unchecked predefined questions on questions page and delete just gives "404 not found" status. 
Those questions don't seem the have "conference_id"
![screenshot from 2015-03-05 18 38 43](https://cloud.githubusercontent.com/assets/7680662/6505436/44a01ff8-c368-11e4-837d-4988ec4c56c6.png)

Those are global questions so they are not editable anyway. Do their edit buttons are suppose to be disabled? I am not sure what (can? :update, q) is doing as they return true for all the questions.
```Ruby
= link_to 'Edit', edit_admin_conference_question_path(@conference.short_title, q),
                    class: 'btn btn-primary', disabled: !(can? :update, q)
```

Can someone please help?

differentreality:
can? :update, q   is checking if the current user can run the action update for that particular question.

We are using cancan gem for checking that:
https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities

Now as to why this is happenig, it has to do with the fact that (I assume) the logged in user is the first user created, which makes this user an admin.
https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L121

I can explain more if you need me to.

sonalkr132:
How should I go about fixing it? Since those questions are global even first admin user is not suppose to be able to edit them.
Should it be like that? One can't even delete them once they were marked global while creating them.

differentreality:
I believe an admin should be able to edit questions.
Some global questions are inserted by seeds file during installation, an admin might want to modify them to better depict the specific conference.  Moreover, by editing a global question, we can remove the global attribute in case it was wrongfully enabled.

@sonalkr132 do you plan on working on the changes? And if so, do you understand why this error exists?

sonalkr132:
@differentreality  Sorry, I didn't notice you replied. 
I guess we should remove the logic of checking whether the questions are global from controller to cancan.

Error exists because of this line of edit view:
```Ruby
= semantic_form_for(@question, url: admin_conference_question_path(@conference.short_title, @question.id)) do |f|
```
Those questions don't have `@conference` associated with them. However, I am not sure how to solve it.

PR #648, opened 2015-04-23
Only get registrations with answered question

differentreality:

Issue #551, opened 2014-11-26
Refactor Questions


hennevogel:
The current data structure and UI of questions is below par and very hard to understand. In light of #549 which @ChrisBr will probably do in the near future we should refactor this.

differentreality:
@hennevogel can you please elaborate on your logic regarding the association of Answer and Registration? (Assuming we are talking about a user's answer to a registration question)

PR #676, opened 2015-06-13
Questions: changes in layout and workflow, bug fixing

differentreality:
https://github.com/openSUSE/osem/issues/551
Includes https://github.com/openSUSE/osem/pull/648
Fixes https://github.com/openSUSE/osem/issues/613

Small change in layout of _form.
Small change in workflow: Question can be added without answers. But a question can be associated with a conference only after answers are added.


app/controllers/admin/questions_controller.rb
```diff
@@ -1,17 +1,16 @@
...
     load_and_authorize_resource :conference, find_by: :short_title
-    load_and_authorize_resource through: :conference, except: [:new, :create]
+    load_and_authorize_resource except: [:create]
 
     def index
       authorize! :index, Question.new(conference_id: @conference.id)
-      @questions = Question.where(global: true).all | Question.where(conference_id: @conference.id)
-      @questions_conference = @conference.questions
-      @new_question = @conference.questions.new
+      @questions = Question.where(global: true).all | Question.where(conference_id: @conference.id) | @conference.questions
```
ChrisBr: What's the difference between 
``Question.where(conference_id: @conference.id) ``
and 
``@conference.questions``

app/controllers/admin/questions_controller.rb
```diff
@@ -22,45 +21,76 @@ def new
...
+    def edit; end
 
     # PUT questions/1
     def update
-      if @question.update_attributes(params[:question])
-        redirect_to(admin_conference_questions_path(conference_id: @conference.short_title), notice: "Question '#{@question.title}' for #{@conference.short_title} successfully updated.")
+      @question.assign_attributes(params[:question])
+
+      if @question.question_type == QuestionType.find_by(title: 'Yes/No')
+        @question.answers = [ Answer.find_or_create_by(title: 'Yes'), Answer.find_or_create_by(title: 'No') ]
```
ChrisBr: DRY, see line 31! 
Can't we do this in a callback?

app/controllers/admin/questions_controller.rb
```diff
@@ -22,45 +21,76 @@ def new
...
+      end
+
+      if @question.save
+        if @question.answers.blank?
```
ChrisBr: Huh? You save it and then delete it? And without telling the user what happened?

ChrisBr:
You also need to rebase! Otherwise good work!

ChrisBr:
What's the actual state of this one?

differentreality:
Henne had suggested a full refactoring (https://github.com/openSUSE/osem/issues/551) which I don't know how to implement after a certain point (hence my last comment in the issue). So I am not sure which way we want to proceed. We also very briefly discussed the option of using a gem, but we haven't quite decided on one, or had any further communication about the whole issue.

hennevogel:
The whole Question/Answer thing is a mess I would like to get rid of by using some gem to build questionnaires. I'm not sure of the implications for this PR...

Rolence:
You can't just get rid of it and use gems to generate questions. This will not be in flow with the grand design  of the whole system

ChrisBr:
@differentreality @hennevogel what's the consent of this PR and the Question/Answer thing?

ChrisBr:
If there is no consent yet, can we just close the PR for now?

hennevogel:
We can yes

app/controllers/admin/questions_controller.rb
differentreality: https://github.com/openSUSE/osem/blob/master/app/models/question.rb#L3
From the Question model:
`has_and_belongs_to_many :conferences`
This relationship is used for enabled questions (which will show up in registration)

The conference_id attribute is a manual one, used to know which questions were created for that conference, even if they are not enabled.

Issue #675, opened 2015-06-13
proposal subtitle and difficulty only shown during /edit


meonkeys:
Perhaps this was intentional (if so, what's the intent?), but proposal subtitle and difficulty are only shown during `/edit`.

Here's the `/edit` page:

![subtitle and difficulty only when editing](https://cloud.githubusercontent.com/assets/50639/8144836/2161055e-11a4-11e5-933a-a7a571a55b07.png)

And here's the `/new` page:

![screencapture-osem-seagl-org-conference-seagl2015-proposal-new-1434208695590](https://cloud.githubusercontent.com/assets/50639/8144854/edfd8bdc-11a4-11e5-9025-dfb782b87beb.png)

differentreality:
Yep that is intentional indeed, so that when you are prompt to the proposal form for the very first time we don't overwhelm you with a lot of information, just the required one. That's in the concept of keeping it simple and fast for speakers to submit their proposal.

Then, if you want to add more stuff, you edit the proposal and you have a lot more options (which could be filled out after the proposal is accepted).

meonkeys:
Ah ok, makes sense. Thanks!

PR #677, opened 2015-06-13
Allow a proposal to be created with a Subtitle and Difficulty Level.

kormoc:
When a proposal is being submitted, we want to allow the submitter to
fill out the form once with all the required data. This change unifies
the forms for both editing and creating proposals. This allows us to
ensure that features are added in parallel to both workflows.

We also reduce the number of duplicate lines by removing duplicates
from the new.html.haml file.

We also increase the column size in css to have both pages show the
same width forms.

### DIFF
--- a/app/views/proposal/_proposal_form.html.haml
+++ b/app/views/proposal/_proposal_form.html.haml
@@ -1,4 +1,8 @@
 = semantic_form_for(@event, url: @url) do |f|
+
+  - if current_user.nil?
+    = render partial: 'devise/shared/sign_up_form_embedded'
+
   = f.inputs name: 'Proposal Information' do
     = f.input :title, as: :string, required: true
 
@@ -39,8 +43,10 @@
 
     = f.input :require_registration, label: 'Require participants to register to your event'
 
-    - if current_user.has_any_role? :admin, { name: :organizer, resource: @conference }, { name: :cfp, resource: @conference }
-      = f.input :is_highlight
+    -# We have to verify current_user in this context as this partial form is used both for new proposals and editing existing proposals
+    - if current_user
+      - if current_user.has_any_role? :admin, { name: :organizer, resource: @conference }, { name: :cfp, resource: @conference }
+        = f.input :is_highlight
 
     %p.text-right
       = link_to '#description',  "data-toggle"=>"collapse" do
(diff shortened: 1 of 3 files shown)


meonkeys:
Conversation on this ticket from #seagl on Freenode:

13:48 <@meonkeys> kormoc: re https://github.com/openSUSE/osem/pull/677 - nice! you saw https://github.com/openSUSE/osem/issues/675#issuecomment-111719467 , yes?
13:48 <@kormoc> Ooh, I missed that
13:49 <@kormoc> IMHO, if it's required information, we need it shown on the main form. Otherwise a fair number of proposals just won't have that data
13:50 <@kormoc> IT also seems silly to me to require people to create, save, and then edit right away
13:52 <@kormoc> I wonder if we built theme support if that would fix it. Then conferences could write CSS rules to hide/show fields for new/edit pages
13:53 <@meonkeys> Well, what @differentreality is saying about the intent is that subtitle and difficulty aren't required until the talk is accepted
13:54 <@kormoc> Hrm. We have required it in the past, as we want to balance out the talks with levels as well as topics
13:54 <@kormoc> it would kinda suck to accept a bunch of talks and then they all end up expert
13:55 < differentreality> on a sidenote: once a proposal is submitted, there is a TODO list in the index page, where it states what else you can/should add
13:55 <@meonkeys> aye, I did appreciate the TODO list
13:56 <@meonkeys> kormoc: good point about the issue of not knowing difficulty up front
13:56 <@kormoc> Yeah, I do like the todo list
13:57 <@meonkeys> I like showing the additional two fields up front because IMHO the major barrier to submitting a talk is having to sit down, pull up a web page, and fill out a form. A couple extra fields won't turn me off, but having to come back again another day and do more work might make me give up.
13:59 <@kormoc> @differentreality, What do you feel about using custom css/themes to allow customizing the forms?
13:59 <@kormoc> it would be easy enough to do display:none on elements on default when creating, but then we could override them
13:59 < differentreality> this is btw a very recent changes we did to the layout of the proposal, till then submitter would only have to come back to edit the proposal to add a commercial (that's slides from slideshare, video from youtube etc)

ChrisBr:
@kormoc Can you paste screenshots please?

I will checkout this tomorrow!

meonkeys:
@ChrisBr Do you mean screenshots of what the UI looks like _with_ this patch?

See #675 for screenshots of the look & feel on `master` (before this patch is applied).

hennevogel:
The only way to 'fix' this is that we make configurable what is required for an Event.

ChrisBr:
@meonkeys Yeah, what it looks like with this patch applied (already know how it looks on master)!

Hm, the intention was that we dropped everything from the new page which is not 'mandatory' to make the UI clear! Not sure if we want to make a step 'back' ...

eximious:
Maybe a way to split the difference would to leave a default "no level specified" and let folks fill it out if they want to when they submit their proposal, but not require it?

hennevogel:
That still would mean you have to parse and understand all of these options. This will turn a lot of people away, while accepting the talk with minimum options and requesting additional information afterward will not. 

Don't believe this? Check the exceptional works of [@lukew](http://www.lukew.com/ff/entry.asp?1868) :-)

Issue #590, opened 2015-02-02
Abstract word counting does not seem to work


scarabeusiv:
When entering abstract for talk on openSUSE conference 2015 the abstract words are not counted. Regardless of the content I see there "You have used 0 words. Abstracts must be between 0 and 250 words. "

The browser is chromium 40 and only extension is adblock, so no disabled scripting.

v0dro:
I'll do this. Could you please assign me to it?

hennevogel:
No need to. Just do it :neckbeard:

v0dro:
I've zeroed the bug to [this line](https://github.com/openSUSE/osem/blob/master/app/assets/javascripts/osem.js#L169).

The funny thing is, when I copy paste the entire `$("#event_abstract").bind` function into the browser, the abstract count is updated properly! (exact same code)

Any idea why this should happen?

v0dro:
It works now.

Please evaluate.

differentreality:
Fixed in #678

PR #678, opened 2015-06-14
Fix the abstrct count on the proposal create and edit pages

kormoc:
Due to the DOM not being ready when osem.js is executed, the
bind statements do not actually bind on the correct elements.
We just wrap the binds in a document ready event which allows
the elements to fully load and the events to fire correctly.

I also bind on more then just keyup, to catch times when people
paste content into the field. The input event is only supported
by newer browsers, but is entirely a noop for older browsers.

### DIFF
--- a/app/assets/javascripts/osem.js
+++ b/app/assets/javascripts/osem.js
@@ -157,24 +157,27 @@ function word_count(text, divId, maxcount) {
     }
 };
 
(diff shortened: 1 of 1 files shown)


differentreality:
Fixes https://github.com/openSUSE/osem/issues/590 
:+1:

meonkeys:
:rocket: :rose: :guitar:

PR #679, opened 2015-06-14
Rework bootstrapSwitch

differentreality:
I am moving the js code for switch outside of the $(document).ready function, which results in us not being able to click on the switch button in events#index, but that's not necessarily a problem, since index page is not supposed to change stuff anyway.
event#edit works fine.

I have also re-written the js code for switch so that it can be used in any place, for any model.
When the switch-checkbox is introduced in the view, one needs to pass a method and a url (where the switch should redirect to). If we have all the data given into that URL, the switch can work regrdless of the model or attribute that is being changed.

### DIFF
--- a/app/views/admin/events/_proposal.html.haml
+++ b/app/views/admin/events/_proposal.html.haml
@@ -5,7 +5,7 @@
       %br
       %small
         = @event.subtitle
-      = link_to "Edit", edit_admin_conference_event_path(@conference.short_title, @event), :class => "btn btn-mini btn-default pull-right"
+      = link_to 'Edit', edit_admin_conference_event_path(@conference.short_title, @event), class: 'btn btn-mini btn-primary pull-right'
 
 .row
   .col-md-12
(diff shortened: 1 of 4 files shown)


app/views/admin/events/_proposal.html.haml
```diff
@@ -83,6 +84,17 @@
                                             @event,
                                             event: { difficulty_level_id: difficulty.id }),
                 method: :patch
+      %tr
+        %td
+          %b Requires Registration
```
ChrisBr: Why wasn't that included before?

ChrisBr:
@differentreality basically this LGTM but I haven't checked out the branch and tested it (but I trust you have ;)). Would you mind to wrap up your PR comment in the git commit message?

app/views/admin/events/_proposal.html.haml
differentreality: No particular reason. I think it should be there and I should have added it long before.

differentreality:
@ChrisBr yeah, but 2 sets of eyes are always better. Which is why I think positive reviews are better to be based on checking out the branch and actually seeing/using stuff. 
I will merge after you (or someone else for that matter) do so, alright?

differentreality:
I didn't think it was very much relevant to the commit itself, to be honest. Of course I can if you think it is useful there.

(I assume you refer to the second paragraph, which I have explained a bit more in the wiki: https://github.com/openSUSE/osem/wiki/Bootstrap-Switch-Checkbox)

hennevogel:
:+1: go :fireworks:

Issue #684, opened 2015-06-16
Postgresql error


milovanderlinden:
When going to an event admin page [http://localhost:3000/admin/conference/testconference](http://localhost:3000/admin/conference/testconferenc), a pg error is thrown:

    PG::GroupingError: ERROR:  column "events.id" must appear in the GROUP BY clause or be used in an aggregate function
    LINE 1: SELECT "events".* FROM "events" WHERE "events"."conference_i...

To me this looks like a valid error as you cannot select * when using a group by in postgres.

milovanderlinden:
Application trace:

app/models/conference.rb:837:in `calculate_program_minutes'
app/models/conference.rb:395:in `current_program_minutes'
app/models/conference.rb:404:in `current_program_hours'
app/controllers/admin/conference_controller.rb:107:in `show'

milovanderlinden:
Some light on the issue [http://stackoverflow.com/questions/16418504/pgerror-in-group-by-clause](http://stackoverflow.com/questions/16418504/pgerror-in-group-by-clause)

hennevogel:
We don't really support PostGre, I'm sure there are lots of other places where the APP goes wrong. Sorry :disappointed:

milovanderlinden:
No worries. I will make postgresql supported in my fork as I need it for a heroku deploy of the app. Please let me know if you would like to add support for postgresql to your branch somewhere in the future.

ChrisBr:
@milovanderlinden Great! Let us know how OSEM works for heroku! We planned a deploy to heroku feature some time ago but never finished it ... :(

hennevogel:
@milovanderlinden awesome! :bouquet:

milovanderlinden:
It works pretty well. I also discovered an issue with campaigns. Will fix that when I find some time later. Progress can be seen on my dev environment at http://young-woodland-2376.herokuapp.com/

PR #685, opened 2015-06-16
Fix for #684

milovanderlinden:
By defining the fields that are in the group by with a select, the issue where the pg driver defaults to * when no fields are defined gets fixed.

Please accept this pull-request.

Kind regards,

Milo

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -356,7 +356,7 @@ class Conference < ActiveRecord::Base
   # ====Returns
   # * +hash+ -> hash
   def self.event_distribution
-    calculate_event_distribution_hash(Event.group(:state).count)
+    calculate_event_distribution_hash(Event.select(:state).group(:state).count)
   end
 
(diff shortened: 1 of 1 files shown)


ChrisBr:
@milovanderlinden Thanks for your contribution! Have you tried this fix also with MySQL/MariaDB and SQLite?

ChrisBr:
See #684 
We only support MySQL/Maria DB ...

hennevogel:
As @milovanderlinden want's to add PG complete support, let's get this stuff merged :-)

ChrisBr:
@hennevogel Do we want to merge this into master or first in a separate branch for heroku support?

hennevogel:
This is not about heroku support. This is about making SQL calls DB agnostic :-) So this should go into master. We should also think about how we can test this...

hennevogel:
Anyway. Let's get this in

PR #693, opened 2015-07-02
Attendance btn

raluka:


app/views/admin/registrations/index.html.haml
```diff
@@ -44,24 +44,14 @@
               %td
                 = link_to 'Questions','#', class: 'btn btn-success question-btn', 'data-id' => index, 'data-name' => registration.name
             %td
+              = check_box_tag @conference.short_title, registration.id, true, class: 'switch-checkbox', method: :patch,
+                data: { size: "large",
```
ChrisBr: Try 
```
size: "small"
```
And as the button doesn't belong to the button group anymore I would consider an additional table column.

ChrisBr:
Okay @raluka, good work so far :clap: 

What's now left is to give the button some functionality! As yesterday discussed I would refactor the absent/present actions in the registrations controller to one function ```toogle_attendance```.
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/registrations_controller.rb#L39

raluka:
Thanks Chris, I will modify size for button, and redactor those 2 methods.

ChrisBr:
Additional you should consider to squash your two git commits together. 
http://davidwalsh.name/squash-commits-git

But please be very careful, you can easily destroy your work :) If you're not certainly sure what you're doing you can ask us in IRC :)

app/views/admin/registrations/index.html.haml
differentreality: The visually wrong alignment is caused by the line-height of the div. We might want to look into explicitely defining it in our css? What do you think @ChrisBr ?

+1 on the additional column OR just include it to the .btn-group, I think logically it makes sense to have it there.

differentreality:
https://github.com/openSUSE/osem/pull/697

PR #697, opened 2015-07-15
Issue#690: fixed Attendance Button

raluka:


differentreality:
Do you think it would make sense to keep the notice/alert messages in the controller?

app/controllers/admin/registrations_controller.rb
```diff
@@ -11,7 +11,8 @@ def index
       @attended = @conference.registrations.where('attended = ?', true).count
     end
 
-    def edit; end
+    def edit;
```
ChrisBr: Please keep the old syntax 

```
```

ChrisBr:
I really like the refactoring of the absent & present actions :+1:

ChrisBr:
@differentreality Hm, yes a visual notification would definitely makes sense ... But as we now use AJAX I'm not sure if we can keep it in the controller ...

ChrisBr:
Otherwise LGTM

differentreality:
@ChrisBr  what about sth like https://github.com/leonid-shevtsov/unobtrusive_flash ?

hennevogel:
The button changes. What other notification do you need? :-)

differentreality:
@hennevogel I was thinking flash message, so that we know which user/registration we affected, since this is in the index page and in case of wrong user it could be proven very helpful to see it on the message.

ChrisBr:
@hennevogel yes, sure! But if sth goes wrong in the backend the button will change too ...
But I agree, we shouldn't Overengineer this ... we can add the notification in a later PR!

app/views/admin/registrations/index.html.haml
```diff
@@ -44,24 +44,15 @@
               %td
                 = link_to 'Questions','#', class: 'btn btn-success question-btn', 'data-id' => index, 'data-name' => registration.name
             %td
+              = check_box_tag "#{@conference.short_title}_#{registration.id}", registration.id, registration.attended,
+                class: 'switch-checkbox', method: :patch,
+                url: toggle_attendance_admin_conference_registration_path(@conference.short_title, id: registration.id)+"?attended=",
+                data: { size: "large",
```
ChrisBr: You should use size: 'small' like in #693 discussed ... Large still looks a little bit ugly :(

ChrisBr:
Okay, when you finally fix the size and squash your commits together we're ready to :ship: 

Good work :clap: :clap: :clap:

app/views/admin/registrations/index.html.haml
raluka: ok, will do it

differentreality:
@ChrisBr even with 'small' the button doesnt look very good... 

screenshot

raluka:
Large button vs small button. Any option need a bit of CSS to make it acceptable.
large_button
small_button

hennevogel:
Changing the appearance of the bootstrap-switch buttons is not the scope of this pull request.

PR #699, opened 2015-07-23
comment notification

raluka:
Sending comment notification to specific users. Please give us your opinion.

### DIFF
--- a/app/assets/stylesheets/osem.css.scss
+++ b/app/assets/stylesheets/osem.css.scss
@@ -1,13 +1,13 @@
 html {
-    position: relative;
-    min-height: 100%;
+  position: relative;
+  min-height: 100%;
 }
 
 body {
-    /* Margin bottom by 2 times the footer height */
-    margin-bottom: 60px;
-    /* Margin bottom by navbar height */
-    padding-top: 60px;
+  /* Margin bottom by 2 times the footer height */
+  margin-bottom: 60px;
+  /* Margin bottom by navbar height */
+  padding-top: 60px;
 }
 
 #content {
@@ -15,23 +15,23 @@ body {
 }
 
 #footer {
-    position: absolute;
-    bottom: 0;
-    width: 100%;
-    /* Set the fixed height of the footer here */
-    height: 60px;
-    background-color: #f5f5f5;
-    .container {
-      padding: 15px;
-    }
+  position: absolute;
+  bottom: 0;
+  width: 100%;
+  /* Set the fixed height of the footer here */
+  height: 60px;
+  background-color: #f5f5f5;
+  .container {
+    padding: 15px;
+  }
 }
 
 .nav-tabs {
-    margin-bottom: 15px;
+  margin-bottom: 15px;
 }
 
 .img-center {
-    margin: 0 auto;
+  margin: 0 auto;
 }
 
 /* centered columns styles */
@@ -71,4 +71,14 @@ p.comment-body {
 
 #account-already {
   font-size: 0.6em;
+}
+
+/* comments views pane: use padding-bottom before next comment box */
+.panel.panel-default .panel-body .box {
+  padding-bottom: 20px;
+}
+
+/* comments views pane: use padding-bottom after each comment */
+.panel.panel-default .panel-body .box p {
+  padding-bottom: 20px;
 }
--- a/app/assets/stylesheets/osem.css.scss
+++ b/app/assets/stylesheets/osem.css.scss
@@ -74,11 +74,11 @@ p.comment-body {
 }
 
 /* comments views pane: use padding-bottom before next comment box */
-.panel.panel-default .panel-body .box {
+.panel.panel-default .panel-body .notifications {
   padding-bottom: 20px;
 }
 
 /* comments views pane: use padding-bottom after each comment */
-.panel.panel-default .panel-body .box p {
+.panel.panel-default .panel-body .notifications p {
   padding-bottom: 20px;
 }
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -1,0 +1,9 @@
+module Admin
+  class CommentsController < Admin::BaseController
+    load_and_authorize_resource
+
+    def index
+      @ordered_events = Event.order(:title).all
+    end
+  end
+end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -3,7 +3,8 @@ module Admin
     load_and_authorize_resource
 
     def index
-      @conferences_available = Conference.with_roles([:admin, :organizer, :cfp], current_user)
+      @comments = Comment.accessible_by(current_ability).order(created_at: :desc).group_by { |comment| comment.commentable.conference }
+      @another_comments = Hash[@comments.map{|conference, comments| [conference, comments.group_by {|comment| comment.commentable}]}]
     end
   end
 end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -3,17 +3,24 @@ module Admin
     load_and_authorize_resource
 
     def index
+      # All available comments, grouped and sorted
       @comments = grouped_comments(accessible_ordered_comments)
+
+      # Grouped, sorted, available comments, posted since current_user last login
       @unread_comments = grouped_comments(accessible_ordered_comments.find_since_last_login(current_user))
+
+      # Grouped, sorted, available comments, posted by current_user
       @posted_comments = grouped_comments(accessible_ordered_comments.find_comments_by_user(current_user))
     end
 
     private
 
+    # Returning all available comments, ordered by created_at: :desc and by event.title
     def accessible_ordered_comments
       Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
     end
 
+# Grouping all comments by conference, and by event. It returns {:conference => {:event => [{comment_2}, {comment_1 }]}}
     def grouped_comments(remarks)
       remarks.group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
     end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -12,8 +12,8 @@ module Admin
       Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
     end
 
-    def grouped_comments(comments)
-      comments.group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
+    def grouped_comments(remarks)
+      remarks.group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
     end
   end
 end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -1,6 +1,8 @@
 module Admin
   class CommentsController < Admin::BaseController
+    load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource
+
 
     def index
       @ordered_events = Event.order(:title).all
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -8,6 +8,8 @@ module Admin
       @posted_comments = grouped_comments(accessible_ordered_comments.find_comments_by_user(current_user))
     end
 
+    private
+
     def accessible_ordered_comments
       Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
     end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -1,11 +1,9 @@
 module Admin
   class CommentsController < Admin::BaseController
-    load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource
 
-
     def index
-      @ordered_events = Event.order(:title).all
+      @conferences_available = Conference.with_roles([:admin, :organizer, :cfp], current_user)
     end
   end
 end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -3,7 +3,7 @@ module Admin
     load_and_authorize_resource
 
     def index
-      @comments = Comment.accessible_by(current_ability).order(created_at: :desc).group_by { |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by {|comment| comment.commentable}]}.to_h
+      @comments = Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC').group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
     end
   end
 end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -3,7 +3,17 @@ module Admin
     load_and_authorize_resource
 
     def index
-      @comments = Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC').group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
+      @comments = grouped_comments(accessible_ordered_comments)
+      @unread_comments = grouped_comments(accessible_ordered_comments.find_since_last_login(current_user))
+      @posted_comments = grouped_comments(accessible_ordered_comments.find_comments_by_user(current_user))
+    end
+
+    def accessible_ordered_comments
+      Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
+    end
+
+    def grouped_comments(comments)
+      comments.group_by{ |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by{|comment| comment.commentable}]}.to_h
     end
   end
 end
--- a/app/controllers/admin/comments_controller.rb
+++ b/app/controllers/admin/comments_controller.rb
@@ -3,8 +3,7 @@ module Admin
     load_and_authorize_resource
 
     def index
-      @comments = Comment.accessible_by(current_ability).order(created_at: :desc).group_by { |comment| comment.commentable.conference }
-      @another_comments = Hash[@comments.map{|conference, comments| [conference, comments.group_by {|comment| comment.commentable}]}]
+      @comments = Comment.accessible_by(current_ability).order(created_at: :desc).group_by { |comment| comment.commentable.conference }.map {|conference, comments| [conference, comments.group_by {|comment| comment.commentable}]}.to_h
     end
   end
 end
--- a/app/controllers/admin/conference_controller.rb
+++ b/app/controllers/admin/conference_controller.rb
@@ -27,7 +27,6 @@ module Admin
       @recent_users = User.limit(5).order(created_at: :desc)
       @recent_events = Event.limit(5).order(created_at: :desc)
       @recent_registrations = Registration.limit(5).order(created_at: :desc)
-
       @top_submitter = Conference.get_top_submitter
 
       @submissions = {}
--- a/app/controllers/admin/conference_controller.rb
+++ b/app/controllers/admin/conference_controller.rb
@@ -27,6 +27,7 @@ module Admin
       @recent_users = User.limit(5).order(created_at: :desc)
       @recent_events = Event.limit(5).order(created_at: :desc)
       @recent_registrations = Registration.limit(5).order(created_at: :desc)
+
       @top_submitter = Conference.get_top_submitter
 
       @submissions = {}
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -263,10 +263,6 @@ module ApplicationHelper
     end
   end
 
-  def can_manage_comments(conference)
-    (current_user.has_role? :organizer, conference) || (current_user.has_role? :cfp, conference)
-  end
-
   def sign_in_path
     if CONFIG['authentication']['ichain']['enabled']
       new_user_ichain_session_path
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -278,4 +278,8 @@ module ApplicationHelper
       new_user_registration_path
     end
   end
+
+  def unread_notifications(user)
+    @unread_notifications = Comment.find_since_last_login(user)
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -280,6 +280,6 @@ module ApplicationHelper
   end
 
   def unread_notifications(user)
-    Comment.accessible_by(current_ability).find_since_last_login(user).limit(5)
+    Comment.accessible_by(current_ability).find_since_last_login(user)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -263,6 +263,10 @@ module ApplicationHelper
     end
   end
 
+  def can_manage_comments(conference)
+    (current_user.has_role? :organizer, conference) || (current_user.has_role? :cfp, conference)
+  end
+
   def sign_in_path
     if CONFIG['authentication']['ichain']['enabled']
       new_user_ichain_session_path
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -280,6 +280,6 @@ module ApplicationHelper
   end
 
   def unread_notifications(user)
-    Comment.accessible_by(current_ability).find_since_last_login(current_user).limit(5)
+    Comment.accessible_by(current_ability).find_since_last_login(user).limit(5)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -280,6 +280,7 @@ module ApplicationHelper
   end
 
   def unread_notifications(user)
-    Comment.find_since_last_login(user)
+    available_conferences_ids = Conference.with_roles([:admin, :organizer, :cfp], user).pluck(:id)
+    Comment.find_since_last_login(user).where(commentable_type: 'Event', commentable_id: Event.where(conference_id: available_conferences_ids))
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,14 @@
 module ApplicationHelper
+  # Set resource_name for devise so that we can call the devise help links (views/devise/shared/_links) from anywhere (eg sign_up form in proposal#new)
+  def resource_name
+    :user
+  end
+
+  # Set devise_mapping for devise so that we can call the devise help links (views/devise/shared/_links) from anywhere (eg sign_up form in proposal#new)
+  def devise_mapping
+    @devise_mapping ||= Devise.mappings[:user]
+  end
+
   def pluralize_without_count(count, noun, text = nil)
     if count != 0
       count == 1 ? "#{noun}#{text}" : "#{noun.pluralize}#{text}"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -280,6 +280,6 @@ module ApplicationHelper
   end
 
   def unread_notifications(user)
-    @unread_notifications = Comment.find_since_last_login(user)
+    Comment.find_since_last_login(user)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,13 +19,11 @@ module ApplicationHelper
   end
 
   def event_progress_color(progress)
-    progress = progress.to_i
-
-    case progress
+    case progress.to_i
     when 100 then 'progress-bar-success'
-    when 86 then 'progress-bar-info'
+    when 85 then 'progress-bar-info'
     when 71 then 'progress-bar-warning'
-    when 14, 29, 43, 57 then 'progress-bar-danger'
+    when 14, 28, 42, 57 then 'progress-bar-danger'
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -280,7 +280,6 @@ module ApplicationHelper
   end
 
   def unread_notifications(user)
-    available_conferences_ids = Conference.with_roles([:admin, :organizer, :cfp], user).pluck(:id)
-    Comment.find_since_last_login(user).where(commentable_type: 'Event', commentable_id: Event.where(conference_id: available_conferences_ids))
+    Comment.accessible_by(current_ability).find_since_last_login(current_user).limit(5)
   end
 end
--- a/app/mailers/mailbot.rb
+++ b/app/mailers/mailbot.rb
@@ -80,6 +80,22 @@ class Mailbot < ActionMailer::Base
     end
   end
 
+  def send_notification_email_for_comment(comment)
+    @comment = comment
+    @event = @comment.commentable
+    @conference = @event.conference
+    recipients = User.comment_notifiable(@conference) # with scope
+    recipients.each do |user|
+      @user = user
+      mail(to: @user.email,
+           from: @conference.contact.email,
+           reply_to: @conference.contact.email,
+           template_path: 'admin/emails',
+           template_name: 'comment_template',
+           subject: "New comment has been posted for #{@event.title}")
+    end
+  end
+
   def build_email(conference, to, subject, body)
     mail(to: to,
          from: conference.contact.email,
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -137,6 +137,7 @@ class Ability
     can :manage, Sponsor, conference_id: conf_ids_for_organizer
     can :manage, SponsorshipLevel, conference_id: conf_ids_for_organizer
     can :manage, Ticket, conference_id: conf_ids_for_organizer
+    can :manage, Comment, conference_id: conf_ids_for_organizer
   end
 
   def signed_in_with_cfp_role(user)
@@ -155,6 +156,7 @@ class Ability
     can :manage, CallForPaper, conference_id: conf_ids_for_cfp
     can :manage, Commercial, commercialable_type: 'Event',
                              commercialable_id: Event.where(conference_id: conf_ids_for_cfp).pluck(:id)
+    can :manage, Comment, conference_id: conf_ids_for_cfp
   end
 
   def signed_in_with_info_desk_role(user)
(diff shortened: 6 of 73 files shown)


app/mailers/mailbot.rb
```diff
@@ -80,6 +80,20 @@ def send_on_call_for_papers_dates_updates(conference)
     end
   end
 
+  def notification_email(conference, event, comment)
+    recipients = User.joins(:roles).where('roles.name IN (?)', [:organizer, :cfp]).where('roles.resource_id = ?', conference.id).select(:name, :email).distinct
```
ChrisBr: I'd prefer to use a scope here. Not sure if there already exists one for organizer / cfp but we could for sure add one.

http://guides.rubyonrails.org/active_record_querying.html#scopes

differentreality: For educative purposes (as in, you haven't done anything wrong, I am just suggesting easier/faster/different options):
1) for the roles we use Rolify gem, which includes the following capability:
User.with_role :organizer, Conference.second

More details: https://github.com/RolifyCommunity/rolify/wiki/Usage

2) You can use where like this too:
Conference.where(title: 'test', short_title: 'test')
or
Conference.where(id: [3,4])

More details: http://guides.rubyonrails.org/active_record_querying.html

app/models/email_settings.rb
```diff
@@ -49,11 +50,17 @@ def get_values(conference, user, event = nil)
       h['registration_end_date'] = conference.registration_period.end_date
     end
 
-    if !event.nil?
+    if event
```
ChrisBr: Good catch :)

db/migrate/20150724182915_add_subject_and_comment_template_for_email_settings.rb
```diff
@@ -0,0 +1,23 @@
...
+User {comment_user} posted a new comment for event {eventtitle} of {conference} .
+
+" {comment_body} "
+
+To reply to this comment, please go to {comment_reply}
+
+Best wishes,
+{conference} Team
+)
+    EmailSettings.all.each do |email_setting|
```
ChrisBr: @raluka You only set the subject & template for existing email settings! That would mean when you create a new conference the related subject & template would be nil...

There're several ways to fix this:

1) Set default values in your migration
```
add_column :email_settings, :comment_subject, :string, :default => comment_subject
```

2) Set default values when we build the email settings for a new conference
https://github.com/openSUSE/osem/blob/master/app/models/conference.rb#L928

3) Add a feature to enter & change the the subject & template
https://github.com/openSUSE/osem/blob/master/app/views/admin/emails/index.html.haml

I'd go for the first or second implementation, I don't see much value in changing the comment email ...
But when we go for one of the first two solutions, the question is if we need to store the template & subject in the database at all?

@raluka Can you follow :)

raluka: @ChrisBr solved in last commit

db/schema.rb
```diff
@@ -187,6 +187,8 @@
     t.string   "call_for_papers_dates_updates_subject"
     t.text     "call_for_papers_schedule_public_template"
     t.text     "call_for_papers_dates_updates_template"
+    t.string   "comment_subject"
```
ChrisBr: @raluka have you executed the new migration locally? Because there're no changes in your schema file...

ChrisBr:
@raluka You should also add edit fields to the email view:
https://github.com/openSUSE/osem/blob/master/app/views/admin/emails/index.html.haml

Otherwise it's not necessary to store the templates in the database at all (which could also be an option and we should discuss that here...)

db/schema.rb
raluka: I didn't do it, damn. I'm fixing it now

db/migrate/20150724182915_add_subject_and_comment_template_for_email_settings.rb
```diff
@@ -0,0 +1,24 @@
+class AddSubjectAndCommentTemplateForEmailSettings < ActiveRecord::Migration
```
hennevogel: Why do you create these in the DB? And not in [app/views/admin/emails/index.html.haml](https://github.com/openSUSE/osem/blob/master/app/views/admin/emails/index.html.haml) like we do for the other templates?

ChrisBr:
@raluka We have two options:

1. Store the template & subject in the database and implement in the email settings index page to change the text. (that makes the template very customisable)
2. Store the database for instance in an email view (this way it's not customisable because you need to touch the source code when you want to change the template)
So the question we should now answer is: Is it really necessary to have customized templates for comment notification?

db/migrate/20150724182915_add_subject_and_comment_template_for_email_settings.rb
ChrisBr: You mean the defaults? Yep, that could also be an option but first we should answer the questions I raised in my other comment ...

ChrisBr:
@raluka Except the schema file it looks good so far :clap: :clap: :clap: 
(Just had a brief look so far, will review tomorrow morning in detail!)

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150415121038) do
+ActiveRecord::Schema.define(version: 20150729174340) do
```
ChrisBr: Probably a leftover from the previous changes ... 
We don't need this because we have no changes on the DB

raluka: That changed because of previous migration, yes.

ChrisBr:
@raluka Commit e69cbd6686b6fe00127e65cf116fdb85dcbbc04b introduced again an different db schema version ... Maybe you need to reset your database?

app/models/email_settings.rb
```diff
@@ -67,6 +74,14 @@ def generate_email_on_conf_updates(conference, user, conf_update_template)
     parse_template(conf_update_template, values)
   end
 
+  def generate_email_on_comment_create(conference, event, comment, user)
+    values = get_values(conference, user, event, comment)
+    comment_template = File.read("#{Rails.root}/app/views/admin/emails/comment_template.text.erb")
```
ChrisBr: Don't really like this approach of loading the view ...
@hennevogel what do you think? Any better idea?

ChrisBr:
Okay, looks good so far for me but we should wait for suggestions from @hennevogel & @differentreality.

Finally I'd recommend to squash all the commits in this PR together. Because you now have 18 commits for a small change. I'll try to explain:

```
git reset HEAD~18 --soft
```
this will reset your head 18 commits (your number of commits in this PR) back but will preserve your changes / work.

```
git commit -m "Implement comment notification feature"
```
this will commit all of your changes to a new commit.

```
git push origin view_comment_notification -f
```
this will push the new commit to your branch. Please note the -f at the end of the commit which stays for force and is necessary because you rewrite the history of the branch.

Can you follow?

Please be aware with git, you can easily destroy your whole work ;) Maybe you can ask a coach to help you with this task?

raluka:
We need to implement also "unread notification when login", and squash everything into one, @ChrisBr . Regarding `generate_email_on_comment_create` method, I am open to suggestions for re-factoring. My approach is a beginner one.

app/models/email_settings.rb
hennevogel: Why don't you use [mailer views](http://guides.rubyonrails.org/action_mailer_basics.html#mailer-views)?

app/views/admin/comments/index.html.haml
```diff
@@ -0,0 +1,25 @@
+%h1 Title here
```
hennevogel: Title here?

app/helpers/admin/comments_helper.rb
```diff
@@ -0,0 +1,7 @@
+module Admin
+  module CommentsHelper
+    def unread_comments
+      Comment.limit(5).order(created_at: :desc)
```
hennevogel: How does this relate to unread? It's the last five.

app/views/layouts/_navigation.html.haml
```diff
@@ -13,17 +13,31 @@
...
+                = render 'layouts/user_menu'
+          - if can? :manage, Conference
+            %ul.nav.navbar-nav.navbar-right
+              %li.dropdown
+                %a.dropdown-toggle{"data-toggle" => "dropdown", :href => "#", id: "unread-comments"}
+                  - if unread_comments
+                    Notifications (#{unread_comments.length})
+                    %b.caret
+                    %ul.dropdown-menu
+                      = render partial: 'layouts/unread_comment', collection: unread_comments
```
hennevogel: Why do you create a partial for one link_to line?

app/assets/javascripts/admin/comments.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
hennevogel: Please do not add empty scaffold files

app/assets/stylesheets/admin/comments.css.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Admin::Comments controller here.
```
hennevogel: Please do not add empty scaffold files

app/models/email_settings.rb
```diff
@@ -10,7 +10,8 @@ class EmailSettings < ActiveRecord::Base
                   :send_on_venue_update, :venue_update_subject, :venue_update_template,
                   :send_on_call_for_papers_dates_updates, :send_on_call_for_papers_schedule_public,
                   :call_for_papers_schedule_public_subject, :call_for_papers_dates_updates_subject,
-                  :call_for_papers_schedule_public_template, :call_for_papers_dates_updates_template
+                  :call_for_papers_schedule_public_template, :call_for_papers_dates_updates_template, :comment_subject,
+                  :comment_template
```
hennevogel: You're not using comment_subject and comment_template anymore are you?

app/views/admin/comments/index.html.haml
```diff
@@ -0,0 +1,25 @@
...
+      %li
+        %a{:href=>"#all_comments", "data-toggle"=>"tab"}
+          %span.fa.fa-check-square
+          Comments
+      %li
+        %a{:href=>"#posted_comments", "data-toggle"=>"tab"}
+          %span.fa.fa-file-text
+          Posted Comments
+    .tab-content
+      .tab-pane.active#unread_comments
```
hennevogel: No content?

test/controllers/admin/comments_controller_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
```
hennevogel: Please do not add empty scaffold files

test/helpers/admin/comments_helper_test.rb
```diff
@@ -0,0 +1,6 @@
+require 'test_helper'
```
hennevogel: Please do not add empty scaffold files

config/routes.rb
```diff
@@ -15,6 +15,7 @@
   namespace :admin do
     resources :users
     resources :people
+    resources :comments, only: [:index, :show]
```
hennevogel: Where is show?

app/views/admin/comments/index.html.haml
```diff
@@ -0,0 +1,25 @@
...
+      .tab-pane#all_comments
+      .tab-pane#posted_comments
+
+%container
+  %ul
```
hennevogel: What is a &lt;container&gt;?

app/models/email_settings.rb
raluka: forgot about that, it will be gone next commit

app/views/admin/comments/index.html.haml
raluka: WIP, I forgot to mention that

app/assets/javascripts/admin/comments.js
raluka: I'll clear them

app/views/layouts/_navigation.html.haml
```diff
@@ -13,17 +13,32 @@
...
+              %a.dropdown-toggle{"data-toggle" => "dropdown", :href => "#", id: "current-user-detail"}
+                - if not current_user.name.blank?
+                  #{current_user.name}
+                -else
+                  #{current_user.email}
+                = image_tag(current_user.gravatar_url(size: '18'), title: "Yo #{current_user.name}!", :alt => '')
+                %b.caret
+              %ul.dropdown-menu
+                = render 'layouts/user_menu'
+          - if can? :manage, Conference
```
ChrisBr: Is can? manage, Conference the right thing here?
Isn't there also a CfP role?

@hennevogel @differentreality

config/routes.rb
```diff
@@ -15,6 +15,7 @@
   namespace :admin do
     resources :users
     resources :people
+    resources :comments
```
ChrisBr:
```
resources :comments
```
creates seven different routes but you only need the index route!

Please change this to 

```
resources :comments, only: [:index]
```

See http://guides.rubyonrails.org/routing.html for more details!
@raluka

app/helpers/application_helper.rb
```diff
@@ -278,4 +278,8 @@ def sign_up_path
       new_user_registration_path
     end
   end
+
+  def unread_notifications(user)
+    @unread_notifications = Comment.find_since_last_login(user)
```
ChrisBr: Is it necessary to assign it to an instance variable instead of just return it?
And is this method necessary at all?

app/views/layouts/_navigation.html.haml
differentreality: Yes, there is a cfp role too.

Based on the changes in models/ability.rb we want users with the role **organizer** and the role **cfp** to be able to manage comments. Is that right @raluka ?

"if can? :manage, Conference" means *if the user can manage some conference (any conference)*. That makes sense to cover the users with the role **organizer** but not the ones with the **cfp** role.

Moreover roles' abilities may change eventually, and this statement depends on the roles that the organizer or a cfp team member has. But what we really care about here is if the user can manage comments, right @raluka ? (regardless of the user's role or ability to manage or not manage a conference)
So I would suggest to check about that:
- if can? :manage, Comment

I also have concerns about the **manage** ability.
Isn't this reallly about being able to **show** the comments? Just because you get the notification and you can see the comment, it doesn't necessarily mean you can delete the comment or edit the comment or even create a new comment. But this is something we can work on afterworks too, as it doesn't directly affect any operation in OSEM.

raluka: if can? :manage Comment is more appropriate here, perfectly true. I'll update code

config/routes.rb
raluka: thanks @ChrisBr , I will change it

differentreality:
There are 2 things wrong with the ability for comments (https://github.com/raluka/osem/blob/view_comment_notification/app/views/layouts/_navigation.html.haml#L28)

1) The Comment model does not have a conference_id attribute, you need to properly define the ability (you can check the other definitions, we have similar case)

2) The above is not the source of the problem you are facing. What happens is that 
can? :manage, Comment
is checking to see if there is any comment that can be managed by the user and this is a bit more abstract than it sounds. Meaning that if you have any definition in ability about the Comment model (such as  **can :manage, Comment, conference_id: conf_ids_for_organizer**) that means that there is (potentially) some comment record that can be managed by the user. And that's why the **can? :manage, Comment** returns true. Please do spend some time reading about it in the manual: https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities

3) You may wonder why this user has access to the ability definition that corresponds to an organizer or a cfp, when the user has neither roles. Perhaps by looking in the ability.rb file you can figure that out on your own. If not, let me shed some more light :) (This also leads to the solution of the problem)

app/models/ability.rb
```diff
@@ -155,6 +158,8 @@ def signed_in_with_cfp_role(user)
     can :manage, CallForPaper, conference_id: conf_ids_for_cfp
     can :manage, Commercial, commercialable_type: 'Event',
                              commercialable_id: Event.where(conference_id: conf_ids_for_cfp).pluck(:id)
+    can [:read, :create], Comment, commentable_type: 'Event',
```
differentreality: I think we should explicitely define the actions we want, **read** is not a controller action, it is an alias the gem offers but the actions it corresponds to could change on a future gem update.

differentreality:
@raluka that is what has been happening before no? Did something change?
After my comment about abilities, have you managed to debug and figure out no 3) of my comment?
If you go through the code in modesl/ability.rb you should be able to see when and why there is an ability defined for the user regarding Comment class (even though there is no actual record of Comment that is being accessible - but that's what we care about the, the class, since you are checking the ability by its class in the layout file).

app/models/ability.rb
```diff
@@ -137,6 +137,8 @@ def signed_in_with_organizer_role(user)
     can :manage, Sponsor, conference_id: conf_ids_for_organizer
     can :manage, SponsorshipLevel, conference_id: conf_ids_for_organizer
     can :manage, Ticket, conference_id: conf_ids_for_organizer
+    can [:index, :create], Comment, commentable_type: 'Event',
```
differentreality: You are authorizing create action but to which controller action does it correspond to?

spec/models/ability_spec.rb
```diff
@@ -198,6 +200,8 @@
       it{ should_not be_able_to(:manage, other_event.difficulty_level) }
       it{ should be_able_to(:manage, event.commercials.first) }
       it{ should_not be_able_to(:manage, other_event.commercials.first) }
+      it{ should be_able_to(:create, event.comment_threads.first) }
```
differentreality: I believe that the create action check succeeds for the wrong reasons. We should discuss this test.

raluka: I agree, let's discuss this tomorrow morning

ChrisBr:
@raluka As you asked in Slack what we suggest you could also test:

In your PR you should test the code you added!

Looking at your PR I don't see any (at least not much) model code, so need to introduce model tests.

You've introduced the Comment Controller, so what about a Controller test? How we do it in OSEM you can see here:
https://github.com/openSUSE/osem/blob/master/spec/controllers/conference_controller_spec.rb
http://everydayrails.com/2012/04/07/testing-series-rspec-controllers.html

Additional you've introduced Mailer code. We don't test the mailer atm in OSEM but if you want you can also have a lookt at this. Here are some resources:
http://guides.rubyonrails.org/action_mailer_basics.html
http://vicfriedman.github.io/blog/2013/08/05/how-to-write-rspec-mailer-tests/
http://blog.lucascaton.com.br/index.php/2010/10/25/how-to-test-mailers-in-rails-3-with-rspec/
(just skimmed the links, so I don't know how useful they are...)

app/models/user.rb
```diff
@@ -26,7 +29,7 @@ class User < ActiveRecord::Base
 
   devise(*devise_modules)
 
-  has_and_belongs_to_many :roles
+    has_and_belongs_to_many :roles
```
differentreality: why do we have an extra space here?

app/models/user.rb
```diff
@@ -11,6 +11,9 @@ class User < ActiveRecord::Base
 
   before_create :setup_role
 
+  # add scope
+  scope :comment_notifiable, ->(conference) {joins(:roles).where('roles.name IN (?)', [:organizer, :cfp]).where('roles.resource_id = ?', conference.id)}
```
differentreality: Even though we don't currently use roles table for roles other than scoped to **Conference**, in my opinion it is still better to fully form our where statement. 
resource_type: 'Conference' and resource_id: conference.id 
or we can just do  resource: conference

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Admin
+  class CommentsController < Admin::BaseController
+    load_and_authorize_resource :conference, find_by: :short_title
```
differentreality: https://github.com/CanCanCommunity/cancancan/wiki/Nested-Resources

Even though this line is technically correct, essentially it will never run because we have no variable conference_id set, as we are navigating to /admin/comments not in admin/conference/comments.

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+    load_and_authorize_resource
+
+
+    def index
+      @ordered_events = Event.order(:title).all
```
differentreality: This fetche all available events, but do we want to show **all** Events? Or just the ones that belong to a conference, for which conference the user has the role organizer or cfp?

app/helpers/application_helper.rb
```diff
@@ -278,4 +278,8 @@ def sign_up_path
       new_user_registration_path
     end
   end
+
+  def unread_notifications(user)
+    Comment.find_since_last_login(user)
```
differentreality: Do we really need to define a helper function for 1 line of code?

app/models/user.rb
raluka: by mistake, I will change it back

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Admin
+  class CommentsController < Admin::BaseController
+    load_and_authorize_resource
+
+    def index
+      @conferences_available = Conference.with_roles([:admin, :organizer, :cfp], current_user)
```
differentreality: Unfortunately there is no role named **admin**. Please check again the available roles: organizer, cfp, info_desk, volunteers_coordinator. 
In addition, the User model has an attribute called **is_admin**.

app/views/layouts/_user_menu.html.haml
```diff
@@ -21,7 +21,10 @@
     = link_to(destroy_user_session_path, :method=>'delete') do
       %span.fa.fa-minus
       Sign out
-- if can? :manage, Conference
+- if current_user.is_admin || (current_user.has_any_role? :admin, { :name => :organizer, :resource => :any },
```
differentreality: Please check your syntax: http://breakthebit.org/post/8453341914/ruby-19-and-the-new-hash-syntax

We discussed extensively about this implementation and I thought we agreed it is best to use cancancan methods to check for abilities?

app/helpers/application_helper.rb
```diff
@@ -278,4 +278,9 @@ def sign_up_path
       new_user_registration_path
     end
   end
+
+  def unread_notifications(user)
+    available_conferences_ids = Conference.with_roles([:admin, :organizer, :cfp], user).pluck(:id)
+    Comment.find_since_last_login(user).where(commentable_type: 'Event', commentable_id: Event.where(conference_id: available_conferences_ids))
```
differentreality: I think that here you could simply use Comment.accessible_by(current_ability) which would make all this much more understandable.

app/controllers/admin/comments_controller.rb
differentreality: I think it would make more sense for the index action of the CommentsController to set a @comments variable with the comments that we want available (those comments are Comment.accessible_by(current_ability), and I think it should work fine).

ChrisBr:
@raluka except the points @differentreality stated out, it LGTM so far :)

Do you plan to add some additional tests? Unfortunately the PR is still failing because of decreasing test coverage ...

app/controllers/admin/comments_controller.rb
nTraum: Hi,
I'm helping out @raluka working on this. We have a hard time coming up with a query that enables us to display the comments grouped by events title. Could you give us a hint on how your implementation would look like? :)

differentreality: @nTraum could you show us what you have tried so far so that we can guide you? Or elaborate on exactly what is giving you a hard time? Are you working on **Comment.accessible_by(current_ability)** or something else?

app/controllers/admin/conference_controller.rb
```diff
@@ -27,7 +27,6 @@ def index
       @recent_users = User.limit(5).order(created_at: :desc)
       @recent_events = Event.limit(5).order(created_at: :desc)
       @recent_registrations = Registration.limit(5).order(created_at: :desc)
-
```
differentreality: Shall we remove this file from the commit since there is no actual change to it?

app/controllers/admin/comments_controller.rb
raluka: @differentreality, me and @nTraum discussed Friday that, starting from `@comments = Comment.accessible_by(current_ability)`,  how can we sort comments by event_title, and after that call @comments in the view, in such manner that they will be displayed like Github does. So that was about the above question.
In the meantime, I think I found an workaround solution, and please tell me if this is the right one.
Starting from accessible comments, I take events ids, get all events with those ids, sorted after title:
```events_ids = Comment.accessible_by(current_ability).pluck(:comment able_id)
@events = Event.where(id: events_ids).order(title: :desc)
```
In this manner, it seems much easier to iterate through events and show them in the view as:
```
@events.each do |event|
  = event.title
  - event.comment_threads.each do |comment|
    = comment.body
```

raluka: @differentreality, me and @nTraum discussed Friday that, starting from `@comments = Comment.accessible_by(current_ability)`,  how can we sort comments by event_title, and after that call @comments in the view, in such manner that they will be displayed like Github does. So that was about the above question.
In the meantime, I think I found an workaround solution, and please tell me if this is the right one.
Starting from accessible comments, I take events ids, get all events with those ids, sorted after title:
````events_ids = Comment.accessible_by(current_ability).pluck(:comment able_id)
@events = Event.where(id: events_ids).order(title: :desc)
````
In this manner, it seems much easier to iterate through events and show them in the view as:
````@events.each do |event|
  = event.title
  - event.comment_threads.each do |comment|
    = comment.body
```

app/views/layouts/_user_menu.html.haml
raluka: @differentreality : in this case, should I ask `if can? :manage, Conference || can? :manage, CallForPaper || can? :manage, Registration || can? :manage, Vposition` ?? because after I modified `signed_in_with_roles(user)` in ability.rb, the link is not visible with `if can? :manage, Conference` condition

differentreality: @raluka could you describe in simple words what we want to achieve here? Because I think it makes perfect sense that the :manage option does not work... But please try to explain what we want to show and when, so that we can debug this!

raluka: In simple words: Administration link is not visible anymore for users with roles of :cfp, :info_desk, :volunteer_coordinator. Before changing `signed_in_with_roles(user)` in ability.rb (line:87), these users were able to see the link. What we want to achieve here? make the link visible again for users with one of above mentioned roles.

differentreality: Very nice! Those users, with these roles, must be able to see the link 'Administration' which redirects to /admin/conference. And why they able to see the link? What ability do they have that tells us that they should be seeing and accessing the link 'Administration' ?

raluka: they have the ability of managing different  parts of a Conference.

differentreality: Yes that is what is happening in simplified words, but technically speaking? Show me the code that assigns the necessary access!

raluka: you mean, the code that is in first reply?

differentreality: I mean the specific code that allows for the users you mentioned to have access to the conference, and that defines what kind of access they have. Those lines tell me nothing about a user with cfp being able to manage the conference model

raluka: that's because only organizer can manage the conference model. cfp can manage at most the event model. still, he should be able to see admin/conference page.

differentreality: All of your questions are answered in the ability.rb file. You need to read that file, understand **exactly** what each role can do.  manage the conference and see the admin/conference page are two very different things!  Once you know exactly which abilities each role has, you ll know exactly which ability each role has regarding the Conference model. Once you figure that out, you ll immediately be able to move on!

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      @posted_comments = grouped_comments(accessible_ordered_comments.find_comments_by_user(current_user))
+    end
+
+    private
+
+    def accessible_ordered_comments
+      Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
+    end
+
+    def grouped_comments(remarks)
```
ChrisBr: I'd prefer an example how the output looks in a comment in the function! It's hard to understand on a first glance ..
@differentreality what do you think?

app/controllers/admin/comments_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    def index
+      @comments = grouped_comments(accessible_ordered_comments)
+      @unread_comments = grouped_comments(accessible_ordered_comments.find_since_last_login(current_user))
+      @posted_comments = grouped_comments(accessible_ordered_comments.find_comments_by_user(current_user))
+    end
+
+    private
+
+    def accessible_ordered_comments
+      Comment.accessible_by(current_ability).joins('INNER JOIN events ON commentable_id = events.id').order('events.title', 'comments.created_at DESC')
```
ChrisBr: Not tried but would 
```
Comment.accessible_by(current_ability).joins(:comment).order('events.title', 'comments.created_at DESC')
```
not work?

app/controllers/admin/comments_controller.rb
differentreality: I think that there is no point in having the private functions.

We can just split the code of **index** into separate lines.

I also don't like the manual INNER JOIN. @raluka didn't we end up finding a solution for this that did not require this manual inner join? I hope you have notes of your previous code :)

spec/controllers/admin/comments_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  context 'logged in as admin, organizer or cfp' do
+    before :each do
+      sign_in(organizer)
+    end
+    describe 'GET #index' do
+      it 'populates a hash with comments' do
+        comment
```
ChrisBr: I assume you need comment to instantiate a comment factory?
You could move this also to the before block as you call in all of your tests!

spec/controllers/admin/comments_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  let(:participant) { create(:user) }
+  let(:event) { create(:event, conference: conference) }
+  let(:comment) { create(:comment, commentable_type: 'Event', commentable_id: event.id) }
+
+  context 'not logged in user' do
+    describe 'GET #index' do
+      it 'renders the :index template' do
+        comment
+        get :index
+        expect(response).to redirect_to(user_session_path)
```
ChrisBr: Is there no flash when a not logged in user tries to access a protected area?

app/controllers/admin/comments_controller.rb
differentreality: I think not, because comments do not have a relatioinship with events (I assume you meant joins(:events)?). There is commentable which in our case is only referring to Event model.

app/views/layouts/_user_menu.html.haml
```diff
@@ -21,7 +21,7 @@
     = link_to(destroy_user_session_path, :method=>'delete') do
       %span.fa.fa-minus
       Sign out
-- if can? :manage, Conference
+- if can? :access, Admin
```
differentreality: It seems to me that this file would be better structured if the links in it appeared (or not) independently of any logic there is behind our application.
That way even if we change some logic in the application, we don't have to worry about touching the code for this menu.

This if statement *implies* that if you have access to the admin area you can properly access the links inside the if statement (links redirect to **show** action for conference and **new** action for conference), which is of course not true.

Even though our current application logic might implicitely prevent us from showing the **new** link when it wouldn't be accessible, in my opinion, this is more of a happy coincidence rather than an explicit statement with code.

raluka: Actually, this line ```https://github.com/raluka/osem/blob/view_comment_notification/app/views/layouts/_user_menu.html.haml#L27``` is about **index**, not **show**, and it is following the logic from Admin::ConferenceController  #index. But I will add an extra condition for "Create New" link, for the future.

spec/controllers/admin/comments_controller_spec.rb
raluka: there is no flash when a non-logged user tries to access protected area, but it is a redirect to login page

spec/controllers/admin/comments_controller_spec.rb
raluka: I'll do that, thanks for tip!

spec/controllers/admin/comments_controller_spec.rb
differentreality: No, https://github.com/openSUSE/osem/blob/master/app/controllers/admin/base_controller.rb#L7
@ChrisBr perhaps we should change that in the future?

app/controllers/admin/comments_controller.rb
raluka: Related of INNER JOIN, if there is any other elegant method to ask database for comments sorted by event title and date of creation, I will be happy to know it and to change my query. Our discussion about this subject ended with you telling  that I should rather do the query in view, instead of controller. 
But I believe that the logic of code can stay in controller in this case, not in view, and I preferred to create 2 private methods, one for the query into data base, the other one for grouping the comments by conference and by events. I assumed it also that the code will be more readable, compared with the case when we are asking everything in a single chained  call.
Of course, if you prefer any other way to do this, please tell me and I'll do it.

differentreality:
Functionally speaking, I think this is good to go.
Do you plan on making any other changes in any part of the code? I remember you mentioning something about changing the view?

app/controllers/admin/comments_controller.rb
raluka: `Comment.accessible_by(current_ability).joins(:events).order('events.title', 'comments.created_at DESC')` is not working, if this was the suggestion.

differentreality:
Is coveralls complaining about coverage in the whole app and not just this PR?
I managed to get it up in https://github.com/openSUSE/osem/pull/676 but only by adding brand new tests that weren't there before.

raluka:
Also, Travis is complaining about some rubocop offenses, in a file I never modified. Should I also solve those offenses, in order to pass tests?

app/assets/stylesheets/osem.css.scss
```diff
@@ -71,4 +71,14 @@ p.comment-body {
 
 #account-already {
   font-size: 0.6em;
-}
\ No newline at end of file
+}
+
+/* comments views pane: use padding-bottom before next comment box */
+.panel.panel-default .panel-body .box {
```
hennevogel: box is a very bad class name and will likely clash with other things in the future...

app/models/ability.rb
```diff
@@ -88,12 +88,13 @@ def signed_in_with_roles(user)
...
-    signed_in_with_cfp_role(user)
-    signed_in_with_info_desk_role(user)
-    signed_in_with_volunteers_coordinator_role(user)
+    signed_in_with_organizer_role(user) if user.has_role? :organizer, :any
+    signed_in_with_cfp_role(user) if user.has_role? :cfp, :any
+    signed_in_with_info_desk_role(user) if user.has_role? :info_desk, :any
+    signed_in_with_volunteers_coordinator_role(user) if user.has_role? :volunteer_coordinator, :any
 
     # for users with any role
+    can :access, Admin
```
hennevogel: Users with any role can access Admin? Why?

hennevogel:
I concur with @differentreality, functionality wise this is awesome stuff. I'll merge it manually and remove the nitpicks here and there. Thank you both!

app/models/ability.rb
differentreality: That was happening already actually, all roles (organizer, cfp, info desk, volunteers coordinator) have access to the admin area (index Conference, which is inherited by https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L36). That's all the Admin option does (just a manual thing we thought of so that we can easily check if a user has any of those roles instead of typing a long if statement, https://github.com/raluka/osem/blob/view_comment_notification/app/models/ability.rb#L97)

PR #700, opened 2015-07-27
Roles updated

differentreality:
#583


differentreality:
This is also ready for review

screenshot-osem - chromium
screenshot-osem - chromium-1

app/controllers/admin/conference_controller.rb
```diff
@@ -169,29 +173,12 @@ def edit
...
-      render 'roles', formats: [:js]
-    end
-
-    def remove_user
-      @selection = params[:role]
-      @user.revoke @selection, @conference
-
-      @role_users = get_users(@selection)
-      render 'roles', formats: [:js]
+    def create_roles
```
ChrisBr: Hm, I don't like this in the controller ... 
What about do this in a [callback](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html)?

ChrisBr: Of course, you would need to rewrite your conditional in line ``72`` also...

app/controllers/admin/conference_controller.rb
```diff
@@ -67,11 +66,16 @@ def create
...
       if @conference.valid?
         @conference.save
+        flash[:notice] = 'Conference was successfully created.'
+
         # user that creates the conference becomes organizer of that conference
-        current_user.add_role :organizer, @conference
-        redirect_to(admin_conference_path(id: @conference.short_title),
-                    notice: 'Conference was successfully created.')
+        unless create_roles && (current_user.add_role :organizer, @conference)
+         flash[:alert] = 'Could not create roles properly!'
```
ChrisBr: Do we than an alert and a notice flash?
I don't like that ...

app/controllers/admin/conference_controller.rb
```diff
@@ -67,11 +66,16 @@ def create
 
       if @conference.valid?
```
ChrisBr: If you touch this action anyway, what about refactor this to 
``
@conference.save
``
instead of
``
if @conference.valid?
  @conference.save
  ...
``

it's done twice ...

app/controllers/admin/roles_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def edit
+      @users = @role.users
+    end
+
+    def update
+      role_name = @role.name
+
+      if @role.name == 'organizer' && params[:role][:user_ids].reject(&:empty?).blank?
```
ChrisBr: Hm, looks like a validation?

app/controllers/admin/roles_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        redirect_to admin_conference_role_path(@conference.short_title, @role.name)
+      else
+        flash[:error] = 'Could not update role! ' + @role.errors.full_messages.to_sentence
+        @role.name = role_name
+        render :edit
+      end
+    end
+
+    def toggle_user
+      if params[:user] && (user_email = params[:user][:email]) && user_email.present?
```
ChrisBr: Please move 
``
user_email = params[:user][:email]
``
out of the conditional! This is really hard to understand ...

app/controllers/admin/roles_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        user = User.find_by(email: user_email)
+        if user
+          state = params[:user][:state]
+          if state == 'false'
+            @role.users.delete(user)
+            flash[:notice] = "User #{user.email} removed successfully"
+          else
+            @role.users << user unless @role.users.include? user
```
ChrisBr: This looks like a validation
``unless @role.users.include? user``

app/controllers/admin/roles_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+        flash[:error] = 'You must enter a search word!'
+      end
+      redirect_to admin_conference_role_path(@conference.short_title, @role.name)
+    end
+
+    protected
+
+    def set_selection
+      # Set 'organizer' as default role, when there is no other selection
+      params[:id] ? (@selection = params[:id].parameterize.underscore) : (@selection = 'organizer')
```
ChrisBr: This is more readable
``
@selection = params[:id] ? params[:id].parameterize.underscore : 'organizer'
``

app/models/role.rb
```diff
@@ -1,9 +1,25 @@
 class Role < ActiveRecord::Base
-  has_and_belongs_to_many :users
   belongs_to :resource, polymorphic: true
+  has_and_belongs_to_many :users
 
   scopify
 
-  LABELS = ['Attendee', 'Volunteer', 'Speaker', 'Sponsor', 'Press', 'Keynote Speaker']
```
ChrisBr: Why did you remove it?

ChrisBr:
And you'd need to rebase!
Otherwise good work @differentreality

app/models/role.rb
differentreality: Roles will only be about actionable roles and we will not have labels mixed with roles.

app/controllers/admin/roles_controller.rb
differentreality: Is it? I am not seeing it.
I just don't want to excecute code that does not need to be executed.

Re-writing it another (simpler?) way anyway!

app/controllers/admin/conference_controller.rb
differentreality: Sth goes terribly wrong when I put it in after_save.
I will further investigate. Ideas welcome.

hennevogel: Can you tell us what goes wrong?

differentreality: First of all the fact that I don't know what this is doing here again... It is moved in models/conference.rb (see below)

hennevogel: Heh, so remove it here please :-)

hennevogel:
Alright, let's get this in!

PR #704, opened 2015-08-06
Fix ordering for event.average_voting in events#index

differentreality:
https://github.com/openSUSE/osem/issues/703

PR #705, opened 2015-08-07
added Voted? column to admin/events view

altsalt:
Background and textual notification to easily see which proposals a user has or has not voted upon.

Requesting merge.

### DIFF
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -34,6 +34,8 @@
           %b Difficulty
         %th
           %b State
+        %th
+          %b Voted?
       - @events.each do |event|
         %tr
           %td
@@ -147,3 +149,14 @@
                   %span.caret
                 %ul.dropdown-menu{:role=>"menu"}
                   = render 'change_state_dropdown', event: event
+
+          - if event.voted?(event, current_user)
+            - bgcolor = ""
+          - else
+            - bgcolor = "#F7819F"
+          %td{:style=>"background-color: #{bgcolor}"}
+            - if event.voted?(event, current_user)
+              Yes
+            - else
+              Not yet
+
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -34,8 +34,6 @@
           %b Difficulty
         %th
           %b State
-        %th
-          %b Voted?
       - @events.each do |event|
         %tr
           %td
@@ -56,6 +54,12 @@
                     = javascript_tag "$('label[avgrate=true]').prevAll().andSelf().addClass('bright');"
                   - else
                     = label_tag "label_rating", "", :class => "avgrating"
+                %br
+                - voted = event.voted?(event, current_user)
+                - if voted
+                  Your rating: #{voted.rating}
+                - else
+                  Not rated
               - else
                 0/#{@conference.call_for_paper.rating}
               %br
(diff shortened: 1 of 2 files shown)


app/views/admin/events/index.html.haml
```diff
@@ -34,6 +34,8 @@
           %b Difficulty
         %th
           %b State
+        %th
+          %b Voted?
```
hennevogel: Not very clear what this means. Voted on at all? Voted on by you?

hennevogel:
Otherwise, nice. Thanks! :+1:

altsalt:
I was trying to avoid using too much horizontal space and the text of "Yes" and "Not yet" seem self-explanatory to me.

Thoughts on re-wording?

hennevogel:
Well if you're trying to save horizontal space you should have put this into the 'rating' column with

```
- if event.voted?(event, current_user)
  Your Vote: number
- else
  %span{style = "background-color: #{bgcolor}" }
    =link_to "Vote" event_path(event)
```

altsalt:
I re-did the patch as suggested, using numbers not stars.
Dropped the background color as differences in text are noticeable.

p.s. I have also patched in pull #704

app/views/admin/events/index.html.haml
```diff
@@ -34,6 +34,8 @@
           %b Difficulty
         %th
           %b State
+        %th
```
differentreality: The new column needs to go too with the current implementation

differentreality:
I agree with Henne about not horizontally expanding.
However a visual indication of not-voted proposals could be useful. Perhals a success/warning label (http://getbootstrap.com/components/#labels)?

altsalt:
Good catch on the extra column!

I wrapped both notifications in a label, thoughts?

differentreality:
screenshot-15

differentreality:
I like the label indication, @hennevogel you?

I would change the wording though to "Your vote:" and "No vote" (because 'You haven't voted' is too long)

screenshot-19

altsalt:
"Haven't voted" is 13 characters
"Not voted" is 9 characters
"Not rated" is 9 characters
"Your vote: #" is 12 charaters

I don't think the word length matters too much as long as it is is clear.
Personally, I think "Your vote: #" and "Not rated" read best.

differentreality:
"Not rated" sounds to me as if the event has not rating at all.
"No vote" would use the same word origin as "Your vote" and I think it is more likely to be less confusing about the fact that it corresponds to this particular user's vote. Does this only make sense to me?

hennevogel:
We're not in the character saving business. We are in the guiding-users-to-do-useful-things business :-) So if you think about what this label should convey it's something like:

*You haven't voted yet, please click here immediately to do so!*

Now you can abbreviate this in any way as long as it does not loose it's meaning. My personal preference would be: 'Vote now'

differentreality:
@hennevogel this PR contains changes which we have already merged. Github says we can go ahead and merge this, but I'd like to re-confirm that with you. Should we merge or have the PR rebased?

hennevogel:
If you're not sure click the 'command line instructions' link below and see what happens if you pull it in manually :-)

differentreality:
Events with 0 average rating have **not** been rated by the current_user either.

Should we move the code for the label **after** the else code?

hennevogel:
What is this PR waiting for?

differentreality:
I ll merge it and do a grooming fix later for the label to appear when no one has voted.

Issue #649, opened 2015-04-23
Wrong days remaining label when submitting a new proposal


koslibpro:
Hey,

When opened the proposals page for the conference via OSEM (https://events.opensuse.org/conference/osc15/proposal), it showed a subzero number of days remaining (in my case -2). It is suggested to fix this by just showing 0 days remaining in such case, or completely changing the sentence shown there.

Keep up the good work,
kostas

ChrisBr:
@koslibpro Normally you shouldn't be able to access this page at all! Thanks for reporting!

koslibpro:
oh then have also in mind that it was not a bookmark or something, i accessed it completely normally via the app flow

ChrisBr:
@koslibpro From which page? On the root page and the conference page there are no links available anymore!

koslibpro:
Noticed it now, i cannot see the link directly, so i have to assume that the link showed up in there after navigating a bit within the app. Trying now to reproduce it, I'll let you know.

differentreality:
@ChrisBr  he is in proposal#index not proposal#new and I think the link on the top right user-menu stays either way (because you *need* to be able to see the list of your submitted proposals even after CFP is over. CfP will be over, afterwards your submission will be accepted.)

ChrisBr:
@differentreality hm, yeah you're right! But obviously we hide the link to the proposal#index page...

differentreality:
Fixed in #708

PR #708, opened 2015-08-24
Issue #649: Wrong days remaining label when submitting a new proposal

raluka:


app/views/proposal/index.html.haml
```diff
@@ -97,5 +97,5 @@
                   class: 'btn btn-default', id: "edit_proposal_#{event.id}"
   .row
     .col-md-12
-      - if @conference.cfp_open? || (current_user.has_role? :organizer, @conference)
+      - if @conference.cfp_open?
```
ChrisBr: Why do you remove
```
|| (current_user.has_role? :organizer, @conference)
```

app/views/admin/call_for_papers/show.html.haml
```diff
@@ -19,8 +19,12 @@
         %dt
           Days Left:
         %dd
-          - days = (@call_for_paper.end_date - Date.today).to_i
+          - if @conference.cfp_open?
```
ChrisBr: Hm, maybe we should move this to a helper or model method?

ChrisBr:
Except my two comments it LGTM!
Thanks @raluka

app/views/proposal/index.html.haml
raluka: With that condition, if you have role of organizer for conference, you can still see "New Proposal" button if call_for_papers period is closed. Also, you can submit new proposal after call_for_papers period is closed.

ChrisBr: I know! But as an Organizer I want to be able to submit proposals after the CFP is closed. Not sure if we already cover this in the admin menu.

@hennevogel @differentreality what do you think?

raluka: ok

raluka:
@ChrisBr , if everything is ok here, please let me know, so I squash commits. Thanks

app/views/conference/_call_for_paper.html.haml
```diff
@@ -25,14 +25,17 @@
         and closes
         %em
           = @conference.call_for_paper.end_date.strftime('%A, %B %-d. %Y.')
-        That means you have only
-        - days = (@conference.call_for_paper.end_date - Date.today).to_i
-        %b
-          = pluralize(days, 'days')
+        - if @conference.cfp_open?
+          That means you have only
+          - days = @conference.call_for_paper.remaining_days
```
ChrisBr: You don't need to store this in a variable. Just 

```
= pluralize(@conference.call_for_paper.remaining_days, 'days')
```

should be ok!

app/views/proposal/_encouragement_text.html.haml
```diff
@@ -6,18 +6,20 @@
...
     The submission period has begun
     %em
       = @conference.call_for_paper.start_date.strftime('%A, %B %-d. %Y')
     and closes
     %em
       = @conference.call_for_paper.end_date.strftime('%A, %B %-d. %Y.')
-    That means you have only
-    - days = (@conference.call_for_paper.end_date - Date.today).to_i
+    That means you have
+    - days = @conference.call_for_paper.remaining_days
```
ChrisBr: Same here!

app/views/admin/call_for_papers/show.html.haml
```diff
@@ -19,8 +19,12 @@
...
+            - days = @call_for_paper.remaining_days
+          - else
+            - days = 0
```
ChrisBr: I'd move the zero check to the helper, so you don't need to check it in the view!

Sth like 

```
def remaining_days(date = Date.today)
  result = (self.end_date - date).to_i
  result > 0 ? result : 0
end
```

ChrisBr:
Otherwise LGTM!

Squash the commits and we can :ship: it 
:clap:

ChrisBr:
Awesome @raluka!
Thank you :bouquet:

PR #712, opened 2015-09-23
Support JSONP exports of conference schedules for embedding

kormoc:
Currently, it's rather difficult to use OSEM and embed a schedule
into another website. We currently use OSEM for the backend, but
still use a different system for our main site. This change
allows us to dynamically embed the schedule in our main website
as well as allow us a simple way to get data for the use in
mobile applications or similar.

This limits the JSONP support to only GETS on json formatted
requests. This prevents any XSS even though we dont require
a authenticity token.

### DIFF
--- a/app/controllers/conference_controller.rb
+++ b/app/controllers/conference_controller.rb
@@ -1,6 +1,13 @@
 class ConferenceController < ApplicationController
   before_filter :respond_to_options
   load_and_authorize_resource find_by: :short_title
+  skip_before_filter :verify_authenticity_token, if: :json_request?
+
+  def json_request?
+    return false if request.request_method != 'GET'
+    return false if !request.format.json?
+    true
+  end
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).order('start_date ASC')
@@ -18,6 +25,11 @@ class ConferenceController < ApplicationController
       @today = Date.current.strftime('%Y-%m-%d')
     else
       @today = @conference.start_date.strftime('%Y-%m-%d')
+    end
+
+    respond_to do |format|
+        format.html
+        format.json { render json: @conference.as_json, callback: params[:callback] }
     end
   end
 
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -132,6 +132,73 @@ class Conference < ActiveRecord::Base
   end
 
   ##
+  # Builds out the model for returning a json response
+  #
+  #
+  def as_json(_options={})
+    super(
+        only: [
+            :description,
+            :end_date,
+            :logo,
+            :short_title,
+            :start_date,
+            :title,
+            ],
+        include: {
+            difficulty_levels: {
+                only: [
+                    :description,
+                    :id,
+                    :title,
+                ]
+            },
+            event_types: {
+                only: [
+                    :description,
+                    :id,
+                    :length,
+                    :title,
+                ]
+            },
+            rooms: {
+                only: [
+                    :name,
+                    :size,
+                ],
+                include: {
+                    events: {
+                        only: [
+                            :abstract,
+                            :description,
+                            :difficulty_level_id,
+                            :event_type_id,
+                            :guid,
+                            :is_highlight,
+                            :require_registration,
+                            :start_time,
+                            :subtitle,
+                            :title,
+                            :track_id,
+                            ],
+                        methods: [
+                            :speaker_names,
+                            ],
+                    },
+                },
+            },
+            tracks: {
+                only: [
+                    :description,
+                    :id,
+                    :name,
+                ]
+            },
+        }
+        )
+  end
+
+  ##
   # Checks if the user is registered to the conference
   #
   # ====Args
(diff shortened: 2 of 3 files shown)


app/controllers/conference_controller.rb
```diff
@@ -1,6 +1,13 @@
 class ConferenceController < ApplicationController
   before_filter :respond_to_options
   load_and_authorize_resource find_by: :short_title
+  skip_before_filter :verify_authenticity_token, if: :json_request?
+
+  def json_request?
+    return false if request.request_method != 'GET'
+    return false if not request.format.json?
+    true
```
ex-nerd: is it still ruby convention to leave off the `return`?  so weird...

app/serializers/conference_serializer.rb
```diff
@@ -20,7 +20,9 @@ def revision
   # FIXME: adjusting the format the DIRTY way, for oSC13.
   # If you think this is ugly, don't look at the methods below
```
meonkeys: Heh.

meonkeys:
:shipit:

Disclaimer: I've got no RoR skillz and don't know the OSEM codebase well.

app/controllers/conference_controller.rb
```diff
@@ -1,6 +1,13 @@
 class ConferenceController < ApplicationController
   before_filter :respond_to_options
   load_and_authorize_resource find_by: :short_title
+  skip_before_filter :verify_authenticity_token, if: :json_request?
+
+  def json_request?
+    return false if request.request_method != 'GET'
+    return false if not request.format.json?
```
ChrisBr:
```
unless request.format.json?
```

app/controllers/conference_controller.rb
```diff
@@ -19,6 +26,11 @@ def schedule
     else
       @today = @conference.start_date.strftime('%Y-%m-%d')
     end
+
+    respond_to do |format|
+        format.html
+        format.json { render :json => @conference.as_json(), :callback => params[:callback] }
```
ChrisBr: You should use the new Ruby hash syntax:

```
json: @conference.as_json()
```

app/models/conference.rb
```diff
@@ -132,6 +132,73 @@ def date_range_string
   end
 
   ##
+  # Builds out the model for returning a json response
+  #
+  #
+  def as_json(options={})
+    super(
+        :only => [
```
ChrisBr: Here as well!

ChrisBr:
@meonkeys thanks for your contribution :)

Hm, the only thing I worry is that we would publish events which aren't confirmed yet (but I'd be okay with that)... And I'm not sure if the as_json function is the right way to go ...

Additional, you have some style violations. You can run ```rubocop``` on your local machine to see the violations.

@hennevogel @differentreality what do you think?

meonkeys:
@kormoc ping

differentreality:
@ChrisBr Does it really show events that are not confirmed?
From what I am seeing it shows rooms, and then events that belong to those rooms. And since this is json format for schedule that makes sense. 
But in fact even if an event is confirmed, but not scheduled (assigned to room), it won't show.

Trying out the code locally alsoc concludes to this. Or I am missing sth :)

differentreality:
@kormoc would you git rebase opensuse/master and then auto fix the syntax with 
**bundle exec rubocop -Dc .rubocop.yml -a**   please?

kormoc:
Sorry for the branch churn. Rebasing mistype. The current PR should be good to go, with all the syntax fixes.

app/controllers/conference_controller.rb
```diff
@@ -1,6 +1,13 @@
 class ConferenceController < ApplicationController
   before_filter :respond_to_options
   load_and_authorize_resource find_by: :short_title
+  skip_before_filter :verify_authenticity_token, if: :json_request?
```
hennevogel: Why is this needed? verify_authenticity_token is for verification of the CSRF token for posting things, which you don't do.

differentreality: http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html

hennevogel: Ah there is an exception for ajax. Didn't know that. Please limit the skipping to the methods you really need it for. Not for the whole controller

app/models/conference.rb
```diff
@@ -132,6 +132,73 @@ def date_range_string
   end
 
   ##
+  # Builds out the model for returning a json response
+  #
+  #
+  def as_json(_options={})
```
hennevogel: You do know about [serialization](http://api.rubyonrails.org/classes/ActiveModel/Serialization.html) right?

hennevogel:
@kormoc Apart from the mentioned things I keep wondering what you are missing in the already existing API? You can query all the information you want.

kormoc:
@hennevogel None of the current APIs I'm aware of allow jsonp formats. We currently embed the schedule on a different domain name and use client side javascript to get the updated schedule and to display it on the page. This allows us to use a fully static hosting solution for our main page but allow dynamic data on the sections that need it.

For example, http://seagl.org/schedule/2015.html is currently using this patch.

differentreality:
@kormoc are you saying we could not integrate your changes into https://osem.seagl.org/api/v1/conferences instead of having something new?

differentreality:
I talked with @kormoc on IRC, we can integrate his changes into our api instead.

But as seagl conf is coming up, he won't be able to work on it soon.

differentreality:
Changes will be integrated in #729

meonkeys:
:dancer:

Issue #686, opened 2015-06-16
Postalcode requires number


milovanderlinden:
But international postal codes are not always only numbers.

ChrisBr:
Thanks for the hint! This is really ugly. We should never store a value like postalcode as an integer ...

https://github.com/openSUSE/osem/blob/master/db/schema.rb#L518

@skriesch How about this one? Aren't you a migration expert now? ;)

differentreality:
Fixed in #713

PR #713, opened 2015-09-29
Postalcode requires number - not anymore!

raluka:
Issue #686: Postalcode requires number

### DIFF
--- a/db/migrate/20150929142405_change_postalcode_format_in_venues.rb
+++ b/db/migrate/20150929142405_change_postalcode_format_in_venues.rb
@@ -1,5 +1,9 @@
 class ChangePostalcodeFormatInVenues < ActiveRecord::Migration
-  def change
+  def up
     change_column :venues, :postalcode, :string
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration.new('Cannot reverse migration.')
   end
 end
--- a/db/migrate/20150929142405_change_postalcode_format_in_venues.rb
+++ b/db/migrate/20150929142405_change_postalcode_format_in_venues.rb
@@ -1,0 +1,9 @@
+class ChangePostalcodeFormatInVenues < ActiveRecord::Migration
+  def up
+    change_column :venues, :postalcode, :string
+  end
+
+  def down
+    change_column :venues, :postalcode, :integer
+  end
+end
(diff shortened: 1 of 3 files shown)


differentreality:
@raluka you are not supposed to manually edit the schema file. 

You need to add a migration with your change, run the migration, and the changes in schema will automatically appear.

http://edgeguides.rubyonrails.org/active_record_migrations.html
Check out change_column

raluka:
It seems I forgot to add migration file to Git repo :). I'll add it tomorrow morning. @differentreality , do you have anything against using up/down methods for migration? Or do you prefer change method?

differentreality:
@raluka we don't really have a choice in this case, rails can't guess the previous type we were using :)(what's the opposite of the line of code in your migration? - We can't be sure)

differentreality:
Very nice @raluka !
(Tip: it is helpful for you and everyone, if in the description of your PR you add the link to the issue, https://github.com/openSUSE/osem/issues/686)

@ChrisBr  @hennevogel do you think we should validate acceptable characters in postalcode field? Numbers, letters, dash (-) ?

hennevogel:
Nah who knows what's true for world wide postal codes...

ChrisBr:
LGTM

db/migrate/20150929142405_change_postalcode_format_in_venues.rb
```diff
@@ -0,0 +1,9 @@
+class ChangePostalcodeFormatInVenues < ActiveRecord::Migration
+  def up
+    change_column :venues, :postalcode, :string
+  end
+
+  def down
+    change_column :venues, :postalcode, :integer
```
differentreality: This won't work. If the field already has a string value, turning that field into integer type raises an error.
If we force going back to integer type, we need to alter/delete the existing postalcode (or at least its non number parts). Otherwise this is an irreversible migration.

@hennevogel your input? Reverse the migration by deleting non number chars or make it irreversible?

raluka:
I rolled back previous migration, and made an irreversible one.

differentreality:
That looks good, but now db/schema.rb has the changes of another PR too :/

raluka:
@differentreality fixed

hennevogel:
Alright. Let's roll :-)

PR #715, opened 2015-10-06
Issue #603: Refactor EmailSettings attributes

raluka:
EmailSetting incosistency #603


differentreality:
Since we are at it, what about we also uniform the name of the variables?
Like:
call_for_papers_dates_updates_body -> call_for_papers_dates_updated_body
venue_update_body -> venue_updated_body
updated_conference_dates_body -> conference_dates_updated_body
updated_conference_registration_dates_body -> conference_registration_dates_updated_body

What do you think?

db/schema.rb
```diff
@@ -515,7 +515,7 @@
     t.integer  "photo_file_size"
     t.datetime "photo_updated_at"
     t.string   "street"
-    t.integer  "postalcode"
+    t.string   "postalcode"
```
differentreality: That is not relevant to this PR

hennevogel:
I very much like the consistency!

raluka:
@differentreality , @hennevogel what do you think about this PR? Should I improve it  more?

differentreality:
It looks good, but letter_opener does not work (for me?) when it comes to emails for things happening outside of devise. Should it be working? I would have liked to try this PR out.

raluka:
@differentreality, what exactly did not work?

differentreality:
@raluka when you accept a proposal, for example, there is an email sent to the submitter (assuming the email settings params are set). The same happens when you reject a proposal, or when you register to a conference etc. 
I don't see that email being sent to me through letter_opener, neither in master nor in your PR. So even though everything seems nice, I can't really try it out in practice.

raluka:
Checkboxes for sending notifications are disabled by default, they have to be enabled  in order to send emails for updates.

differentreality:
@raluka yes I know. So?

raluka:
Also, start rake jobs:work. With these 2 conditions, sending notifications works for me.

hennevogel:
So let's get this in, thanks :girl:!

PR #716, opened 2015-10-14
Abilities: Minor changes & grooming

differentreality:
Venue has show action, not index.
Make the ability to schedule an Event more clear.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -154,7 +154,7 @@ class Ability
     can :manage, DifficultyLevel, conference_id: conf_ids_for_cfp
     can :manage, EmailSettings, conference_id: conf_ids_for_cfp
     can :manage, Room, conference_id: conf_ids_for_cfp
-    can :index, Venue, conference_id: conf_ids_for_cfp
+    can :show, Venue, conference_id: conf_ids_for_cfp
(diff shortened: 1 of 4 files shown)


app/models/ability.rb
```diff
@@ -127,6 +127,7 @@ def signed_in_with_organizer_role(user)
     can :manage, Vday, conference_id: conf_ids_for_organizer
     can :manage, CallForPaper, conference_id: conf_ids_for_organizer
     can :manage, Event, conference_id: conf_ids_for_organizer
+    can :schedule, Event, conference_id: conf_ids_for_organizer
```
hennevogel: `:manage` means everything

differentreality: Yes indeed, and in fact there does not even exist a function called 'schedule' in the controller. 
The reason I thought we could do it like this is merely for clarity.
You can schedule an event if you can update its start time. Perhaps that is not 100% clear to everyone.
If you think it does not really add any value, I ll remove it.

hennevogel: It doesn't. Either you explicitly allow only specific methods and remove `:manage` or you use `:manage`. Everything else is just misleading IMHO

differentreality:
Ready for review!

ChrisBr:
LGTM :+1:

Issue #553, opened 2014-11-28
Introduce strong parameters 


ChrisBr:
As already described in 
https://github.com/openSUSE/osem/blob/master/Gemfile#L96
we should introduce strong parameters.
Till now it wasn't really necessary but with the new is_highlight (#552) attribute in events we have an attribute which normal users shouldn't be able to alter.

PR #719, opened 2015-10-16
Introduce strong parameters

differentreality:
https://github.com/openSUSE/osem/issues/553

* removes gem 'protected attributes'
* removes attr_accessible from models
* adds model_params function in controllers and permits all attributes previously present in attr_accessible (we perhaps not need them all)
* removes params[:model] from controllers and substitutes it with model_params
* removes user related code from proposal#update and admin/events#update, as we no longer change user data in update function

Notes:
  1.  Rspec tests always need to include model parameters. Sometimes we would just do ```patch :update, id: user.id``` but that is no longer enough because the update action references the user_params, which in this case would not exist and that would rais an error. We need to do ```patch :update, id: user.id, user: { name: 'something' }``` instead


app/controllers/admin/users_controller.rb
```diff
@@ -35,5 +35,12 @@ def update
     end
 
     def edit; end
+
+    private
+
+    def user_params
+      params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
```
ChrisBr: Updating users is only allowed by admins, right? Because if this would be possible for other roles we have to think about the is_admin attribute!

app/controllers/ticket_purchases_controller.rb
```diff
@@ -29,4 +29,10 @@ def destroy
                         "#{@ticket_purchases.errors.full_messages.join('. ')}."
     end
   end
+
+  private
+
+  def ticket_purchase_params
+    params.require(:ticket_purchase).permit(:ticket_id, :user_id, :conference_id, :paid, :quantity)
```
ChrisBr: Please remove ``paid``! A user shouldn't be able to mark a purchase paid ;)

ChrisBr:
Beside my nitpicks it LGTM :+1: 
Awesome work @differentreality, this issue is open since a loooong time ...

hennevogel:
And please rebase :-)

app/controllers/admin/users_controller.rb
differentreality: Yes, and I am also able to edit my own user, of course.
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L67

I am not 100% sure is_admin attribute requires more attention. Please share if you have more thoughts.

differentreality:
Any other comments? :)

ChrisBr:
@differentreality why is this still open? Together with #738 we would get rid of all hakiri issues :+1:

Issue #720, opened 2015-10-17
[production][unknown] ActionController::RoutingError: uninitialized constant Admin::StatsController


differentreality:
unknown threw this 1 times
<pre>
ActionController::RoutingError: uninitialized constant Admin::StatsController
</pre>
[https://events.opensuse.org/admin/conference/osc14/stats](https://events.opensuse.org/admin/conference/osc14/stats)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca99dc71fa00b1000002/problems/5620f0e1764b23000c00000a)

differentreality:
Noticed that the route for stats is still there, and should be removed too.

Shriyanshagro:
@differentreality , The link you have provided isn't available now. So is this issue still open , If yes, can you please elaborate it more.

differentreality:
There isn't much to elaborate on really. There is a routing error about statscontroller. To debug the error:
- Check the routes: is there something relevant about stats?
- Check the controllers, is there a statscontroller?

4bic:
@differentreality @ChrisBr  , [this pr](https://github.com/openSUSE/osem/pull/897)  should clean up the routes

differentreality:
Closed via https://github.com/openSUSE/osem/issues/720

Issue #234, opened 2014-06-12
Clear indication of accept/reject event without email


differentreality:
The option 'without' email should be obvious to the cfp team at all times.

The cfp team member that will do the event acceptances/rejections does not necessarily know or remember if there is an email template in the email settings. 
Clear indication of whether or not the submitter of the proposal will be notified about the CFP team's actions is so very important.

(As a side note, probably for a separate issue, the email templates should not be activated unless they have content.)

ChrisBr:
@differentreality @hennevogel 
Ping

Currently it's this way implemented:

No Mail template generated:
- Accept Event

Mail template generated:
- Accept Event
- Accept Event (Without email)

PR #721, opened 2015-10-17
Style

differentreality:
* Fixes hakiri warning about protect_from_forgery with: :exception
* Removes stats route https://github.com/openSUSE/osem/issues/720 
* Controllers witih separate line for flash[:error] / flash[:notice] and then redirect_to
* Allows for ticket with price of 0 (could be useful for some conferences)
* Indicate whether or not an email is sent on accept/reject of an event https://github.com/openSUSE/osem/issues/234

### DIFF
--- a/app/controllers/admin/call_for_papers_controller.rb
+++ b/app/controllers/admin/call_for_papers_controller.rb
@@ -15,8 +15,8 @@ module Admin
       @call_for_paper = @conference.build_call_for_paper(call_for_paper_params)
 
       if @call_for_paper.save
-        redirect_to admin_conference_call_for_paper_path,
-                    notice: 'Call for papers successfully created.'
+        flash[:notice] = 'Call for papers successfully created.'
+        redirect_to admin_conference_call_for_paper_path
       else
         flash[:error] = "Creating the call for papers failed. #{@call_for_paper.errors.full_messages.join('. ')}."
         render :new
@@ -34,8 +34,9 @@ module Admin
       if @cfp.update_attributes(params[:call_for_paper])
         Mailbot.delay.send_on_call_for_papers_dates_updates(@conference) if send_mail_on_cfp_dates_updates
         Mailbot.delay.send_on_schedule_public(@conference) if send_mail_on_schedule_public
-        redirect_to(admin_conference_call_for_paper_path(@conference.short_title),
-                    notice: 'Call for papers successfully updated.')
+
+        flash[:notice] = 'Call for papers successfully updated.'
+        redirect_to admin_conference_call_for_paper_path(@conference.short_title)
       else
         flash[:error] = "Updating call for papers failed. #{@cfp.errors.to_a.join('. ')}."
         render :new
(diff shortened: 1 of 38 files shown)


app/controllers/admin/call_for_papers_controller.rb
```diff
@@ -15,8 +15,8 @@ def create
       @call_for_paper = @conference.build_call_for_paper(call_for_paper_params)
 
       if @call_for_paper.save
-        redirect_to admin_conference_call_for_paper_path,
-                    notice: 'Call for papers successfully created.'
+        flash[:notice] = 'Call for papers successfully created.'
+        redirect_to admin_conference_call_for_paper_path
```
hennevogel: Why?

differentreality: Why splitting them in 2 lines?  
* Conformity: one style in all our code
* Clarity: each line does 1 specific thing, and it is clear what each line of code does
* Easier to read and understand
* Single style, as guideline, for new contributors

Part of our code has 1 single line with the redirect that includes the notice/alert message (app/controllers/admin/call_for_papers_controller.rb) and part of the code has them separated in 2 lines (https://github.com/openSUSE/osem/blob/master/app/controllers/admin/call_for_papers_controller.rb#L21). Some of the code is somewhere in between with just the redirect_to than inclues the message, but with a few break lines in between (https://github.com/openSUSE/osem/blob/master/app/controllers/admin/call_for_papers_controller.rb#L18). And that's even in one single file.

Part of the code has ***redirect_to admin_conference_call_for_paper_path, notice: 'text'*** and part of it has the same thing but with a parenthesis ***redirect_to(admin_conference_call_for_paper_path, notice: 'text')***

In my opinion, it would be nice to have it all coded the same way. 
And I think that separation of the message from the redirection code is easier to read and graps while browsing through lines of code.

hennevogel:
I can live with the `redirect_to()` => `redirect_to` changes but the flash changes I find strange. What's the reasoning behind those?

differentreality:
Same reasoning actually.

Part of the code has one style (https://github.com/openSUSE/osem/blob/master/app/controllers/admin/contacts_controller.rb#L15) and part of it has the other (https://github.com/openSUSE/osem/blob/master/app/controllers/admin/campaigns_controller.rb#L15). There are more than a few examples like that throughout our code. 
**If** we decide we want to have the same style, we have to change either one or the other.

For me having 2 separate lines for code that does 2 different things (one line for the message shown, one line for the redirection) is more structured and helps me read code faster. 
I am sure not everyone feels that way. Which styling would you prefer, if any?

hennevogel:
I prefer the one created by the rails generators which is 

`redirect_to @object, notice: 'Message'`

because this is the 'standard' and people are used to it.

differentreality:
There is one more thing. That 'standard' syntax only works with notice and alert, but we have been using error instead of alert so far.

differentreality:
Suggestions?

hennevogel:
Add a flash type of error? 

http://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html#method-i-add_flash_types

ChrisBr:
@differentreality can you please drop fafc06dd6e40dd219e988c948a7cf0a50ae5a878
(fixed in 81a43cc2b6ad4577c3fb34f8605ee01a6c1e7993) and rebase?

ChrisBr:
What's the consent of your discussion with @hennevogel ?

differentreality:
I will create a new PR.

PR #723, opened 2015-10-22
Fix update rails and dependencies (v.4.2.4)

raluka:


app/models/datatable.rb
```diff
@@ -74,7 +74,7 @@ def per_page
   def sort_order
     colnum = 0
     sort_by = []
-    while true
+    loop do
```
ChrisBr: Is this recommended by rubocop?
Anyway, I don't like ``while true`` nor ``loop do`` ...

PR #724, opened 2015-10-22
Fix update rails and dependencies (from v.4.1 to v.4.2.4)

raluka:
After gems update, **Rubocop** complained about multiple offenses, so I used auto-correct command.


config/initializers/wrap_parameters.rb
```diff
@@ -1,14 +1,14 @@
...
+  wrap_parameters format: [:json] if respond_to?(:wrap_parameters)
 end
 
-# Disable root element in JSON by default.
-ActiveSupport.on_load(:active_record) do
-  self.include_root_in_json = false
-end
+# To enable root element in JSON for ActiveRecord objects.
+# ActiveSupport.on_load(:active_record) do
+#  self.include_root_in_json = true
```
ChrisBr: Why is it disabled now?

raluka: It is disabled  by default  in Rails 4.2, and it can be enabled if necessarily.

ChrisBr:
I guess  #732 supersedes this PR? So we can close this one I'd say!

Issue #417, opened 2014-08-01
Refactor Conference: Program


hennevogel:
* has_one Program
 * has_many Rooms()
 * has_many Tracks()
 * has_many DifficultyLevels()
 * has_many Events()
 * has_many EventTypes()

ChrisBr:
@hennevogel same here, refactoring is done?

hennevogel:
Nope all of them are still somewhere else.

differentreality:
Program to also include call_for_papers?

hennevogel:
The call for papers stuff should be inside Program I think

differentreality:
I am thinking this would also result in a change in routes, eg:
admin_conference_events_path -> admin_conference_program_events_path

differentreality:
All done here?

hennevogel:
Jawoll! :-)

PR #728, opened 2015-10-24
Introduce Program 

differentreality:
https://github.com/openSUSE/osem/issues/417
Program includes cfp, rooms, tracks, events, event_types, difficulty_levels.

* Rename call_for_paper -> to cfp
* After conference creation, a program is automatically created too
* **Routes** changed, eg. admin_conference_events_path -> admin_conference_program_events_path
* Program can be managed by the same roles as CallForPaper was (that's organizer and cfp)
* ahoy/program.rb model authomatically created
* **Methods** still in **Cfp** model (the ones that explicitely  concern the call for papers, and not the program of the conference in general): 
  * notify_on_cfp_date_update?
  * weeks
  * start_week
  * end_week
  * remaining_days
* Methods moved in **Program**:
  *  rating_enabled? (new method)
  * cfp_open?
  * notify_on_schedule_public?
 
* Program vs Cfp: A conference can have a program (submitted proposals, rating of proposals, schedule, tracks, etc) without opening a public Call for Papers.
* Rename schedule_changes -> schedule_fluid
* Rooms belong to venue
* Added validation check for conference_id and venue_id: https://github.com/differentreality/osem/blob/refactor_conference_program/app/models/venue.rb#L8 and https://github.com/differentreality/osem/blob/refactor_conference_program/app/models/room.rb#L9
* New ability: do not delete venue, if there are rooms assigned to events https://github.com/differentreality/osem/blob/refactor_conference_program/app/models/ability.rb#L109

I also stumbled upon a difference in handling not null constraint in mysql and sqlite. Plz check last commit.


hennevogel:
I'm not sure about rooms to the program. A venue has rooms, a program not. And a room can have no program but a program must have a room.

differentreality:
@hennevogel I think a program could have many rooms, but I don't think this is very relevant because a scheduled event **must** have a room assigned to it.
So, there can be a program with CFP, but if there are no submitted proposals, which are then accepted, and confirmed, and scheduled, why does a program need a room (or rooms)?

So perhaps we should not link a program with rooms at all?

differentreality:
I reworked it so that rooms belong to venue.
I also added some comments to my initial comment of the PR.

differentreality:
This is ready for review!

hennevogel:
Merged manually

PR #729, opened 2015-10-24
API with more options

differentreality:
includes #712

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -1,15 +1,12 @@
 module Api
   module V1
     class RoomsController < Api::BaseController
+      load_resource :conference, find_by: :short_title
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          rooms = Room.all
-        else
-          conference = Conference.find_by_guid(params[:conference_id])
-          rooms = conference.rooms
-        end
+        @conference ? (rooms = @conference.rooms) : (rooms = Room.all)
+
         respond_with rooms
       end
     end
--- a/app/controllers/api/v1/tracks_controller.rb
+++ b/app/controllers/api/v1/tracks_controller.rb
@@ -1,15 +1,12 @@
 module Api
   module V1
     class TracksController < Api::BaseController
+      load_resource :conference, find_by: :short_title
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          tracks = Track.all
-        else
-          tracks = Track.joins(:conference)
-          tracks = tracks.where(conferences: { guid: params[:conference_id] })
-        end
+        @conference ? (tracks = @conference.tracks) : (tracks = Track.all)
+
         respond_with tracks
       end
     end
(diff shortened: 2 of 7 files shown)


app/serializers/conference_serializer.rb
```diff
@@ -20,16 +68,18 @@ def revision
...
-  # FIXME: just giving suseconferenceclient something to play with
-  def description
-    'openSUSE Conference 2013 - Power to the Geeko'
-  end
-
-  # FIXME: same than the former
-  def url
-    'https://conference.opensuse.org/'
-  end
+#   # FIXME: just giving suseconferenceclient something to play with
```
hennevogel: Why commented this stuff?

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -1,15 +1,12 @@
...
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          rooms = Room.all
-        else
-          conference = Conference.find_by_guid(params[:conference_id])
-          rooms = conference.rooms
-        end
+        @conference ? (rooms = @conference.rooms) : (rooms = Room.all)
```
hennevogel: I wonder if this makes sense. Listing all the rooms in the DB...

app/controllers/api/v1/tracks_controller.rb
```diff
@@ -1,15 +1,12 @@
...
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          tracks = Track.all
-        else
-          tracks = Track.joins(:conference)
-          tracks = tracks.where(conferences: { guid: params[:conference_id] })
-        end
+        @conference ? (tracks = @conference.tracks) : (tracks = Track.all)
```
hennevogel: Same here. Who could be interested in all tracks in the DB?

app/serializers/conference_serializer.rb
```diff
@@ -20,16 +68,18 @@ def revision
   # FIXME: adjusting the format the DIRTY way, for oSC13.
   # If you think this is ugly, don't look at the methods below
```
hennevogel: I doubt this comment has any meaning anymore...

app/serializers/conference_serializer.rb
differentreality: Ctrl+D instead of Ctrl+S I guess :)

app/controllers/api/v1/rooms_controller.rb
differentreality: It was possible to list all rooms before, so if there is any chance someone used that data to do sth, if we remove it now, things will break.

I think there is small possibility that someone might need that info. The only scenario I can think of is multiple conferences being co-hosted, 1 OSEM instance, but they want to extract overall data for all conferences to do sth with that data.

I'd more focus on what's the harm keeping it?

differentreality:
Ready for review!

ChrisBr:
LGTM if @hennevogel is okay with it!

app/controllers/api/v1/rooms_controller.rb
hennevogel: Cruft that nobody needs

hennevogel:
Anyway. Let's get this in :-)

Issue #755, opened 2016-01-20
Date Range Validation issue while creating a Conference


chinmaydd:
It is currently possible to create a conference with a `start_date` being greater than the `end_date`.<br>
![screenshot from 2016-01-21 00 28 07](https://cloud.githubusercontent.com/assets/6960646/12459888/08fd2e34-bfd7-11e5-8799-5da32b92513d.png)

PR #756, opened 2016-01-20
Fixes Start date being greater than End date

chinmaydd:
This patch aims at fixing #755 .

Added model validation for checking the same. Test case is updated in the file `conference_spec.rb`.


ChrisBr:
LGTM!

Thanks for your contribution :+1:

PR #762, opened 2016-02-01
Add registration limit to conference

### LINKED ISSUE #761: Feature request: Registration limit to conference
Some conference has a limit seat / space so it would be great a configuration option to limit the amount of registrations to a conference. 

If the conference is full,  a message should be present to the user to indicate this. 

(a better workflow will be a waiting list, but I think is part of a subsequent ticket if this limit feature is included)

davidsf:
Add a registration limit to a conference. Default is 0 that means unlimited registration limit.

If the registration limit is exceeded, redirects the user to the root_path with a alert. 

fix #761

### DIFF
--- a/app/controllers/admin/conference_controller.rb
+++ b/app/controllers/admin/conference_controller.rb
@@ -211,7 +211,7 @@ module Admin
                                          :vpositions_attributes, :use_volunteers, :color,
                                          :sponsorship_levels_attributes, :sponsors_attributes,
                                          :photos_attributes, :targets, :targets_attributes,
-                                         :campaigns, :campaigns_attributes)
+                                         :campaigns, :campaigns_attributes, :registration_limit)
     end
 
     def get_users(role_name)
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -20,6 +20,11 @@ class ConferenceRegistrationsController < ApplicationController
       redirect_to edit_conference_conference_registrations_path(@conference.short_title)
     end
 
+    if @conference.registration_limit_exceeded?
+      redirect_to root_path, alert: "Sorry, registration limit exceeded for #{@conference.title}"
+      return
+    end
+
     @registration = Registration.new
 
     # @user variable needs to be set so that _sign_up_form_embedded works properly
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -65,6 +65,7 @@ class Conference < ActiveRecord::Base
 
   validates_uniqueness_of :short_title
   validates_format_of :short_title, with: /\A[a-zA-Z0-9_-]*\z/
+  validates :registration_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
 
   # This validation is needed since a conference with a start date greater than the end date is not possible
   validate :valid_date_range?
@@ -534,6 +535,10 @@ class Conference < ActiveRecord::Base
     email_settings.send_on_conference_registration_dates_updated &&
     !email_settings.conference_registration_dates_updated_subject.blank? &&
     email_settings.conference_registration_dates_updated_body
+  end
+
+  def registration_limit_exceeded?
+    registration_limit > 0 && registrations.count >= registration_limit
   end
 
   private
--- a/app/views/admin/conference/edit.html.haml
+++ b/app/views/admin/conference/edit.html.haml
@@ -18,5 +18,6 @@
         = f.input :timezone, :as => :time_zone, :hint => "The conference time zone"
         = f.input :start_date, :as => :string, :input_html => { :id => "conference-start-datepicker", :readonly => "readonly"  }
         = f.input :end_date, :as => :string, :input_html => { :id => "conference-end-datepicker", :readonly => "readonly"  }
+      = f.inputs name: "Registrations" do
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
       = f.action :submit, :as => :button, :button_html => {:class => "btn btn-primary"}
-
--- a/app/views/conference/_conference_details.html.haml
+++ b/app/views/conference/_conference_details.html.haml
@@ -27,7 +27,9 @@
               - if conference.user_registered?(current_user)
                 = link_to "My Registration", conference_conference_registrations_path(conference.short_title), :class =>"btn btn-default"
               - else
-                = link_to "Register", new_conference_conference_registrations_path(conference.short_title), :class =>"btn btn-default"
+                = link_to "Register", new_conference_conference_registrations_path(conference.short_title), class: "btn btn-default", disabled: conference.registration_limit_exceeded?
+                - if conference.registration_limit_exceeded?
+                  Sorry, no places left
             - if !current_user.nil? && current_user.proposal_count(conference) > 0
               = link_to "My Proposals", conference_program_proposal_index_path(conference.short_title), :class =>"btn btn-default"
             - elsif conference.program.cfp_open?
--- a/app/views/conference/_registration.html.haml
+++ b/app/views/conference/_registration.html.haml
@@ -6,19 +6,25 @@
   .row
     .col-md-12.text-center
       %h1 Registration
-      - if @conference.tickets.empty?
-        %p.lead
-          Going to
-          = @conference.short_title
-          is free of charge.
+
+      - if @conference.registration_limit_exceeded?
         %p
-          We only ask you to register yourself until
-          = @conference.registration_period.end_date.strftime('%A, %B %-d. %Y')
-          so we can plan for the right amount of people.
+          Sorry, the conference registration limit has exceeded
       - else
-        %p
-          The registration period ends on
-          = @conference.registration_period.end_date.strftime('%A, %B %-d. %Y')
-      %p.cta-button
+        - if @conference.tickets.empty?
+          %p.lead
+            Going to
+            = @conference.short_title
+            is free of charge.
+          %p
+            We only ask you to register yourself until
+            = @conference.registration_period.end_date.strftime('%A, %B %-d. %Y')
+            so we can plan for the right amount of people.
+            %p.cta-button
+        - else
+          %p
+            The registration period ends on
+            = @conference.registration_period.end_date.strftime('%A, %B %-d. %Y')
+          %p.cta-button
         = link_to(new_conference_conference_registrations_path(@conference.short_title), class: 'btn btn-lg btn-success') do
           Register Now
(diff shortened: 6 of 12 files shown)


app/models/conference.rb
```diff
@@ -580,6 +580,10 @@ def notify_on_registration_dates_changed?
     email_settings.conference_registration_dates_updated_body
   end
 
+  def registration_limit_exceeded?
+    registration_limit > 0 && registrations.count==registration_limit
```
ChrisBr: ``
registrations.count == registration_limit
``

app/controllers/conference_registrations_controller.rb
```diff
@@ -20,6 +20,11 @@ def new
       redirect_to edit_conference_conference_registrations_path(@conference.short_title)
     end
 
+    if @conference.registration_limit_exceeded?
```
ChrisBr: Don't! Please use a validation for it!

app/controllers/admin/conference_controller.rb
```diff
@@ -207,7 +207,7 @@ def conference_params
                                          :vpositions_attributes, :use_volunteers, :color,
                                          :sponsorship_levels_attributes, :sponsors_attributes,
                                          :photos_attributes, :targets, :targets_attributes,
-                                         :campaigns, :campaigns_attributes)
+                                         :campaigns, :campaigns_attributes, :registration_limit)
```
ChrisBr: I'm missing validations for registration limit!

app/views/admin/conference/edit.html.haml
```diff
@@ -18,5 +18,6 @@
         = f.input :timezone, :as => :time_zone, :hint => "The conference time zone"
         = f.input :start_date, :as => :string, :input_html => { :id => "conference-start-datepicker", :readonly => "readonly"  }
         = f.input :end_date, :as => :string, :input_html => { :id => "conference-end-datepicker", :readonly => "readonly"  }
+      = f.inputs :name => "Registrations" do
+        = f.input :registration_limit, :as => :string, :hint => "Limit the number of registrations to the conference (0 no limit)"
```
ChrisBr: Please use html5 input validation:
type="number" min="0" max="999"

ChrisBr:
@davidsf thanks for the contribution! See my comments :+1:

app/controllers/conference_registrations_controller.rb
davidsf: Sorry but in conference_registrations_controller I'm not updating the conference object, so I need to check it the limit has exceeded before show the registration form. Maybe there are another way to do it that for sure I don't know :)

ChrisBr: Ah, good point!

Nevertheless, I'd add a validation to the registration model so it's checked on creation. Either a custom validation, sth like this:

``
validate :registration_limit_not_exceeded
 
  def :registration_limit_not_exceeded
    if registration_limit_exceeded
      errors.add(:base, "Registration limit exceeded.")
    end
  end
``

Or you could also have a look at http://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validates_associated (though, never used it)

Furthermore you should hide the registration button on the front page and the splash page. Showing the button and simply redirect back to the root_path is not a good user experience from my point of view.

If we want to keep this code in the controller needs to be discussed, probably it makes sense for the case the link is distributed to social media etc...

davidsf: The registration_limit_exceed method is on conference model. IMHO, I think is conference responsibility to check the limit, not the registration model. 

I have hidden the registration button if limit exceded in the home page. 

As soon as you merge my other PR, I change the _registration partial in the conference show.

ChrisBr: Anyway, we can discuss if the validation belongs to conference or registration! At least I would like to have a validation which prohibits to have more registrations than specified in :registration_limit attribute, either in registration or conference model! 

The other PR is merged!

davidsf: Ok, I have added the validation in conference model and added a msg to the user in _registration in conference show when the limit has exceeded. 

If I miss anything let me know :)

app/views/conference/_conference_details.html.haml
```diff
@@ -26,7 +26,7 @@
             - if conference.registration_open?
               - if conference.user_registered?(current_user)
                 = link_to "My Registration", conference_conference_registrations_path(conference.short_title), :class =>"btn btn-default"
-              - else
+              - elsif not conference.registration_limit_exceeded?
```
ChrisBr: Please use `!conference.registration_limit_exceeded?` instead of `not`
http://ruby-doc.org/core-2.1.3/doc/syntax/precedence_rdoc.html

ChrisBr: I'd also add an else that there are no places left or something similiar!

app/models/conference.rb
```diff
@@ -95,9 +95,11 @@ def confirmed
 
   validates_uniqueness_of :short_title
   validates_format_of :short_title, with: /\A[a-zA-Z0-9_-]*\z/
+  validates :registration_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
```
ChrisBr: Can you also please add a test it's not possible to set registration limit < 0?

app/models/conference.rb
```diff
@@ -95,9 +95,11 @@ def confirmed
...
 
   # This validation is needed since a conference with a start date greater than the end date is not possible
   validate :valid_date_range?
+  validate :registration_limit_not_exceed
```
ChrisBr: Can you also add a test for this validation? Something like this

```
subject.registration_limit = 0

reg = subject.registraton.new

expect(reg.errors.full_messages).to eq(['Registration limit exceeded'])
```

davidsf: I have added the tests. I need to change the comparision in registrations_limit_exceeded? to >= instead of == so the validation to work.

ChrisBr: Aha, and now you see that the validation belongs to the registration and not to the conference! Because when the limit is reached you shouldn't be able to create a registration / the registration should be invalid. With your test you're able to create the registration though it isn't valid, instead the conference is now invalid. 

So, can you please move the validation to the registration and test this instead.

Otherwise, it looks good to me! Really good idea and implementation, thanks :+1:

ChrisBr:
Ah, and can you please rebase (you have the changes from your other PR in  app/views/conference/_registration.html.haml) and squash your commits together into one commit? Thanks :)

differentreality:
Don't you think the message on the home page is a bit vague/confusing as the registration button is no longer there at all (not even as disabled)?

screenshot-osem - chromium-6

app/controllers/conference_registrations_controller.rb
```diff
@@ -20,6 +20,11 @@ def new
...
+      redirect_to root_path, alert: 'Registration limit exceed'
```
differentreality: Shall we make the message a bit more specific?
"Sorry, registration limit exceeded for #{@conference.title}."

davidsf: Yes, that message it's be better.

davidsf:
@differentreality In the homepage as I see it's the first time that show a message instead of a button, so I put the message the shorter without the button. 

Maybe the disabled button with the message below should be more clear.

app/models/conference.rb
davidsf: I have moved the validation to registration model. It seems that there are not a clear solution of this type of validations, I have used one that works and it seems relative clear and it's tested.

davidsf:
@ChrisBr I'm not to sure how to do the rebase and put all in one commit.  Can you help me a bit ? Thanks :)

differentreality:
Please first fix the styling errors shown in travis. Then you can do    git rebase origin/master, you will then have to do a force push: git push -f

davidsf:
Squashed all commits in one.

app/views/admin/conference/edit.html.haml
```diff
@@ -18,5 +18,6 @@
         = f.input :timezone, :as => :time_zone, :hint => "The conference time zone"
         = f.input :start_date, :as => :string, :input_html => { :id => "conference-start-datepicker", :readonly => "readonly"  }
         = f.input :end_date, :as => :string, :input_html => { :id => "conference-end-datepicker", :readonly => "readonly"  }
+      = f.inputs :name => "Registrations" do
+        = f.input :registration_limit, :as => :number, :in => 0..9999, :hint => "Limit the number of registrations to the conference (0 no limit)"
```
differentreality: Can you please use the new ruby syntax?
http://breakthebit.org/post/8453341914/ruby-19-and-the-new-hash-syntax

app/views/conference/_registration.html.haml
```diff
@@ -6,19 +6,25 @@
...
-          Going to
-          = @conference.short_title
-          is free of charge.
+
+      - if @conference.registration_limit_exceeded?
         %p
-          We only ask you to register yourself until
-          = @conference.registration_period.end_date.strftime('%A, %B %-d. %Y')
-          so we can plan for the right amount of people.
+          Sorry, the conference registration limit has exceed
```
differentreality: Oops little typo here exceed -> exceeded

app/views/conference/_conference_details.html.haml
```diff
@@ -26,8 +26,12 @@
             - if conference.registration_open?
               - if conference.user_registered?(current_user)
                 = link_to "My Registration", conference_conference_registrations_path(conference.short_title), :class =>"btn btn-default"
-              - else
+              - elsif !conference.registration_limit_exceeded?
                 = link_to "Register", new_conference_conference_registrations_path(conference.short_title), :class =>"btn btn-default"
```
differentreality: You could simply add at the end of this line:

disabled: registration_limit_exceeded?

And move the if statement on the next line so that you show the message, if needed.

spec/models/registration_spec.rb
```diff
@@ -0,0 +1,26 @@
+#!/bin/env ruby
+# encoding: utf-8
+require 'spec_helper'
+require 'ahoy'
```
differentreality: Why do we need to require ahoy in this particular test?

differentreality:
Good work @davidsf ! 
My new comments are about styling, not functionality. It would be awesome if you would apply the styling as well, before we merge the feature.

davidsf:
@differentreality  Done all that you commented. There are files with a mix of old a new ruby syntax, maybe a good task for another ticket :)

differentreality:
@davidsf Indeed it is :) Would be nice to have everything in the new syntax.

Thank you the feature!

ChrisBr:
@differentreality what's left here?

@davidsf sorry, wasn't very responsive the last week ... :(

differentreality:
@ChrisBr nothing.

Sorry for the delay. Thank you for the code @davidsf ;)

spec/models/registration_spec.rb
```diff
@@ -0,0 +1,25 @@
+#!/bin/env ruby
+# encoding: utf-8
```
sonalkr132: Why this?

PR #765, opened 2016-02-08
Travis: Enable IRC notification, disable addons

hennevogel:


differentreality:
:+1:  in #osem irc channel plz  :-)

PR #768, opened 2016-02-08
Update rubocop todo file

hennevogel:


.rubocop_todo.yml
```diff
@@ -1,137 +1,476 @@
...
+# `rubocop --auto-gen-config`
+# on 2016-02-08 14:42:45 +0100 using RuboCop version 0.37.0.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 3
+# Offense count: 2
+Lint/DuplicatedKey:
```
differentreality: Could you please elaborate on what this does?

.rubocop_todo.yml
```diff
@@ -1,137 +1,476 @@
...
+  Exclude:
+    - 'spec/support/save_feature_failures.rb'
+
+# Offense count: 13
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles, Include.
+# SupportedStyles: action, filter
+# Include: app/controllers/**/*.rb
+Rails/ActionFilter:
+  Exclude:
```
differentreality: Why are we excluding the following ones?

.rubocop_todo.yml
hennevogel: What exactly? The cop?

.rubocop_todo.yml
hennevogel: Because they have offenses? The general idea of the rubocop_todo.yml is that you record which offenses you have where and then you can either explicitly disable or enable the cops in your rubocop.yml and remove the setting in rubocop_todo.yml

.rubocop_todo.yml
differentreality: Yeap, is it for sth like this line?
https://github.com/openSUSE/osem/blob/master/db/migrate/20140801164901_move_conference_media_to_commercial.rb#L13

.rubocop_todo.yml
differentreality: Ah so all exluded files are the ones that do have the offense and when we will enable the cop we will *only* have to fix offenses in these particular files? 
After that they won't remain as excluded but will be fixed?

hennevogel: Yes, once you move the cops setting over to rubocop.yml you fix all the offenses. Then you don't have the need to exclude anything anymore...

.rubocop_todo.yml
```diff
@@ -1,137 +1,476 @@
...
+    - 'app/models/conference.rb'
+
+# Offense count: 84
+Metrics/AbcSize:
+  Max: 75
+
+# Offense count: 7
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 514
```
differentreality: For the record, that was previously 524 in style/ClassLength

app/models/conference.rb
```diff
@@ -1,6 +1,5 @@
 ##
 # This class represents a conference
-# rubocop:disable Style/ClassLength
 class Conference < ActiveRecord::Base
```
differentreality: Do we need to keep that with the new name `Metrics/ClassLength` ?

differentreality:
:+1:

.rubocop_todo.yml
hennevogel: Yes exactly. rubocop has a nice [default config file](https://github.com/bbatsov/rubocop/blob/5023b234cc3f6657b4f9798914e08923a0a72f8d/config/enabled.yml#L972) with descriptions for everything...

app/models/conference.rb
hennevogel: I would want to try to make sure we're not making this larger. Or if we have to, increase the limit in the config file.

PR #769, opened 2016-02-08
Gems: Update MySQL2, switch assets to rails-assets.org

hennevogel:


davidsf:
In https://github.com/rails-assets/rails-assets/issues/291 says that rails-assets.org will shutdown by the end of this year. There are some forks but the future is not really clear for the project.

hennevogel:
The end of this year is very far away :-)

differentreality:
:+1:

PR #772, opened 2016-02-08
Fix room_params in scheduler update

### LINKED ISSUE #771: Can't schedule any event
In current master, it not possible to schedule an event. When you drop the event, the ajax call throws an error 500:

 NoMethodError in Admin::SchedulesController#update
  undefined method `permit' for "o6dXt50iZ1cpueI7QH5EOA":String
  Extracted source (around line #68):
  params.require(:room).permit(:guid)

davidsf:
Fix room_params in scheduler update. fix #771

### DIFF
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -69,7 +69,7 @@ module Admin
     end
 
     def room_params
-      params.require(:room).permit(:guid)
+      params.require(:room)
     end
   end
 end


differentreality:
It might not be clear at first glance why this is true based on our current setup. 
The parameters passed from schedule do not include any attribute for the models. They just give the guid straight away (see server log below).

Perhaps a thing to look into during refactoring of schedule? So that parameters are actually formed like the following? @hennevogel ?

`room: { guid: 'myguid' }`

In that form we are able to properly allow parameters in controller.

Server log:

`Started PUT "/admin/conference/osc15/schedule" for ::1 at 2016-02-09 00:14:12 +0200
Processing by Admin::SchedulesController#update as JSON
  Parameters: {"event"=>"TUxlBLoPZHSgtdhX-jWQ9w", "room"=>"lpakGFP148ar3Lq_VrmIZQ", "date"=>"2015-05-01", "time"=>"9:15", "conference_id"=>"osc15"}`

hennevogel:
Maybe this will make sense in the refactoring, I don't know yet. This fix is needed anyway for now :-) Thanks @davidsf :bouquet:

PR #774, opened 2016-02-11
Put the config name as title in layouts.

davidsf:


ChrisBr:
@davidsf good idea! Thanks :+1:

PR #778, opened 2016-02-13
Add more images to venue

### LINKED ISSUE #777: Commercial for the venue
It would be great if in the venue section the user can add more images, not just one.

davidsf:
Add venue_photos model and venue_photos admin controller in order to add more images to Venues and allow to delete it from the venue form.

Show the additional photos in splashscreen. Added fancybox gem for zoom when click on the additional photo.

fix #777

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -127,6 +127,9 @@ gem 'bootstrap-switch-rails', '~> 3.0.0'
 
 gem 'ruby-oembed'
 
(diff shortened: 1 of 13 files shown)


hennevogel:
The idea is nice but we are against uploading too many files into OSEM. There are tons of ways to include image slide shows, that's why we have the Commercial model. This should get used instead.

PR #779, opened 2016-02-16
Update spring-commands-rspec and springify bin/rspec

sonalkr132:
Running test was really slow, hence I wanted to added spring. Turns out, you guys already have it there, just no one uses it (I guess?).

**NOTE:** use `spring rspec` to run tests with spring.


ChrisBr:
Because it's default in rails 4.1 ...
I don't mind, @hennevogel what do you think?

sonalkr132:
I have added `bin/*` to rubocop cause it contains code we didn't write. Also, rubocop would never shut up about following line (asking us to make it `==`, when spring intended to use `=`):
```Ruby
if match = Bundler.default_lockfile.read.match(/^GEM$.*?^    spring \((.*?)\)$.*?^$/m)
```

hennevogel:
Let's try it :-)

PR #781, opened 2016-02-16
If the conference has only one day, not put the day twice in splashpage

davidsf:
If the conference has only one day, not put the day twice in splashpage


hennevogel:
Awesome! Please also adopt the inline documentation for this method :-)

davidsf:
@hennevogel ah, ok. Done :)

hennevogel:
Thanks @davidsf :bouquet:

PR #783, opened 2016-02-17
Fix error 500 in user proposal confirmation

### LINKED ISSUE #782: Error 500 in proposal confirmation
When user confirms a proposal, the web throw an error 500 with log:

```
NameError (undefined local variable or method `conference' for #<Event:0x0000000ab11110>):
  app/models/event.rb:114:in `process_confirmation'
  app/controllers/proposal_controller.rb:97:in `confirm'
` ``

davidsf:
Fix error 500 in user proposal confirmation. fix #782


differentreality:
:+1:

PR #784, opened 2016-02-18
Send notif only for build on canonical repo and enable build on all branches

sonalkr132:
There is no point of sending notification for build on forks.
Also, a contributor would expect that travis builds all the branches. Usual workflow is that they create a new branch for the feature/fix they are working on.


sonalkr132:
Sorry, limit notification only to main repo doesn't exist on travis (yet!)  :cry:

hennevogel:
I wouldn't want to have it anyway :-)

PR #786, opened 2016-02-19
Speed up test boot time

sonalkr132:
This significantly improves start time of tests. It is also the recommended
way: https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#linting-factories

Read more at: thoughtbot/factory_girl#772

lint was going out to internet to check for links validity. I got following error from lint when I was offline:
> The following factories are invalid: (FactoryGirl::InvalidFactoryError)
* event_commercial - Validation failed: Network is unreachable - connect(2) for "www.youtube.com" port 80 (ActiveRecord::RecordInvalid)

For me it brings down time of running `rspec  spec/controllers/conference_controller_spec.rb` form 30+ sec to 4.33 sec (without spring in both cases)

EDIT:
- [x]  Add the rake lint task
- [x] Make it possible to disable running FactoryGirl.lint 
- [x] Mock the external request

### DIFF
--- a/spec/support/external_request.rb
+++ b/spec/support/external_request.rb
@@ -1,0 +1,29 @@
+# Mock external requests to youtube
+require 'webmock/rspec'
+WebMock.disable_net_connect!(allow_localhost: true)
+
+RSpec.configure do |config|
+  config.before(:each) do
+    mock_commercial_request
+  end
+end
+
+def mock_commercial_request
+  response = {
+    author_name: 'Confreaks',
+    html: '<iframe width=\'560\' height=\'315\' src=\'https://www.youtube.com/embed/BTTygyxuGj8?feature=oembed\' frameborder=\'0\' allowfullscreen></iframe>',
+    thumbnail_width: 480,
+    thumbnail_url: 'https://i.ytimg.com/vi/BTTygyxuGj8/hqdefault.jpg',
+    provider_name: 'YouTube',
+    width: 459,
+    type: 'video',
+    provider_url: 'http://www.youtube.com/',
+    version: '1.0',
+    thumbnail_height: 360,
+    title: 'RailsConf 2014 - Closing Keynote by Aaron Patterson',
+    author_url: 'https://www.youtube.com/user/Confreaks',
+    height: 344
+  }
+  WebMock.stub_request(:get, /.*youtube.*/).
+    to_return(status: 200, body: response.to_json, headers: {})
+end
--- a/Gemfile
+++ b/Gemfile
@@ -179,6 +179,8 @@ group :test do
   # Extracted from RSpec 3 stub_model and mock_model
   gem 'rspec-activemodel-mocks'
   gem 'timecop'
+  # Mock external requests
+  gem 'webmock'
 end
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -130,6 +130,8 @@ GEM
       simplecov (>= 0.7)
       term-ansicolor
       thor
+    crack (0.4.2)
+      safe_yaml (~> 1.0.0)
     currencies (0.4.2)
     daemons (1.1.9)
     database_cleaner (1.3.0)
@@ -409,6 +411,7 @@ GEM
     ruby-openid (2.5.0)
     ruby-progressbar (1.7.5)
     rubyzip (1.0.0)
+    safe_yaml (1.0.4)
     sass (3.2.19)
     sass-rails (4.0.4)
       railties (>= 4.0.0, < 5.0)
@@ -464,6 +467,9 @@ GEM
       binding_of_caller (>= 0.7.2)
       railties (>= 4.0)
       sprockets-rails (>= 2.0, < 4.0)
+    webmock (1.20.4)
+      addressable (>= 2.3.6)
+      crack (>= 0.3.2)
     websocket-driver (0.3.3)
     whenever (0.9.2)
       activesupport (>= 2.3.4)
@@ -553,6 +559,7 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   web-console (~> 2.0)
+  webmock
   whenever
 
 BUNDLED WITH
--- a/config/config.yml.example
+++ b/config/config.yml.example
@@ -15,6 +15,10 @@ defaults: &defaults
     ichain:
       enabled: false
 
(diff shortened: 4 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -51,12 +51,6 @@
   #     --seed 1234
   config.order = 'random'
 
-  # Setting up DB cleaning to maintain empty rows
```
hennevogel: Erm what has this todo with factory girl linting?

hennevogel:
And in general, if we have a Factory that goes out to the internet for linting we need to fix this factory, not work around it. @ChrisBr?

spec/spec_helper.rb
sonalkr132: Nothing really. I was going through spec_helper.rb and found this, hence I removed it as well. Should I send a different PR with just this change?

sonalkr132:
> if we have a Factory that goes out to the internet for linting we need to fix this factory, not work around it

It is not a work around of lint going to internet. I mentioned it as a reason of why test setup was slow. This PR doesn't fix that issue It introduces recommended way of setting up lint and speeds up running tests generally.
yes, we will eventually need to fix lint going to internet as well. Rspec has way of blocking urls when running feature specs, I think we can do somethin similar for fix it.

spec/spec_helper.rb
hennevogel: I guess in the light of my overall comment you can forget about this...

hennevogel:
If linting is fast we don't need to split it from running the test suite...

hennevogel:
I guess the most valuable thing would be stubbing this one request with [webmock](https://github.com/bblimke/webmock).

hennevogel:
My recommendation:

1. [webmock](https://github.com/bblimke/webmock) the commercial request
2. Add the rake lint task
3. Make it possible to disable running FactoryGirl.lint in spec/support/factory_girl.rb with an ENV variable

sonalkr132:
@hennevogel I think you are suggesting something like:

```Ruby
RSpec.configure do |config|
   ENV[OSM_FACTORY_LINT] || = true
   config.before(:suite) do
      FactoryGirl.lint if  ENV[OSM_FACTORY_LINT] 
   end
 end
```

Instead of an ENV, we can add this setting to `config.yml.example`? What do you think?

spec/spec_helper.rb
hennevogel: Alright this looks good now except this change. Please get rid of it.

sonalkr132:
@hennevogel I am mocking all the requests individually. I don't think I should have to do that. Please checkout this question: http://stackoverflow.com/questions/35535617/factorygirl-used-with-webmock

I couldn't really find any documentation on mocking factories.

spec/support/external_request.rb
```diff
@@ -0,0 +1,25 @@
+RSpec.configure do |config|
+  config.before(:suite) do
+    mock_request
+  end
+end
+
+def mock_request
```
ChrisBr: I'd name this more explicit, this does not only mock_request, this mocks the requests going to youtube / the commercial requests.

ChrisBr:
The factory goes to the internet because we check in the validation if the commercial is valid:
https://github.com/openSUSE/osem/blob/master/app/models/commercial.rb#L23

The only way to not go to the internet would be to skip the validation in the test. However, I also think webmock is the way to go.

In general, slow tests are bad but I wouldn't spent to much time in this to make it a few seconds faster. If it's now 25 seconds faster (as you stated in your PR), great. If it's 5 seconds faster (as you stated in your blog, by the way, a little bit polemic), I wouldn't spent to much time in it...

sonalkr132:
> . If it's now 25 seconds faster (as you stated in your PR), great. If it's 5 seconds faster (as you stated in your blog, by the way, a little bit polemic)

Here, I am saying running test is appox 25 seconds faster (for me) cause now validation is using the webmock and not making a real request. Figure I have reported here largely depends on speed of internet connection.

Time I have mentioned in my blog has nothing much to do with this PR other than the fact that I take 5 seconds as lower limit for test runs (for single file).

>  I wouldn't spend to much time in it

I think even 5 seconds gain matters a lot. I want to work on test coverage. I will be running single tests multiple times and saving just those 5 seconds matter a lot to me. I shouldn't have to wait for 5 seconds when the actual test is running just for a couple of milliseconds.

Please feel free to take my branch for a spin. Set `OSEM_FACTORY_LINT` = false in your config.yml and run a single test file (for ex: `rspec spec/controllers/conference_controller_spec.rb`). Run the same test on current master, I hope you will be find this PR worth the effort.

spec/spec_helper.rb
```diff
@@ -22,6 +22,10 @@
 require 'capybara/poltergeist'
 require 'phantomjs'
 
+# Mock exernal requests
```
hennevogel: This belongs to `spec/support/webmock`

spec/factories/commercials.rb
```diff
@@ -11,5 +11,7 @@
     factory :event_commercial do
       association :commercialable, factory: :event
     end
+
+    after(:build) { mock_commercial_request }
```
hennevogel: Why is this needed here and the other 3 places? Why is one time not enough?

sonalkr132: Checkout [this comment](https://github.com/openSUSE/osem/pull/786#issuecomment-187099047).

sonalkr132:
@hennevogel I have removed `extra mock_commercial_request` call following this: https://robots.thoughtbot.com/how-to-stub-external-services-in-tests

ChrisBr:
@hennevogel what do you think? Let's give it a try?

sonalkr132:
I think we should set `factory_girl_lint` to false as default, since we have added rake task to travis.

Right now, it is true and lint will be run both as rake task and then as before suit.

.travis.yml
```diff
@@ -20,5 +20,6 @@ before_script:
   - cp config/secrets.yml.example config/secrets.yml
   - RAILS_ENV=test bundle exec rake db:migrate --trace
 script:
+  - 'bundle exec rake factory_girl:lint'
```
hennevogel: Please drop this

sonalkr132: done!

hennevogel:
:+1: good work @sonalkr132

Issue #791, opened 2016-02-24
Users can register even when registration period has ended


sonalkr132:
Steps to reproduce:

* Create an event
* Set its registration period which is already over (say 2016-02-08 to 2016-02-08)
* Sign out and sign in as different user
* Go to `conference/<event_short_title>/register`
* Register to event

I think beside adding validation to `Registration` in relation to `RegistrationPeriod` we need to add a before filter to Registration controller as well.

Related: #788

PR #795, opened 2016-02-26
Fix authorization of registration controller and add tests

sonalkr132:
Test of new action serves as test of before_filter :registration_period_inclusion.
Change in proposal spec cause conference has to have valid registration
period for successful redirect to registration after confirm proposal.

Closes: #791

I feel like there is no need of adding validation to model as requests with invalid registration objects (wrt to registration period) won't ever reach registration model. Should I add it anyway?

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -1,7 +1,7 @@
 class ConferenceRegistrationsController < ApplicationController
   before_filter :authenticate_user!, except: [:new, :create]
   load_resource :conference, find_by: :short_title
-  authorize_resource :conference_registrations, class: Registration
+  authorize_resource :conference_registrations, class: Registration, except: [:new, :create]
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
@@ -25,8 +25,9 @@ class ConferenceRegistrationsController < ApplicationController
       return
     end
 
-    @registration = Registration.new
-
+    @registration = Registration.new(conference_id: @conference.id)
+    # make sure that conference is open for registration
+    authorize! :new, @registration
     # @user variable needs to be set so that _sign_up_form_embedded works properly
     @user = @registration.build_user
   end
@@ -50,6 +51,7 @@ class ConferenceRegistrationsController < ApplicationController
     end
 
     @registration.user = @user
+    authorize! :create, @registration
 
     if @registration.save
       # Trigger ahoy event
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -5,6 +5,9 @@ class ConferenceRegistrationsController < ApplicationController
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
+    @registration = Registration.new(conference_id: @conference.id)
+    authorize! :new, @registration
+
     # Redirect to registration edit when user is already registered
     if @conference.user_registered?(current_user)
       redirect_to edit_conference_conference_registrations_path(@conference.short_title)
@@ -15,19 +18,11 @@ class ConferenceRegistrationsController < ApplicationController
       return
     end
 
-    # avoid openid sign_in to redirect to register/new when the sign_in user had already a registration
-    if current_user && @conference.user_registered?(current_user)
-      redirect_to edit_conference_conference_registrations_path(@conference.short_title)
-    end
-
     if @conference.registration_limit_exceeded?
       redirect_to root_path, alert: "Sorry, registration limit exceeded for #{@conference.title}"
       return
     end
 
-    @registration = Registration.new(conference_id: @conference.id)
-    # make sure that conference is open for registration
-    authorize! :new, @registration
     # @user variable needs to be set so that _sign_up_form_embedded works properly
     @user = @registration.build_user
   end
@@ -91,9 +86,9 @@ class ConferenceRegistrationsController < ApplicationController
       redirect_to root_path,
                   notice: "You are not registered for #{@conference.title} anymore!"
     else
-      redirect_to root_path,
-                  error: "Could not update your registration for #{@conference.title}: "\
-                  "#{@registration.errors.full_messages.join('. ')}."
+      redirect_to conference_conference_registrations_path(@conference.short_title),
+                  flash: { error: "Could not delete your registration for #{@conference.title}: "\
+                  "#{@registration.errors.full_messages.join('. ')}." }
     end
   end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -46,8 +46,12 @@ class Ability
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can :show, User
     unless CONFIG['authentication']['ichain']['enabled']
-      can [:show, :create], Registration do |registration|
+      can :show, Registration do |registration|
         registration.new_record?
+      end
+
+      can [:new, :create], Registration do |registration|
+        registration.conference.registration_open? && registration.new_record?
       end
       can :show, Event do |event|
         event.new_record?
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -72,6 +72,10 @@ class Ability
 
     can :manage, Registration, user_id: user.id
 
+    can [:new, :create], Registration do |registration|
+      registration.conference.registration_open?
+    end
+
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
 
--- a/spec/controllers/conference_registration_controller_spec.rb
+++ b/spec/controllers/conference_registration_controller_spec.rb
@@ -1,0 +1,284 @@
+require 'spec_helper'
+
+describe ConferenceRegistrationsController, type: :controller do
+  let!(:first_user) { create(:user) }
+  let(:conference) { create(:conference) }
+  let(:conference_with_open_registration) { create(:conference) }
+  let!(:open_registration_period) { create(:registration_period, conference: conference_with_open_registration, start_date: Date.current - 6.days) }
+  let(:user) { create(:user) }
+
+  context 'user is not signed in' do
+    describe 'GET #new' do
+      context 'ichain is enabled' do
+        before do
+          CONFIG['authentication']['ichain']['enabled'] = true
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        after { CONFIG['authentication']['ichain']['enabled'] = false }
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+      end
+    end
+
+    describe 'POST #create' do
+      before do
+        post :create, user: attributes_for(:user),
+                      registration: attributes_for(:registration),
+                      conference_id: conference_with_open_registration.short_title
+      end
+
+      it 'assigns user variable' do
+        expect(assigns(:user)).not_to be_nil
+      end
+
+      it 'signs in registration user' do
+        expect(controller.current_user).not_to be_nil
+      end
+
+      it 'shows success message in flash notice' do
+        expect(flash[:notice]).to match('You are now registered and will be receiving E-Mail notifications.')
+      end
+
+      it 'redirects to registration show path' do
+        expect(response).to redirect_to conference_conference_registrations_path(conference_with_open_registration.short_title)
+      end
+
+      it 'creates a new registration' do
+        expect(Registration.count).to eq 1
+      end
+    end
+  end
+
+  context 'user is signed in' do
+    before { sign_in(user) }
+
+    describe 'GET #new' do
+      context 'user is registered to conference' do
+        before do
+          create(:registration, user: user, conference: conference_with_open_registration)
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to edit registration page' do
+          expect(response).to redirect_to edit_conference_conference_registrations_path(conference_with_open_registration.short_title)
+        end
+      end
+
+      context 'registration limit has reached' do
+        before do
+          conference_with_open_registration.update_attributes(registration_limit: 1)
+          create(:registration, conference: conference_with_open_registration)
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:alert]).to match "Sorry, registration limit exceeded for #{conference_with_open_registration.title}"
+        end
+      end
+
+      context 'successful request' do
+        before do
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns registration and user variables' do
+          expect(assigns(:registration)).to be_instance_of(Registration)
+          expect(assigns(:user)).to be_instance_of(User)
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+      end
+    end
+
+    describe 'GET #show' do
+      before do
+        @registration = create(:registration, conference: conference_with_open_registration, user: user)
+      end
+
+      context 'user has purchased a ticket' do
+        before do
+          @ticket = create(:ticket, conference: conference_with_open_registration)
+          @purchased_ticket = create(:ticket_purchase, conference: conference_with_open_registration,
+                                                       user: user,
+                                                       ticket: @ticket)
+          get :show, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'renders the show template' do
+          expect(response).to render_template('show')
+        end
+
+        it 'assigns workshops variable' do
+          expect(assigns(:workshops)).to eq @registration.workshops
+        end
+
+        it 'assigns price of purchased tickets to total_price and purchased tickets to tickets' do
+          expect(assigns(:total_price)).to eq Money.new(10000, 'USD')
+          expect(assigns(:tickets)).to match_array [@purchased_ticket]
+        end
+      end
+
+      context 'user has not purchased any ticket' do
+        before do
+          get :show, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns 0 dollars to total_price and empty array to tickets variables' do
+          expect(assigns(:total_price)).to eq Money.new(0, 'USD')
+          expect(assigns(:tickets)).to match_array []
+        end
+      end
+    end
+
+    describe 'GET #edit' do
+      before do
+        create(:registration, conference: conference_with_open_registration, user: user)
+        get :edit, conference_id: conference_with_open_registration.short_title
+      end
+
+      it 'renders the edit template' do
+        expect(response).to render_template('edit')
+      end
+    end
+
+    describe 'POST #create' do
+      context "tickets are available and user hasn't bought any" do
+        before do
+          create(:ticket, conference: conference_with_open_registration)
+          post :create, registration: attributes_for(:registration),
+                        conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns user variable' do
+          expect(assigns(:user)).to eq user
+        end
+
+        it 'redirects to conference tickets path' do
+          expect(response).to redirect_to conference_tickets_path(conference_with_open_registration.short_title)
+        end
+
+        it 'creates a new registration' do
+          expect(Registration.count).to eq 1
+        end
+      end
+
+      context 'registration save fails' do
+        before do
+          allow_any_instance_of(Registration).to receive(:save).and_return(false)
+          post :create, registration: attributes_for(:registration),
+                        conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+
+        it 'does not create registration' do
+          expect(Registration.count).to eq 0
+        end
+      end
+    end
+
+    describe 'PATCH #update' do
+      before do
+        @registration = create(:registration,
+                               conference: conference_with_open_registration,
+                               user: user,
+                               arrival: Date.new(2014, 04, 25))
+      end
+
+      context 'updates successfully' do
+        before do
+          patch :update, registration: attributes_for(:registration, arrival: Date.new(2014, 04, 29)),
+                         conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to registration show path' do
+          expect(response).to redirect_to conference_conference_registrations_path(conference_with_open_registration.short_title)
+        end
+
+        it 'shows success message in flash notice' do
+          expect(flash[:notice]).to match('Registration was successfully updated.')
+        end
+
+        it 'updates the registration' do
+          @registration.reload
+          expect(@registration.arrival).to eq Date.new(2014, 04, 29)
+        end
+      end
+
+      context 'update fails' do
+        before do
+          allow_any_instance_of(Registration).to receive(:update_attributes).and_return(false)
+          patch :update, registration: attributes_for(:registration, arrival: Date.new(2014, 04, 27)),
+                         conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'renders edit template' do
+          expect(response).to render_template('edit')
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:error]).to include 'Could not update your registration'
+        end
+
+        it 'does not update the registration' do
+          @registration.reload
+          expect(@registration.arrival).to eq Date.new(2014, 04, 25)
+        end
+      end
+    end
+
+    describe 'DELETE #destroy' do
+      before do
+        @registration = create(:registration, conference: conference_with_open_registration, user: user)
+      end
+
+      context 'deletes successfully' do
+        before do
+          delete :destroy, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+
+        it 'shows success message in flash notice' do
+          expect(flash[:notice]).to match("You are not registered for #{conference_with_open_registration.title} anymore!")
+        end
+
+        it 'deletes the registration' do
+          expect(Registration.count).to eq 0
+        end
+      end
+
+      context 'delete fails' do
+        before do
+          allow_any_instance_of(Registration).to receive(:destroy).and_return(false)
+          delete :destroy, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to registration show path' do
+          expect(response).to redirect_to conference_conference_registrations_path(conference_with_open_registration.short_title)
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:error]).to include 'Could not delete your registration'
+        end
+
+        it 'does not delete the registration' do
+          expect(Registration.last).to eq @registration
+        end
+      end
+    end
+  end
+end
--- a/spec/controllers/conference_registration_controller_spec.rb
+++ b/spec/controllers/conference_registration_controller_spec.rb
@@ -9,6 +9,22 @@ describe ConferenceRegistrationsController, type: :controller do
 
   context 'user is not signed in' do
     describe 'GET #new' do
+      context 'ichain is disabled' do
+        before do
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns conference, registration and user variable' do
+          expect(assigns(:conference)).to eq conference_with_open_registration
+          expect(assigns(:registration)).to be_instance_of Registration
+          expect(assigns(:user)).to be_instance_of User
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+      end
+
       context 'ichain is enabled' do
         before do
           CONFIG['authentication']['ichain']['enabled'] = true
@@ -19,6 +35,10 @@ describe ConferenceRegistrationsController, type: :controller do
 
         it 'redirects to root path' do
           expect(response).to redirect_to root_path
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:alert]).to match 'You are not authorized to access this page. Maybe you need to sign in?'
         end
       end
     end
@@ -56,6 +76,21 @@ describe ConferenceRegistrationsController, type: :controller do
     before { sign_in(user) }
 
     describe 'GET #new' do
+      context 'successful request' do
+        before do
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns registration and user variables' do
+          expect(assigns(:registration)).to be_instance_of(Registration)
+          expect(assigns(:user)).to be_instance_of(User)
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+      end
+
       context 'user is registered to conference' do
         before do
           create(:registration, user: user, conference: conference_with_open_registration)
@@ -82,26 +117,27 @@ describe ConferenceRegistrationsController, type: :controller do
           expect(flash[:alert]).to match "Sorry, registration limit exceeded for #{conference_with_open_registration.title}"
         end
       end
-
-      context 'successful request' do
-        before do
-          get :new, conference_id: conference_with_open_registration.short_title
-        end
-
-        it 'assigns registration and user variables' do
-          expect(assigns(:registration)).to be_instance_of(Registration)
-          expect(assigns(:user)).to be_instance_of(User)
-        end
-
-        it 'renders the new template' do
-          expect(response).to render_template('new')
-        end
-      end
     end
 
     describe 'GET #show' do
       before do
         @registration = create(:registration, conference: conference_with_open_registration, user: user)
+      end
+
+      context 'successful request' do
+        before do
+          get :show, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'assigns conference, registration and workshops variables' do
+          expect(assigns(:conference)).to eq conference_with_open_registration
+          expect(assigns(:registration)).to eq @registration
+          expect(assigns(:workshops)).to eq @registration.workshops
+        end
+
+        it 'renders the show template' do
+          expect(response).to render_template('show')
+        end
       end
 
       context 'user has purchased a ticket' do
@@ -111,14 +147,6 @@ describe ConferenceRegistrationsController, type: :controller do
                                                        user: user,
                                                        ticket: @ticket)
           get :show, conference_id: conference_with_open_registration.short_title
-        end
-
-        it 'renders the show template' do
-          expect(response).to render_template('show')
-        end
-
-        it 'assigns workshops variable' do
-          expect(assigns(:workshops)).to eq @registration.workshops
         end
 
         it 'assigns price of purchased tickets to total_price and purchased tickets to tickets' do
@@ -141,8 +169,13 @@ describe ConferenceRegistrationsController, type: :controller do
 
     describe 'GET #edit' do
       before do
-        create(:registration, conference: conference_with_open_registration, user: user)
+        @registration = create(:registration, conference: conference_with_open_registration, user: user)
         get :edit, conference_id: conference_with_open_registration.short_title
+      end
+
+      it 'assigns conference and registration variable' do
+        expect(assigns(:conference)).to eq conference_with_open_registration
+        expect(assigns(:registration)).to eq @registration
       end
 
       it 'renders the edit template' do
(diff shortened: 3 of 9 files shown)


differentreality:
I feel like we should have proper authorization for this.

sonalkr132:
@differentreality I am sorry but I don't understand what do you mean by *proper authorization*. Can you please explain?

differentreality:
We are using cancancan for authorization https://github.com/CanCanCommunity/cancancan

I believe that we should properly define when the user can access the Registration model in models/ability.rb, so that the action **new** is not accessible when we are not within the registration period.

sonalkr132:
Thank you. I will make the changes. I think it makes more sense too.
Shouldn't we limit **create** as well?

differentreality:
I think so

sonalkr132:
I have updated my PR with suggested change.

app/controllers/conference_registrations_controller.rb
```diff
@@ -1,10 +1,11 @@
 class ConferenceRegistrationsController < ApplicationController
   before_filter :authenticate_user!, except: [:new, :create]
   load_resource :conference, find_by: :short_title
-  authorize_resource :conference_registrations, class: Registration
+  authorize_resource :conference_registrations, class: Registration, except: [:new, :create]
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
+    authorize! :new, Registration.new(conference_id: @conference.id)
```
differentreality: At the end of the action we define @registration, perhaps we should simply authorize that?

app/controllers/conference_registrations_controller.rb
```diff
@@ -40,6 +41,7 @@ def show
   def edit; end
 
   def create
+    authorize! :create, Registration.new(conference_id: @conference.id)
     @registration = @conference.registrations.new(registration_params)
```
differentreality: **create** action also needs to authorize that the registration has a user.

differentreality:
You should also add relevant tests to https://github.com/openSUSE/osem/blob/master/spec/features/conference_registration_spec.rb

In addition, in your tests I think you should follow the natural logic:
_I create a conference, then I create a registration_period for that conference, etc._
Even though you might be achieving the same end result in the tests, a few months from now, I think, it will be difficult to read.

app/controllers/conference_registrations_controller.rb
sonalkr132: hey, doesn't validation of registration for [presence of user](https://github.com/openSUSE/osem/blob/master/app/models/registration.rb#L26) cover us here?

sonalkr132:
@differentreality I have updated my PR. Please let me know if you can think of more test scenarios.

Regarding:
> create action also needs to authorize that the registration has a user.

I haven't added auth for presence of user, cause I think it is covered in validation of registration. I hope this is what you meant.

app/models/ability.rb
```diff
@@ -50,8 +50,10 @@ def not_signed_in
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can :show, User
     unless CONFIG['authentication']['ichain']['enabled']
-      can [:show, :create], Registration do |registration|
-        registration.new_record?
+      can :show, Registration
+
```
differentreality: You need to maintain the previous condition of `registration.new_record?` in the new abilities you define.

app/controllers/conference_registrations_controller.rb
```diff
@@ -1,10 +1,14 @@
 class ConferenceRegistrationsController < ApplicationController
   before_filter :authenticate_user!, except: [:new, :create]
   load_resource :conference, find_by: :short_title
-  authorize_resource :conference_registrations, class: Registration
+  authorize_resource :conference_registrations, class: Registration, except: [:new, :create]
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
+    @registration = Registration.new(conference_id: @conference.id)
+    # make sure that conference is open for registration
```
differentreality: That is not exactly what the following line does. The following does checkes the authorization of @registration, it checks if there is proper access to the action 'new' for @registration. Checking if the registration of the conference is open, is one of the defined abilities but there are / there can be more. So I think the comment is a bit misleading.

app/controllers/conference_registrations_controller.rb
```diff
@@ -89,9 +87,9 @@ def destroy
       redirect_to root_path,
                   notice: "You are not registered for #{@conference.title} anymore!"
     else
-      redirect_to root_path,
-                  error: "Could not update your registration for #{@conference.title}: "\
-                  "#{@registration.errors.full_messages.join('. ')}."
+      redirect_to conference_conference_registrations_path(@conference.short_title),
```
differentreality: Please add a feature tests for this change.

spec/features/proposal_spec.rb
```diff
@@ -1,7 +1,8 @@
 require 'spec_helper'
 
 feature Event do
-  let!(:conference) { create(:conference) }
+  let(:registration_period) { create(:registration_period, start_date: Date.current) }
+  let!(:conference) { create(:conference, registration_period: registration_period) }
```
differentreality: Can you please define conference before you define the registration_period? It would be more clear what you are doing, because the logic is: I create a conference, then I create a registration_period for this conference.

spec/models/ability_spec.rb
```diff
@@ -35,6 +35,11 @@
 
     # Test abilities for not signed in users
     context 'when user is not signed in' do
+      let(:open_registration_period) { create(:registration_period, start_date: Date.current - 6.days) }
+      let(:open_registration_conf) { create(:conference, registration_period: open_registration_period) }
```
differentreality: I would suggest a variable name that better indicates what it is about. Eg
`let(:conference_with_open_registration)`

Same for the following.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
+require 'spec_helper'
+
+describe ConferenceRegistrationsController, type: :controller do
+  let(:conference) { create(:conference) }
+  let(:open_registration_period) { create(:registration_period, start_date: Date.current - 6.days) }
+  let(:open_registration_conf) { create(:conference, registration_period: open_registration_period) }
+  let(:user) { create(:user) }
```
differentreality: This user is an admin.

spec/factories/registration.rb
```diff
@@ -3,6 +3,7 @@
 FactoryGirl.define do
   factory :registration do
     user
-    conference
+    conference { create(:conference, registration_period: create(:registration_period, start_date: 3.days.ago)) }
```
differentreality: I don't like this much. I feel that the details regarding the conference and its registration_period need to be explicitely and visibly defined in each test.
@hennevogel  @ChrisBr ?

app/models/ability.rb
sonalkr132: I have restored the check for `new` and `create` method, however I don't understand why should we checking `new_record?` for `show` method? I think it just shouldn't exist here as we can't show registration unless user is registered and signed in.

spec/factories/registration.rb
sonalkr132: My reasoning behind adding this here was that a registration object can't be created unless the conference is open. It is kinda implicit that if we are creating a registration we meant to do it for an open conference.

Related thought:
Adding a *validation* to `registration` model for `conference` being open is not a bad idea in this situation (where we were able to create invalid registrations for closed conferences). I didn't add the validation cause things work just fine without it, except in tests. In tests we can potentially create registration, which would be invalid. Hence, to avoid adding validation to registration model which we would have never used I added this here. I guess best practice would be to add the validation anyway?
Let me know what do you guys think.

sonalkr132:
@differentreality I have made most of the changes you suggested and left a few comments. Thank for you review!

app/controllers/conference_registrations_controller.rb
sonalkr132: done!

differentreality:
We should probably provision for confirmed speakers to be able to register even after the registration period (they will be there anyway).
All that might collide with registration_limit, we should address it (in a new PR?).

sonalkr132:
> We should probably provision for confirmed speakers to be able to register even after the registration period (they will be there anyway)

Yeah, it makes sense. It definitely should be a new PR. However, in this one I do want to fix `:show` of registration. Please take a look at [this comment.](https://github.com/openSUSE/osem/pull/795#discussion_r54533346)

ChrisBr:
Please rebase to fix Hakiri issues.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
+require 'spec_helper'
+
+describe ConferenceRegistrationsController, type: :controller do
+  let(:conference) { create(:conference) }
+  let(:conference_with_open_registration) { create(:conference, registration_period: open_registration_period) }
+  let(:open_registration_period) { create(:registration_period, start_date: Date.current - 6.days) }
```
ChrisBr: Switch these two lines as you use :open_registration_period to create the :conference_with_open_registration

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+      end
+    end
+
+    describe 'GET #show' do
+      before do
```
ChrisBr: I'd add a context with_ticket and one without_ticket purchased.

app/controllers/conference_registrations_controller.rb
ChrisBr: Why do you redirect to conference_conference_registrations_path? conference_conference_registration_path would be better from my POV...

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,255 @@
...
+      end
+
+      context 'updates successfully' do
+        before do
+          patch :update, registration: attributes_for(:registration, arrival: Date.new(2014, 04, 29)),
+                         conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to registration show path' do
+          expect(response).to redirect_to conference_conference_registrations_path(conference_with_open_registration.short_title)
```
ChrisBr: conference_conference_registration_path

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: > Can you please define conference before you define the registration_period? It would be more clear what you are doing, because the logic is: I create a conference, then I create a registration_period for this conference. 
The same applies for other tests.
@differentreality 

[outdated diff](https://github.com/openSUSE/osem/pull/795#discussion_r54487416) :cry:

spec/factories/registration.rb
ChrisBr: I don't have a strong preference here ...
@hennevogel

spec/features/conference_registration_spec.rb
```diff
@@ -49,5 +65,16 @@
         expect(conference.user_registered?(participant)).to be(true)
       end
     end
+
+    context 'registration is closed' do
+      let(:conference_with_closed_registration) { create(:conference, registration_period: create(:registration_period)) }
+
+      scenario 'registers for a conference', feature: true do
+        participant.is_admin = false
```
ChrisBr: Why is participant an admin? Is it needed in the other tests? If not, I'd prefer to change this and remove the line!

spec/features/proposal_spec.rb
```diff
@@ -127,6 +128,7 @@
     end
 
     scenario 'confirms a proposal', feature: true, js: true do
+      registration_period # opens conference for registration
```
ChrisBr: ``
let!(:registration_period) { create(:registration_period, conference: conference, start_date: Date.current) }
``
(note the exclamation mark) does not work?

spec/features/conference_registration_spec.rb
sonalkr132: Chekout #799 I think it is know issue that first user is admin in text env too. You guys have normally created a `first_user` in other specs to get around this.

I have to use it cause, even `create(:user, is_admin: false)` make no difference :sweat:

sonalkr132: So it is either this or, I make a user to throw away like other specs. I think we shouldn't have to do either.. hence the issue.

differentreality: In the beginning you can do 
let!(:first_user) { create(:user) }

So that the rest of the code does not make us wonder why.

app/controllers/conference_registrations_controller.rb
sonalkr132: > Why do you redirect to conference_conference_registrations_path?

on GET request `conference_conference_registrations_path` goes to show action.
```
conference_conference_registrations_path	    POST	/conference/:conference_id/register(.:format)	conference_registrations#create
new_conference_conference_registrations_path	GET	/conference/:conference_id/register/new(.:format)	conference_registrations#new
edit_conference_conference_registrations_path	GET	/conference/:conference_id/register/edit(.:format)	conference_registrations#edit
                                                GET	/conference/:conference_id/register(.:format)	conference_registrations#show
                                                PATCH	/conference/:conference_id/register(.:format)	conference_registrations#update
                                                PUT	/conference/:conference_id/register(.:format)	conference_registrations#update
                                                DELETE	/conference/:conference_id/register(.:format)	conference_registrations#destroy
```
**show** method:
```Ruby
def show
    @workshops = @registration.workshops
    @total_price = Ticket.total_price(@conference, current_user)
    @tickets = current_user.ticket_purchases.where(conference_id: @conference.id)
  end
```
I think we need to change the name of `def show` to `def index`?

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: done!

spec/features/proposal_spec.rb
sonalkr132: Thank you. I have changed it.

app/models/ability.rb
differentreality: I don't quite recall myself why we allowed show method for new records of Registration and Event. You could open an issue about it, so that we can investigate it in more depth.
Let's keep it as it was till we do.

spec/factories/registration.rb
differentreality: We can limit access to Registration with authorization, which also allows us to be flexible with organizers/admins, whom we may want to allow to create a registration after the deadline is over. Should that be enough without a Registration validation?

differentreality: I would still like to see registration_period being created and properly defined in each test, and not in the factory.
Note: The factories of conference and registration_period currently have the right set of date ranges.

app/controllers/conference_registrations_controller.rb
sonalkr132: ping!

differentreality: This is a singular resource (http://guides.rubyonrails.org/routing.html, 2.5 Singular Resources):

```
A singular resourceful route generates these helpers:

new_geocoder_path returns /geocoder/new
edit_geocoder_path returns /geocoder/edit
geocoder_path returns /geocoder
```

So `conference_conference_registration_path` for **show** action.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+          create(:registration, user: user, conference: conference_with_open_registration)
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to edit registration page' do
+          expect(response).to redirect_to edit_conference_conference_registrations_path(conference_with_open_registration.short_title)
+        end
+      end
+
+      context 'ichain is enabled and user is not signed in' do
```
differentreality: We are supposed to be in the context 'user is signed in'

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+        it 'renders the show template' do
+          expect(response).to render_template('show')
+        end
+
+        it 'assigns workshops variable' do
+          expect(assigns(:workshops)).to eq @registration.workshops
+        end
+
+        it 'assigns price of purchased tickets to total_price and purchased tickets to tickets' do
+          expect(assigns(:total_price)).to eq @ticket.price * @purchased_ticket.quantity
```
differentreality: Make it explicit. Instead of the multiplication, simply put the right number here.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,276 @@
...
+          expect(flash[:notice]).to match('Registration was successfully updated.')
+        end
+
+        it 'updates the registration' do
+          @registration.reload
+          expect(@registration.arrival).to eq Date.new(2014, 04, 29)
+        end
+      end
+
+      context 'update fails' do
```
differentreality: Aren't you going to check if the arrival attribute has been set or not?

spec/features/conference_registration_spec.rb
```diff
@@ -49,5 +66,15 @@
...
       end
     end
+
+    context 'registration is closed' do
+      let(:conference_with_closed_registration) { create(:conference, registration_period: create(:registration_period)) }
+
+      scenario 'registers for a conference', feature: true do
+        visit new_conference_conference_registrations_path(conference_with_closed_registration.short_title)
+        expect(current_path).to eq(root_path)
+        expect(flash).to eq 'You are not authorized to access this page.'
```
differentreality: Why is this the message?

spec/features/proposal_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 feature Event do
   let!(:conference) { create(:conference) }
+  let!(:registration_period) { create(:registration_period, conference: conference, start_date: Date.current) }
```
differentreality: Why do we need that?

spec/models/registration_spec.rb
```diff
@@ -1,24 +1,70 @@
...
+    it { should have_and_belong_to_many(:qanswers) }
+    it { should have_and_belong_to_many(:vchoices) }
+    it { should have_many(:events_registrations) }
+    it { should have_many(:workshops) }
+  end
+
+  describe 'callbacks' do
+    after { subject.run_callbacks(:create) }
+
+    it 'sets week of registration' do
```
differentreality: Why are you not testing the functionality?

spec/features/conference_registration_spec.rb
sonalkr132: > Why is this the message?

When registration is closed, request for new registration get caught by authorization of those requests. Thats why flash message is about denied access. I think we can customize error message, how closed registration is not the only time auth would fail. I guess we can be more generic:
```Ruby
authorize! :new, @registration, message: "You can't register to this conference"
```

spec/features/proposal_spec.rb
sonalkr132: Project confirmation gets redirected to new registration page. We need conference to be open for registration for the redirect.

We won't need it, after we have solved #802

spec/models/registration_spec.rb
sonalkr132: Callbacks are private methods (Testing them breaks encapsulation). Also, `subscribe_to_conference` and `send_registration_mail` are just delegating responsibility to  `Subscription` and `Mailbot` classes. So, `Subscription` and `Mailbot` classes should make sure that it does what is was asked to (ie we need to test it those classes).

Do you want me to test them regardless?

spec/features/proposal_spec.rb
differentreality: :+1: you are right, the relevant test should also check if the current_path is correct.

spec/features/conference_registration_spec.rb
differentreality: Right, in that case why do we still have code
https://github.com/sonalkr132/osem/blob/registration_period/app/controllers/conference_registrations_controller.rb#L22 ?

app/controllers/conference_registrations_controller.rb
differentreality: Upon IRC discussion, our routes define `resource :conference_registrations` in plural.  We leave the route as it is, and we will change the whole route to singular in a different PR.

spec/features/conference_registration_spec.rb
sonalkr132: On IRC discussion we decided to leave things as it is for now.

sonalkr132:
@differentreality I have made changes in 6a1cdbe. Please let me know if I missed something.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+        get :edit, conference_id: conference_with_open_registration.short_title
+      end
+
+      it 'renders the edit template' do
+        expect(response).to render_template('edit')
+      end
+    end
+
+    describe 'POST #create' do
+      context "tickets are avilable and user hasn't bought any" do
```
differentreality: available typo

differentreality: You need to test if a registration is actually created

sonalkr132: The path (39-41-43-48-49-51) which leads to successful registration was already covered in [test for POST #create for unsigned user](https://github.com/openSUSE/osem/pull/795/files#diff-f594354d23f55f0c31fc2d9fbd49250eR25). I have added it here too tho.

spec/features/proposal_spec.rb
differentreality: Are you going to add the test ??

spec/models/registration_spec.rb
```diff
@@ -1,24 +1,79 @@
...
+      it 'sets week of registration' do
+        expect(subject).to receive(:set_week)
+        expect(subject.week).to eq 18
+      end
+    end
+
+    describe '#subscribe_to_conference' do
+      it 'subscribes to conference' do
+        expect(subject).to receive(:subscribe_to_conference)
+        expect(subject.user.subscribed?(subject.conference)).to be_truthy
```
differentreality: user.subscribed? returns either true or false. We want this to be true. Explicitely, not truthy.

sonalkr132: `be_true` is same as `be_truthy`

https://github.com/rspec/rspec-expectations/issues/283

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,283 @@
+require 'spec_helper'
+
+describe ConferenceRegistrationsController, type: :controller do
+  let(:conference) { create(:conference) }
+  let(:conference_with_open_registration) { create(:conference) }
+  let!(:open_registration_period) { create(:registration_period, conference: conference_with_open_registration, start_date: Date.current - 6.days) }
+  let(:user) { create(:user) }
```
differentreality: What about creating a first user before that one?

spec/models/registration_spec.rb
differentreality: @sonalkr132 this value needs to calculate to either true or false. We need to test if it is true or false. We should not test if it is not false or nil.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+describe ConferenceRegistrationsController, type: :controller do
+  let!(:first_user) { create(:user) }
+  let(:conference) { create(:conference) }
+  let(:conference_with_open_registration) { create(:conference) }
+  let!(:open_registration_period) { create(:registration_period, conference: conference_with_open_registration, start_date: Date.current - 6.days) }
+  let(:user) { create(:user) }
+
+  context 'user is not signed in' do
+    describe 'GET #new' do
+      context 'ichain is enabled' do
```
differentreality: Do we need to cover what happens for `GET #new` when the user is not signed in and ichain is not enabled?

sonalkr132: yes! For [following else](https://github.com/openSUSE/osem/blob/master/app/controllers/conference_registrations_controller.rb#L13-L16):
```Ruby
 # Redirect to registration edit when user is already registered
    if @conference.user_registered?(current_user)
      redirect_to edit_conference_conference_registrations_path(@conference.short_title)
      return
    # ichain does not allow us to create users during registration
    elsif CONFIG['authentication']['ichain']['enabled'] && !current_user
      redirect_to root_path, alert: 'You need to sign in or sign up before continuing.'
      return
    end
```

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+        expect(Registration.count).to eq 1
+      end
+    end
+  end
+
+  context 'user is signed in' do
+    before { sign_in(user) }
+
+    describe 'GET #new' do
+      context 'user is registered to conference' do
```
differentreality: And if not?

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+        get :edit, conference_id: conference_with_open_registration.short_title
+      end
+
+      it 'renders the edit template' do
+        expect(response).to render_template('edit')
+      end
+    end
+
+    describe 'POST #create' do
+      context "tickets are available and user hasn't bought any" do
```
differentreality: The context and the tests in it don't match a lot. IMO the successful creation of a registration happens anyway, so it should not be tested inside this context.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+        it 'assigns user variable' do
+          expect(assigns(:user)).to eq user
+        end
+
+        it 'redirects to conference tickets path' do
+          expect(response).to redirect_to conference_tickets_path(conference_with_open_registration.short_title)
+        end
+
+        it 'creates a new registration' do
+          expect(Registration.count).to eq 1
```
differentreality: expected.to change { Registration.count }.by 1

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+      context 'user has purchased a ticket' do
+        before do
+          @ticket = create(:ticket, conference: conference_with_open_registration)
+          @purchased_ticket = create(:ticket_purchase, conference: conference_with_open_registration,
+                                                       user: user,
+                                                       ticket: @ticket)
+          get :show, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'renders the show template' do
```
differentreality: This should not be inside this context.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+        end
+
+        it 'assigns 0 dollars to total_price and empty array to tickets variables' do
+          expect(assigns(:total_price)).to eq Money.new(0, 'USD')
+          expect(assigns(:tickets)).to match_array []
+        end
+      end
+    end
+
+    describe 'GET #edit' do
```
differentreality: You haven't tested if it tries to edit the right object.

spec/controllers/conference_registration_controller_spec.rb
differentreality: What if there are no tickets?

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+          patch :update, registration: attributes_for(:registration, arrival: Date.new(2014, 04, 27)),
+                         conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'renders edit template' do
+          expect(response).to render_template('edit')
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:error]).to include 'Could not update your registration'
```
differentreality: Your error message must include this part of the code too: `for #{@conference.title}:`

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+        before do
+          delete :destroy, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+
+        it 'shows success message in flash notice' do
+          expect(flash[:notice]).to match("You are not registered for #{conference_with_open_registration.title} anymore!")
```
differentreality: Test that conference title explicitely!

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,284 @@
...
+          allow_any_instance_of(Registration).to receive(:destroy).and_return(false)
+          delete :destroy, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to registration show path' do
+          expect(response).to redirect_to conference_conference_registrations_path(conference_with_open_registration.short_title)
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:error]).to include 'Could not delete your registration'
```
differentreality: Complete message.

spec/features/conference_registration_spec.rb
```diff
@@ -28,13 +29,29 @@
...
+          click_link 'My Registration'
+        end
+
+        scenario 'deletion of registration is successful' do
+          expect(current_path).to eq(conference_conference_registrations_path(conference.short_title))
+
+          click_link 'Unregister'
+          expect(conference.user_registered?(participant)).to be(false)
+          expect(current_path).to eq root_path
+          expect(flash).to eq "You are not registered for #{conference.title} anymore!"
```
differentreality: Explicit conference title please :-)

app/controllers/conference_registrations_controller.rb
```diff
@@ -15,18 +18,11 @@ def new
       return
     end
 
-    # avoid openid sign_in to redirect to register/new when the sign_in user had already a registration
-    if current_user && @conference.user_registered?(current_user)
-      redirect_to edit_conference_conference_registrations_path(@conference.short_title)
-    end
-
     if @conference.registration_limit_exceeded?
```
differentreality: Do we still need that?

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: So psuedo code would be `new` would be something like this:
```
if user is registered
   # do something
elsif ichain is enabled and user is not signed in
  # do something else
end

if user is signed in and registerd
   # do something
end

if registration limit has reached
   # do something
end

# do something if none of the above if blocks already redirect request 
```
So the way things are set up, there really isn't a context for `user is not registered`. [context](https://github.com/openSUSE/osem/pull/795/files#diff-f594354d23f55f0c31fc2d9fbd49250eR86) of `successful request` probably the closet thing.

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: > when the user is not signed in and ichain is not enabled?

We aren't dealing with that case in `new` action.

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: [done](https://github.com/openSUSE/osem/pull/795/files#diff-f594354d23f55f0c31fc2d9fbd49250eR234)!

spec/features/proposal_spec.rb
sonalkr132: [done](https://github.com/openSUSE/osem/pull/795/files#diff-8519356ff4df1e8926f17ba7f3ccf7e5R138)!

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: I think you are suggesting that it is a common thing for both `tickets available` and `tickets not available`. How do I introduce this common context tho?

spec/controllers/conference_registration_controller_spec.rb
differentreality: Why not?

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: [done](https://github.com/openSUSE/osem/pull/795/files#diff-f594354d23f55f0c31fc2d9fbd49250eR172)!

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,331 @@
...
+        it 'assigns conference, registration and user variable' do
+          expect(assigns(:conference)).to eq conference_with_open_registration
+          expect(assigns(:registration)).to be_instance_of Registration
+          expect(assigns(:user)).to be_instance_of User
+        end
+
+        it 'renders the new template' do
+          expect(response).to render_template('new')
+        end
+      end
```
differentreality: flash message?

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,331 @@
...
+          CONFIG['authentication']['ichain']['enabled'] = true
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        after { CONFIG['authentication']['ichain']['enabled'] = false }
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+      end
```
differentreality: flash message?

app/models/ability.rb
```diff
@@ -46,9 +46,13 @@ def not_signed_in
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can :show, User
     unless CONFIG['authentication']['ichain']['enabled']
-      can [:show, :create], Registration do |registration|
+      can :show, Registration do |registration|
         registration.new_record?
       end
+
+      can [:new, :create], Registration do |registration|
+        registration.conference.registration_open? && registration.new_record?
```
differentreality: What about a signed_in user? Are we going to limit access to new and create actions, unless registration is open?

differentreality:
Please spend some time to go through your test suite and make sure that you are covering all test cases and that your tests cover all the new code you have written.

sonalkr132:
@differentreality, I have added rule for sign_in user. Thank you pointing it out.
I think I have tested the changes I made in codebase:
* change in `def destory` redirect path: [Feature test](https://github.com/sonalkr132/osem/blob/registration_period/spec/features/conference_registration_spec.rb#L47-L54), [controller test](https://github.com/sonalkr132/osem/blob/registration_period/spec/controllers/conference_registration_controller_spec.rb#L315-L333)
* Modification in `:new` and `:create` authorization rule for registration: [Ability](https://github.com/sonalkr132/osem/blob/registration_period/spec/models/ability_spec.rb#L64-L67) [test](https://github.com/sonalkr132/osem/blob/registration_period/spec/models/ability_spec.rb#L91-L92)

spec/features/conference_registration_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'spec_helper'
 
 feature Registration do
+  let!(:first_user) { create(:user) } # first user is admin
   let!(:conference) { create(:conference, registration_period: create(:registration_period, start_date: 3.days.ago)) }
```
differentreality: No longer needed

differentreality:
For the models tests let's do it like this:
https://github.com/openSUSE/osem/blob/master/spec/models/event_spec.rb

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+          expect(response).to redirect_to root_path
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:alert]).to match 'You are not authorized to access this page. Maybe you need to sign in?'
+        end
+      end
+    end
+
+    describe 'POST #create' do
```
differentreality: What happens when it fails?

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+
+        it 'creates a new registration' do
+          expect do
+            post :create, registration: attributes_for(:registration),
+                          conference_id: conference_with_open_registration.short_title
+          end.to change{ Registration.count }.by 1
+        end
+      end
+
+      context 'tickets are available' do
```
differentreality: There is one more case:
tickets available, user registers to conference, user buys ticket, user unregisters for conference (ticket is still bought. Note: we don't have a confirmation message for unregistering), user registers again for the conference

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+          create(:registration, conference: conference_with_open_registration)
+          get :new, conference_id: conference_with_open_registration.short_title
+        end
+
+        it 'redirects to root path' do
+          expect(response).to redirect_to root_path
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:alert]).to match "Sorry, registration limit exceeded for #{conference_with_open_registration.title}"
```
differentreality: explicit conference title

differentreality:
The controller test should be checking that the registration in `new` action belongs to the right conference and in `create` action that it belongs to the right conference and it is for the right user.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+
+      it 'renders the edit template' do
+        expect(response).to render_template('edit')
+      end
+    end
+
+    describe 'POST #create' do
+      context 'successfully registers' do
+        it 'assigns user variable' do
+          post :create, registration: attributes_for(:registration),
```
differentreality: Put this in a `before` block

spec/controllers/conference_registration_controller_spec.rb
sonalkr132: @differentreality I have added it in context of `signed in user` [here](https://github.com/openSUSE/osem/pull/795/files#diff-f594354d23f55f0c31fc2d9fbd49250eR225). Do you want me to add the same thing here too?

differentreality: yes

sonalkr132:
Will send smaller PRs. Closing this.

PR #796, opened 2016-02-27
Add users#show view spec

differentreality:


ChrisBr:
@differentreality @hennevogel do we really want to continue with view specs? From my POV I think view specs are overhead, they often break, you often have to adapt them ... I'd really focus on model and feature tests (and probably controller tests)....

hennevogel:
Yeah I would also like to only address unti tests (rspec speak: Model + Controller tests) and Workflows (rspec speak: Feature tests)

PR #801, opened 2016-02-29
Model and controller tests for Target

sonalkr132:


app/controllers/admin/targets_controller.rb
```diff
@@ -39,9 +37,9 @@ def destroy
         redirect_to(admin_conference_targets_path(conference_id: @conference.short_title),
                     notice: 'Target successfully destroyed.')
       else
-        redirect_to(admin_conference_targets_path(conference_id: @conference.short_title),
-                    error: 'Target was successfully destroyed.' \
-                    "#{@target.errors.full_messages.join('. ')}.")
+        redirect_to admin_conference_targets_path(conference_id: @conference.short_title),
+                    flash:  { error: "Could not delete target for #{@conference.title}: "\
+                    "#{@target.errors.full_messages.join('. ')}." }
```
differentreality: Why do we need flash?

app/models/target.rb
```diff
@@ -34,8 +34,6 @@ def get_progress
         conference.registrations.where('created_at < ?', due_date).count
       when Target.units[:program_minutes]
         conference.current_program_minutes
-      else
-        0
```
differentreality: Not sure you should remove that! @ChrisBr ?

spec/factories/targets.rb
```diff
@@ -6,5 +6,6 @@
     target_count 100
     unit Target.units[:submissions]
     conference
+    campaign
   end
```
differentreality: Creating a target does not require you to have a campaign, so I don't think we should have that in the factory.

app/controllers/admin/targets_controller.rb
sonalkr132: By default rails only provides `:notice` and `:alert`. For everything else we need to use the general purpose flash. (Source: [Rails API#Flash](http://guides.rubyonrails.org/action_controller_overview.html#the-flash))

app/models/target.rb
sonalkr132: > Not sure you should remove that!

A simple answer would be that I couldn't write test for that case, hence I removed it. If we can't write test for something, we shouldn't have it in our code.

More detailed one:
We have `switch` statement over `unit`. `unit` can never have values other than first three cases as we validate `unit` for inclusion in:
```Ruby

  def self.units
    {
      registrations: 'Registration',
      submissions: 'Submission',
      program_minutes: 'Program minute'
    }
  end

 validates :unit, allow_nil: false, inclusion: { in: Target.units.values }
```

spec/factories/targets.rb
sonalkr132: > Creating a target does not require you to have a campaign

Yes, it doesn't. However, I needed campaign to write tests for following lines:
```Ruby
    if unit == Target.units[:submissions]
      numerator = campaign.submissions_count
    elsif unit == Target.units[:registrations]
      numerator = campaign.registrations_count
    # more branch
    end
```

I suppose another way of achieving the same would be something like:
```Ruby
 registration_target = { create(:target, target_count: 10, unit: Target.units[:registrations]) }
 campaign = create(:campaign, conference:registration_target.conference )
 campaign.targets = registration_target
```
I think adding campaign to factory is cleaner. Let me know what do you prefer and I will change it.

app/models/target.rb
ChrisBr: sounds valid!

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+    describe 'GET #new' do
+      before { get :new, conference_id: conference.short_title }
+
+      it 'renders new template' do
+        expect(response).to render_template('new')
+      end
+
+      it 'assigns registration and user variables' do
+        expect(assigns(:target)).to be_instance_of(Target)
```
ChrisBr: The test name does not match the actual test ...

sonalkr132: sorry :sweat_smile: Fixed it!

app/controllers/admin/targets_controller.rb
differentreality: Yes, you are right, I thought we had implemented that already. Why don't we add error flash type instead? So that our redirect is cleaner and looks like the ones with alert or notice?
http://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html#method-i-add_flash_types

spec/factories/targets.rb
ChrisBr: Basically I'm with @differentreality, you shouldn't add fields / attributes to the factory which are not necessary to get a valid factory. However, cleaner code is also a valid point. How about [inheritance](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#inheritance)?

differentreality: For target factory we should keep only necessary attributes/relationships.
However, you can introduce a new factory, eg target_full which will have more relationships, and you can work on that. Check https://github.com/openSUSE/osem/blob/master/spec/factories/conferences.rb

ChrisBr:
Otherwise LGTM :+1: 
(Please rebase to get rid of the Hakiri issues)

app/models/target.rb
differentreality: @ChrisBr I think it would be a better practice to keep the `else`. In case something goes wrong with our validations, we won't end up with nil value, but we will have 0 instead.

spec/factories/targets.rb
sonalkr132: Thank you. I have used `campaign_target` as name of new factory. Let me know if you can think of better name.

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+describe Admin::TargetsController, type: :controller do
+  let(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let(:target) { create(:target, conference: conference) }
+
+  context 'user is admin' do
+    before { sign_in admin }
+
+    describe 'GET #index' do
```
differentreality: You could also add a test that checks if it assigns the right conference value.

differentreality: You should also check if @targets have the proper value.

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+      end
+    end
+
+    describe 'POST #create' do
+      context 'saves successfuly' do
+        before do
+          post :create, target: attributes_for(:target), conference_id: conference.short_title
+        end
+
+        it 'redirects to admin target show path' do
```
differentreality: It redirects to index path

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+          expect(flash[:error]).to match("Creating target failed: #{target.errors.full_messages.join('. ')}.")
+        end
+
+        it 'does not create new target' do
+          expect(Target.count).to eq 0
+        end
+      end
+    end
+
+    describe 'GET #edit' do
```
differentreality: You should also check that it tries to edit the right target object.

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+        it 'renders edit template' do
+          expect(response).to render_template('edit')
+        end
+
+        it 'shows error in flash message' do
+          expect(flash[:error]).to match("Target update failed: #{target.errors.full_messages.join('. ')}.")
+        end
+
+        it 'does not update target' do
+          expect(target.target_count).not_to eq 2
```
differentreality: Please make that explicit.
You can define the target_count in your let() definition, so that we explicitely now the original value, and then we can check if the target.target_count still holds its original value.

spec/models/target_spec.rb
```diff
@@ -32,74 +36,95 @@
...
     it 'returns zero if there are no registrations' do
-      conference = build(:conference)
-      target = build(:target, target_count: 10, unit: Target.units[:registrations])
-      conference.targets = [target]
-
-      expect(target.get_progress).to eq('0')
+      expect(registration_target.get_progress).to eq('0')
     end
 
     it 'returns 10 if there one registrations of 10' do
```
differentreality: Since you are at it, can we also properly rewrite this text?

spec/models/target_spec.rb
```diff
@@ -32,74 +36,95 @@
...
-      conference.targets = [target]
-      conference.program.events = [event]
+      expect(program_minutes_target.get_progress).to eq('10')
+    end
+  end
+
+  describe '#get_campaign' do
+    let(:campaign_registration_target) { create(:campaign_target, target_count: 10, unit: Target.units[:registrations]) }
+    let(:campaign_submission_target) { create(:campaign_target, target_count: 10, unit: Target.units[:submissions]) }
+    let(:campaign_program_minutes_target) { create(:campaign_target, target_count: 300, unit: Target.units[:program_minutes]) }
```
differentreality: Not that this is wrong, but since you are not doing anything different with these variables, you could have simply defined `let(:campaign) { create(:campaign }` in the beginning of the file and then associate campaign with registration_target, submission_target and program_minutes_target, and then you would not need to define more objects here.

spec/models/target_spec.rb
```diff
@@ -32,74 +36,95 @@
...
-      event = create(:event)
+      program_minutes_target.conference.program.events = [event]
 
-      conference.targets = [target]
-      conference.program.events = [event]
+      expect(program_minutes_target.get_progress).to eq('10')
+    end
+  end
+
+  describe '#get_campaign' do
```
differentreality: While the progress is the most important value in this hash, the test should be checking the overal functionality of `get_campaign`: if the function actually returns a hash, which is called result and contains all the values it is supposed to.

spec/models/target_spec.rb
```diff
@@ -32,74 +36,112 @@
...
-      expect(target.get_progress).to eq('0')
+      expect(program_minutes_target.get_progress).to eq('10')
     end
+  end
 
-    it 'returns 10 if there one submissions of 10' do
-      conference = create(:conference)
-      target = create(:target, target_count: 10, unit: Target.units[:submissions])
-      event = create(:event)
+  def prepapre_result_hash(target, value, progress)
```
sonalkr132: I didn't want to do this cause we are mostly duplicating the logic/code of model in test.

differentreality: do what?

sonalkr132: sorry, I forgot to quote you:
> While the progress is the most important value in this hash, the test should be checking the overal functionality of get_campaign: if the function actually returns a hash, which is called result and contains all the values it is supposed to.

I have added it here (with prepapre_result_hash function). I didn't want to do it cause it duplicates logic.

spec/models/target_spec.rb
```diff
@@ -32,74 +36,127 @@
...
-      target = create(:target, target_count: 10, unit: Target.units[:submissions])
-      event = create(:event)
+    it 'returns zero if there are no submissions' do
+      expect(submission_target.get_progress).to eq('0')
+    end
 
-      conference.targets = [target]
-      conference.program.events = [event]
+    it 'returns 10 if there is one submission of 10' do
+      submission_target.conference.program.events = [event]
```
differentreality: I would create the event here. This line gets very confusing. 
You can just do `create(:event, program: conference.program)`, you can define the conference in the beginning and create targets of this particular conference. Same for line 81.

app/models/target.rb
ChrisBr: Agreed!

sonalkr132: Do you guys want me to add it back? I would still say that
> In case something goes wrong with our validations, we won't end up with nil value

is not something we should worry about. Cause if we do, we will end up with `if-else` blocks all over the code case. For ex (from user model):
```Ruby
validates :email, presence: true

validates :username, presence: true

def self.find_for_auth(auth, current_user = nil)
  # omitted code//

  if user.new_record?
    if auth.info.email
      user.email = auth.info.email
    else
      # do something else
    end
    user.name = auth.info.name
    if auth.info.username
      user.username = auth.info.username
    else
      # do something else
    end
    user.password = Devise.friendly_token[0, 20]
    user.skip_confirmation!
  end

  user
end
```

sonalkr132:
**context**: test for `get_campaign`
@differentreality I have changed it back to my original implementation. If you can think of dealing with all the date-time(`created_at`, `due_date` and `days_left`: values depend on time tests are run) inconsistency in following hash:
```Ruby
result = {
     # code omitted...
      'created_at' => created_at,
      'progress' => progress,
      'days_left' => days_left,
    }

def days_left
    (due_date - Date.today).to_i
end
```
please let me know.
My opinion is that effort it would take to stub the `days_left` and preset `created_at` and `due_date` is not worth it. We would end up with tests which would be too complicated and we won't be testing much anyway.

app/models/target.rb
differentreality: Yes, leave those 2 lines as they were.

spec/controllers/admin/targets_controller_spec.rb
differentreality: The fact that `target.target_count` does not equal 2, does not necessarily mean that it is equal it is original value. What is the value that `target.target_count` is supposed to have? That is what you should test against.

differentreality:
In my opinion, when testing https://github.com/openSUSE/osem/blob/master/app/models/target.rb#L49, we are testing:
- if it returns a hash
- if the hash has all the keys it should have
- if the values of those keys are correct

Even if you let created_at be implicit, you can still check about the others. 
Eg. If you do `program_minutes_target.due_date = Date.today + 5`, regardless of the actual date of each machine `days_left` will still be 5.

sonalkr132:
@differentreality I have made the changes you asked.

spec/models/target_spec.rb
```diff
@@ -32,74 +35,130 @@
...
+  describe '#get_progress' do
+    it 'returns zero if there are no registrations' do
+      expect(registration_target.get_progress).to eq('0')
     end
 
-    it 'returns zero if there are no submissions' do
-      conference = create(:conference)
-      target = build(:target, target_count: 10, unit: Target.units[:submissions])
-      conference.targets = [target]
+    it 'returns 10 if there is one registration of 10' do
```
differentreality: We discussed on IRC how to phrase this better.
`returns 10, when there is 1 registration and the target is 10`

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+    before { sign_in admin }
+
+    describe 'GET #index' do
+      before { get :index, conference_id: conference.short_title }
+
+      it 'renders index template' do
+        expect(response).to render_template('index')
+      end
+
+      it 'assigns target and conference variable' do
```
differentreality: assigns @targets and @conference variable**s**

The index action uses the variable in plural (@targets).
Both variables come from cancancan authorization in the controller.
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/targets_controller.rb#L3
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/targets_controller.rb#L4

spec/controllers/admin/targets_controller_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+    end
+
+    describe 'GET #edit' do
+      before { get :edit, conference_id: conference.short_title, id: target.id }
+
+      it 'renders edit template' do
+        expect(response).to render_template('edit')
+      end
+
+      it 'edits right target object' do
```
differentreality: It is best if you use the same phrasing as in other similar cases, eg. `assigns @target variable`

spec/models/target_spec.rb
```diff
@@ -32,74 +35,130 @@
...
+  describe '#get_campaign' do
+    context 'submissions' do
+      before do
+        submission_target.campaign = create(:campaign, name: 'Submission Campaign', conference: submission_target.conference)
+        submission_target.created_at = Time.utc(2014, 5, 10)
+        submission_target.due_date = Date.today + 4.days
+        allow(submission_target.campaign).to receive(:submissions_count) { 20 }
+      end
+
+      it 'returns hash has correct key-value pairs' do
```
differentreality: We should phrase that better.

sonalkr132: For example?
Returns a hash with values of the corresponding campaign?

sonalkr132:
updated! Not sure what is going wrong with coverall :confused:

spec/models/target_spec.rb
```diff
@@ -32,74 +35,130 @@
...
+        allow(program_minutes_target.conference).to receive(:current_program_minutes) { 20 }
+      end
+
+      it 'returns a hash with values of the corresponding campaign program minutes' do
+        result = {
+          'target_name' => "300 Program minutes by #{Date.today + 4.days}",
+          'campaign_name' => 'Program Campaign',
+          'value' => 20,
+          'unit' => 'Program minute',
+          'created_at' => Time.utc(2014, 5, 10).in_time_zone,
```
differentreality: @hennevogel is this how we should test the `created_at` value? Or should this be simply implicit:
`'created_at' => program_minutes_target.created_at`

hennevogel: I actually can't remember what this does. @ChrisBr might as this is his code :-)

differentreality:
:+1:

PR #805, opened 2016-03-01
Better description to avoid sign up failures

chashmeetsingh:


differentreality:
I agree with the concept, but I think we should use hint instead of placeholder.

chashmeetsingh:
@differentreality isn't this the best way to express?

differentreality:
Yes

chashmeetsingh:
#813 sorry had deleted the repo by mistake @differentreality

PR #806, opened 2016-03-01
Redirect to root if conference over

### LINKED ISSUE #791: Users can register even when registration period has ended

chashmeetsingh:
Fixes #791

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -5,6 +5,14 @@ class ConferenceRegistrationsController < ApplicationController
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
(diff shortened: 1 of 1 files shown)


differentreality:
@chashmeetsingh sorry we are already working on fixing this in #795 

Take a look at the changes of https://github.com/openSUSE/osem/pull/795/files to see how we use authorization to allow or disallow access to specific actions of a controller.

PR #790, opened 2016-02-22
Add program to event relation in user/show.

davidsf:
Fix error 500. Add program to event relation in user/show. fix #789

PR #812, opened 2016-03-03
Feature test for users#show

differentreality:
Feature test, instead of view test #796 
Catches #790


spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
 require 'spec_helper'
 
 feature User do
+  let(:user) { create(:user, name: 'My Name', biography: 'This is my biography!') }
 
-  shared_examples 'admin ability' do
+  describe 'users#show' do
+    let!(:conference) { create(:conference, title: 'My conference') }
```
ChrisBr: is the exclamation mark necessary?

spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
...
+    let!(:unconfirmed_event) { create(:event, state: 'unconfirmed', title: 'This is my test event title!', program: conference.program) }
```
ChrisBr: same here?

spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
...
 
-  end
+    before :each do
+      unconfirmed_event.event_users = [create(:event_user, user: user, event_role: 'submitter')]
```
ChrisBr: let!(:unconfirmed_event) { create(:event, state: 'unconfirmed', title: 'This is my test event title!', program: conference.program, users: [create(:event_user, user: user, event_role: 'submitter')]) }
does not work?

spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
...
+      another_conference = create(:conference, title: 'My another conference')
+
+      another_event = create(:event, state: 'confirmed', title: 'This is my another test event title!', program: another_conference.program)
+
+      another_event.event_users = [create(:event_user, user: user, event_role: 'submitter')]
 
-  describe 'admin' do
-    it_behaves_like 'admin ability', :admin
+      visit user_path(user)
+      expect(page).to have_selector('h3', text: 'My Name presents 3 Events:', count: 1)
```
ChrisBr: Personally I'm not a big fan of these expect statements ...
(We're doing the same in the old obs tests and we're having a hard time to debug ...)

spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
...
+    let!(:unconfirmed_event) { create(:event, state: 'unconfirmed', title: 'This is my test event title!', program: conference.program) }
 
-  end
+    before :each do
+      unconfirmed_event.event_users = [create(:event_user, user: user, event_role: 'submitter')]
+    end
+
+    it 'shows user information' do
+      visit user_path(user)
+      expect(page).to have_content 'My Name'
```
ChrisBr: Personally I'm not a big fan of these expect statements ...
(We're doing the same in the old obs tests and we're having a hard time to debug ...)

spec/features/user_spec.rb
```diff
@@ -1,12 +1,56 @@
...
+      expect(page).to have_content 'This is my biography!'
+      expect(page).to have_selector('img')
+      expect(page).to_not have_content 'My Name presents'
+    end
+
+    it 'does not show event information for unconfirmed events' do
+      expect(page).to_not have_selector('h3', text: 'My Name presents 1 Event:')
```
ChrisBr: Personally I'm not a big fan of these expect statements ...
(We're doing the same in the old obs tests and we're having a hard time to debug ...)

spec/features/user_spec.rb
ChrisBr: And to be honest, I'm not sure if this is really a "feature"...

I'd prefer to test a feature to sign up / sign in and than you can test these on the way ... But just wrapping a view test into a feature test, I don't like it :/

(But just my POV)

ChrisBr:

differentreality:
@ChrisBr can you elaborate on what you have in mind? What I am trying to do is catch this with a test: https://github.com/openSUSE/osem/pull/790/files

ChrisBr:
For instance just clicking the link (however, the test should already fail when you only access the page due to a nil exception).

Testing several 
```
expect(page).to have_content 'Foo'
expect(page).to have_content 'Bar'
```

Testing single have_content to test that the page renders is fine.

is just moving view test to a feature test. Every time you just change the wording in the view you have to adapt this feature test, not good!

Just my personal opinion, @hennevogel what do you think?

spec/features/user_spec.rb
hennevogel: Me neither. There is no real value in these kinds of tests.

hennevogel:
Yes this is not testing any of our functionality and we should test our functionality not which text we write on some page.

* For Models we should test methods validations and callbacks we have written as [model specs](https://relishapp.com/rspec/rspec-rails/v/3-4/docs/model-specs)
* For Controllers we test methods that we have written as [controller specs](https://relishapp.com/rspec/rspec-rails/v/3-4/docs/controller-specs)
* Our major workflows (things like Sign Up, Creating/Updating a Conference, Submitting a Proposal etc) we test with [feature specs](https://relishapp.com/rspec/rspec-rails/v/3-4/docs/feature-specs)

hennevogel:
If we can agree on the above roles we can close this...

differentreality:
I am updating it with controller tests. However I am not sure how I could catch #790 with a controller test.

PR #820, opened 2016-03-04
update rolify

differentreality:
It updates rolify and cancancan gems to latest version.

Removes `has_and_belongs_to_many :roles` from **User** model.
Makes necessary changes according to that deletion.

*Note: If we need to find all the users that have role x, we need to execute:
`User.with_role x.name, x.resource`


app/controllers/admin/conference_controller.rb
```diff
@@ -217,8 +217,8 @@ def conference_params
     def get_users(role_name)
       @role_users = {}
       # Initialize @role variable, so that view can show the role description
-      @role = Role.where(name: role_name, resource: @conference)
-      @role.blank? ? @role_users[role_name] = @role : @role_users[role_name] = @role.first.users
+      @role = Role.where(name: role_name, resource: @conference) || []
```
hennevogel: Doe the || trigger on an empty relation? I don't think so.

differentreality: Nop, you are right. I was tyring it with `find_by`, when I added the `|| []` part.

config/initializers/rolify.rb
```diff
@@ -4,5 +4,5 @@
 
   # Dynamic shortcuts for User class (user.is_admin? like methods). Default is: false
   # Enable this feature _after_ running rake db:migrate as it relies on the roles table
-  config.use_dynamic_shortcuts
+  # config.use_dynamic_shortcuts
```
hennevogel: What's this disabling? The methods?

db/schema.rb
```diff
@@ -11,307 +11,310 @@
...
 
-ActiveRecord::Schema.define(version: 20160201221411) do
+ActiveRecord::Schema.define(version: 20160226133808) do
 
   create_table "ahoy_events", force: :cascade do |t|
     t.uuid     "visit_id",   limit: 16
-    t.integer  "user_id"
-    t.string   "name"
-    t.text     "properties"
+    t.integer  "user_id",    limit: 4
```
hennevogel: Hm what causes this change? Sounds strange...

db/schema.rb
```diff
@@ -11,307 +11,310 @@
...
 
   create_table "campaigns", force: :cascade do |t|
-    t.integer  "conference_id"
-    t.string   "name"
-    t.string   "utm_source"
-    t.string   "utm_medium"
-    t.string   "utm_term"
-    t.string   "utm_content"
-    t.string   "utm_campaign"
+    t.integer  "conference_id", limit: 4
```
hennevogel: Same here. 4? Sounds very limited.

config/initializers/rolify.rb
differentreality: Yes, which are some extra methods rolify provides, such as:
user.is_organizer_of? Conference.last

But we are using the simple method:
user.has_role? :organizer, Conference.last

db/schema.rb
hennevogel: Looks like this comes from your DB. Can you recreate it from scratch please?

hennevogel: @differentreality?

differentreality:
Even after I run `rake db:setup`,I still see the limit for both mysql and sqlite.

hennevogel:
I guess that's a thing then. :eyes:  Thanks for checking

Gemfile.lock
```diff
@@ -334,8 +334,8 @@ GEM
       railties (= 4.2.5.2)
       sprockets-rails
     rails-assets-date.format (1.2.3)
-    rails-assets-holderjs (2.9.1)
-    rails-assets-jquery (2.2.0)
+    rails-assets-holderjs (2.9.3)
```
sonalkr132: Why did you update holderjs? [2.9.3 is on rubygems.org](https://rubygems.org/gems/holder_rails/versions/2.9.3) but not on [rails-assets.org](https://rails-assets.org/).  May be we should downgrade for now?

PR #822, opened 2016-03-04
ProposalController#create refactor

### LINKED ISSUE #597: Refactor
https://github.com/openSUSE/osem/blob/master/app/controllers/proposal_controller.rb#L24

ProposalController#create to be more simple and understandable. 
Tests should stay green, code should be simplified

chocoelho:
Resolves #597 by using method extraction.

### DIFF
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -148,11 +148,9 @@ class ProposalController < ApplicationController
     @event.event_users.new(user: current_user,
                            event_role: 'speaker')
 
-    unless @event.save
-      flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
-      render action: 'new'
-      return
-    end
+    return unless !@event.save
+    flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
+    render action: 'new'
   end
 end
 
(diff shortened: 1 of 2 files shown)


hennevogel:
What's the value of extracting this code into methods if those methods are only used once?

chocoelho:
It was a way to simplify the create method, user creation makes more sense to be extracted rather than proposal submission. User creation has more chances to be reused than proposal submission. Maybe moving it to UsersController will be better.

chocoelho:
I think user creation should be used as `before_action` only when create from proposal is called.

hennevogel:
You should cross bridges when we you reach them, not before :-) It's a common mistake to program for the 'future' while making the 'present' unpleasant.  

Once you're in need of using the code behind `create_user_if_not_signed_in` or `proposal_submission` twice (some argue 3 times) you can start to think about moving this out.
If you don't have the need, the only thing you do by moving this to methods, is making it harder to understand/read. People who maintain this code (a.k.a. reading it in 2+ years) will have to jump around in the file just to find out what `proposal_submission` *actually* means.

Another thing you should learn about is where things belong to. The only thing that belongs to helpers is code that supports views. Before filters don't.

hennevogel:
I'm closing this, please don't let this discourage you. Keep up the good work! :v:

PR #823, opened 2016-03-04
Some Vagrant fixes

hennevogel:
* The whole boostrap script runs in /vagrant
* Make sure that we don't overwrite existing databases
  even if there is no config for them


ChrisBr:
Cool :+1:

PR #826, opened 2016-03-04
Add controller test for users#show

differentreality:


sonalkr132:
user controller doesn't need to authenticate users? I think we are falling back to authorization for permission `edit` and `update` actions, however shouldn't we authenticate anyway?

differentreality:
@sonalkr132 why are you saying it is not authenticating users?

sonalkr132:
[User controller](https://github.com/openSUSE/osem/blob/master/app/controllers/users_controller.rb#L1) doesn't have following line:
```Ruby
before_filter :authenticate_user!
```

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    it 'assigns [] to @events, when user does not have any submissions' do
+      expect(assigns(:events)).to eq []
+    end
+
+    it 'assigns the correct value to @events, when the user has submissions' do
+      conference = create(:conference)
+      event = create(:event, state: 'confirmed', program: conference.program)
+      event.event_users = [create(:event_user, user: user, event_role: 'submitter')]
```
ChrisBr: ` event.event_users << create(:event_user, user: user, event_role: 'submitter')`
should also work!

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      sign_in user
+      get :edit, id: user.id
+      expect(assigns(:user)).to eq user
+    end
+  end
+
+  describe 'PATCH #update' do
+    context 'with valid attributes' do
+      before :each do
+        sign_in user
```
ChrisBr: Isn't it possible to move the `sign_in user` in a shared before block?

ChrisBr:
Otherwise LGTM :+1:

differentreality:
@sonalkr132 what about line 2?
https://github.com/openSUSE/osem/blob/master/app/controllers/users_controller.rb#L2

sonalkr132:
> @sonalkr132 what about line 2?

That is for authorization not authentication.

spec/controllers/users_controller_spec.rb
differentreality: It won't save us any lines, as I still need the individual before blocks. And since I have the before block anyway, it makes it more clear that I sign_in the user, and we don't have to search for it elsewhere (it is just 1 line).

differentreality:
@ChrisBr take a look

I will squash my commits when we are good to go (don't merge with wip commits plz :-) )

spec/controllers/users_controller_spec.rb
ChrisBr: Ok

ChrisBr:
LGTM, please squash your commits and I'll merge 
:+1:

PR #827, opened 2016-03-05
Feature: Add/change Track for Event

hennevogel:
oSC16 is using special EventTypes to workaround this limitation. I don't see why this shouldn't be possible...

### DIFF
--- a/app/views/proposal/_tooltip.html.haml
+++ b/app/views/proposal/_tooltip.html.haml
@@ -3,33 +3,21 @@
   %li{'class'=>class_for_todo(progress_status['registered'])}
     %span{'class'=>icon_for_todo(progress_status['registered'])}
       - if progress_status['registered']
-        = link_to 'Edit your registration', edit_conference_conference_registrations_path(event.program.conference.short_title)
+        = link_to 'Register to the conference', edit_conference_conference_registrations_path(event.program.conference.short_title)
       - else
         = link_to 'Register to the conference', new_conference_conference_registrations_path(event.program.conference.short_title)
-
   %li{'class'=>class_for_todo(progress_status['biography'])}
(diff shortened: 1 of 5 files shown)


app/views/proposal/_tooltip.html.haml
```diff
@@ -3,33 +3,21 @@
   %li{'class'=>class_for_todo(progress_status['registered'])}
     %span{'class'=>icon_for_todo(progress_status['registered'])}
       - if progress_status['registered']
```
differentreality: Did you mean to remove the if statement too?

app/views/proposal/_tooltip.html.haml
```diff
@@ -3,33 +3,21 @@
...
-        = link_to 'Edit your registration', edit_conference_conference_registrations_path(event.program.conference.short_title)
+        = link_to 'Register to the conference', edit_conference_conference_registrations_path(event.program.conference.short_title)
```
ChrisBr: Why? If `if progress_status['registered']` you're already registered!

ChrisBr:
Otherwise LGTM

app/views/proposal/_tooltip.html.haml
hennevogel: Indeed

hennevogel: Oh no, the path is different. This is correct as it is

app/views/proposal/_tooltip.html.haml
hennevogel: The TODO list content should not change. The TODO is always 'Register to the conference' no matter if it's done or not.

hennevogel:
So can we merge this?

ChrisBr:
Yeah, let's :ship: it ...

PR #837, opened 2016-03-06
Tests for room model and controller

sonalkr132:
I dropped the idea of testing rest of things associated with ticket cause they will be undergo major change soon enough.

### DIFF
--- a/spec/controllers/admin/rooms_controller_spec.rb
+++ b/spec/controllers/admin/rooms_controller_spec.rb
@@ -1,0 +1,171 @@
+require 'spec_helper'
+
+describe Admin::RoomsController do
+  let!(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let!(:venue) { create(:venue, conference: conference) }
+  let(:room) { create(:room, venue: venue, size: 4) }
+
+  context 'admin is signed in' do
+    before { sign_in admin }
+
+    describe 'GET #index' do
+      before { get :index, conference_id: conference.short_title }
+
+      it 'assigns conference, venue and rooms variables' do
+        expect(assigns(:conference)).to eq conference
+        expect(assigns(:venue)).to eq venue
+        expect(assigns(:rooms)).to eq [room]
+      end
+
+      it 'renders index template' do
+        expect(response).to render_template('index')
+      end
+    end
+
+    describe 'GET #edit' do
+      before { get :edit, conference_id: conference.short_title, id: room.id }
+
+      it 'renders edit template' do
+        expect(response).to render_template('edit')
+      end
+
+      it 'assigns room variable' do
+        expect(assigns(:room)).to eq room
+      end
+    end
+
+    describe 'GET #new' do
+      before { get :new, conference_id: conference.short_title }
+
+      it 'renders new template' do
+        expect(response).to render_template('new')
+      end
+
+      it 'assigns room variable' do
+        expect(assigns(:room)).to be_instance_of(Room)
+      end
+    end
+
+    describe 'POST #create' do
+      context 'saves successfuly' do
+        before do
+          post :create, room: attributes_for(:room), conference_id: conference.short_title
+        end
+
+        it 'redirects to admin room index path' do
+          expect(response).to redirect_to admin_conference_venue_rooms_path(conference_id: conference.short_title)
+        end
+
+        it 'shows success message in flash notice' do
+          expect(flash[:notice]).to match('Room successfully created.')
+        end
+
+        it 'creates new room' do
+          expect(Room.count).to eq 1
+        end
+      end
+
+      context 'save fails' do
+        before do
+          allow_any_instance_of(Room).to receive(:save).and_return(false)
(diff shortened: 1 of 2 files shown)


spec/controllers/admin/rooms_controller_spec.rb
```diff
@@ -0,0 +1,171 @@
+require 'spec_helper'
+
+describe Admin::RoomsController do
+  let(:admin) { create(:user) }
```
differentreality: You imply here the user will  be an admin because it is the first user. But you can make it explicit, as we have a factory called admin, instead of user. It will make more sense when reading the test, and the test will not fail if we change the code with the first user.

spec/controllers/admin/rooms_controller_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  let(:conference) { create(:conference) }
+  let!(:venue) { create(:venue, conference: conference) }
+  let(:room) { create(:room, venue: venue, size: 4) }
+
+  context 'admin is signed in' do
```
differentreality: We also need to test what happens if a simple user is signed in.

spec/models/ticket_spec.rb
```diff
@@ -5,7 +5,7 @@
   let(:ticket) { create(:ticket, price: 50, conference: conference) }
```
differentreality: Let us explicitely know the currency too. When you test later for same currency in conference, I do not want to have to go search in the factory about which currency this ticket has. Since currency matters for the test you need to let me know. Either here when you create the ticket, or you need to set the currency of the ticket later, before you use ticket variable for the specific test.

spec/models/ticket_spec.rb
```diff
@@ -43,11 +59,29 @@
...
+      before { ticket_purchase.update_attributes(paid: true) }
+
+      it 'returns true' do
+        expect(ticket.paid?(user)).to eq(true)
+      end
+    end
+
+    context 'user has not paid' do
+      it 'returns false' do
+        expect(ticket.paid?(user)).to eq(true)
```
differentreality: But you check if it returns true, and it succeeds. Why?

sonalkr132: Sorry, went a little overboard with copy and paste :stuck_out_tongue_closed_eyes: The method in question is:
```Ruby
def paid?(user)
    ticket_purchases.where(user_id: user.id, paid: false).count == 0
end
```

So, if `ticket_purchases` is empty, its count will be 0 and the test passes. I have changed the function on the assumption that between a given `ticket` and a `user` there can be only one `ticket_purchase`. I hope it makes sense.

spec/models/ticket_spec.rb
```diff
@@ -68,7 +102,7 @@
              user: user,
              ticket: ticket,
              quantity: 20)
-      expect(ticket.total_price(user)).to eq(Money.new(20 * ticket.price_cents, 'USD'))
+      expect(ticket.total_price(user)).to eq(Money.new(100000, 'USD'))
```
differentreality: Is this value supposed to be 20 (quantity in ticket purchases) * 50 (price of ticket) ?

sonalkr132: yes, but in cents. so 20*5000 = 100000

spec/models/room_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  describe 'association' do
+    it { should belong_to(:venue) }
+    it { should have_many(:events).dependent(:nullify) }
+  end
+
+  describe 'callback' do
+    after { subject.run_callbacks(:create) }
+
+    it 'sets guid' do
```
differentreality: Name this tests 
`it '#generate_guid'`

spec/controllers/admin/rooms_controller_spec.rb
```diff
@@ -0,0 +1,188 @@
+require 'spec_helper'
+
+describe Admin::RoomsController do
+  let!(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let!(:venue) { create(:venue, conference: conference) }
+  let(:room) { create(:room, venue: venue, size: 4) }
+
+  context 'non admin user is signed in' do
```
hennevogel: This is actually testing `BaseController#verify_user_admin` and not RoomsController...

sonalkr132: I added it here after [this diff](https://github.com/openSUSE/osem/pull/837#diff-for-comment-55135403).

hennevogel: It's still wrong :-) Please test `BaseController#verify_user_admin` instead of all the descendants.

spec/controllers/admin/rooms_controller_spec.rb
sonalkr132: Checkout [this discussion](https://github.com/openSUSE/osem/pull/837#discussion_r55192474).

differentreality:
:+1:

PR #838, opened 2016-03-06
Tests for ticket model

sonalkr132:
Change in error message of validation cause previously it was:
"Price currency Currency is different.."
`paid?` method returned true even when `ticket_purchases` is empty for
a given user, hence the change.

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -18,7 +18,7 @@ class Ticket < ActiveRecord::Base
   end
 
   def paid?(user)
-    ticket_purchases.where(user_id: user.id, paid: false).count == 0
+    ticket_purchases.find_by(user: user, paid: true).present?
   end
 
   def quantity_bought_by(user)
@@ -52,7 +52,7 @@ class Ticket < ActiveRecord::Base
 
   def tickets_of_conference_have_same_currency
     unless Ticket.where(conference_id: conference_id).all?{|t| t.price_currency == self.price_currency }
-      errors.add(:price_currency, 'Currency is different from the exist ticktes of this conference.')
+      errors.add(:price_currency, 'is different from the existing tickets of this conference.')
     end
   end
 end
--- a/spec/models/ticket_spec.rb
+++ b/spec/models/ticket_spec.rb
@@ -2,10 +2,10 @@ require 'spec_helper'
 
 describe Ticket do
   let(:conference) { create(:conference) }
-  let(:ticket) { create(:ticket, price: 50, conference: conference) }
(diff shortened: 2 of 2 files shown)


app/models/ticket.rb
```diff
@@ -52,7 +54,7 @@ def self.total_price(conference, user)
 
   def tickets_of_conference_have_same_currency
     unless Ticket.where(conference_id: conference_id).all?{|t| t.price_currency == self.price_currency }
-      errors.add(:price_currency, 'Currency is different from the exist ticktes of this conference.')
+      errors.add(:price_currency, 'is different from the exist ticktes of this conference.')
```
differentreality: typo: from the existing tickets

spec/models/ticket_spec.rb
```diff
@@ -77,16 +111,36 @@
...
-             quantity: 20)
-      expect(Ticket.total_price(conference, user)).to eq(Money.new(20 * ticket.price_cents, 'USD'))
-    end
-
-    it 'returns zero if the user has not bought this ticket' do
-      expect(Ticket.total_price(conference, user)).to eq(Money.new(0, 'USD'))
+    let(:diversity_supporter_ticket) { create(:ticket, conference: conference, price: 500) }
+
+    describe 'user has bought' do
+      context 'no ticktes' do
```
differentreality: typo

app/models/ticket.rb
```diff
@@ -18,7 +18,9 @@ def bought?(user)
   end
 
   def paid?(user)
```
differentreality: What about something like this?
`ticket_purchases.find_by(user: user, paid: true).present?`

spec/models/ticket_spec.rb
```diff
@@ -77,16 +111,36 @@
   end
 
   describe 'self#total_price' do
-    it 'returns the correct value if the user has bought this ticket' do
```
differentreality: #self.total_price

app/models/ticket.rb
sonalkr132: Thanks :yellow_heart:

ChrisBr: Where do we use ``paid?``? When you change the method logic you should also check if we still use it correct!

ChrisBr:
Otherwise LGTM

app/models/ticket.rb
differentreality: https://github.com/openSUSE/osem/blob/master/app/views/admin/tickets/show.html.haml#L34

differentreality:
:+1:

PR #847, opened 2016-03-07
Update country_select gem

ChrisBr:
and removed git repository from Gemfile. Caused bundle install error in vagrant box.


hennevogel:
LGTM

Issue #811, opened 2016-03-03
Ticket description is not working


sonalkr132:
Steps to reproduce:
* Make a conference (say: goto)
* Go to new tickets page (url: http://localhost:3000/admin/conference/goto/tickets/new )
* Click in description text area, the markdown toolbar will appear and textarea goes out of focus.
* Keep clicking in it until the world doesn't make any sense :smile_cat: 
![screenshot from 2016-03-03 17-10-24](https://cloud.githubusercontent.com/assets/7680662/13493254/04036f08-e163-11e5-8f76-d929dc443b11.png)

Console says:
> Uncaught TypeError: this.$textarea.on(...).on(...).on is not a function

pointing to[ these lines](https://github.com/openSUSE/osem/blob/37697517df1a34a5d62773713307a93a0039d98a/vendor/assets/javascripts/bootstrap-markdown.js#L143-L147).

arun1595:
[https://github.com/toopay/bootstrap-markdown/issues/215](url)

rishabhs95:
this issue has been resolved in the pull request #839 . please review the request.

sonalkr132:
closed by #849

PR #849, opened 2016-03-08
Fix markdown editors bug #811 using asset gem.

nishanthvijayan:
As suggested by @hennevogel I've replaced [bootstrap-markdown.js](https://github.com/toopay/bootstrap-markdown) , [to-markdown.js](https://github.com/domchristie/to-markdown) and markdown.js with ther corresponding asset gems thereby upgrading bootstrap-markdown as well.This fixes issue #811


sonalkr132:
I think [markdown.js](https://github.com/openSUSE/osem/blob/master/vendor/assets/javascripts/markdown.js) and [to-markdown.js](https://github.com/openSUSE/osem/blob/master/vendor/assets/javascripts/to-markdown.js)  files can be removed as well.

nishanthvijayan:
I did remove them from vendor/assets/javascripts/  and instead included  gems 'rails-assets-to-markdown' and 'rails-assets-markdown'

sonalkr132:
Yeah. Sorry!

nishanthvijayan:
On a related note, [rails-assets guidelines](https://rails-assets.org/) states that it requires bundler version to be  >= 1.8.4 . Shouldn't this be enforced in the Gemfile ?

hennevogel:
@nishanthvijayan it should

Gemfile
```diff
@@ -1,5 +1,8 @@
 source 'https://rubygems.org'
 
+# rails-assets requires >= 1.8.4
+gem 'bundler', '>= 1.8.4'
```
hennevogel: This will only install bundler in the bundle but not force the version it's built with :-)

hennevogel: Something like this in the Gemfile sounds about right:

if Gem::Version.new(Bundler::VERSION) < Gem::Version.new('1.8.4')
  abort "Bundler version >= 1.8.4 is required"
end

nishanthvijayan: Aah true..
But they have explicitly mentioned to include:
`gem 'bundler', '>= 1.8.4'`
in the Gemfile

screenshot from 2016-03-08 20 32 09

hennevogel: Well that doesn't make it right :-)

nishanthvijayan: So can I make another commit with what you suggested?

hennevogel: Please do so :-)

hennevogel:
Wonderful. Another cruft cleaned up. Thank you!

PR #851, opened 2016-03-09
Replace add userby email textinput in admin/roles#show with searchabl…

nishanthvijayan:
I've replaced the 'add user by email' text input in admin/roles#show with a searchable combobox listing email ID's of all users.I believe this is more user friendly.I've used the asset gem of [bootstrap-select](https://github.com/silviomoreto/bootstrap-select) for this purpose.

select1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -91,6 +91,8 @@ source 'https://rails-assets.org' do
   gem 'rails-assets-bootstrap-markdown'
   gem 'rails-assets-to-markdown'
   gem 'rails-assets-markdown'
+  # for searchable selects
+  gem 'rails-assets-bootstrap-select'
 end
 
 # as date picker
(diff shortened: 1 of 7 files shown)


ChrisBr:
@nishanthvijayan does it scale?

nishanthvijayan:
I didn't test it with more than a few users.So I don't know.Since the plugin is popular I was hoping it would be scalable.How should I test if it scales?

nishanthvijayan:
I'll try and fix the build problems as well.

hennevogel:
User.pluck(:email) won't scale ever :-)

nishanthvijayan:
I'm not sure if I follow.I thought .pluck was the fastest way if only one attribute needs to be fetched.

hennevogel:
It is the fastest. That doesn't mean its fast when you have 50.000 users :-)

nishanthvijayan:
Right :)  Is there any other way to go about this?I'm drawing a blank. :\

hennevogel:
How about autocompletion? You input at least 4 characters and it starts to search for Users :-)

PR #852, opened 2016-03-09
Fix variable @program typo in programs#show

differentreality:
Fixes failing test of features/program_spec.rb
I believe it was suceeding before, because the user was not really an organizer, but an admin.

### DIFF
--- a/app/views/admin/programs/show.html.haml
+++ b/app/views/admin/programs/show.html.haml
@@ -53,7 +53,7 @@
 
   .row
     .col-md-12.text-right
-      - if can? :edit, @progam
+      - if can? :edit, @program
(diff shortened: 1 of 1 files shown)


hennevogel:
You should rebase...

spec/factories/users.rb
```diff
@@ -16,6 +16,10 @@
       gravida.
     EOS
 
+    after(:create) do |user|
```
hennevogel: Why is this here?

differentreality: I don't know, some bad git pull on master perhaps.

sonalkr132:
hey, can we get rid bc1d16a from this PR?

ChrisBr:
LGTM, merging... :+1:

Issue #799, opened 2016-02-28
First user in test is admin


sonalkr132:
Steps to reproduce:
* Run following code (say spec/model/admin_spec.rb)
```Ruby

require 'spec_helper'

describe Registration do
  before do
    @user = create(:user)
  end

  it 'user is admin' do
    expect(@user.is_admin).to be_truthy
  end
end
```
* spring rspec spec/model/admin_spec.rb

Test passes! I think it is happening because of hack we do to set first user as admin.
Tests seem to be written with assumption that when we do `create(:user)` it is just a normal user (for example [this conference_registration spec](https://github.com/openSUSE/osem/blob/master/spec/features/conference_registration_spec.rb#L5))

differentreality:
What about we create a user in `config.before(:suite)`?
Our application will always has a first user, and any user creation in the tests will always correspond to a simple user, and not an admin. @hennevogel ?

sonalkr132:
I haven't given much time to find where exactly is the hack for this admin user setup, is it some migration file?

I think we can get around it by putting the hack in an `if` block which doesn't run when `ENV` is test.

hennevogel:
This is a feature of the ap. If a test assumes something wrong, it's wrong. If the factory is wrong, it's wrong.

PR #854, opened 2016-03-09
Enforce non-admin attribute of user factory

sonalkr132:
Build will pass after #852 gets merged.

TODO:
- [x] remove extra `first_user` from specs.

closes: #799

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,4 +1,18 @@
 # Read about factories at https://github.com/thoughtbot/factory_girl
+
+# It is a feature of our app that first signed up user is admin. This property
+# is set in a before create callback `setup_role` in user model.
+# We want to override this behavior when we create user factories. We want
+# `create(:user)` to create non-admin user by default. We are enforcing it
+# with `after create` blocks in following factories.
+#
+# Following commands won't work:
+#       `create(:user, is_admin: true)`
+#       `create(:admin, is_admin: false)`
+#
+# For a non-admin user, use: `create(:user)`
+# For an admin user, use: `create(:admin)
+
 FactoryGirl.define do
   factory :user do
     sequence(:email) { |n| "example#{n}@example.com" }
@@ -16,8 +30,19 @@ FactoryGirl.define do
       gravida.
     EOS
 
+    after(:create) do |user|
+      user.is_admin = false
+      # save with bang cause we want change in DB and not just in object instance
+      user.save!
+    end
+
     factory :admin do
-      is_admin true
+      # admin factory needs its own after create block or else after create
+      # of user factory will override `is_admin` value.
+      after(:create) do |user|
+        user.is_admin = true
+        user.save!
+      end
     end
   end
 end
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -3,8 +3,7 @@ require 'cancan/matchers'
 
 describe 'User' do
   describe 'Abilities' do
-    # automatically becomes admin
-    let!(:first_user) { create(:user) }
+    let!(:admin) { create(:admin) }
 
     # see https://github.com/CanCanCommunity/cancancan/wiki/Testing-Abilities
     subject(:ability){ Ability.new(user) }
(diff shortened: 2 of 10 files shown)


spec/factories/users.rb
```diff
@@ -1,4 +1,20 @@
 # Read about factories at https://github.com/thoughtbot/factory_girl
+
+# It is a feature of our app that first signed up user is admin. This property
+# is set in a before create callback `setup_role` in user model.
+# We want to override this behavior when we create user factories. We want
+# `create(:user)` to create non-admin user by default. We are enforcing it
+# with `after create` blocks in following factories.
+#
+# Following commands won't work:
```
differentreality: won't work for the first user.

sonalkr132: It won't work for any of user factories. `after create` of user/admin factory gets executed all the time.

differentreality: I think what is actually not going to work is:
create(:user, is_admin: true) 
create(:admin, is_admin: false)

differentreality:
The test fail is due to #852

spec/factories/users.rb
sonalkr132: Teah. Technically for rest two, setting `is_admin` won't work either. However, since thats their default it won't matter. Should I remove them?

spec/models/ability_spec.rb
```diff
@@ -210,7 +208,7 @@
       it{ should_not be_able_to(:manage, my_conference.tickets.first) }
       it{ should_not be_able_to(:manage, conference_public.tickets.first) }
 
-      it{ should_not be_able_to(:manage, my_registration) }
+      it{ should be_able_to(:manage, my_registration) }
```
sonalkr132: @differentreality user with cfp role should be able to manage his own registration (my_registration)?

spec/models/ability_spec.rb
```diff
@@ -3,9 +3,7 @@
 
 describe 'User' do
   describe 'Abilities' do
-    # automatically becomes admin
-    let!(:first_user) { create(:user) }
-
+    let(:user) { create(:user) }
```
differentreality: Please be careful, you are creating wrong variables. The variable name was first_user and it was (implicitely) an admin.
We already have below a variable user!

spec/models/ability_spec.rb
differentreality: Every signed in user can manage his/her own registration. I think that this is not what this test is testing, because my_registration was a registration for another user. How come you changed the test?

spec/models/ability_spec.rb
sonalkr132: Shouldn't tests have failed? :fearful:

spec/models/ability_spec.rb
sonalkr132: It is obviously wrong that I used two `user` in same scope.

> How come you changed the test?

It was failing.. I really don't think I understand `ability_specs` :confused: You will either have to take time to explain it to me or you can send a new PR using my first commit. 
I would prefer first, may be we can work on refactoring it and there is other work on authorization we have discussed. At least we should divide it in files(on basis of roles) so that we don't have to scroll 200 lines to see the definition of each variable.

differentreality: OK test was failing, but when a test fails, we don't change the test in order to make it pass. We search to find out what's wrong and a test that was succeeding, now fails.
Have you read a few chapters of https://github.com/CanCanCommunity/cancancan/wiki ? I am obviously more than happy to spend time to take you through whatever you don't get, however I cannot do that unless you first read up and experiment a bit with how cancancan works. When you do, find me on IRC, we will set up some time to discuss in more detail.

sonalkr132:
@differentreality updated!

differentreality:
I think it might be wise, when it comes to roles, to first check if the user.is_admin is actually false. Otherwise spec/features/base_controller_spec.rb and spec/features/ability_spec.rb will give us false positive successes on tests. @hennevogel ?

spec/factories/users.rb
hennevogel: yes

hennevogel:
I don't get your meaning @differentreality? Check what where?

differentreality:
I am referring to all changes in https://github.com/openSUSE/osem/pull/854/files#diff-93fca694bc29540a4901733802ce5ac7L23 and https://github.com/openSUSE/osem/pull/854/files#diff-5520d8caa12f728235617093002a09c6L26

hennevogel:
I don't think we need to test our factories in tests. Sounds like overkill to me.

hennevogel:
I would merge this. @differentreality?

differentreality:
OK

Issue #421, opened 2014-08-01
Refactor Conference: Drops


hennevogel:
* [x] html_export_path
* [x] dietary_choices
* [x] social_events

differentreality:
@hennevogel anything else to drop?

PR #855, opened 2016-03-09
drop dietary choices, social events, and html_export_path

differentreality:
https://github.com/openSUSE/osem/issues/421

### DIFF
--- a/db/migrate/20160309182642_remove_social_events_table.rb
+++ b/db/migrate/20160309182642_remove_social_events_table.rb
@@ -1,0 +1,134 @@
+class RemoveSocialEventsTable < ActiveRecord::Migration
+  class TempConference < ActiveRecord::Base
+    self.table_name = 'conferences'
+
+    has_many :temp_social_events
+  end
+
+  class TempSocialEvent < ActiveRecord::Base
+    self.table_name = 'social_events'
+
+    belongs_to :temp_conference
+    has_and_belongs_to_many :temp_registrations
+  end
+
+  class TempRegistration < ActiveRecord::Base
+    self.table_name = 'registrations'
+
+    belongs_to :temp_conference
+    has_and_belongs_to_many :temp_social_events
+    has_and_belongs_to_many :temp_qanswers
+  end
+
+  class TempRegistrationsSocialEvent < ActiveRecord::Base
+    self.table_name = 'registrations_social_events'
+
+    belongs_to :temp_registrations
+    belongs_to :temp_social_events
+  end
+
+  class TempQuestionType < ActiveRecord::Base
+    self.table_name = 'question_types'
+
+    has_many :temp_questions
+  end
+
+  class TempQuestion < ActiveRecord::Base
+    self.table_name = 'questions'
+
+    has_many :temp_qanswers
+    has_many :temp_answers, through: :temp_qanswers
+    belongs_to :temp_question_type
+    has_and_belongs_to_many :temp_conferences
+  end
+
+  class TempAnswer < ActiveRecord::Base
+    self.table_name = 'answers'
+
+    has_many :temp_qanswers
+    has_many :temp_questions, through: :temp_qanswers
+  end
+
+  class TempQanswer < ActiveRecord::Base
+    self.table_name = 'qanswers'
+
+    belongs_to :temp_question
+    belongs_to :temp_answer
+    has_and_belongs_to_many :temp_registrations
+  end
+
+  class TempConferencesQuestions < ActiveRecord::Base
+    self.table_name = 'conferences_questions'
+  end
+
+  class TempQanswerRegistration < ActiveRecord::Base
+    self.table_name = 'qanswers_registrations'
+  end
+
+  def up
+    # Create Question of 'Multiple Choice' type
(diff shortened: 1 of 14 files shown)


ChrisBr:
Uff, 600 loc ...

Can we somehow split this in three different commits / PR? Otherwise it's really hard to review @differentreality ...

differentreality:
@ChrisBr 600+ changes (if that's what you mean) are from stupid db/schema, it is not going to be helpful on that note if I split into commits or PRs.

db/migrate/20160309182642_remove_social_events_table.rb
```diff
@@ -0,0 +1,134 @@
+class RemoveSocialEventsTable < ActiveRecord::Migration
```
hennevogel: I would very much like us to stop doing these migrations and switch to [rake based data migrations](https://robots.thoughtbot.com/data-migrations-in-rails).

hennevogel: What I mean is in the future :-)

hennevogel:
Have you tested the migration with live data?

differentreality:
Yes, I have.

ChrisBr:
Can you fix the rubocop offense?

db/migrate/20160309182642_remove_social_events_table.rb
ChrisBr: So we will leave this PR as it is? @hennevogel @differentreality

db/migrate/20160309182642_remove_social_events_table.rb
```diff
@@ -0,0 +1,134 @@
...
+    self.table_name = 'conferences_questions'
+  end
+
+  class TempQanswerRegistration < ActiveRecord::Base
+    self.table_name = 'qanswers_registrations'
+  end
+
+  def up
+    # Create Question of 'Multiple Choice' type
+    qtype = TempQuestionType.find_or_create_by!(title: 'Multiple Choice')
```
ChrisBr: Shouldn't we move this line also to line 73? (inside the if statement?)

db/migrate/20160309182642_remove_social_events_table.rb
```diff
@@ -0,0 +1,134 @@
...
+        question = TempQuestion.find_or_initialize_by(title: 'Which of the following social events are you going to attend?',
+                                                      conference_id: conference.id,
+                                                      question_type_id: qtype.id)
+        question.save!
+        # Enable the question for the conference
+        TempConferencesQuestions.find_or_create_by!(conference_id: conference.id,
+                                                    question_id: question.id)
+      end
+
+      TempSocialEvent.where(conference_id: conference.id).each do |social_event|
```
ChrisBr: Shouldn't we move this also into the if statement?

db/migrate/20160309182655_remove_dietary_choices_table.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        qa.save!
+
+        # Associate appropriate answer with registration
+        TempQanswerRegistration.find_or_create_by!(registration_id: registration.id,
+                                                   qanswer_id: qa.id)
+
+        # Move data from other_dietary_choice to other_special_needs
+        registration.other_special_needs = registration.other_special_needs || ''
+        registration.other_special_needs += " Other dietary choice: #{registration.other_dietary_choice}."
```
ChrisBr: If you change this line to
``
registration.other_special_needs << "Other dietary choice: #{registration.other_dietary_choice}."
``

you could drop the line before it (without actually testing it ...)

ChrisBr:
@differentreality with which data have you tested it? Maybe we should test it with an actual database dump of events.o.o? @hennevogel

db/migrate/20160309182642_remove_social_events_table.rb
hennevogel: Yeah this is too much work to drop IMHO :-)

differentreality: Would that be sth like https://github.com/openSUSE/osem/pull/855/commits/f9bf080de292748489e1e3299a0810c961e58211 ?
Would we have to invoke the rake task from within the migration that drops the columns, so that we ensure that the rake task is executed **prior** to the destruction of data?

differentreality:
I have an old dump of events.o.o, happy to get a new one and test it again :)

db/migrate/20160309182642_remove_social_events_table.rb
hennevogel: > Would that be sth like f9bf080

Yes exactly.
> Would we have to invoke the rake task from within the migration

This is the thing. Data migration is part of the deployment, not part of the development. So for us, we would add that new rake task to `config/deploy.rb` and for other deployments we would have to come up with a way to describe how to do this.

This also means that dropping the old schema should not happen in the same commit.

ChrisBr:
@differentreality why does CI fail?
Have you tried the migration with the new db dump?

differentreality:
I hadn't removed some references of social events and dietary choices, let's see now. 
Migrations seem to run successfully on the newest db dump I got.

hennevogel:
Okay lets get this in

Issue #821, opened 2016-03-04
Unregistering from conference does not delete ticket purchase


sonalkr132:
Steps to reproduce:
* Make a conference and then tickets.
* Sign in with a different user, register to conference and  buy tickets
* Unregister form the conference 
* Sign in with admin and go to tickets area (url: http://localhost:3000/admin/conference/<conf>/tickets)
* You can still see tickets are there
* Sign in again with other user and register, now you have tickets you purchases last time you registered.

EDIT: Of course you can just use, rails c or pry instead.

chashmeetsingh:
@sonalkr132 can't reproduce

sonalkr132:
@chashmeetsingh get on #osem and hit me up. I will help you reproduce it.

ChrisBr:
I can reproduce. Yeah, if you unregister I'd expect that it deletes the ticket purchases ...

rishabhs95:
can i work on this issue?

rishabhs95:
worked on this issue according to its tests, sending the PR.

rishabhs95:
please review the PR #856

differentreality:
What are we expecting when we implement sales on tickets? 
User registers + pays for a ticket, the ticket cannot be unpaid, even if the user uregisters from the conference. Thus we keep ticket_purchase, since the ticket is already paid for? Inform organizers somehow about paid tickets from unregistered users?

hennevogel:
That is to be determined during the implementation I guess. Don't know the refund options of e.g. stripe

hennevogel:
Fixed in #876

PR #856, opened 2016-03-09
Destroying tickets after user unregisters from a conference

rishabhs95:
once a user unregisters from a conference, the tickets he purchased should be destroyed.
fixing #821 .

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -85,7 +85,7 @@ class ConferenceRegistrationsController < ApplicationController
   end
 
   def destroy
-    tickets_purchased = TicketPurchase.find_by(conference_id: @registration.conference_id)
+    tickets_purchased = TicketPurchase.where(conference_id: @registration.conference_id)
     if @registration.destroy
       if tickets_purchased
         tickets_purchased.destroy
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -87,7 +87,9 @@ class ConferenceRegistrationsController < ApplicationController
   def destroy
     tickets_purchased = TicketPurchase.find_by(conference_id: @registration.conference_id)
     if @registration.destroy
-      tickets_purchased.destroy
+      if tickets_purchased
+        tickets_purchased.destroy
+      end
       redirect_to root_path,
                   notice: "You are not registered for #{@conference.title} anymore!"
     else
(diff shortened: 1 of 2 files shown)


sonalkr132:
hey, don't you think that using a `before destroy` [callback](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html) would make more sense?

Please add tests too.

rishabhs95:
okay, working on it

app/controllers/conference_registrations_controller.rb
```diff
@@ -85,7 +85,11 @@ def update
   end
 
   def destroy
+    tickets_purchased = TicketPurchase.find_by(conference_id: @registration.conference_id)
```
sonalkr132: Please don't use `find_by` either. A registered user can have multiple tickets purchase, use `where`.

ChrisBr:
@sonalkr132 why an ``before_destroy``? I'd go with an ``after_destroy``...

app/controllers/conference_registrations_controller.rb
rishabhs95: yes, that's true... will use `where` in this commit... writing test.

sonalkr132:
@ChrisBr Would it make any difference? Both of them are transnational change.

sonalkr132:
>  why an before_destroy?

If for some reason delete to tickets failed, use of `before_destroy` will rollback the complete transaction and nothing will be deleted. However if we use `after_destroy` we may end up with tickets purchased by unregistered users.

rishabhs95:
for its test here, is some thing like 
`it 'it deletes the conference tickets' do
 expect { delete :destroy, conference_id: conference.short_title }.to change(TicketPurchase, :count).by(-1)  end`
okay?

ChrisBr:
@sonalkr132 I don't get your point ...

ChrisBr:
@rishabhs95 From my POV, yes!

ChrisBr:
@sonalkr132 I just tried it, it does not make any difference if you use ``after`` or ``before`` in this case.

ChrisBr:
@rishabhs95 btw: please also test that it destroys all tickets if you have more than one ticket (e.g. different tickets and different amount of tickets)

rishabhs95:
okay @ChrisBr ... so is the current fix okay with adding new tests and `where` clause?
or should it be modified to `before_destroy` or `after_destroy`?

ChrisBr:
Use the hooks please as suggested!

rishabhs95:
created another PR #870

Issue #818, opened 2016-03-04
Name attribute in registration email not set 


ChrisBr:
The name attribute in registration email not set:

```
-----------------
Dear {name},

Thank you for Registering for oSC16.
-----------------
```

Could be related to #577, maybe we need to switch to :username as :name isn't mandatory anymore! Maybe we also need to check the other mail templates as well!

Shriyanshagro:
tried to solved the issue , check if it works
#https://github.com/Shriyanshagro/osem/commit/17e7f49100269d929e626701b502e9f3a11c35e1

hennevogel:
@Shriyanshagro  And who do you think sets @user? :-)

The variables are fetched in EmailSettings#get_values. There are no instance variables.

Ana06:
I think the problem is here (line 64 of email_settings.rb):

text = text.gsub "{#{key}}", value unless text.blank? || value.blank?

As it is said in issue #577,  name can be blank, and if that is the case {name} is not replaced. A possible solution could be having a function get_name() in User that returns the name if it is not blank. Otherwise it returns the username (or the email or the text 'user'). This function can be used in get_values (in email_settings.rb) to ensure that name is not blank.

hennevogel:
I would find it even cooler to just overload `#name`

sonalkr132:
This can be close since #865 is merged.

Issue #577, opened 2014-12-31
Name attribute empty


differentreality:
After ichain implementation, the 'name' attribute for user is no longer required.

This lets users get created without a value for 'name' attribute.

But we use name, for example recent registrations in admin/conference/index.
Being blank is ugly and withholds the necessary information for no good reason.

http://paste.opensuse.org/89076165

differentreality:
I have concerns about that, I am not sure 'username' is what we are really after here.

Username does not really tell me who is the last person to register to the conference, in the essence that:
if you see that 'differentreality' registered to a conference, that gives you much *different information* than seeing that 'Stella Rouzi' registered to the conference, which is actually a very straight forward info! 
(assuming you don't know me that well so that you can immediately relate my nickname, which may well be my username, to my real name)

ChrisBr:
@differentreality If you ask me, I wouldn't worry too much about it! If you think that this information is really necessary, the only way to fix it is to make name mandatory (and in the same breath I would recommend to split name in first and lastname). But as I said, I think a good and acceptable solution would be to use the username...
@hennevogel What do you think about it?

hennevogel:
Web apps usually have one primary identifier: **either** username or email address. I don't know a single one which requires me to set a Given Name and Surname.

But that does not keep us from showing this **IF** it is available. We can also update it from the Access Manager data (like we do it with the email address).

ChrisBr:
@hennevogel @differentreality Facebook for instance requires for instance a first and surname ;)

That's a good point, showing the full name if there is one otherwise the username!
The Access Manager is an option, but not all instances use it (eg Owncloud)! If they don't use it, the name is still empty!

hennevogel:
@ChrisBr this would be in [user#for_ichain_username](https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L73) so it wouldn't matter for installations with ichain disabled...

ChrisBr:
@hennevogel Yes, I know! Just wanted to let you know that this opportunity doesn't exist if ichain isn't enabled!

differentreality:
Same problem in registrations list. Really looks bad and if the email does not contain the name of the user it is even more difficult for the Info Desk people to find the user in that registration list.

hennevogel:
[...]

PR #857, opened 2016-03-09
Name attribute in registration email not set

Ana06:
Issue #818

I have added a function get_name() in User that returns the name if it is not blank. Otherwise it returns the username. This function is used in get_values (in email_settings.rb) to ensure that name is not blank and therefore {name} is replaced in emails.

This can also solve issue #577 (maybe it is needed to use get_name in some other places)

### DIFF
--- a/app/views/admin/programs/show.html.haml
+++ b/app/views/admin/programs/show.html.haml
@@ -53,7 +53,7 @@
 
   .row
     .col-md-12.text-right
-      - if can? :edit, @progam
+      - if can? :edit, @program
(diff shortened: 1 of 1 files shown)


ChrisBr:
Hi @Ana06,

as @hennevogel stated in #818, overriding the #name attribute would be cooler. Note that you have to use ```read_attribute(:name)``` to not recursively fetch the name (see [stackoverflow)](http://stackoverflow.com/questions/21835116/how-can-i-overwrite-a-getter-method-in-an-activerecord-model).

Otherwise :+1:

Issue #804, opened 2016-03-01
Add flash_type error


differentreality:
In a lot of our controllers we use `flash[:error] = 'message'` and then we redirect.
In other places we just use `redirect_to root_path, notice: 'message'`
To have a common way, we should add the flash_type error (http://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html#method-i-add_flash_types) and rewrite the controllers to use a common style.

AdiChat:
@differentreality  So does that mean that we rewrite the two types of flash messages to the form 
[ redirect_to user_path(@user), warning: "Incomplete profile" ] after adding the flash_type error. So this involves modifying controllers and view. Is this the vision?

differentreality:
Yes, but we want to add the flash_type of error, so when we do `add_flash_types :error`, then we obviously need to use **error** in our messages, not warning.

Which views would you modify with this change?

chashmeetsingh:
@differentreality can I work on this?

differentreality:
Anyone who wants to work on any issue can do so, by assigning himself/herself from the right side of this page where it says **Assignee**.

AdiChat:
@differentreality  It seems views will not be affected by this change as the objective is to attain uniformity in code style.

differentreality:
Closed via #858

PR #858, opened 2016-03-10
Add flash type error.Rewrite controllers to use common flash style wi…

nishanthvijayan:
As recommended in #804 , added flash_type :error and rewrote all redirect_to statements to use the same style

### DIFF
--- a/app/controllers/admin/call_for_papers_controller.rb
+++ b/app/controllers/admin/call_for_papers_controller.rb
@@ -34,8 +34,8 @@ module Admin
       if @cfp.update_attributes(call_for_paper_params)
         Mailbot.delay.send_on_call_for_papers_dates_updated(@conference) if send_mail_on_cfp_dates_updated
         Mailbot.delay.send_on_schedule_public(@conference) if send_mail_on_schedule_public
-        redirect_to(admin_conference_call_for_paper_path(@conference.short_title),
-                    notice: 'Call for papers successfully updated.')
+        redirect_to admin_conference_call_for_paper_path(@conference.short_title),
+                    notice: 'Call for papers successfully updated.'
       else
         flash[:error] = "Updating call for papers failed. #{@cfp.errors.to_a.join('. ')}."
         render :new
--- a/app/controllers/admin/campaigns_controller.rb
+++ b/app/controllers/admin/campaigns_controller.rb
@@ -12,8 +12,8 @@ module Admin
       @campaign.attributes = campaign_params
 
       if @conference.save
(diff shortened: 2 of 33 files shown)


sonalkr132:
@nishanthvijayan Can we [change](https://github.com/openSUSE/osem/pull/858/files#diff-d438f64a44599197d7a8f71b0a65e039R18) [them](https://github.com/openSUSE/osem/pull/858/files#diff-1403510ba5f98202f033151a8787c089R28) [all](https://github.com/openSUSE/osem/search?p=1&q=flash+error) please?

nishanthvijayan:
The remaining instances of flash[:notice / :alert / :error] are being used with render and redirect_back_or_to.
They don't seem to support passing flash as argument.

nishanthvijayan:
I've also left out 2 other cases where replacing flash seemed impractical/impossible.See [here](https://github.com/openSUSE/osem/blob/af168f0f557a551022c361cc5aa9a039bec94c2d/app/controllers/admin/roles_controller.rb#L51) and [here](https://github.com/openSUSE/osem/blob/056df982b822e0a41b6937bd3e0c412574d4b481/app/controllers/admin/questions_controller.rb#L68)

sonalkr132:
> They don't seem to support passing flash as argument.

[Interesting](https://rails.lighthouseapp.com/projects/8994/tickets/4911-patch-add-notice-alert-and-flash-as-options-to-render)  :astonished:

nishanthvijayan:
@differentreality  I've made the changes like you asked.

ChrisBr:
@nishanthvijayan sorry for the delay, I'll review today!

ChrisBr:
@nishanthvijayan good work :+1: 

However, I recognized that we use different flash types for the same actions/failures. For instance, sometimes we use ``:error`` and sometimes ``:alert`` when updating / saving goes wrong! It would be nice if this would also be consistent.

From my POV: ``:notice`` for successfully saving / creating is fine. For failure while saving / updating I'd go with an ``alert`` type. @hennevogel @differentreality what do you think?

differentreality:
I would go for error on failure to update/save. And I am  OK with some alert types we have for cancan exception, or when the user is not signed in.

ChrisBr:
@nishanthvijayan so please change all update/save failures to ``:error``! Thanks!

nishanthvijayan:
@ChrisBr  Done.

ChrisBr:
@nishanthvijayan Awesome! Thanks a lot :+1:

PR #861, opened 2016-03-10
Fix overflow of proposal edit in admin area

sonalkr132:
Edit of proposal in admin area is reusing partial `proposal/form`.
This PR changes:
screenshot from 2016-03-10 17-40-21

to:
screenshot from 2016-03-10 17-47-46


hennevogel:
Looks good to me

PR #865, opened 2016-03-10
Name attribute in registration email not set

Ana06:
Issues #818 and #577. Pull request #857.

I agree that it is better to override the #name attribute. I have read that `attr_accessor_with_default` is deprecated, so I think it is better not to use it. I have used `self[:name]` instead of `read_attribute(:name)` as it is preferred according to the [Rails Style Guide](https://github.com/bbatsov/rails-style-guide#read-attribute).

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -33,6 +33,13 @@ describe User do
     expect(another_user.roles[1]).to eq(cfp_role)
   end
 
+  describe '#name' do
+    it 'returns the username as name if there is not name' do
+      user = create(:user, name: nil)
+      expect(user.name).to eq(user.username)
+    end
+  end
+
   describe '#has_role?' do
     describe 'when user has a role' do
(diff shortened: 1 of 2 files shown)


ChrisBr:
Cool :clap: 

If you add a test we can :ship: it ...

Ana06:
Done! :wink:

spec/models/user_spec.rb
```diff
@@ -33,6 +33,13 @@
     expect(another_user.roles[1]).to eq(cfp_role)
   end
 
+  describe '#name' do
+    it 'returns the username as name if there is not name' do
```
sonalkr132: It would nice if we have test for case where it `returns the name if name is present`.

hennevogel: I doubt that we need to test `||` that extensively...

hennevogel:
Thank you for your contributions @Ana06 :bouquet:

PR #866, opened 2016-03-11
Add before_destroy callback to cancel role destroy.solves #864

### LINKED ISSUE #864: Removing all users from a role removes the role
Removing all users  from a role (other that organizer) deletes the corresponding role as well.

nishanthvijayan:
Removing all users from a role was deleting role as well.This fixes this issue by cancelling destroy through a before_destroy callback
Fixes #864

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -4,7 +4,8 @@ class Conference < ActiveRecord::Base
   require 'uri'
   serialize :events_per_week, Hash
   # Needed to call 'Conference.with_role' in /models/ability.rb
-  resourcify
+  # Dependent destroy will fail as roles#destroy will be cancelled,hence delete_all
+  resourcify :roles, dependent: :delete_all
 
   default_scope { order('start_date DESC') }
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -1,10 +1,17 @@
 class Role < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
   has_and_belongs_to_many :users, join_table: :users_roles
-
+  before_destroy :cancel
   scopify
 
   validates :name, presence: true
 
   validates :name, uniqueness: { scope: :resource }
+
+  private
+
+  # Needed to ensure that removing all user from role doesn't remove role.
+  def cancel
+    false
+  end
 end
--- a/lib/tasks/roles.rake
+++ b/lib/tasks/roles.rake
@@ -1,0 +1,15 @@
+namespace :roles do
+  desc 'Adds back deleted roles to all conferences'
+  task add: :environment do
+
+    Conference.all.each do |c|
+      Role.where(name: 'organizer', resource: c).first_or_create(description: 'For the organizers of the conference (who shall have full access)')
+      Role.where(name: 'cfp', resource: c).first_or_create(description: 'For the members of the CfP team')
+      Role.where(name: 'info_desk', resource: c).first_or_create(description: 'For the members of the Info Desk team')
+      Role.where(name: 'volunteers_coordinator', resource: c).first_or_create(description: 'For the people in charge of volunteers')
+    end
+
+    puts 'All done!'
+  end
+end
+
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -5,10 +5,19 @@ describe Role do
   let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let!(:cfp_role) { Role.find_by(name: 'cfp', resource: conference) }
   let!(:organizer) { create(:user, role_ids: organizer_role.id) }
+  let(:user) { create(:user) }
 
   it 'get_users' do
     expect(organizer_role.users).to include organizer
     expect(organizer_role.users.count).to eq 1
     expect(cfp_role.users).to eq []
+  end
+
+  it 'does not delete role when last user of role is removed' do
+    user.add_role :cfp, conference
+    expect(cfp_role.users).to include user
+    user.remove_role :cfp, conference
+    expect(cfp_role.users).to eq []
+    expect(conference.roles.count).to eq 4
   end
 end


nishanthvijayan:
@differentreality I've made the necessary fixes.
Checkout [resourcify options](https://github.com/RolifyCommunity/rolify/blob/d44ec217c4cfb13d0b84bd6829a3a9f5f59aa2da/lib/rolify.rb#L47)

sonalkr132:
We should add tests so that it doesn't happen again.

app/models/role.rb
```diff
@@ -1,10 +1,16 @@
 class Role < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
   has_and_belongs_to_many :users, join_table: :users_roles
-
+  before_destroy :cancel
```
sonalkr132: If we are just returning false, we can do oneliner `before_destroy { false }`. Also, can we add please add comment about why we are doing this?

app/models/role.rb
```diff
@@ -1,10 +1,16 @@
...
   scopify
 
   validates :name, presence: true
 
   validates :name, uniqueness: { scope: :resource }
+
+  private
+
```
differentreality: Add a comment about what the `cancel` function is / what it does.

app/models/conference.rb
```diff
@@ -4,7 +4,7 @@ class Conference < ActiveRecord::Base
   require 'uri'
   serialize :events_per_week, Hash
   # Needed to call 'Conference.with_role' in /models/ability.rb
```
differentreality: Add another line of comment briefly explaining why we need the options to resourcify

nishanthvijayan:
@differentreality  I've added the comments and feature test to ckeck for regression

differentreality:
I am wondering if we should be checking that all conferences have all roles available (and if not, creating them).

nishanthvijayan:
I could add a scenario to visits roles index page and verify that all  4 roles are listed.Would that be enough?

differentreality:
What we would test is if a conference has 4 roles available (which would be organier, cfp, etc).
That is already being tested in admin/conferences controller test for `POST #create`

What I meant was that we check the integrity of the DB with a migration. So that if there are conferences that were impacted from rolify's behavior (considering that we are not able to create a new role anymore), then we should make sure that we add the missing roles back. So that every conference in the DB has 4 roles available.

app/models/role.rb
hennevogel: And this effectively makes it impossible to destroy roles. Why?

differentreality: We want to override destroying roles, because rolify by default deletes a role when its last user is removed.
From the app we are not able to create or delete a role, because we have fixed available roles for all conferences. And we need these roles to exist, even without users.

hennevogel: And there is no way to tell rolify not to do that other then making it impossible in the model?

nishanthvijayan: Seems that way.Rolify docs don't mention anything.

differentreality: @hennevogel I have not found a way to do it so far, no.

nishanthvijayan:
Sorry for the late reply.
I don't think I understand your idea.Is this to fix the issue for those who are already using osem/developing osem?Is it necessary?

differentreality:
Yes, exactly. 
Right now we cannot create a role from its interface. So, until today if someone were to add users to a role, and then remove all those roles, that role would be deleted. So there might be conferences that end up not having all 4 roles.
As such IMO we need to address that issue as well, otherwise organizers do not have an alternative to create the deleted role again.

nishanthvijayan:
@differentreality Done.
henne suggested that I write a rake task instead of a migration.

lib/tasks/roles.rake
```diff
@@ -0,0 +1,11 @@
+namespace :roles do
+  desc "Adds back deleted roles to all conferences"
+  task :add => :environment do
```
differentreality: add: :environment

lib/tasks/roles.rake
```diff
@@ -0,0 +1,11 @@
+namespace :roles do
+  desc "Adds back deleted roles to all conferences"
```
differentreality: single quotes

lib/tasks/roles.rake
```diff
@@ -0,0 +1,11 @@
...
+  task :add => :environment do
+    Conference.all.each do |c|
+      Role.where(name: 'organizer', resource: c).first_or_create(description: 'For the organizers of the conference (who shall have full access)')
+      Role.where(name: 'cfp', resource: c).first_or_create(description: 'For the members of the CfP team')
+      Role.where(name: 'info_desk', resource: c).first_or_create(description: 'For the members of the Info Desk team')
+      Role.where(name: 'volunteers_coordinator', resource: c).first_or_create(description: 'For the people in charge of volunteers')
+    end
+  end
+end
```
differentreality: empty line at the end of file

nishanthvijayan:
Done.

lib/tasks/roles.rake
differentreality: Would be nice to add a message at the end, like:

`puts 'All done!'`

differentreality:
I guess we should be testing if the rake task is doing what it's supposed to do, @hennevogel ?

nishanthvijayan:
I've fixed the style issues.Are tests needed as well?

hennevogel:
I doubt we need tests for the rake task. it's simple enough...

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+
+feature Role do
+  let!(:conference) { create(:conference) }
```
ChrisBr: Are the exclamation marks really necessary?

ChrisBr: Btw: does this really need to be a feature test? From my POV a feature test is a little bit overhead and a simple model test should also do the trick ...

differentreality:
@nishanthvijayan ping

nishanthvijayan:
@differentreality what do you suggest? should I change the feature test to model test?

differentreality:
@nishanthvijayan yes, let's see it :)

nishanthvijayan:
Done.

spec/models/role_spec.rb
```diff
@@ -5,10 +5,19 @@
...
   let!(:organizer) { create(:user, role_ids: organizer_role.id) }
+  let(:user) { create(:user) }
 
   it 'get_users' do
     expect(organizer_role.users).to include organizer
     expect(organizer_role.users.count).to eq 1
     expect(cfp_role.users).to eq []
   end
+
+  it 'adds and removes user' do
```
differentreality: We are actually testing something else here: 'does not delete role when last user is removed' (or sth along those lines)

nishanthvijayan:
@differentreality: updated.

differentreality:
Awesome @nishanthvijayan, keep up the good work!

PR #867, opened 2016-03-11
WIP: Proposal controller tests

sonalkr132:
**#in_progress**

### DIFF
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -46,31 +46,28 @@ class ProposalController < ApplicationController
                            event_role: 'submitter')
     @event.event_users.new(user: current_user,
                            event_role: 'speaker')
+    if @event.save
+      ahoy.track 'Event submission', title: 'New submission'
 
-    unless @event.save
+      flash[:notice] = 'Proposal was successfully submitted.'
+      redirect_to conference_program_proposal_index_path(@conference.short_title)
+    else
       flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
       render action: 'new'
-      return
     end
-
-    ahoy.track 'Event submission', title: 'New submission'
-
-    flash[:notice] = 'Proposal was successfully submitted.'
-    redirect_to conference_program_proposal_index_path(@conference.short_title)
   end
 
   def update
     authorize! :update, @event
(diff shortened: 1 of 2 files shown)


spec/controllers/proposal_controller_spec.rb
```diff
@@ -0,0 +1,205 @@
...
+        end
+      end
+    end
+
+    describe 'PATCH #confirm' do
+      context 'confirmed successfully' do
+        context 'user have registered for the conference' do
+          before do
+            request.env['HTTP_REFERER'] = '/'
+            # change state of event to accepted
```
sonalkr132: hey, how do I change state of event from new to accepted? `event.accept!` says:
> ArgumentError: wrong number of arguments (0 for 1)

differentreality:
@sonalkr132 sth like this? https://github.com/openSUSE/osem/blob/master/spec/models/ability_spec.rb#L28

app/controllers/proposal_controller.rb
```diff
@@ -46,31 +46,28 @@ def create
...
-      render action: 'new'
-      return
+      render action: 'edit'
     end
-
-    redirect_to(conference_program_proposal_index_path(conference_id: @conference.short_title),
-                notice: 'Proposal was successfully updated.')
   end
 
   def destroy
```
sonalkr132: Shouldn't this be named `withdraw`? This method also making `DELETE` request for updating the record when it should make `PATCH` request, not very RESTful.

sonalkr132: To save the event in `destroy` method [we are doing](https://github.com/openSUSE/osem/pull/867/files#diff-8c13e9cfe1f5edf7d4923e633a08f790R84):
```
@event.save(validate: false)
```

When for reset (`withdraw`, `restart` etc), we are doing:
```
if @event.save
   # do something
else
  # do something else
end
```

I think we should be consistent. Should I change `destroy` save to `if else` block too?

hennevogel: `destroy -> withdraw` and `if else` sounds like a plan yes

hennevogel:
Please rebase and fix your commit message. Thanks

sonalkr132:
@hennevogel This PR won't be complete until #872 gets fixed. I am skipping tests here. If I add them now, I will have to change everything again.

hennevogel:
Alright lets close this then

Issue #868, opened 2016-03-12
user can submit proposals event when cfp is not specified


sonalkr132:
Steps to reproduce:
* Create a conference (say: goto) and make sure that you don't specify call for papers
* Sign out and sign with a non-admin user
* Go to new proposal path of your new conference (path: `http://localhost:3000/conference/goto/program/proposal/new`)
* You shouldn't really be able to see the form but you do. Fill the proposal and submit. Again you shouldn't be really be able to submit but it will.

I think it is not desired behavior and instead of[ following line](https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L76-L78):
```
   can :manage, Event do |event|
      event.users.include?(user)
    end
```
We should check for `event.program.cfp_open?` for `:new` and `:create` action.

Blocking: #867

PR #872, opened 2016-03-12
Allow submission of proposal only if program has cfp

sonalkr132:
close: #868

### DIFF
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -2,9 +2,11 @@ class ProposalController < ApplicationController
   before_filter :authenticate_user!, except: [:show, :new, :create]
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true
-  load_and_authorize_resource :event, parent: false, through: :program
+  load_and_authorize_resource :event, parent: false, through: :program, except: [:new, :create]
 
   def index
+    @event = @program.events.new
+    @event.event_users.new(user: current_user, event_role: 'submitter')
     @events = current_user.proposals(@conference)
   end
 
@@ -14,6 +16,9 @@ class ProposalController < ApplicationController
   end
 
   def new
+    @event = @program.events.new
+    @event.event_users.new(user: current_user, event_role: 'submitter') if current_user
+    authorize! :new, @event
     @user = User.new
     @url = conference_program_proposal_index_path(@conference.short_title)
   end
@@ -46,6 +51,7 @@ class ProposalController < ApplicationController
                            event_role: 'submitter')
     @event.event_users.new(user: current_user,
                            event_role: 'speaker')
+    authorize! :new, @event
 
     unless @event.save
       flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -76,8 +76,12 @@ class Ability
     can :manage, Event do |event|
       event.users.include?(user)
     end
-    # can create an event until the last day of a conference
-    can :create, Event, program_id: Conference.where('end_date >= ?', Date.today).map { |conference| conference.program.id}.compact
+
+    # cannot create an event if program does not have open cfp
+    cannot [:new, :create], Event do |event|
+      user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
+      !event.program.cfp_open? || !event.new_record? || !user_inclusion
+    end
 
     # can manage the commercials of their own events
     can :manage, Commercial, commercialable_type: 'Event', commercialable_id: user.events.pluck(:id)
--- a/app/views/proposal/index.html.haml
+++ b/app/views/proposal/index.html.haml
@@ -98,5 +98,5 @@
                   class: 'btn btn-default', id: "edit_proposal_#{event.id}"
   .row
     .col-md-12
-      - if can? :create, @program.events.new
+      - if can? :create, @event
         = link_to "New Proposal", new_conference_program_proposal_path(@conference.short_title), :class => "btn btn-success pull-right"
--- a/spec/factories/events.rb
+++ b/spec/factories/events.rb
@@ -31,7 +31,7 @@ FactoryGirl.define do
       libero quis porta ultricies. Fusce pulvinar accumsan lobortis.
     EOS
     after(:build) do |event|
-      event.event_users << build(:submitter)
+      event.event_users << build(:submitter) unless event.submitter # so that we don't have two submitters
       # set an event_type if none is passed to the factory.
       # needs to be created here because otherwise it doesn't belong to the
       # same conference as the event
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -32,6 +32,9 @@ describe 'User' do
 
     let(:registration) { create(:registration) }
 
+    let(:program_with_cfp) { create(:program, cfp: create(:cfp)) }
+    let(:program_without_cfp) { create(:program) }
+
     # Test abilities for not signed in users
     context 'when user is not signed in' do
       it{ should be_able_to(:index, Conference)}
@@ -58,7 +61,9 @@ describe 'User' do
       it{ should be_able_to(:show, Registration.new)}
       it{ should_not be_able_to(:manage, registration)}
 
-      it{ should be_able_to(:create, Event)}
+      it{ should be_able_to(:new, Event.new(program: program_with_cfp)) }
+      it{ should_not be_able_to(:new, Event.new(program: program_without_cfp)) }
+      it{ should_not be_able_to(:create, Event.new(program: program_without_cfp))}
       it{ should be_able_to(:show, Event.new)}
 
       it{ should_not be_able_to(:manage, :any)}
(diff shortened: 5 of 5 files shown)


app/models/ability.rb
```diff
@@ -73,12 +73,15 @@ def signed_in(user)
 
     can [:create, :destroy], Subscription, user_id: user.id
 
-    can :manage, Event do |event|
+    can [:index, :show, :edit, :destroy, :update, :confirm, :restart], Event do |event|
```
differentreality: Keep :manage and then disable :new and :create

app/models/ability.rb
```diff
@@ -73,12 +73,15 @@ def signed_in(user)
...
       event.users.include?(user)
     end
-    # can create an event until the last day of a conference
-    can :create, Event, program_id: Conference.where('end_date >= ?', Date.today).map { |conference| conference.program.id}.compact
 
+    # can create an event until the last day of a conference
+    # can create an event if program has call for papers defined
+    can [:new, :create], Event do |event|
+      event.program.cfp_open? && event.new_record? && event.program_id == Conference.where('end_date >= ?', Date.today).map { |conference| conference.program.id}.compact
```
differentreality: I am not sure we need the event.program_id.
I know it comes from a model validation, but the model does not validate against cfp_open? and cfp is already before the end of the conference.

sonalkr132: > I am not sure we need the event.program_id

Yeah, we don't. `event.program.cfp_open?` will ensure that cfp is open and is present. cfp has validation for `before end of conference`.

app/models/ability.rb
```diff
@@ -76,9 +76,11 @@ def signed_in(user)
     can :manage, Event do |event|
       event.users.include?(user)
     end
-    # can create an event until the last day of a conference
-    can :create, Event, program_id: Conference.where('end_date >= ?', Date.today).map { |conference| conference.program.id}.compact
 
+    # can create an event if program has call for papers defined
+    can [:new, :create], Event do |event|
+      event.program.cfp_open? && event.new_record?
```
differentreality: Why do we need `&& event.new_record?` here too? And why don't we allow a user to create an Event only for his/her own user?

differentreality:
Take a closer look to your tests please.

app/models/ability.rb
sonalkr132: > Why do we need && event.new_record? here too?

Does `:manage` part already cover us there?

> why don't we allow a user to create an Event only for his/her own user?

An event only for his/her own user? I don't understand what do you mean by that.

hennevogel: `can :manage, Event` covers all cases. `can [:new, :create], Event` [overwrites](https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence) this with rules not dependent on users.

sonalkr132: @hennevogel, so my current implementation is okay? Since refining the rule overwrites the previous one and not supplements it, I should have a check for `new_record?`. Also, `:new` and `:create` action are independent of user ownership.

hennevogel: Yes it's okay.

hennevogel:
@differentreality what about the tests?

sonalkr132:
@differentreality  This PR ensures that users can't submit proposals unless program has cfp, I have added ability test for them:
```Ruby
it{ should_not be_able_to(:new, Event.new(program: program_without_cfp)) }
it{ should_not be_able_to(:create, Event.new(program: program_without_cfp)) }
```

I am not sure what am I missing :confused:

differentreality:
We should be testing that the user creates a new Event when there is an open_cfp `and` the event.users include the user. And that the user cannot create/new//manage an Event for another user, even if the cfp is open.

app/models/ability.rb
differentreality: Hmm what I understood is that a user can create an Event, where the user is included in event.users, even if the cfp_open? is false.

From  https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence

> Adding can rules do not override prior rules, but instead are logically or'ed.

 (And the example below that)
Shouldn't we add negative tests to ensure that this does not happen?

sonalkr132:
> We should be testing that the user creates a new Event when there is an open_cfp.

Doesn't following test cover `:new` and `:create`:
```Ruby
it{ should be_able_to(:manage, user_event_with_cfp) }
```

> and the event.users include the user.

I don't think it should be responsibility of this PR. `user` is added to `event.users` in controller and I have tested for that in controller tests I am adding in #867 

>  And that the user cannot create/new//manage an Event for another user, even if the cfp is open.

Again, I think it is responsibility of controller tests to ensure that `current_user` can't make events for a different user.
 `:new` and `:create` rules I have written does not have any dependence on `user`.

app/models/ability.rb
sonalkr132: >> Adding can rules do not override prior rules, but instead are logically or'ed.

I guess this does mean that we can lose `&& event.new_record?`. This also means that `cannot` overrides but `can` supplements. Probably `cannot` will make more sense here?

```Ruby
 can :manage, Event do |event|
    event.users.include?(user)
 end

cannot [:new, :create], Event do |event|
   !event.program.cfp_open?
end
```

hennevogel:
> I think it is responsibility of controller tests to ensure that current_user can't make events for a different user.

No, that is something for the ability spec

sonalkr132:
This is what I tried:
```Ruby
# ability_spec
let(:user) { create(:user) }     # let username: username3
let(:user2) { create(:user) }   # let username: username4
let(:event_user2) { create(:submitter, user: user2) }

it{ should_not be_able_to(:new, Event.new(program: program_with_cfp, event_users: [event_user2])) }
```
I changed event factory, so that we don't end up creating two submitters:
```
# previously: event.event_users << build(:submitter)
event.event_users ||= build(:submitter)
```

Now, if I check `user` object in ability model file, I see that it is being set to `user2` (username4) and not not `user` (username3). So something is wrong with test, I am not sure what. I am looking into it, please if you know something about it, it would be really helpful.

hennevogel:
This is what you tried **where**?

sonalkr132:
> This is what you tried where?

@hennevogel forget it! I made some progress.

> And that the user cannot create/new//manage an Event for another user.

What that means is that now instead of:
```Ruby
# This would fail cause event.users would not include any users. ie: event.users = [] 
- if can? :create, @program.events.new
```
We will have to:
```Ruby
- if can? :create, @program.events.new(users: [current_user])
```
However, we allow submission of proposal even if user is not signed in. If user is not signed in there is no `user` object to decide whether that non-existent user is included in `event.users`.

app/controllers/proposal_controller.rb
```diff
@@ -2,9 +2,11 @@ class ProposalController < ApplicationController
   before_filter :authenticate_user!, except: [:show, :new, :create]
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true
-  load_and_authorize_resource :event, parent: false, through: :program
+  load_and_authorize_resource :event, parent: false, through: :program, except: [:new, :create]
 
   def index
+    @event = @program.events.new
+    @event.event_users.new(user: current_user, event_role: 'submitter')
```
sonalkr132: I am assigning `@event` here cause in index template we have:
```Ruby
- if can? :create, @event
     = link_to "New Proposal", new_conference_program_proposal_path(@conference.short_title), :class => "btn btn-success pull-right"
```

app/controllers/proposal_controller.rb
```diff
@@ -14,6 +16,9 @@ def show
   end
 
   def new
+    @event = @program.events.new
+    @event.event_users.new(user: current_user, event_role: 'submitter') if current_user
```
sonalkr132: I am using `if current_user` cause same `new` action will be used for non-signed in users[1] and we don't have `current_user` there.

[1] we allow proposal submission for non-signed in user.

app/models/ability.rb
```diff
@@ -76,8 +76,12 @@ def signed_in(user)
...
       event.users.include?(user)
     end
-    # can create an event until the last day of a conference
-    can :create, Event, program_id: Conference.where('end_date >= ?', Date.today).map { |conference| conference.program.id}.compact
+
+    # cannot create an event if program does not have open cfp
+    cannot [:new, :create], Event do |event|
+      user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
+      !event.program.cfp_open? || !event.new_record? || !user_inclusion
+    end
```
sonalkr132: I have to use combination of `can` followed by `cannot` (overrides previous can rule), cause:
```Ruby
can [:new, :create], Event do |event|
   event.program.cfp_open?
end
```
^ will logically OR the two rules, we want to authorize if `AND` rule passes.

app/controllers/proposal_controller.rb
hennevogel: We do? I doubt that.

sonalkr132: @hennevogel Am I missing something here? I can do this on master:
screenshot from 2016-03-23 01-23-58

Code [1] [2] seems to be set up this way too.

[1] https://github.com/openSUSE/osem/blob/master/app/views/proposal/new.html.haml#L25
[2] https://github.com/openSUSE/osem/blob/master/app/controllers/proposal_controller.rb#L17

hennevogel: Of course. Sorry confused `new` with `create`

hennevogel:
Alright. Lets :rocket:

PR #875, opened 2016-03-13
Fix API endpoints conferences/id/rooms/ & tracks/ issue

### LINKED ISSUE #874: API endpoints /api/v1/conferences/shorttitle/rooms/  and tracks/  fails with error
Steps to reproduce:
1. Create a conference,say testconf (short title)
2. Add a track
3. Add a venue and few rooms to that venue

Now  go to /api/v1/conferences/testconf/rooms/
It raises  error: undefined method `venue' for nil:NilClass

Similarly try /api/v1/conferences/testconf/tracks
it raises error: undefined method `tracks'

nishanthvijayan:
Fixes #874

### DIFF
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -8,8 +8,7 @@ module Api
         if params[:conference_id].blank?
           rooms = Room.all
         else
-          conference = Conference.find_by_guid(params[:conference_id])
-          rooms = conference.venue.rooms if conference.venue
+          @conference.venue ? (rooms = @conference.venue.rooms) : (rooms = [])
         end
         respond_with rooms
       end
--- a/app/controllers/api/v1/rooms_controller.rb
+++ b/app/controllers/api/v1/rooms_controller.rb
@@ -5,12 +5,11 @@ module Api
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          rooms = Room.all
+        if @conference
+          respond_with @conference.venue ? @conference.venue.rooms : Room.none
         else
-          @conference.venue ? (rooms = @conference.venue.rooms) : (rooms = [])
+          respond_with Room.all
         end
-        respond_with rooms
       end
     end
   end
--- a/app/controllers/api/v1/tracks_controller.rb
+++ b/app/controllers/api/v1/tracks_controller.rb
@@ -5,7 +5,7 @@ module Api
       respond_to :json
 
       def index
-        @conference ? (tracks = @conference.tracks) : (tracks = Track.all)
+        @conference ? (tracks = @conference.program.tracks) : (tracks = Track.all)
 
         respond_with tracks
       end
--- a/app/controllers/api/v1/tracks_controller.rb
+++ b/app/controllers/api/v1/tracks_controller.rb
@@ -5,7 +5,7 @@ module Api
       respond_to :json
 
       def index
-        @conference ? (tracks = @conference.program.tracks) : (tracks = Track.all)
+        tracks = @conference ? @conference.program.tracks : Track.all
 
         respond_with tracks
       end
--- a/spec/controllers/api/v1/rooms_controller_spec.rb
+++ b/spec/controllers/api/v1/rooms_controller_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper'
+
+describe Api::V1::RoomsController do
+  let!(:conference) { create(:conference) }
+  let!(:venue) { create(:venue, conference: conference) }
+  let!(:conference_room) { create(:room, name: 'Conference Room', venue: venue) }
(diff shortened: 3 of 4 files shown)


differentreality:
Yeap that looks good! Shall we test it?

nishanthvijayan:
Sure.I'll start right away.

nishanthvijayan:
@differentreality  Done.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -8,8 +8,7 @@ def index
         if params[:conference_id].blank?
```
differentreality: Actually this can just be `if @conference` and invert the code of if/else.

spec/requests/api/v1/tracks_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+
+describe 'API tracks' do
+  let!(:conference) { create(:conference) }
+  let!(:track1) { create(:track, program_id: conference.program.id) }
```
differentreality: This track is supposed to have the name 'Example Track', but to check that I would have to go back to the factory. You can set the name of the track here and call it 'Conference Track' or something like that, so that you can also lose the comment on line 34. The name of the variable can also be something like track_of_conference instead of track1, so that it makes more sense.

nishanthvijayan:
@differentreality  I've cleaned up the tests.

differentreality:
@hennevogel do we need to keep the API tests in spec/requests instead of spec/controllers?

hennevogel:
No. `spec/requests` is for [request specs](https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec). These are controller tests.

app/controllers/api/v1/rooms_controller.rb
```diff
@@ -5,11 +5,10 @@ class RoomsController < Api::BaseController
       respond_to :json
 
       def index
-        if params[:conference_id].blank?
-          rooms = Room.all
+        if @conference
+          @conference.venue ? (rooms = @conference.venue.rooms) : (rooms = [])
```
ChrisBr: please don't assign variables in conditionals. This is really hard to read! Instead do sth like:

``
rooms = @conference.venue ? ...
``

Furthermore please don't return an empty Array, instead use ``Room.none`` to stick to the ActiveRecord API.

Long answer, short: Please change the line to:

```
rooms = @conference.venue ? @conference.venue.rooms : Room.none
```

app/controllers/api/v1/tracks_controller.rb
```diff
@@ -5,7 +5,7 @@ class TracksController < Api::BaseController
       respond_to :json
 
       def index
-        @conference ? (tracks = @conference.tracks) : (tracks = Track.all)
+        @conference ? (tracks = @conference.program.tracks) : (tracks = Track.all)
```
ChrisBr: Same here: Please don't assign variables in a conditional statement!

nishanthvijayan: OK.It was already in that form.So I assumed it was okay.I'll fix everything ASAP

app/controllers/api/v1/rooms_controller.rb
nishanthvijayan: Changing it as you suggested would lead to a rubocop offence 
[Style/ConditionalAssignment: Use the return of the conditional for variable assignment and comparison](http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/ConditionalAssignment)
How about this?
```
rooms = if @conference
      @conference.venue ? @conference.venue.rooms : Room.none
    else
      Room.all
    end
```

ChrisBr: Hm, I don't like the rubocop suggestions ...
Do we need the rooms variable at all?

What about?
```
if @conference
  respond_with @conference.venue ? @conference.venue.rooms : Room.none
else
  respond_with Room.all
end

(note that this is untested...)
```

app/controllers/api/v1/tracks_controller.rb
ChrisBr: Okay! Ever heard from the [boyscout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule)? ;)

app/controllers/api/v1/rooms_controller.rb
nishanthvijayan: Yup that works.

app/controllers/api/v1/tracks_controller.rb
nishanthvijayan: Sorry my bad.

ChrisBr:
@nishanthvijayan cool :+1: Good work!

PR #880, opened 2016-03-15
Update active_model_serializers.Fix conference serializer and api tests

### LINKED ISSUE #879: API controllers are not using their corresponding serializers.
The API controllers are not using their corresponding serializers, even when they are specified explicitly.
Fixing the above will break api/v1/conferences as its serializer has some errors.

nishanthvijayan:
The controllers are not using the serializers, because of a bug in the version of active_model_serializer that we currently use.Hence the update.
API tests were fixed to work with the new API response structure.
Fixes bugs in conference_serializer as well.
Fixes #879


ChrisBr:
LGTM! Thanks @nishanthvijayan :+1:

Issue #881, opened 2016-03-15
events#index datatable state not persistent


hennevogel:
You sort the datatable of events by `Type`, then change the `Track` of one of the `Events` with the dropdown. The subsequent reload of `events#index` sorts the table by `id` again.

The datatable should save the state.

rishabhs95:
interested to work with this bug... @hennevogel

hennevogel:
No need to ask. Just go ahead :-)

rishabhs95:
okay :)

differentreality:
Fixed in #882

PR #882, opened 2016-03-15
enabling state save feature in datatables

rishabhs95:
fixing bug #881.
enabling stateSave feature of data tables.


ChrisBr:
Cool! Just tested it and works for me!

:+1:

PR #884, opened 2016-03-16
Remove datatable functionality from Roles index table.

nishanthvijayan:
The table in roles#index view has no need of datatable features as the number of roles is just 4..

roles_datatabe_without
roles_datatable_with


hennevogel:
LGTM. Thanks

Issue #802, opened 2016-03-01
ProposalController#confirm: Redirect of unregistered user to registration page


sonalkr132:
Following is the code of concern:

``` Ruby
def confirm
    #  code unrelated to this issue is omitted.
    if @conference.user_registered?(current_user)
      redirect_to(conference_program_proposal_index_path(@conference.short_title),
                  notice: 'The proposal was confirmed.')
    else
      redirect_to(new_conference_conference_registrations_path(conference_id: @conference.short_title),
                  alert: 'The proposal was confirmed. Please register to attend the conference.')
    end
  end
```

The redirect of unregistered user to new registration page (else part) should only happen if conference is open for registration.

differentreality:
We should also consider if we want https://github.com/openSUSE/osem/issues/829 or not

AEtherC0r3:
While working on #829 i will add the final touches so this can be closed too

PR #886, opened 2016-03-17
Fix authorization of new and create in ConferenceRegistrationsController

sonalkr132:
Closes: #791
Change in proposal spec cause conference has to have valid registration period for successful redirect to registration after confirm proposal.

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -1,10 +1,13 @@
 class ConferenceRegistrationsController < ApplicationController
   before_filter :authenticate_user!, except: [:new, :create]
   load_resource :conference, find_by: :short_title
-  authorize_resource :conference_registrations, class: Registration
+  authorize_resource :conference_registrations, class: Registration, except: [:new, :create]
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
+    @registration = Registration.new(conference_id: @conference.id)
+    authorize! :new, @registration
+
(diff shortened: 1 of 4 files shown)


hennevogel:
Phew. I don't understand your description. Please elaborate

sonalkr132:
@hennevogel Proposal confirmation gets redirected to new registration page. We need conference to be open for registration for the redirect.

We won't need it, after we have solved #802

hennevogel:
Do we want to allow this? 

1. Someone unregistered submits a Talk to our conference
2. We allow registration until next Friday
3. Someone unregistered confirms his event on Monday

@differentreality @ChrisBr opinions?

differentreality:
That workflow is currently possible. The thing is that if a speaker confirms, that means the speaker is going to attend the conference. It could make sense that the speaker is able to register.

It could, however, mean that people registered after the registration period closes, don't get to have some things that are time-sensitive, ie. badges that had to be printed on a certain date right after registration period closed. 
Of course the discussion is only about confirmed speakers, and you would expect that organizers would provision for their speakers separately.

But that is not related to this particular PR now, is it? 
Unless we also want to integrate provision for confirmed speakers within this PR, it makes total sense not to redirect to a page where the user has no access. And the user should not have access to new registration page, unless registration period is open.

hennevogel:
Well this workflow is what what this PR fixes. I think we should think about this whole cycle more. I'll record an issue

hennevogel:
But let's merge this for now

Issue #877, opened 2016-03-14
incorrect number of tickets shown in admin tickets view


rishabhs95:
admin tickets view shows incorrect number of tickets purchased by users.

for example, two users buy different number of tickets:
<img width="342" alt="screen shot 2016-03-14 at 6 07 59 pm" src="https://cloud.githubusercontent.com/assets/10002834/13745695/79fabc38-ea15-11e5-93ca-971915a31429.png">
<img width="538" alt="screen shot 2016-03-14 at 6 08 23 pm" src="https://cloud.githubusercontent.com/assets/10002834/13745697/7b0664c4-ea15-11e5-979c-fead4a583863.png">

still admin gets the information that number of tickets sold = 2 (number of users who purchased the tickets)

<img width="881" alt="screen shot 2016-03-14 at 6 07 34 pm" src="https://cloud.githubusercontent.com/assets/10002834/13745700/7cf275f2-ea15-11e5-8b39-acd4c9de75bf.png">

shouldn't it show the total quantity of tickets sold for each ticket?

rishabhs95:
working with this bug

rishabhs95:
@ChrisBr the sold value in admin's ticket view for a conference should show the total number of tickets sold right? 
like in above example, it should've been : 
- 36 for 1$ ticket
- 28 for 10$ ticket
- 33 for 90$ ticket

differentreality:
The table header says **sold**, but the value it shows is `= ticket.buyers.count`
Perhaps we should have column for `Sold` and another column for `Buyers`

rishabhs95:
okay @differentreality ... i will do that  :)

hennevogel:
Who cares how many Buyers there are?

rishabhs95:
i think that the main thing is to display the correct number of total tickets bought... but we may display the number of buyers if it is required.
which would be better?

hennevogel:
You can also display a squirrel eating a nut
 
![tumblr_n6ie66wsvo1rpe379o1_r1_500-1](https://cloud.githubusercontent.com/assets/514785/13846347/ac5a2bea-ec46-11e5-9759-56f1a6d21094.gif)

The question is **why** would you? ;-) This is an APP. Display data that is relevant for the people using it :-) In my opinion this is

Ticket | Price | Sold | Total Amount
------------ | ------------- | ------------- | -------------
Supporter | $15 | 10 | $150

rishabhs95:
[...]

PR #889, opened 2016-03-18
correcting ticket's details view for admin.

rishabhs95:
fixing bug #877 .

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -48,6 +48,14 @@ class Ticket < ActiveRecord::Base
     result ? result : Money.new(0, 'USD')
   end
 
+  def tickets_sold
+    ticket_purchases.sum(:quantity)
+  end
+
+  def tickets_turnover
+    tickets_sold * price
+  end
+
   private
 
   def tickets_of_conference_have_same_currency
--- a/app/views/admin/tickets/index.html.haml
+++ b/app/views/admin/tickets/index.html.haml
@@ -10,9 +10,9 @@
       %table.table.table-hover#tickets
         %thead
           %th Title
-          %th Description
           %th Price
           %th Sold
+          %th Turnover
           %th Actions
         %tbody
           - @conference.tickets.each do |ticket|
@@ -21,11 +21,11 @@
                 = link_to(admin_conference_ticket_path(@conference.short_title, ticket.id)) do
                   = ticket.title
               %td
-                = truncate(ticket.description)
-              %td
                 = humanized_money_with_symbol ticket.price
               %td
-                = ticket.buyers.count
+                = ticket.tickets_sold
+              %td
+                = humanized_money_with_symbol ticket.tickets_turnover
               %td
                 .btn-group
                   = link_to 'Edit', edit_admin_conference_ticket_path(@conference.short_title, ticket.id),
--- a/app/views/admin/tickets/show.html.haml
+++ b/app/views/admin/tickets/show.html.haml
@@ -2,10 +2,11 @@
   .col-md-12
     .page-header
       %h1
-        = @ticket.title
-        Ticket
-        %small
-          = humanized_money_with_symbol @ticket.price
+        %div{"data-placement" => "left", "data-toggle" => "tooltip", :title => @ticket.description}
+          = @ticket.title
+          Ticket
+          %small
+            = humanized_money_with_symbol @ticket.price
       %p.text-muted
         People who bought this ticket
 .row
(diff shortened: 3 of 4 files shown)


app/views/admin/tickets/index.html.haml
```diff
@@ -10,9 +10,9 @@
       %table.table.table-hover#tickets
         %thead
           %th Title
-          %th Description
           %th Price
           %th Sold
+          %th Total Price
```
hennevogel: It's not the total price. It's the **Turnover**

app/views/admin/tickets/index.html.haml
```diff
@@ -21,11 +21,11 @@
...
                   = ticket.title
               %td
-                = truncate(ticket.description)
-              %td
                 = humanized_money_with_symbol ticket.price
               %td
-                = ticket.buyers.count
+                = ticket.ticket_purchases.inject(0){|sum,tp| sum+tp.quantity}
+              %td
+                = humanized_money_with_symbol ticket.ticket_purchases.inject(0){|sum,tp| sum+tp.quantity}*ticket.price
```
hennevogel: DRY this and move it to two methods (`#sold` and `#turnover`) in `Ticket`

app/views/admin/tickets/show.html.haml
```diff
@@ -6,6 +6,11 @@
         Ticket
         %small
           = humanized_money_with_symbol @ticket.price
+      - if @ticket.description?
```
hennevogel: This is potentially very large. Make this a [tooltip or popover](http://getbootstrap.com/javascript/#tooltips) on `@ticket.title`

app/controllers/admin/tickets_controller.rb
```diff
@@ -2,6 +2,7 @@ module Admin
   class TicketsController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :ticket, through: :conference
+    helper_method :tickets_sold, :tickets_turnover
```
ChrisBr: Huh? Why? Why not a model method?

app/assets/javascripts/application.js
```diff
@@ -49,4 +49,7 @@ $(document).ready(function() {
     $('body').smoothScroll({
         delegateSelector: 'a.smoothscroll'
     });
+    $(function () {
```
ChrisBr: Please don't to this in the application.js 
``
For performance reasons, the Tooltip and Popover data-apis are opt-in, meaning you must initialize them yourself.
``

rishabhs95: should i shift it to `osem-tickets.js` then?

app/controllers/admin/tickets_controller.rb
rishabhs95: yes... i just noticed it is better to shift most of the logic to models :+1:

app/models/ticket.rb
```diff
@@ -48,6 +48,13 @@ def self.total_price(conference, user)
     result ? result : Money.new(0, 'USD')
   end
 
+  def self.tickets_turnover(ticket_id, conference_id)
```
ChrisBr: Ever heard from instance methods ;)
http://www.railstips.org/blog/archives/2009/05/11/class-and-instance-methods-in-ruby/

app/models/ticket_purchase.rb
```diff
@@ -53,4 +53,10 @@ def self.update_quantity(conference, quantity, ticket, user)
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  def self.tickets_sold(ticket_id, conference_id)
```
ChrisBr: This method also belongs to ticket and probably you can do the inject in a database query for better performance ...

app/models/ticket.rb
```diff
@@ -48,6 +48,15 @@ def self.total_price(conference, user)
     result ? result : Money.new(0, 'USD')
   end
 
+  def tickets_sold
+    TicketPurchase.where(ticket_id: id, conference_id: conference_id).inject(0) { |a, e| a + e.quantity }
```
ChrisBr: ``ticket_purchases.sum(:quantity)`` should also work!
(Note it's untested...)

app/models/ticket.rb
```diff
@@ -48,6 +48,15 @@ def self.total_price(conference, user)
...
+  end
+
+  def tickets_turnover
+    price = Money.new(price_cents, price_currency)
```
ChrisBr: I assume that you don't need this line, just ``tickets_sold * price`` should be fine!

rishabhs95:
pushed with all suggested changes

ChrisBr:
LGTM

Issue #598, opened 2015-02-10
Make use of Rails I18n data format in EventSerializer


kalabiyau:
https://github.com/openSUSE/osem/blob/master/app/serializers/event_serializer.rb#L6

We need to simplify usage of data formatting using http://guides.rubyonrails.org/i18n.html#adding-date-time-formats

Shriyanshagro:
@kalabiyau , I guess I can solve the issue , but can you please elaborate it more. What exactly is the requirement ??

kalabiyau:
@Shriyanshagro we have at the location a string 

```ruby
t.blank? ? '' : %{ #{t.time.strftime('%Y-%m-%dT%H:%M:%S')}#{t.formatted_offset(false)} }
```

which is formatting of datetime done in php manner
Instead we should rather rely on `I18n.l` and format defined in locale file

Shriyanshagro:
[https://github.com/Shriyanshagro/osem/commit/775e3d8bcab5d0e8399541ccfb2c6de53dee473a](url)
using of gem I18n , made data format Internationalized in EventSerializer.

sonalkr132:
We can close this, now we are using I18n data format in EventSerializer. @kalabiyau Please open a new issue with your recent suggestions.

PR #892, opened 2016-03-18
Make use of Rails I18n data format in EventSerializer #598

Shriyanshagro:
Internationalized data format , Can be used in future , handling configuration for different locales.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,10 @@ gem 'rolify'
 # to show flash messages from ajax requests
 gem 'unobtrusive_flash', '>=3'
 
+
+# Internationalizing gem , use gem according to version of your rails application
+gem 'rails-i18n', '~> 4.0.0' # For 4.0.x
+
 # as state machine
 gem 'transitions', :require => %w( transitions active_record/transitions )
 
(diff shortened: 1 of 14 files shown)


Gemfile
```diff
@@ -40,6 +40,12 @@ gem 'rolify'
 # to show flash messages from ajax requests
 gem 'unobtrusive_flash', '>=3'
 
+
+# Internationalizing gem , use gem according to version of your rails application
+# gem 'rails-i18n', '~> 5.0.0.beta1' # For 5.0.0.beta1
+gem 'rails-i18n', '~> 4.0.0' # For 4.0.x
+# gem 'rails-i18n', '~> 3.0.0' # For 3.x
```
hennevogel: We know which rails version we use don't we? :-)

Shriyanshagro: Yeah we know, I just comment it out in case if we ever switch to higher version.
So should I remove commented lines ??

hennevogel: Yes you should

Shriyanshagro: task done :-)

hennevogel:
respect_the_rubicop

Issue #860, opened 2016-03-10
Ask proposal submitter to add track only if conf has one


sonalkr132:
Steps to reproduce:
* Make a conference and make sure that you haven't specified any track
* Set cfp dates of the conf
* Go to home page and submit a proposal
* Click on `Complete your proposal`, tooltip will appear with todo list asking you to `add a track` (even if there is no track to add)
* Click on `add a track`, you will go to edit page of proposal. You won't see the option to add track.

I wish it was as easy as adding an if statement to `_tooltip.html.haml`. [proposal status](https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L187) needs to be considered too.

sonalkr132:
Fixed after merge of #893

PR #893, opened 2016-03-19
Proposal submitter should be asked to add track only if conf has track

nishanthvijayan:
Solves #860 
progress_status['track'] will be:  
- nil if conf doesn't have track
-    false if conf has track but the event doesn't
-    true if conf and event  has a track

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -29,11 +29,15 @@ module ApplicationHelper
   end
 
   def event_progress_color(progress)
-    case progress.to_i
-    when 100 then 'progress-bar-success'
-    when 85 then 'progress-bar-info'
-    when 71 then 'progress-bar-warning'
-    when 14, 28, 42, 57 then 'progress-bar-danger'
+    progress = progress.to_i
+    if progress == 100
+      'progress-bar-success'
+    elsif progress >= 85
+      'progress-bar-info'
+    elsif progress >= 71
+      'progress-bar-warning'
+    else
+      'progress-bar-danger'
     end
   end
 
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -190,7 +190,7 @@ class Event < ActiveRecord::Base
       commercials: self.commercials.any?,
       biography: !self.submitter.biography.blank?,
       subtitle: !self.subtitle.blank?,
-      track: !self.track.blank?,
+      track: (!self.track.blank? unless self.program.tracks.empty?),
       difficulty_level: !self.difficulty_level.blank?,
       title: true,
       abstract: true
@@ -204,7 +204,7 @@ class Event < ActiveRecord::Base
   # * +String+ -> Progress in Percent
   def calculate_progress
     result = self.progress_status
-    (100 * result.values.count(true) / result.values.count).to_s
+    (100 * result.values.count(true) / result.values.compact.count).to_s
   end
 
   private
--- a/app/views/proposal/_tooltip.html.haml
+++ b/app/views/proposal/_tooltip.html.haml
@@ -15,9 +15,10 @@
   %li{'class'=>class_for_todo(progress_status['commercials'])}
     %span{'class'=>icon_for_todo(progress_status['commercials'])}
       = link_to 'Add a commercial', edit_conference_program_proposal_path(event.program.conference.short_title, event, anchor: 'commercials-content')
-  %li{'class'=>class_for_todo(progress_status['track'])}
-    %span{'class'=>icon_for_todo(progress_status['track'])}
-      = link_to 'Add a track', edit_conference_program_proposal_path(event.program.conference.short_title, event)
+  - unless progress_status['track'].nil?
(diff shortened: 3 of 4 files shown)


app/models/event.rb
```diff
@@ -190,7 +190,7 @@ def progress_status
       commercials: self.commercials.any?,
       biography: !self.submitter.biography.blank?,
       subtitle: !self.subtitle.blank?,
-      track: !self.track.blank?,
+      track: (!self.track.blank? if !self.program.tracks.empty?),
```
ChrisBr: `unless self.program.tracks.empty?`

app/helpers/application_helper.rb
```diff
@@ -29,11 +29,15 @@ def event_status_icon(event)
   end
 
   def event_progress_color(progress)
-    case progress.to_i
-    when 100 then 'progress-bar-success'
-    when 85 then 'progress-bar-info'
-    when 71 then 'progress-bar-warning'
-    when 14, 28, 42, 57 then 'progress-bar-danger'
+    progress = progress.to_i
```
ChrisBr: Why do you change this?

ChrisBr:
Basically I understand the idea and I like it, however, I don't understand why all these changes are necessary? Just checking in the view and don't display the link anymore should do the trick ...

app/views/proposal/index.html.haml
```diff
@@ -76,7 +76,7 @@
                 - progress_percentage = event.calculate_progress
                 .progress
                   %div{class: "progress-bar #{event_progress_color(progress_percentage)}", style: "width:#{progress_percentage}%;"}
-                    = event.progress_status.reject{ |_key, value| value }.length
+                    = event.progress_status.reject{ |_key, value| value or value.nil? }.length
                     left
```
nishanthvijayan: This was changed so that the display text of the progress bar does not count "Add a track" if no track is present.

nishanthvijayan:
That's not enough.The progress bar text and progress bar color should also not consider "Add a track" if no track is present.Hence the other changes.

app/models/event.rb
```diff
@@ -204,7 +204,7 @@ def progress_status
   # * +String+ -> Progress in Percent
   def calculate_progress
     result = self.progress_status
-    (100 * result.values.count(true) / result.values.count).to_s
+    (100 * result.values.count(true) / result.values.compact.count).to_s
```
nishanthvijayan: Consider a scenario where a conf has no tracks and submitter has performed all the tasks presented in the tooltip.In this case, the progress bar should be at 100%.
So when calculating progress percentage, "Add a track" task should be ignored if no track is available for the conference.

app/helpers/application_helper.rb
nishanthvijayan: The numbers being checked are wrong and the progress bar has been defaulting to its default color.

app/models/event.rb
nishanthvijayan: Done.

app/helpers/application_helper.rb
differentreality: You could rewrite case like this http://stackoverflow.com/a/6208739
@ChrisBr which way would you prefer?

app/views/proposal/_tooltip.html.haml
```diff
@@ -15,9 +15,10 @@
   %li{'class'=>class_for_todo(progress_status['commercials'])}
     %span{'class'=>icon_for_todo(progress_status['commercials'])}
       = link_to 'Add a commercial', edit_conference_program_proposal_path(event.program.conference.short_title, event, anchor: 'commercials-content')
-  %li{'class'=>class_for_todo(progress_status['track'])}
-    %span{'class'=>icon_for_todo(progress_status['track'])}
-      = link_to 'Add a track', edit_conference_program_proposal_path(event.program.conference.short_title, event)
+  - if !progress_status['track'].nil?
```
differentreality: - unless progress_status['track'].nil?

app/helpers/application_helper.rb
ChrisBr: No, I'm fine with this solution!

app/models/event.rb
ChrisBr: Makes sense!

app/views/proposal/index.html.haml
```diff
@@ -76,7 +76,7 @@
                 - progress_percentage = event.calculate_progress
                 .progress
                   %div{class: "progress-bar #{event_progress_color(progress_percentage)}", style: "width:#{progress_percentage}%;"}
-                    = event.progress_status.reject{ |_key, value| value }.length
+                    = event.progress_status.reject{ |_key, value| value or value.nil? }.length
```
ChrisBr: `|_key, value| value || value.nil?`

http://phrogz.net/programmingruby/language.html#table_18.4

(However, rubocop should complain about this ...)

nishanthvijayan: Done

hennevogel:
LGTM

Issue #888, opened 2016-03-18
NoMethodError in Admin::EventsController#comment


sonalkr132:
Steps to reproduce:
* Make a conference (say `goto`), assign cfp dates
* Visit home page and submit a proposal to `goto`
* Visit `events` show page through admin area (url: http://localhost:3000/admin/conference/goto/program/events/1`)
* Click on `comments (0)`, a textarea will appear.
* Fill in your comment and click `Add comment`

You will get:
> NoMethodError in Admin::EventsController#comment
undefined method `permit' for "sdfsdfsd":String

pointing to following [line](https://github.com/openSUSE/osem/blob/813658fb196cd00c2a95da419e68158d663b1343/app/controllers/admin/events_controller.rb#L179):
```Ruby
    def comment_params
        params.require(:comment).permit(:commentable, :body, :user_id)  # <---
    end
```

differentreality:
Closed in #894

Rolence:
it works.
![fixed_comment_view](https://cloud.githubusercontent.com/assets/11888675/13980636/fef56af2-f0df-11e5-9a56-fbb0e45916bf.png)

PR #894, opened 2016-03-20
rewrite comments form, use strong parameters with comments

differentreality:
https://github.com/openSUSE/osem/issues/888

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -94,7 +94,9 @@ module Admin
     end
 
     def comment
-      comment = Comment.build_from(@event, current_user.id, comment_params)
+      comment = Comment.new(comment_params)
+      comment.commentable = @event
+      comment.user_id = current_user.id
       comment.save!
       if !params[:parent].nil?
         comment.move_to_child_of(params[:parent])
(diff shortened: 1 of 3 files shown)


app/controllers/admin/events_controller.rb
```diff
@@ -94,7 +94,9 @@ def edit
     end
 
     def comment
-      comment = Comment.build_from(@event, current_user.id, comment_params)
+      comment = Comment.new(comment_params)
+      comment.commentable = @event
+      comment.user_id = current_user.id
```
hennevogel: Hm just for my understanding: What's the difference?

differentreality: .build_from expects an event instance, an id of user, and a string. Instead of passing a string I am passing comment_params, which permits body attribute (body attribute holds the string of the actual comment).
All that to implement strong params for the controller. 
(And it also makes it easier for the future to add new fields to Comment model.)

differentreality: In fact, we could also eliminate user_id from controller and pass it with a hidden field from the view form.

hennevogel:
Alright. LGTM!

PR #897, opened 2016-03-22
Update routes.rb

4bic:
removed index get '/stats' => 'stats#index' as it references to a non-existent statscontroller


differentreality:
:+1: Thanks for contributing @4bic!

PR #901, opened 2016-03-23
Add test for conference registration controller show and destroy action

sonalkr132:
After failed delete, user should be redirected to registration show and not
root.
A conference can have only one registration corresponding to each user, hence:
TODO (in a different PR):  change conference_registrations to singular resource


hennevogel:
LGTM

PR #904, opened 2016-03-23
Made use of Rails I18n data format in EventSerializer #598

Shriyanshagro:
Internationalized data format , Can be used in future , handling configuration for different locales.

### DIFF
--- a/app/serializers/event_serializer.rb
+++ b/app/serializers/event_serializer.rb
@@ -5,7 +5,7 @@ class EventSerializer < ActiveModel::Serializer
 
   def date
     t = object.start_time
-    t.blank? ? '' : %{ #{t.time.strftime('%Y-%m-%dT%H:%M:%S')}#{t.formatted_offset(false)} }
+    t.blank? ? '' : %{ #{l t, format: :short}#{t.formatted_offset(false)} }
   end
 
   def speaker_ids
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,9 @@ gem 'rails-observers'
 # for tracking data changes
 gem 'paper_trail'
 
+# Internationalizing gem , use gem according to version of your rails application
+gem 'rails-i18n', '~> 4.0.0' # For 4.0.x
+
 # as authentification framework
 gem 'devise'
 gem 'devise_ichain_authenticatable'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -357,6 +357,9 @@ GEM
       rails-deprecated_sanitizer (>= 1.0.1)
     rails-html-sanitizer (1.0.3)
       loofah (~> 2.0)
+    rails-i18n (4.0.8)
+      i18n (~> 0.7)
+      railties (~> 4.0)
     rails-observers (0.1.2)
       activemodel (~> 4.0)
     railties (4.2.5.2)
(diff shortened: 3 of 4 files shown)


ChrisBr:
@Shriyanshagro what's about #892?

Shriyanshagro:
@ChrisBr I found some merging issue with the PR #892 , so I tried re-writing the code again .
But it too didn't work. Also on #892 unnecessarily many merge have been provoked , so it's better to remove #892 .

hennevogel:
LGTM

app/serializers/event_serializer.rb
```diff
@@ -5,7 +5,7 @@ class EventSerializer < ActiveModel::Serializer
 
   def date
     t = object.start_time
-    t.blank? ? '' : %{ #{t.time.strftime('%Y-%m-%dT%H:%M:%S')}#{t.formatted_offset(false)} }
+    t.blank? ? '' : %{ #{l t, format: :short}#{t.formatted_offset(false)} }
```
differentreality: undefined method `l' for #<EventSerializer:0x007fe87e8f5728>
Anyone else gets that?

nishanthvijayan: Yup. Same here

nishanthvijayan: http://stackoverflow.com/a/11724935

Shriyanshagro:
@differentreality , one can visit the links to get the purpose of 'l'
[http://guides.rubyonrails.org/i18n.html#adding-date-time-formats](url)
[http://guides.rubyonrails.org/i18n.html#the-public-i18n-api](url)

differentreality:
Sure but it is not working for me when I have scheduled events. Have you tried it? Can you give me a screenshot of the result you see?
(Also please fix the links in your comments)

kalabiyau:
@differentreality @Shriyanshagro 

I would like to re-phrase that issue. It looks we are on the wrong track here.

1) Serializers are only exposed in the API, hence we have mobile app affected by them
2) let's look into what was produced before PR

```
irb(main):002:0>  %{ #{Time.now.strftime('%Y-%m-%dT%H:%M:%S')}#{Time.now.formatted_offset(false)} }
=> " 2016-04-04T09:28:59+0200 "
```

Generally wrong string, which is not really js oriented.

Instead of messing around with specific format for dates we better can do:


```diff
-  attributes :guid, :title, :length, :date, :language, :abstract, :speaker_ids, :type, :room, :track
-
-  def date
-    t = object.start_time
-    t.blank? ? '' : %{ #{l t, format: :short}#{t.formatted_offset(false)} }
-  end
+  attributes :guid, :title, :length, :language, :abstract,
+             :speaker_ids, :type, :room, :track, :start_time
```

where 

1) we replace ambiguous date attribute to very straightforward start_time
2) we rely on correct Ruby->JSON serialization process which returns in case of present
start_time - `2014-04-30T22:16:20.171Z` which is perfectly valid timestamp with TZ info. Easy to consume, and no need for additional wrappers and formats.

Here is a draft spec covering the proposal

```diff
+require 'spec_helper'
+
+describe EventSerializer do
+  subject { build(:event) }
+
+  describe '#start_time' do
+
+    it 'is empty when event has no start_time' do
+      subject.start_time = nil
+      actual = JSON.parse(described_class.new(subject).to_json)['event']
+      expect(actual['start_time']).to be nil
+    end
+
+    it 'is set to formatted JSON time string when start_time is populated' do
+      defined_time = Time.now
+      subject.start_time = defined_time
+      actual = JSON.parse(described_class.new(subject).to_json)['event']
+      expect(Time.parse(actual['start_time']).utc).to be_within(1.second).of defined_time.utc
+    end
+  end
+
+end
```

Issue #760, opened 2016-02-01
Configure language of events


davidsf:
In some conference the events could be in various languages. It would be great that admins could configure the accepted languages in the backend and the user in the proposal selects the language of his event.

Shriyanshagro:
Good Idea ..
I can work on it . Can someone guide me on it ??
@ChrisBr @differentreality

hennevogel:
@Shriyanshagro If you want guidance come up with concrete questions please :-) What is it you need guidance for?

Ana06:
I'll do it! :wink: 
Where should I put the languages configuration in the backend? I think that it can be another page inside the program section of the backend.
How should admins select the language? I think that a list of check-boxes would be easy to use, specially if a lot of languages want to be selected.

hennevogel:
Yes `Program` sounds like the object to hold this info. I think for starters this can be a simple text field with comma separated value of ISO 639-1 language codes and a select for `Event`

PR #905, opened 2016-03-23
Languages

Ana06:
Issue #760

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -219,6 +219,7 @@ GEM
     i18n_data (0.7.0)
     inversion (0.12.3)
       loggability (~> 0.4)
+    iso-639 (0.2.5)
     jquery-datatables-rails (2.2.3)
       jquery-rails
       sass-rails
@@ -560,6 +561,7 @@ DEPENDENCIES
   guard-rspec (~> 4.2.8)
   haml-rails
   hoptoad_notifier (~> 2.3)
+  iso-639
   jquery-datatables-rails (~> 2.2.1)
   jquery-rails
   leaflet-rails
@@ -615,6 +617,3 @@ DEPENDENCIES
   web-console (~> 2.0)
   webmock
   whenever
-
-BUNDLED WITH
-   1.11.2
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -21,10 +21,12 @@ class ProposalController < ApplicationController
     authorize! :new, @event
     @user = User.new
     @url = conference_program_proposal_index_path(@conference.short_title)
+    @languages = @program.languages_list
   end
 
   def edit
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
+    @languages = @program.languages_list
   end
 
   def create
@@ -152,7 +154,7 @@ class ProposalController < ApplicationController
   def event_params
     params.require(:event).permit(:event_type_id, :track_id, :difficulty_level_id,
                                   :title, :subtitle, :abstract, :description,
-                                  :require_registration, :max_attendees)
+                                  :require_registration, :max_attendees, :language)
   end
 
   def user_params
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -52,6 +52,7 @@ class Program < ActiveRecord::Base
 
   before_create :create_event_types
   before_create :create_difficulty_levels
+  validate :check_languages_format
 
   ##
   # Checcks if the program has rating enabled
@@ -83,6 +84,10 @@ class Program < ActiveRecord::Base
     (!conference.email_settings.program_schedule_public_subject.blank? && !conference.email_settings.program_schedule_public_body.blank?)
   end
 
+  def languages_list
+    self.languages.split(',').map {|l| ISO_639.find(l).english_name} if self.languages.present?
+  end
+
   private
 
   ##
@@ -112,5 +117,21 @@ class Program < ActiveRecord::Base
                                                 description: 'Events require expert knowledge of the topic.',
                                                 color: '#EF6E69')
     true
+  end
+
+  ##
+  # Check if languages string has the right format. Used as validation.
+  #
+  def check_languages_format
+    return unless self.languages.present?
+    # All white spaces are removed to allow languages to be separated by ',' and ', '. The languages string without spaces is saved
+    self.languages = self.languages.delete(' ').downcase
+    errors.add(:languages, 'must be two letters separated by commas') && return unless
+    self.languages.match(/^$|(\A[a-z][a-z](,[a-z][a-z])*\z)/).present?
+    languages_array = self.languages.split(',')
+    # We check that languages are not repeated
+    errors.add(:languages, "can't be repeated") && return unless languages_array.uniq!.nil?
+    # We check if every language is a valid ISO 639-1 language
+    errors.add(:languages, 'must be ISO 639-1 valid codes') unless languages_array.select{ |x| ISO_639.find(x).nil? }.empty?
   end
 end
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -22,6 +22,9 @@
           %b Submitter
         %th
           %b Speaker
+        -if @program.languages.present?
+          %th
+            %b Language
         %th
           %b Requires Registration
         %th
@@ -82,6 +85,10 @@
               = link_to speaker.name, admin_user_path(speaker)
             - else
               Unknown speaker
+
+          -if @program.languages.present?
+            %td
+              = event.language
 
           %td.text-center{'data-order' => "#{event.require_registration}"}
             = check_box_tag @conference.short_title, event.id, event.require_registration,
--- a/app/views/admin/programs/_form.html.haml
+++ b/app/views/admin/programs/_form.html.haml
@@ -4,9 +4,10 @@
       %h1 Program
 .row
   .col-md-8
-    = semantic_form_for(@program, :url => admin_conference_program_path(@conference.short_title),:html => {:multipart => true}) do |f|
+    = semantic_form_for(@program, url: admin_conference_program_path(@conference.short_title), html: {multipart: true}) do |f|
       = f.input :schedule_public, label: "Show Schedule on the home and splash page"
       = f.input :schedule_fluid, label: "Allow submitters to change their event after it is scheduled"
-      = f.input :rating, :hint => "Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals."
+      = f.input :rating, hint: 'Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals.'
+      = f.input :languages, hint: "Enter the languages allowed for events as values of #{link_to('ISO 639-1', 'http://www.loc.gov/standards/iso639-2/php/code_list.php', target: "_blank")} language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages.".html_safe
       %p.text-right
-        = f.action :submit, :as => :button, :button_html => {:class => "btn btn-primary"}
+        = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}
--- a/app/views/proposal/_proposal_form.html.haml
+++ b/app/views/proposal/_proposal_form.html.haml
@@ -14,6 +14,11 @@
       data: { min_words: type.minimum_abstract_length, max_words: type.maximum_abstract_length }]},
       include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' }
 
+    - if @program.languages.present?
+      = f.input :language, as: :select,
+        collection: @languages,
+        include_blank: false, label: 'Language', input_html: { class: 'select-help-toggle' }
+
     - @conference.program.event_types.each do |event_type|
       %span{ class: 'help-block select-help-text event_event_type_id collapse', id: "#{event_type.id}-help" }
         = event_type.description
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -266,6 +266,7 @@ ActiveRecord::Schema.define(version: 20160427104236) do
     t.boolean  "schedule_fluid",  default: false
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.string   "languages"
   end
 
   create_table "qanswers", force: :cascade do |t|
--- a/spec/models/program_spec.rb
+++ b/spec/models/program_spec.rb
@@ -58,4 +58,37 @@ describe Program do
       expect(conference.program.difficulty_levels.count).to eq 3
     end
   end
+
+  describe 'languages' do
+    it "is not valid if languages aren't two letters separated by commas" do
+      program.languages = 'eng, De es'
+      expect(program.valid?).to eq false
+      expect(program.errors[:languages]).to eq ['must be two letters separated by commas']
+    end
+
+    it 'is not valid if languages are repeated' do
+      program.languages = 'en,de,es,en'
+      expect(program.valid?).to eq false
+      expect(program.errors[:languages]).to eq ["can't be repeated"]
+    end
+
+    it "is not valid if languages aren't ISO 639-1 valid codes" do
+      program.languages = 'en,hh,yu,zi,oo'
+      expect(program.valid?).to eq false
+      expect(program.errors[:languages]).to eq ['must be ISO 639-1 valid codes']
+    end
+
+    it 'is valid otherwise' do
+      program.languages = 'en,De, ES, ru,el'
+      expect(program.valid?).to eq true
+    end
+  end
+
+  describe '#languages_list' do
+    it 'returns the list of readable languages' do
+      program.languages = 'en,de,fr,ru,zh'
+      expect(program.languages_list).to eq %w(English German French Russian Chinese)
+    end
+  end
+
 end
--- a/Gemfile
+++ b/Gemfile
@@ -80,6 +80,9 @@ gem 'cocoon'
 # as the JavaScript library
 gem 'jquery-rails'
 
+# for languages validation
+gem 'iso-639'
+
 # frontend javascripts
 source 'https://rails-assets.org' do
   # for placeholder images
--- a/app/controllers/admin/programs_controller.rb
+++ b/app/controllers/admin/programs_controller.rb
@@ -26,7 +26,7 @@ module Admin
     private
 
     def program_params
-      params.require(:program).permit(:rating, :schedule_public, :schedule_fluid)
+      params.require(:program).permit(:rating, :schedule_public, :schedule_fluid, :languages)
     end
   end
 end
--- a/app/views/admin/events/_proposal.html.haml
+++ b/app/views/admin/events/_proposal.html.haml
@@ -98,6 +98,14 @@
                                               off_text: 'No' }
           - if @event.require_registration
             = registered_text(@event)
+
+      -if @program.languages.present?
+        %tr
+          %td
+            %b Language
+          %td
+            = @event.language
+
       - if !@event.room.nil?
         %tr
           %td
--- a/app/views/admin/programs/show.html.haml
+++ b/app/views/admin/programs/show.html.haml
@@ -32,6 +32,10 @@
         %dd
           = difficulty_levels(@conference)
         %dt
+          Languages:
+        %dd
+          = @program.languages
+        %dt
           Public Schedule
         %dd#schedule_public
           - if @program.schedule_public
--- a/app/views/proposal/new.html.haml
+++ b/app/views/proposal/new.html.haml
@@ -31,6 +31,11 @@
                 data: { min_words: type.minimum_abstract_length, max_words: type.maximum_abstract_length }]},
                 include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' }
 
+              - if @program.languages.present?
+                = f.input :language, as: :select,
+                  collection: @languages,
+                  include_blank: false, label: 'Language', input_html: { class: 'select-help-toggle' }
+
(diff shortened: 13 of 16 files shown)


app/models/program.rb
```diff
@@ -94,4 +95,15 @@ def create_difficulty_levels
...
   end
+
+  ##
+  # Check if languages string has the right format. Used as before_save.
+  #
+  def check_languages_format
+    # All white spaces are removed to allow languages to be separated by ',' and ', ' but only the first option is saved
+    if self.languages.present?
+      self.languages = self.languages.delete(' ')
+      self.languages.match(/^$|(\A[a-z][a-z](,[a-z][a-z])*\z)/).present?
```
ChrisBr: Hm, that means that I could also enter ``foobar`` as language?

ChrisBr:
Your implemenation goes in the direction of labeling proposals. There is a nice gem for this: https://github.com/mbleigh/acts-as-taggable-on

Basically I like the idea, however, I don't like the self coded labeling and also there is no validation if the input is a language.

Also it should be possible to prohibit entering a language as Organizer!

app/models/program.rb
Ana06: No, as the regular expression checks that it is a list of words of two letters. Although you could enter `fo` as language and it is not. I could split the string and check if every language is a real language.

ChrisBr: Does it make sense to add more than one language? I mean usually the speaker presents only in one language (sure there could be subtitles, however, I'd neglect that ...)

ChrisBr:
So from my POV:
- A proposal / event has one language associated
- The Organizers should be able to prohibit selecting a different language, provide languages which the speakers can select (I don't want that the speaker can input every possible language, eg for a Conference in Germany the Organizers want to allow only German and English as presentation language) and also a default language!

ChrisBr:
Maybe this gem could help to solve the problem: https://github.com/scsmith/language_list

app/models/program.rb
Ana06: In the event only one language is selected, but this code is for the program, where the languages that are allowed for events are defined. Then it is a list of languages. There is a selector when you create a proposal with all the languages in this list

ChrisBr:
@hennevogel thoughts?

app/models/program.rb
ChrisBr: Ok

ChrisBr:
Also please provide screenshots

Ana06:
A proposal/event has only one language of a list of languages defined by organizers. The languages are words of two letters as in ISO 639-1. It was suggested by @hennevogel in issue #760 to use ISO 639-1.

The program in the backend looks like that:

program-edit

program

And when you create a new proposal:

proposal-new


There is not a default language. I think it should be a select of the languages in the list in the same place where the list is written. I can add it.

It is not checked if the languages introduced represent a real language. I also think that it could be useful to show `english` instead of `en` when creating a proposal (as most people won't know what `en` means). Both things can be done with language list gem, although it is said that it has functions to retrieve only common languages. Maybe it is better to have a hash with all codes and its language instead of using the gem. I think that using ActsAsTaggableOn gem is not very useful here, although using parse => true instead of split(',') is probably more elegant using add and remove won't be very nice, as every time languages field changes all tags should be removed and the new ones added (maybe not all of them are new, but it would be the same as checking if they are new or not it is not worthwhile). Maybe I'm wrong but I think that this is to simple to use ActsAsTaggableOn gem.

Ana06:
Do the language should be displayed in any other place? I have added in:

events-index

events-show

ChrisBr:
@Ana06 sorry haven't had any time yet, will look into it on the weekend!

Ana06:
Ok! :smile:

app/models/program.rb
```diff
@@ -94,4 +95,15 @@ def create_difficulty_levels
                                                 color: '#EF6E69')
     true
   end
+
+  ##
+  # Check if languages string has the right format. Used as before_save.
+  #
+  def check_languages_format
+    # All white spaces are removed to allow languages to be separated by ',' and ', ' but only the first option is saved
+    if self.languages.present?
```
ChrisBr: Do you need the ``self``? If not, drop it!

app/controllers/proposal_controller.rb
```diff
@@ -16,6 +16,8 @@ def show
   def new
     @user = User.new
     @url = conference_program_proposal_index_path(@conference.short_title)
+    @has_language = @program.languages.present?
```
ChrisBr: Merge these two lines!

ChrisBr: What about the :edit action?

ChrisBr:
Also please fix the RuboCop offence
```
app/models/program.rb:104:5: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression.
```

app/controllers/proposal_controller.rb
Ana06: I defined @has_language because I also need to know if languages is present in the view, but I can change it.
I forgot the :edit action :frowning:

app/models/program.rb
Ana06: I think that `self` is needed

Ana06:
I've made all corrections you have pointed out.
I've also introduced a text indicating that the fist language in the list is the default language (that was already implemented as the first language is the first shown in the selector).
I have found a gem (https://github.com/alphabetum/iso-639) which I think is worthwhile using. I'm using it to valid languages and to translate languages ('en' to 'English'). It looks like that:

backend

new_proposal

Gemfile.lock
```diff
@@ -579,3 +581,6 @@ DEPENDENCIES
   web-console (~> 2.0)
   webmock
   whenever
+
```
ChrisBr: Drop these lines please

app/controllers/proposal_controller.rb
```diff
@@ -16,11 +16,13 @@ def show
   def new
     @user = User.new
     @url = conference_program_proposal_index_path(@conference.short_title)
+    @languages = @program.languages.split(',').map {|l| ISO_639.find(l).english_name} if @program.languages.present?
```
ChrisBr: Please do that in a function in the program model!

app/controllers/proposal_controller.rb
```diff
@@ -16,11 +16,13 @@ def show
...
   end
 
   def edit
     authorize! :edit, @event
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
+    @languages = @program.languages.split(',').map {|l| ISO_639.find(l).english_name} if @program.languages.present?
```
ChrisBr: That's exactly the same line code as in the new action ... so please DRY it in a function!

app/models/program.rb
```diff
@@ -41,6 +41,7 @@ def confirmed
 
   before_create :create_event_types
   before_create :create_difficulty_levels
+  before_save :check_languages_format
```
ChrisBr: Why a before_save hook? What happens if it fails? Does the user get an error message? For me this looks more like a validation ...

ChrisBr:
Please also rebase!

app/controllers/proposal_controller.rb
Ana06: Sorry, I was going to move it to a function but I forgot. :open_mouth:

app/models/program.rb
Ana06: The user get a generic error message. I am going to use validation to include more specific error messages.

Gemfile.lock
```diff
@@ -580,9 +582,7 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   unobtrusive_flash (>= 3)
-  web-console (~> 2.0)
-  webmock
-  whenever
```
differentreality: Why are we removing these 3 lines? Did you do that manually?
I think @ChrisBr 's comment might have been about the lines below this (lines 585-588), as the comment was on the empty line, not on the ones you removed. 
In any case, you cannot remove gems from the Gemfile.lock while you still have them in your Gemfile. That's what's causing travis to fail.

Ana06: I did it manually. I though that it was strange to remove them and I saw that travis failed because of that. I clearly misunderstand what @ChrisBr said. I'll fix it. Thank you. :wink:

app/models/program.rb
```diff
@@ -94,4 +95,29 @@ def create_difficulty_levels
                                                 color: '#EF6E69')
     true
   end
+
+  ##
+  # Check if languages string has the right format. Used as before_save.
```
ChrisBr: The comment is wrong!

app/models/program.rb
```diff
@@ -94,4 +95,29 @@ def create_difficulty_levels
...
+  #
+  def check_languages_format
+    # All white spaces are removed to allow languages to be separated by ',' and ', ' but only the first option is saved
+    return unless self.languages.present?
+    self.languages = self.languages.delete(' ')
+    # We check if every language is a valid ISO 639-1 language
+    errors.add(:languages, 'must be two letters separated by commas') && return unless
+    self.languages.match(/^$|(\A[a-z][a-z](,[a-z][a-z])*\z)/).present?
+    error = false
+    self.languages.split(',').each do |lan|
```
ChrisBr: I just looked briefly into it but you could rewrite it as

```
self.languages.split(',').each do |lan|
  if ISO_639.find(lan).nil?
     errors.add(:languages, 'must be ISO 639-1 valid codes')
     return
end
```
Then you don't need the variable and the extra check after the loop!

app/models/program.rb
```diff
@@ -94,4 +95,29 @@ def create_difficulty_levels
...
+      error = ISO_639.find(lan).nil?
+      next if !error
```
ChrisBr: Ah and furthermore, this won't work! It iterates anyway over the whole Enumerable, what you would need is 
``
break if error
``
(but anyway, I think my other solution would be better)

ChrisBr: As mentioned, I just looked briefly into it, maybe there is another better solution! Will have a deeper look tomorrow!

app/models/program.rb
```diff
@@ -94,4 +95,29 @@ def create_difficulty_levels
...
+    end
+    if error
+      errors.add(:languages, 'must be ISO 639-1 valid codes')
+    end
+  end
+
+  def languages_list
```
ChrisBr: just ```languages``` should be enough!

app/views/admin/events/index.html.haml
```diff
@@ -38,6 +41,7 @@
         %tr
           %td
             = event.id
+
```
ChrisBr: What about this whitespace change?

app/views/admin/events/index.html.haml
```diff
@@ -70,6 +74,7 @@
             - bgcolor="#F7819F"
           - else
             - bgcolor=""
+
```
ChrisBr: What about this whitespace change?

app/views/admin/programs/_form.html.haml
```diff
@@ -8,5 +8,6 @@
       = f.input :schedule_public, label: "Show Schedule on the home and splash page"
       = f.input :schedule_fluid, label: "Allow submitters to change their event after it is scheduled"
       = f.input :rating, :hint => "Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals."
+      = f.input :languages, :hint => "Enter the languages allowed for events as values of ISO 639-1 language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages."
```
ChrisBr: What about a link to a description of the ISO-639?

ChrisBr:
@hennevogel please also have a look, it was your idea ;)

hennevogel:
@ChrisBr I'm fine with your review :-)

app/models/program.rb
vicgalle: I had done it like that but Rubocop doesn't like it:
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/NonLocalExitFromIterator

app/models/program.rb
vicgalle: I am not pretty sure what you mean

app/models/program.rb
Ana06: I had done it like that but Rubocop doesn't like it:
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/NonLocalExitFromIterator

app/views/admin/events/index.html.haml
Ana06: I though It was more readable with it. I can delete it if you want.

app/views/admin/programs/_form.html.haml
Ana06: I'll add it. I suppose you mean this link:

http://www.loc.gov/standards/iso639-2/php/code_list.php

app/models/program.rb
Ana06: I also think it is better but I am not sure if Rubocop would like it.

app/models/program.rb
ChrisBr: Then break out of the loop at least!

app/models/program.rb
ChrisBr: Why not? Break should be totally valid!

ChrisBr:
@Ana06 any update?

app/models/program.rb
Ana06: I meant that Rubocop wouldn't like your other solution. I think Rubocop would like `break`.

Ana06:
Sorry, I couldn't solve it before. I was really busy this week.

I have eliminated the white spaces I added (I had added because I though the code was more readable).

I have changed `next if !error` for `break if error`. I thought both things do the same. In fact, I had never used break or next as I had always used the other way Rubocop doesn't like with return. It seems that Rubocop is happy with that.

I solved a problem with languages_list function.

Last, I added the link to ISO-639.

I hope it is okey now. :smile:

Ana06:
@ChrisBr @hennevogel is there any other change I should make?

hennevogel:
@Ana06 please rebase

Ana06:
Done! :smile:  Why does Travis CI fail?

differentreality:
@Ana06 it previously failed due to phantomjs, I restarted travis and now it has a real error. Please take a look?

differentreality:
@Ana06 The second style error must be the same issue travis has in general. 
Please also rebase again against the latest changes from yesterday.

On a different note, I am not a big fan of organizers entering strings for language manually. Couldn't we rework that into a multiple choice option? With prefilled text.

From the proposal submitter's point of view I think that something like **EN - English**, instead of just EN, or even just **English**, would make it even more straight forward.
If we stick with just initials, it would also be nice to show EN instead of en, I think that's the common way to write language initials, and some styling is nice.
Moreover if we ever wanted to know any laguage statistics accross the conferences we had, it will be impossible when we use different indicators.

ChrisBr:
@Ana06 also please rebase!

Ana06:
@differentreality I don't understand how to use multiple choice option as there are a lot of possible languages.


I have reviewed the code and I realised that capital letters where not allowed when introducing languages in the backend. I have changed it to allow then, although I save then in the same way. I mean, the following options are all allowed:

`en,de,es`
`en, de, es`
`EN,DE,ES`
`EN, de, ES`

But in the four cases `en,de,es` is saved. So there is not any problem with statistics, as English is always saved as `en`.

backend



And that is what proposal submitters see:
submitters
I think that it already does what you have suggested (it shows `English`), or maybe I understood you wrong.

app/views/proposal/_proposal_form.html.haml
```diff
@@ -13,7 +13,12 @@
       collection: @conference.program.event_types.map {|type| ["#{type.title} - #{show_time(type.length)}", type.id,
       data: { min_words: type.minimum_abstract_length, max_words: type.maximum_abstract_length }]},
       include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' }
+      - if @program.languages.present?
 
+    = f.input :language, as: :select,
```
differentreality: Identation doesn't seem right

app/views/admin/programs/_form.html.haml
```diff
@@ -8,5 +8,6 @@
       = f.input :schedule_public, label: "Show Schedule on the home and splash page"
       = f.input :schedule_fluid, label: "Allow submitters to change their event after it is scheduled"
       = f.input :rating, :hint => "Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals."
+      = f.input :languages, :hint => "Enter the languages allowed for events as values of #{link_to('ISO 639-1', 'http://www.loc.gov/standards/iso639-2/php/code_list.php', :target => "_blank")} language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages.".html_safe
       %p.text-right
```
differentreality: Please use the new ruby style

Ana06: What do you mean with _new ruby style_?

differentreality: @Ana06 There is a new coding style for ruby, which we try to follow in new code we write. Googling about it: https://github.com/bbatsov/ruby-style-guide#hash-literals

Ana06: Ok! I'll change it

Ana06:
I have changed to the new ruby style the whole `app/views/admin/programs/_form.html.haml` file

app/models/program.rb
differentreality: Why don't we substitute all that loop just with
`errors.add(:languages, 'must be ISO 639-1 valid codes') if self.languages.split(',').select{ |x| ISO_639.find(x).nil? }`

differentreality:
@Ana06 for future reference, it is enough to make sure you use the new ruby style for the new code you introduce. Otherwise it may get difficult to review both styling changes and new feature implementation. But you can always open a new PR just with styling changes ;)

You should also use single quotes, when you don't need string interpolation.

app/models/program.rb
```diff
@@ -113,4 +118,23 @@ def create_difficulty_levels
                                                 color: '#EF6E69')
     true
   end
+
+  ##
+  # Check if languages string has the right format. Used as validation.
+  #
+  def check_languages_format
+    # All white spaces are removed to allow languages to be separated by ',' and ', ' but only the first option is saved
```
differentreality: This comment should be one line lower, as it refers to L128, not L127
What do you mean only the first option is saved?

app/models/program.rb
```diff
@@ -113,4 +118,23 @@ def create_difficulty_levels
...
+    return unless self.languages.present?
+    self.languages = self.languages.delete(' ').downcase
+    # We check if every language is a valid ISO 639-1 language
```
differentreality: This line does not say what happens to the following line! That comment should be above the line it describes.

app/models/program.rb
Ana06: I mean that `en, de` and `en,de` are both allowed, but it is always saved as `en,de`, removing spaces (I did it like that for being consistent). I will change the comment to make it more clear.

app/models/program.rb
Ana06: I'll revise all comments in this function.

app/models/program.rb
Ana06: @differentreality I think that you solution is much more elegant, although `select{}` will return an array (empty or not) and it will always add the error. But we could check if the array is not empty:

`errors.add(:languages, 'must be ISO 639-1 valid codes') unless self.languages.split(',').select{ |x| ISO_639.find(x).nil? }.empty?`

Ana06:
I have just realised that languages could be repeated. I mean `en,de,en` was accepted. I suppose that we want to avoid that as it would be more difficult to know any language statistics and it would be confusing for the users that register for the conference too.

I have added this check and fixed all things @differentreality told me.

I won't make styling changes while I am implementing new features and I will use single quotes if possible from now on. :wink:

ChrisBr:
@Ana06 It looks good to me :+1: 

However, I'd wait till we released version 1.0 of OSEM before merging this as we don't want to introduce new features before 1.0! Thanks

Ana06:
Ok! :wink:

Should I add any test?

hennevogel:
Always add tests :-)

Ana06:
Ok :wink:

I've started adding tests, but I will add more tests. :smile:

Ana06:
I've added tests in all test files that exists. I don't know if there is any reason to have no tests for some parts. Even If that parts need to have tests I think that it is better not to add them in this pull request to avoid confusion.

spec/models/program_spec.rb
```diff
@@ -58,4 +58,31 @@
       expect(conference.program.difficulty_levels.count).to eq 3
     end
   end
+
+  describe 'validation' do
+    it '#check_languages_format' do
```
ChrisBr: @Ana06 please write for each test case a separate test because otherwise it's hard to debug if one test fails!

ChrisBr:
@Ana06 what is the default language if the user doesn't select one?

ChrisBr:
And please squash your commits!

Ana06:
It is not possible to not select one. The selector has the first language by default, and there is no option to select no language. It could happen that the languages are set after the event is created, in that case the language would be `nil`.
By squash do you mean that I add more things in the same commit instead of doing several commits?

Ana06:
I've squashed the commits and also split the long test. :blush:

db/schema.rb
```diff
@@ -351,9 +352,13 @@
     t.boolean  "include_registrations"
     t.boolean  "include_sponsors"
     t.boolean  "include_lodgings"
+    t.string   "banner_photo_file_name"
```
ChrisBr: Huh?

Please drop these changes!

ChrisBr:
@hennevogel please review! From my side it's good to go!

hennevogel:
If you're okay it's okay

PR #906, opened 2016-03-24
Downgrade holderjs to 2.9.1

sonalkr132:
Latest version avilable on rails-assets.org is 2.9.1

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -336,7 +336,7 @@ GEM
     rails-assets-bootstrap-markdown (2.10.0)
       rails-assets-bootstrap (~> 3)
     rails-assets-date.format (1.2.3)
-    rails-assets-holderjs (2.9.3)
+    rails-assets-holderjs (2.9.1)
     rails-assets-jquery (2.2.1)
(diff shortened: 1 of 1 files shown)


sonalkr132:
> Net::HTTPServiceUnavailable: <h2>This website is under heavy load (queue
full)</h2><p>We're sorry, too many people are accessing this website at the same
time. We're working on this problem. Please try again later.</p>

Great! :anger:

differentreality:
Where do you see that info about the version?

sonalkr132:
@differentreality go to https://rails-assets.org/ and type `holderjs` in search box, it will show all the version available.

Sorry, I can't provide more direct link, they are using js.

sonalkr132:
I think they provide option to rebuild the gem, and import gems which are not available. However they are having major traffic issue. I think we would be able to `bundle install` until they get it together :person_frowning:

sonalkr132:
I scheduled `holderjs` for rebuild, and 2.9.3 is available now.
screenshot from 2016-03-24 23-03-05

We won't need this.

PR #907, opened 2016-03-24
Proposal controller improvements and test

sonalkr132:
General improvements in proposal controller:
* use of `if.. else` for `@object.save` is what we follow in rest of controllers. It also saves us from writing `return`
* On failed update render `edit` template and not `new` template.
* On failed save redirect to index path and not render `new` template (for `withdraw`, `restart` and `confirm`)
* Check for success for save in withdraw (save can fail for reasons other than validation)
* Add authorization to failed user save path cause it was complaining that no authorization was performed in that path.

### DIFF
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -24,13 +24,14 @@ class ProposalController < ApplicationController
   end
 
   def edit
-    authorize! :edit, @event
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
   end
 
   def create
     @url = conference_program_proposal_index_path(@conference.short_title)
 
+    # We allow proposal submission and sign up on same page.
+    # If user is not signed in then first create new user and then sign them in
     unless current_user
       @user = User.new(user_params)
       if @user.save
@@ -47,39 +48,38 @@ class ProposalController < ApplicationController
     @event = Event.new(event_params)
     @event.program = @program
 
+    # User which creates the proposal is both `submitter` and `speaker` of proposal
+    # by default.
+    # TODO: Allow submitter to add speakers to proposals
     @event.event_users.new(user: current_user,
                            event_role: 'submitter')
     @event.event_users.new(user: current_user,
                            event_role: 'speaker')
     authorize! :new, @event
 
-    unless @event.save
+    if @event.save
+      ahoy.track 'Event submission', title: 'New submission'
+      redirect_to conference_program_proposal_index_path(@conference.short_title), notice: 'Proposal was successfully submitted.'
+    else
       flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
       render action: 'new'
-      return
     end
-
-    ahoy.track 'Event submission', title: 'New submission'
-
-    redirect_to conference_program_proposal_index_path(@conference.short_title), notice: 'Proposal was successfully submitted.'
   end
 
   def update
-    authorize! :update, @event
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
-    if !@event.update(event_params)
+    if @event.update(event_params)
+      redirect_to(conference_program_proposal_index_path(conference_id: @conference.short_title),
+                  notice: 'Proposal was successfully updated.')
+    else
       flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
-      render action: 'new'
-      return
+      render action: 'edit'
     end
-
-    redirect_to conference_program_proposal_index_path(conference_id: @conference.short_title),
-                notice: 'Proposal was successfully updated.'
   end
 
-  def destroy
-    authorize! :destroy, @event
+  def withdraw
+    authorize! :update, @event
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
     begin
@@ -89,9 +89,13 @@ class ProposalController < ApplicationController
       return
     end
 
-    @event.save(validate: false)
-    redirect_to conference_program_proposal_index_path(conference_id: @conference.short_title),
-                notice: 'Proposal was successfully withdrawn.'
+    if @event.save
+      redirect_to conference_program_proposal_index_path(conference_id: @conference.short_title),
+                  notice: 'Proposal was successfully withdrawn.'
+    else
+      redirect_to conference_program_proposal_index_path(conference_id: @conference.short_title),
+                  error: "Could not withdraw proposal: #{@event.errors.full_messages.join(', ')}"
+    end
   end
 
   def confirm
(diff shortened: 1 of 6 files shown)


sonalkr132:
closes #597

app/models/ability.rb
```diff
@@ -44,7 +44,7 @@ def not_signed_in
 
     # can view Commercials of confirmed Events
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
-    can :show, User
+    can [:show, :create], User
```
differentreality: Why do we need create here? Where would that authentication happen (ie where do we access the create action of user)?

app/controllers/proposal_controller.rb
```diff
@@ -47,39 +49,38 @@ def create
...
-    redirect_to conference_program_proposal_index_path(@conference.short_title), notice: 'Proposal was successfully submitted.'
   end
 
   def update
-    authorize! :update, @event
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
-    if !@event.update(event_params)
+    if @event.update(event_params)
+      redirect_to(conference_program_proposal_index_path(conference_id: @conference.short_title),
```
differentreality: I'd suggest you keep the previous styling.

app/controllers/proposal_controller.rb
```diff
@@ -89,9 +90,13 @@ def destroy
       return
     end
 
-    @event.save(validate: false)
-    redirect_to conference_program_proposal_index_path(conference_id: @conference.short_title),
-                notice: 'Proposal was successfully withdrawn.'
+    if @event.save
```
differentreality: Do we really need to validate when we withdraw the event? 
If the event is resubmitted, validations will run then.

app/models/ability.rb
sonalkr132: > Where would that authentication happen (ie where do we access the create action of user)?

[here](https://github.com/openSUSE/osem/pull/907/files#diff-8c13e9cfe1f5edf7d4923e633a08f790R37):
```Ruby
  def create
    unless current_user
       @user = User.new(user_params)
       authorize! :create, @user  # <- this line
        if @user.save
          sign_in(@user)
        else
         # this path would have no authorization
         flash[:error] = "Could not save user: #{@user.errors.full_messages.join(', ')}"
         render action: 'new'
         return
       end
     end
     # more code
  end
```
cancancan was complaining that no authorization was performed in that path. It is either this or `skip_authorization_check`

app/controllers/proposal_controller.rb
sonalkr132: > Do we really need to validate when we withdraw the event? 

Not really. I had asked about proposed change [here](https://github.com/openSUSE/osem/pull/867#discussion_r55916850). My reasoning is that:
* if are to use `@event.save(validate: false)`, we should do the same in other similar actions (`confirm` and `restart`). @hennevogel said that I should go with `if..else`
* We should do `if @event.save` cause save can fail for reasons other than validation, for ex: our database connection is lost, running out of memory, some obscure IO/socket error (source: [this blog](https://robots.thoughtbot.com/save-bang-your-head-active-record-will-drive-you-mad))

app/controllers/proposal_controller.rb
sonalkr132: I have reverted this change. Should I revert other ([here](https://github.com/openSUSE/osem/pull/907/files#diff-8c13e9cfe1f5edf7d4923e633a08f790R141) and [here](https://github.com/openSUSE/osem/pull/907/files#diff-8c13e9cfe1f5edf7d4923e633a08f790R61)) similar changes as well? I had made the change cause following is the style we follow in rest of controllers:
```Ruby
if @object.save
  # do something
else 
  # do something else
end
```
instead of:
```Ruby
if !@object.save
   # do something
   return
end

# do something
```

app/controllers/proposal_controller.rb
hennevogel: We want valid data so we validate...

hennevogel:
Why the last commit?

sonalkr132:
@differentreality says that it is easier for everyone if instead editing my git commits, I add extra commits during PR review. Extra commits can be squashed once PR is approved.

hennevogel:
I'm mean the content :-) Why do you change a simple if/else to a double negation?

sonalkr132:
It was asked [here](https://github.com/openSUSE/osem/pull/907#discussion_r57571845).

differentreality:
@sonalkr132 if you are referring to my comment, I commented about styling on the redirect_to() line, not the if line. We recently changed all redirect_to() without parenthesis.

sonalkr132:
@differentreality Sorry :scream: I thought you mean the if else part. I commented too. This is epic!

sonalkr132:
updated! :+1:

app/controllers/proposal_controller.rb
```diff
@@ -24,15 +24,17 @@ def new
   end
 
   def edit
-    authorize! :edit, @event
```
hennevogel: What now?

app/controllers/proposal_controller.rb
```diff
@@ -47,39 +49,38 @@ def create
...
                            event_role: 'speaker')
     authorize! :new, @event
 
-    unless @event.save
+    if @event.save
+      ahoy.track 'Event submission', title: 'New submission'
+      redirect_to conference_program_proposal_index_path(@conference.short_title), notice: 'Proposal was successfully submitted.'
+    else
       flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
       render action: 'new'
```
hennevogel: Now we render :new again?

app/controllers/proposal_controller.rb
sonalkr132: We already have following line on [top](https://github.com/sonalkr132/osem/blob/proposal-test/app/controllers/proposal_controller.rb#L5):
```Ruby
  load_and_authorize_resource :event, parent: false, through: :program, except: [:new, :create]
```
We don't need to authorize again.

app/controllers/proposal_controller.rb
```diff
@@ -99,15 +104,15 @@ def confirm
...
+      @event.confirm
     rescue Transitions::InvalidTransition
       redirect_to :back, error: "Event can't be confirmed"
       return
     end
 
-    if !@event.save
-      flash[:error] = "Could not confirm proposal: #{@event.errors.full_messages.join(', ')}"
-      render action: 'new'
+    unless @event.save
```
hennevogel: Please turn this around.
```
if event.save
  if registered
    redirect to :index
  else
    redirect to :new_reigstration
  end
else
  redirect error
end
```

sonalkr132:
@hennevogel Thanks. It is definitely more beautiful that way. Updated!

hennevogel:
LGTM! Thanks

PR #913, opened 2016-03-25
Mark Events without a room as unscheduled events

### LINKED ISSUE #912: Deleting a room should move its events back to Unscheduled Events
If a room with events scheduled is deleted, its events are not added back to Unscheduled Events in the admin/schedule page.This seems to make it impossible to reschedule those events.

nishanthvijayan:
Fixes #912 
Scheduling an event sets the start_time and room model attrs of that event record.
Room has_many events with dependent set to nullify.So on deleting a room, it'll nullify the room attr of associated events but start_time would still be set.
But in schedule page, events are marked as unscheduled events by checking if start_time is not null.
I've changed this to mark events as unscheduled events if they don't have a room.

### DIFF
--- a/public/javascripts/schedule/schedule.js
+++ b/public/javascripts/schedule/schedule.js
@@ -89,7 +89,7 @@ var Schedule = {
             snapMode: "inner",
             zIndex: 2
         });
-        if (date == "none") {
+        if (date == "none" || vars["room_id"] == null) {
             $('#unscheduled').append(newEvent);
         } else {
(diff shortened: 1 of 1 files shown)


differentreality:
:+1:

differentreality:
@ChrisBr @hennevogel can you think of any reason why we would not want to have an event with start_time set when room is not set?

ChrisBr:
No! LGTM

Issue #595, opened 2015-02-08
Show roles in registration list


differentreality:
Registration list should have an indication for the roles the person has.

It is important for the registration desk, for example, to know if the person is a sponsor or a speaker.

shreyanshabhargav:
@differentreality Hi, I am an interested GSOC participant interested to contribute for FOSSDEM. Can you please give me more insight about this issue ?
Thanks

PR #918, opened 2016-03-26
Display roles in registration list

matthin:
#595


matthin:
I'm not too sure why Hakiri is showing security warning only on my pull request. The warnings are for the files commercials_controller.rb, and application_controller.rb; two files which I didn't edit.

sonalkr132:
@matthin Don't worry. It is not you. Hakiri must have updated and now it is reporting new issues in our existing codebase. Checkout #921 for more info.

differentreality:
This is a very visual change, so please attach screenshots of your change.
user.roles also works and we also have user.get_roles (check it out to see what it produces). In this case we would only need roles of the specific conference we are at. 
I was thinking labels (there can be max 4, with the current roles we use).

matthin:
`user.get_roles` returns a hashmap with each of the user's roles, and the associated conferences. So if the user is part of the "Event Coordinator" role for example, we would get back a hash containing "Event Coordinator" and every conference which has a role called "Event Coordinator"; which I think would be every single conference. I can't think of a way to link that data with a single specific conference. Though, I might not be understanding OSEM's database layout well enough.

`user.roles` would work, but would involve a separate explicit query to get the related conference, due to how the model relationship is setup. Overall, I think it would be more complex due to that explicit query. Though, maybe it would be simpler to read.

Here are some screenshots with different screen sizes:

osem-roles-large

osem-roles-medium
It doesn't really look good on smaller screens. Should I change the label margins based on device-type?

osem-roles-small
As you can see, the labels overlap when viewed on phones. It's the same when in landscape mode. So, I'll need to fix that.

ChrisBr:
We agreed that we don't want to have view tests! Otherwise LGTM

matthin:
Ah, okay. I'll go remove the unneeded view test. I agree that they are mostly just overhead.

hennevogel:
LGTM

Issue #623, opened 2015-04-01
prefill your arrival time with dates of the conference


jospoortvliet:
If you are a good boy/girl and register well in advance for a event with OSEM, you get rewarded by having to scroll forward months and months until you get to the dates of the event ;-)

Most people won't arrive that far in advance so I suggest to open the date picker at least at the month of the event starting, perhaps even pre-select the event dates so people too lazy to fill this in don't get frustrated by it. Heck, make it non-mandatory?

differentreality:
That's a good point.

Arrival/departure dates are not mandatory by the way. Could you double check?

jospoortvliet:
ah, yes, that is not mandatory, I can get to the error page without entering it.

Also, wrong password, invalid email address etc gives proper warnings, so input validation works...

differentreality:
So for this issue what is left is just fast-forward the calendar to conference dates, right?

jospoortvliet:
@differentreality yeah, exactly. It's a minor thing...

hennevogel:
Can be easily done in [app/assets/javascripts/osem-datepickers.js](https://github.com/openSUSE/osem/blob/master/app/assets/javascripts/osem-datepickers.js)

jospoortvliet:
@hennevogel I suggest to make this a juniorjob then... ;-)

eximious:
An alternative would be an option to tell the conference organizers, "I'm attending both days, attending just Friday, or just Saturday" tick box, or however many days your event spans. I think that would be more useful than the exact times for general attendees, even if that remianed an option for speakers to fill out. (Let me know if this would be better as a feature request.)

jospoortvliet:
Yeah, just boxes that say what days you're there is easier, indeed...

ChrisBr:
@eximious @jospoortvliet I like the idea too!

differentreality:
As far as datepicker is concerned, using the option defaultDate sets the input box in that date on initilization (before the user actually selects anything from the calendar)

jospoortvliet:
That would be a nice quick fix. In 3-4 months we'll update our instance and get ready for the next ownCloud Contributor Conference (this one was awesome!), I'm wondering if there is anything I should be aware off - you folks planning to do a release at some point?

jospoortvliet:
actually we're looking at deploying registration for ownCloud conference 2016 already this year - was hoping to rekindle interest in this issue as it is quite a usability PITA...

Shriyanshagro:
[...]

PR #919, opened 2016-03-26
prefilled users arrival time with dates of the conference#623

Shriyanshagro:
Now , arrival/departure time while registering for conference will be prefilled with start_date and end_date of conference.

### DIFF
--- a/app/assets/javascripts/osem-datepickers.js
+++ b/app/assets/javascripts/osem-datepickers.js
@@ -1,10 +1,19 @@
 $(function () {
-  $("#registration-arrival-datepicker, #registration-departure-datepicker").datetimepicker({
+  $("#registration-arrival-datepicker").datetimepicker({
       pickTime: true,
       useCurrent: false,
       minuteStepping: 15,
       sideBySide: true,
-      format: "YYYY-MM-DD HH:mm"
+      format: "YYYY-MM-DD HH:mm",
+      defaultDate :   $("#registration-arrival-datepicker").attr('start_date')
+  });
+  $("#registration-departure-datepicker").datetimepicker({
+      pickTime: true,
+      useCurrent: false,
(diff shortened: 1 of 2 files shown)


differentreality:
I don't think we should prefill those values. Each attendee should indicate this, if he/she wants to. It is a very personal choice when someone will arrive to and leave from a conference. 
I think we are just going to end up with false information if we prefill these fields.

Shriyanshagro:
But on [issue](https://github.com/openSUSE/osem/issues/623) . It was discussed , that there should be prefilled arrival/departure date , so that the people too lazy to fill this in don't get frustrated by it.

differentreality:
That might be the title but it is not what is being discussed in the issue. Also discussion does not equal decision. In any case, I was merely stating my opinion. There are more people whose opinion weighs in, let's see what they think.

Shriyanshagro:
Okay , than lets wait for others to put in their opinion. If you would ask me, I would suggest let it be prefilled with dates (+/-)1 days with actual dates of conference.

ChrisBr:
What about leaving it empty by default but when I click on the datetimepicker the start / end date is preselected?

Shriyanshagro:
@ChrisBr  ,do you mean that, dates would get preselected , only when we click on datetimepicker for first time??

ChrisBr:
@Shriyanshagro for instance

hennevogel:
Yes this is what pre-select means :-)

Shriyanshagro:
So , is this the final idea to implement ??

hennevogel:
Yes

hennevogel:
Please make one pull request or this is not reviewable

Shriyanshagro:
okay . Will create new Pr  with new idea been discussed here.

Issue #632, opened 2015-04-15
Event Registration


hennevogel:
There are two reasons submitter wants to require registration for his event:
1. They can only accommodate a limited amount of people because of limited resources (think Rasberry PI workshop and I have only N machines)
2. Knowing how many people will show up

Both are not handled in the current form. I can not limit the amount of registrations and I can't see how many people registered.

differentreality:
Note: we can see how many people have registered in 
https://github.com/openSUSE/osem/pull/634/files#diff-8fb9d019430ee6b805c281047c1b7a2fR70

PR #923, opened 2016-03-28
Enable event registration for proposals

differentreality:
https://github.com/openSUSE/osem/issues/632

* Enables event registration and setting of max attendees by submitter
* Adds indication to events that require registration
* Shows registration list of people for event
  * Shows who attended the event (organizers can toggle attendance)
* Refactors the event registration during conference registration
* It validates max_attendees attribute of event against room.size, if the event is scheduled when enabling require_registration

TODO: Add PDF/XLS generation of registration list

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -33,7 +33,6 @@ class ConferenceRegistrationsController < ApplicationController
   end
 
   def show
-    @workshops = @registration.workshops
     @total_price = Ticket.total_price(@conference, current_user)
     @tickets = current_user.ticket_purchases.where(conference_id: @conference.id)
   end
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -145,10 +145,14 @@ class ProposalController < ApplicationController
     end
   end
 
+  def registrations; end
+
   private
 
   def event_params
-    params.require(:event).permit(:title, :subtitle, :track_id, :event_type_id, :abstract, :description, :require_registration, :difficulty_level_id)
+    params.require(:event).permit(:event_type_id, :track_id, :difficulty_level_id,
+                                  :title, :subtitle, :abstract, :description,
+                                  :require_registration, :max_attendees)
   end
 
   def user_params
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -14,7 +14,8 @@ class Event < ActiveRecord::Base
   has_many :commercials, as: :commercialable, dependent: :destroy
   belongs_to :event_type
 
-  has_and_belongs_to_many :registrations
+  has_many :events_registrations
+  has_many :registrations, through: :events_registrations
 
   belongs_to :track
   belongs_to :room
@@ -32,6 +33,10 @@ class Event < ActiveRecord::Base
   validates :abstract, presence: true
   validates :event_type, presence: true
   validates :program, presence: true
+  validates :max_attendees, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_nil: true }
+
+  validate :max_attendees_and_require_registration
+  validate :max_attendees_no_more_than_room_size
 
   scope :confirmed, -> { where(state: 'confirmed') }
   scope :highlighted, -> { where(is_highlight: true) }
@@ -62,6 +67,19 @@ class Event < ActiveRecord::Base
     event :reject do
       transitions to: :rejected, from: [:new], on_transition: :process_rejection
     end
+  end
+
+  ##
+  # Checkes if the event has a start_time and a room
+  # ====Returns
+  # * +true+ or +false+
+  def scheduled?
+    room && start_time ? true : false
+  end
+
+  def registration_possible?
+    return false unless max_attendees
+    registrations.count < max_attendees
   end
 
   def voted?(event, user)
@@ -206,6 +224,21 @@ class Event < ActiveRecord::Base
   end
 
   private
+
+  ##
+  # If max_attendees variable is set (higher than 0)
+  # variable require_registration must also be set
+  def max_attendees_and_require_registration
+    errors.add(:require_registration, 'must be enabled, when you set max_attendees') if max_attendees && !require_registration
+    errors.add(:max_attendees, 'must be enabled, when you set require_registration') if require_registration && max_attendees.nil?
+  end
+
+  ##
+  # Do not allow, for the event, more attendees than the size of the room
+  def max_attendees_no_more_than_room_size
+    return unless room && max_attendees_changed?
+    errors.add(:max_attendees, "cannot be more than the room's capacity (#{room.size})") if max_attendees && (max_attendees > room.size)
+  end
 
   def abstract_limit
     # If we don't have an event type, there is no need to count anything
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -8,8 +8,19 @@ class Program < ActiveRecord::Base
   has_many :tracks, dependent: :destroy
   has_many :difficulty_levels, dependent: :destroy
   has_many :events, dependent: :destroy do
-    def workshops
+    def require_registration
       where(require_registration: true, state: :confirmed)
+    end
+
+    def with_registration_open
+      where(require_registration: true, state: :confirmed).
+      map { |e| e if e.max_attendees > e.registrations.count }.compact
+    end
+
+    # All confirmed events of the conference with attribute require_registration
+    # excluding the events the user has already registered to
+    def remaining_for_registration(registration)
+      require_registration - registration.events
     end
 
     def confirmed
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,6 +36,7 @@ class User < ActiveRecord::Base
   has_many :event_users, dependent: :destroy
   has_many :events, -> { uniq }, through: :event_users
   has_many :registrations, dependent: :destroy
+  has_many :events_registrations, through: :registrations
   has_many :ticket_purchases, dependent: :destroy
   has_many :tickets, through: :ticket_purchases, source: :ticket
   has_many :votes, dependent: :destroy
@@ -53,8 +54,31 @@ class User < ActiveRecord::Base
             },
             presence: true
 
+  ##
+  # Checkes if the user attended the event
+  # This is used for events that require registration
+  # The user must have registered to attend the event
+  # Gets an event
+  # === Returns
+  # * +true+ if the user attended the event
+  # * +false+ if the user did not attend the event
+  def attended_event? event
+    event_registration = event.events_registrations.find_by(registration: self.registrations)
+
+    return false unless event_registration.present?
+    event_registration.attended
+  end
+
   def name
     self[:name] || username
+  end
+
+  ##
+  # Checks if a user has registered to an event
+  # ====Returns
+  # * +true+ or +false+
+  def registered_to_event? event
+    event.registrations.pluck(:id).include? self.registrations.find_by(conference_id: event.program.conference.id).id
   end
 
   def subscribed? conference
--- a/app/views/admin/events/registrations.html.haml
+++ b/app/views/admin/events/registrations.html.haml
@@ -1,0 +1,40 @@
+.row
+  .col-md-12
+    .page-header
+      %h1
+        Registrations (#{@event_registrations.length}/#{@event.max_attendees})
+      .text-muted
+        for
+        = @event.title
+
+    - if @event.room && (@event_registrations.length > @event.room.size)
+      %b Attention:
+      You have more registrations than the capacity of the room!
+
+    .well
+      %table.table.table-hover.table-borderd.table-striped.datatable#registrations
+        %thead
+          %th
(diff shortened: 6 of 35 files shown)


differentreality:
screenshot

screenshot-2

screenshot-4

screenshot-3

screenshot-1

screenshot-6

screenshot-7

screenshot-14

ChrisBr:
Please rebase @differentreality

app/models/user.rb
```diff
@@ -57,6 +57,32 @@ def name
...
+  ##
+  # Checks if a user has registered to an event
+  # ====Returns
+  # * +true+ or +false+
+  def registered_to_event? event
+    event.registrations.pluck(:id).include? self.registrations.find_by(conference_id: event.program.conference.id).id
+  end
+
   def subscribed? conference
     self.subscriptions.find_by(conference_id: conference.id).present?
```
dronov: Maybe it could be written a little bit elegant: 

```self.subscriptions.exists? conference_id: conference.id```

File: app/models/event.rb [hunk 2]
```diff
@@ -64,6 +69,27 @@ class Event < ActiveRecord::Base
...
+    "Registered: #{registrations.count}"
+  end
+
+  def registration_possible?
+    return false unless max_attendees
+    registrations.count < max_attendees
+  end
+
   def voted?(event, user)
     event.votes.where('user_id = ?', user).first
```
dronov: I think method ```voted?``` should return true or false. So it would be written like:

```event.votes.where('user_id = ?', user).any?```

**Comment by dronov** [comment]
Hello @differentreality :) I left some comments, maybe them could be useful.

**Comment by hennevogel** [comment]
Please drop the '0' as indicator that capacity is unlimited. Either capacity is >= 1 or not.

**Comment by ChrisBr** [comment]
@dronov please only comment on changes in this PR!

**Comment by ChrisBr** [comment]
@differentreality I really like the screenshots! Awesome :bouquet:

File: app/controllers/conference_registrations_controller.rb [hunk 3]
```diff
@@ -29,15 +29,24 @@ def new
 
     # @user variable needs to be set so that _sign_up_form_embedded works properly
     @user = @registration.build_user
+
+    # Add the events with existing registration at the end
+    @events_registrations = (@conference.program.events.with_registration_open - @registration.events) + @registration.events
```
ChrisBr: What about a scope?

app/controllers/conference_registrations_controller.rb
```diff
@@ -29,15 +29,24 @@ def new
...
   end
 
   def show
-    @workshops = @registration.workshops
+    # All the events the user has registered to
+    @events_registered = @registration.events
+    # All confirmed events of the conference with attribute require_registration
+    # excluding the events the user has already registered to
+    @events_require_registration = (@conference.program.events.require_registration - @registration.events)
```
ChrisBr: Also what about a scope?

app/controllers/conference_registrations_controller.rb
```diff
@@ -29,15 +29,24 @@ def new
...
     @total_price = Ticket.total_price(@conference, current_user)
     @tickets = current_user.ticket_purchases.where(conference_id: @conference.id)
   end
 
-  def edit; end
+  def edit
+    @events_registrations = (@conference.program.events.with_registration_open - @registration.events) + @registration.events
```
ChrisBr: Same for this!

app/controllers/conference_registrations_controller.rb
```diff
@@ -79,6 +88,7 @@ def update
       redirect_to  conference_conference_registrations_path(@conference.short_title),
                    notice: 'Registration was successfully updated.'
     else
+      @events_registrations = (@conference.program.events.with_registration_open - @registration.events) + @registration.events
```
ChrisBr: Same here!

app/models/event.rb
```diff
@@ -64,6 +69,27 @@ class Event < ActiveRecord::Base
     end
   end
 
+  ##
+  # Checkes if the event has a start_time and a room
+  # ====Returns
+  # * +true+ or +false+
+  def scheduled?
+    room && start_time ? true : false
```
ChrisBr: Drop ``? true : false``, ``room && start_time`` is the same!

app/models/event.rb
```diff
@@ -64,6 +69,27 @@ class Event < ActiveRecord::Base
...
+  end
+
+  ##
+  # ====Returns
+  # * +String+ -> number of registrations / max allowed registrations
+  def registered_text
```
ChrisBr: That should be a view helper method and not a model method!

app/models/program.rb
```diff
@@ -8,10 +8,15 @@ class Program < ActiveRecord::Base
...
   has_many :difficulty_levels, dependent: :destroy
   has_many :events, dependent: :destroy do
-    def workshops
+    def require_registration
       where(require_registration: true, state: :confirmed)
     end
 
+    def with_registration_open
+      where(require_registration: true, state: :confirmed).
+      map { |e| e if e.max_attendees && ((e.max_attendees == 0) || (e.max_attendees > e.registrations.count)) }.compact
```
ChrisBr: Is it not possible to do this in the ``where``?

app/models/user.rb
```diff
@@ -57,6 +57,32 @@ def name
     self[:name] || username
   end
 
+  ##
+  # Finds all the events the user is registered to
+  # If a conference is provided, results are limited to events of that confernece
+  # ====Returns
+  # * +ActiveRecord::Associations::CollectionProxy+ -> events
+  def event_registrations conference=nil
+    input_registrations = conference ? self.registrations.where(conference: conference) : self.registrations
```
ChrisBr: This should also be possible in a scope...

app/views/proposal/registrations.html.haml
```diff
@@ -0,0 +1,36 @@
...
+        %table.table.table-hover.table-borderd.table-striped.datatable#registrations
+          %thead
+            %th
+            %th Name
+            %th Email
+            %th Created AT
+            %th Attended
+            %th Attended Conference
+          %tbody
+            - @event_registrations.each_with_index do |event_registration, index|
```
ChrisBr: When you use
```
@event_registrations.each.with_index(1) do |event_registration, index|
```
you could drop the ``index + 1`` in the table row!

app/views/admin/events/registrations.html.haml
```diff
@@ -0,0 +1,40 @@
...
+          %th
+          %th Name
+          %th Email
+          %th Created At
+          %th Attended
+          %th Attended Conference
+        %tbody
+          - @event_registrations.each_with_index do |event_registration, index|
+            %tr
+              %td= index + 1
```
ChrisBr: Same here!

app/models/program.rb
differentreality: Dunno :/ 
(I need to check the registrations count of the event (e.registrations.count))

app/models/event.rb
differentreality: Actually `room && start_time` returns the start_time or nil

app/controllers/conference_registrations_controller.rb
```diff
@@ -33,7 +33,8 @@ def new
   end
 
   def show
-    @workshops = @registration.workshops
+    # All the events the user has registered to
+    @events_registered = @registration.events
```
hennevogel: Why don't you use `@registration.events` directly in the view?

app/controllers/proposal_controller.rb
```diff
@@ -145,10 +145,16 @@ def restart
     end
   end
 
+  def registrations
+    @event_registrations = @event.events_registrations
```
hennevogel: Why don't you use `@event.events_registrations` directly in the view?

app/models/event.rb
```diff
@@ -209,6 +227,21 @@ def calculate_progress
...
+  # If max_attendees variable is set (higher than 0)
+  # variable require_registration must also be set
+  def max_attendees_and_require_registration
+    errors.add(:require_registration, 'must be enabled, when you set max_attendees') if max_attendees && !require_registration
+    errors.add(:max_attendees, 'must be enabled, when you set require_registration') if require_registration && max_attendees.nil?
+  end
+
+  ##
+  # Do not allow, for the event, more attendees than the size of the room
+  def max_attendees_no_more_than_room_size
```
hennevogel: This is correct but I wonder what this does to the scheduling workflow. What happens if an admin assigns an Event to a Room and this validations hits?

app/models/user.rb
```diff
@@ -53,10 +54,51 @@ class User < ActiveRecord::Base
...
+  # This is used for events that require registration
+  # The user must have registered to attend the event
+  # Gets an event
+  # === Returns
+  # * +true+ if the user attended the event
+  # * +false+ if the user did not attend the event
+  def attended_event? event
+    event_registration = event.events_registrations.find_by(registration: self.registrations)
+
+    return false unless event_registration.present?
```
hennevogel: Why not `event.events_registrations.find_by(registration: self.registrations).any?`

app/models/user.rb
hennevogel: In Event yes

spec/models/event_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+include ApplicationHelper
```
hennevogel: Uhm what?

spec/models/event_spec.rb
```diff
@@ -89,6 +137,56 @@
...
+  describe '#scheduled?' do
+    it { expect(event.scheduled?).to eq false }
+    it 'returns true if the event is scheduled' do
+      event.room = create(:room)
+      event.start_time = conference.start_date.to_time
+      expect(event.scheduled?).to eq true
+    end
+  end
+
+  describe '#registered_text' do
```
hennevogel: This should be a [helper spec](https://relishapp.com/rspec/rspec-rails/docs/helper-specs/helper-spec)

spec/models/event_spec.rb
hennevogel: See below why this shouldn't be here...

spec/models/user_spec.rb
```diff
@@ -40,6 +40,28 @@
     end
   end
 
+  describe '#event_registrations' do
+    before :each do
```
hennevogel: Why don't you use `let/let!`?

app/models/user.rb
differentreality: You mean why I use `.present?` instead of `.any?` ? Or why I use a variable? (I use a variable because I am re-using it on the next line - thought it would look cleaner that way).

app/models/event.rb
differentreality: We can overcome this by only adding a validation error if the attribute max_attendees has changed (which does not happen when scheduling).

differentreality:
What's remaining is:
* Max_attendees is set to 10, we have 10 registrations, then max_attendees is reduced to 6. The registrations will then remain.
* Max_attendees is set to 10, the event gets assigned a room and the room.size is 5. What should happen to the max_attendees value and what if there were already 10 registrations (like the previous scenario)

app/models/event.rb
hennevogel: Sounds like a plan yes

app/models/user.rb
hennevogel: So it returns false or a record? I would split that...

hennevogel:
This simple shouldn't be possible. If you have 10 registrations then you can't lower it below 10. You shouldn't be able to schedule an Event with 10 registrations to a room of the size 5.

differentreality:
I thought about that but I am not sure if we should put one more limitation to scheduling an event. It is unlikely, but what if the numbers are bigger and there is simply not a large enough room available? You cannot schedule the Event and you cannot reduce max_attendeeds. It creates a huge block for something that is not so serious.

Alternative: only allow people to register to events that are scheduled (not just confirmed).

hennevogel:
Well if the app can't resolve something (like removing attendees and notifying them about it or something similar) we need to error and tell the people how they can resolve it themselves.

Or we don't allow the data to get into this situations. I think your alternative would do that...

differentreality:
On the other hand, having an idea how many people are interested in a talk is interesting for organizers.
It is perhaps a better idea to address the issues and make event registrations a more complete feature.

1) Max_attendees is 10 and we have 10 registrations to Event. Event Submitter wants to reduce the value of max_attendees. People will need to be unregistered and receive a relevant email.

2) Max_attendees is 100 and we have 80 registrations, but the event is scheduled in a room with capacity of 75 people, last 5 registrations need to be deleted (with email to people and event submitter).

3) Event has attribute require_regitsration, and people already have registered. Event submitter decides to uncheck the require_registration attribute. Registrations should be deleted (with email to people).

Once #932 is finalized, I would introduce some reminder emails to people registered to events (Hey you ve registered to attend X, Y, Z events. If for some reason you cannot make it to these events, please unregister, to free up places for other people who want to attend).

hennevogel:
Anyway. Let's get this now in as it is. This is already large enough :-)

PR #926, opened 2016-03-29
Mailer fixes and tests

sonalkr132:
I need help testing following:
```Ruby
def conference_date_update_mail(conference)
    User.joins(:subscriptions).merge(conference.subscriptions).each do |user|
      build_email(conference,
                  user.email,
                  conference.email_settings.conference_dates_updated_subject,
                  conference.email_settings.generate_email_on_conf_updates(conference, user, conference.email_settings.conference_dates_updated_body))
    end
  end
```
What do I even test here? `do..end` block returns nothing!


ChrisBr:
LGTM

PR #927, opened 2016-03-30
Fix reference to non-existant route path in commercials_controller

nishanthvijayan:
The route edit_conference_proposal_path was removed in [this commit](https://github.com/openSUSE/osem/commit/444356fb)


differentreality:
Should we perhaps add a test case for when the creation and update are not successful? @ChrisBr

hennevogel:
Yes we should

nishanthvijayan:
updated.

differentreality:
I would also test that we are on the right path, which is precisely what we are changing in this PR.

nishanthvijayan:
Done

hennevogel:
Awesome, thanks @nishanthvijayan

PR #933, opened 2016-03-31
Add Docker-Compose support

dronov:
Hello! 

I tried to pack OSEM rails app into docker container and run it in development environment using Docker-Compose tool. I moved ```tzinfo-data``` gem to Gemfile cause container had been broken without it (you can see the solution to except it at Dockerfile). Also I installed nokogiri by strange way, cause bundler had some problems and had not succeed in solving them. 

It would be great if someone will give me some feedback. :)
Thanks a lot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM opensuse:42.1
+RUN zypper -q -n --non-interactive install update-alternatives ruby-devel make gcc gcc-c++ libxml2 libxml2-devel libxslt-devel nodejs screen libmysqld-devel sqlite3-devel imagemagick
+
+ENV APP_HOME /osem
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+ADD Gemfile* $APP_HOME/
+COPY Gemfile Gemfile
+COPY Gemfile.lock Gemfile.lock
+
+RUN echo 'install: --no-format-executable' >> /etc/gemrc
+RUN gem install bundler
+RUN gem install nokogiri -v '1.6.7.2' -- --use-system-libraries
+RUN bundle install
+RUN gem install tzinfo
+RUN gem install tzinfo-data
+
+ADD . $APP_HOME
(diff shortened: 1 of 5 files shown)


ChrisBr:
@kalabiyau that's your thing ;)

dronov:
My thoughts was to built an app with docker for development purposes, so I decided to go the simple way and use sqlite. I am not sure that this is the most correct way, cause I have never used sqlite for development, and maybe I should setup mariadb container and use it in OSEM rails app. 

BTW, @ChrisBr showed this issue https://github.com/openSUSE/mentoring/issues/23
I haven't saw it before submitting this PR :\

hennevogel:
We're already using Vagrant for development. Why don't you [integrate](https://www.vagrantup.com/docs/docker/basics.html) docker instead of doing it standalone? :-)

hennevogel:
@dronov any news on this?

dronov:
@hennevogel Hello! 
I made standalone Dockerfile cause is it the usual way run app. And yeah, I did not know that the Docker could be use as a integration.

Btw, there is a connected issue, so maybe is it ok to pack OSEM rails app into container?

hennevogel:
*usual way* is kind of a weak argument when it comes to containers... I would definitely want to avoid to add another tool to the mix we need to maintain...

dronov:
@hennevogel docker don't have a lot of problems with maintaning project's dependencies. All dependencies fixed in Dockerfile and the're similar at all environments using Docker.

Anyway, I understand you. And sorry if I make some problems, it was not polite. 
Could I take another issue to solve?

ChrisBr:
@dronov yes, sure! Just go ahead, it's open source, we welcome every contribution :+1:

ChrisBr:
@dronov @hennevogel what's the final conclusion of this PR?

hennevogel:
That we either use docker-compose or Vagrant. Not both.

PR #934, opened 2016-03-31
Remove unused call_for_paper model and controller files

nishanthvijayan:
call_for_paper model and controller are not being currently used.

PR #935, opened 2016-04-01
Update coverall to 0.8.13

sonalkr132:
Let's not live in false sense of security!
screenshot from 2016-04-01 08-55-21


sonalkr132:
#934 should also help, actually I hit me to update coveralls after seeing that PR. When it should have been coveralls telling us about existence of such files.

hennevogel:
LGTM

PR #941, opened 2016-04-01
Fixes ability for submitting proposal

ChrisBr:
Submitting proposal wasn't possible anymore for unprivileged users. Introduced in 3e32c58460e5af06c87511c64cdeb3834c401e42

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -58,7 +58,7 @@ class Ability
         event.new_record?
       end
 
-      can [:new, :create], Event do |event|
+      can :new, Event do |event|
         event.program.cfp_open? && event.new_record?
       end
     end
@@ -82,14 +82,19 @@ class Ability
 
     can [:create, :destroy], Subscription, user_id: user.id
 
-    can :manage, Event do |event|
+    can [:read, :update, :destroy], Event do |event|
       event.users.include?(user)
     end
 
-    # cannot create an event if program does not have open cfp
-    cannot [:new, :create], Event do |event|
-      user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
-      !event.program.cfp_open? || !event.new_record? || !user_inclusion
+    can [:new], Event do |event|
+      event.program.cfp_open? && event.new_record?
+    end
+
+    can [:create], Event do |event|
+      # event.users.include?(user) doesn't work here because the event_user object isn't saved when we call authorize!
+      # Checks that only the current_user is in event_users
+      user_inclusion = event.event_users.map { |event_user| event_user.user_id }.uniq - [ user.id ]
+      user_inclusion.empty? && event.program.cfp_open? && event.new_record?
     end
 
     # can manage the commercials of their own events
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -2,7 +2,7 @@ class ProposalController < ApplicationController
   before_filter :authenticate_user!, except: [:show, :new, :create]
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true
-  load_and_authorize_resource :event, parent: false, through: :program, except: [:new, :create]
+  load_and_authorize_resource :event, parent: false, through: :program
 
   def index
     @event = @program.events.new
@@ -16,8 +16,8 @@ class ProposalController < ApplicationController
   end
 
   def new
-    @event = @program.events.new
     @event.event_users.new(user: current_user, event_role: 'submitter') if current_user
+    @event.event_users.new(user: current_user, event_role: 'speaker') if current_user
     authorize! :new, @event
     @user = User.new
     @url = conference_program_proposal_index_path(@conference.short_title)
@@ -45,18 +45,7 @@ class ProposalController < ApplicationController
     end
 
     params[:event].delete :user
-
-    @event = Event.new(event_params)
-    @event.program = @program
-
-    # User which creates the proposal is both `submitter` and `speaker` of proposal
-    # by default.
-    # TODO: Allow submitter to add speakers to proposals
-    @event.event_users.new(user: current_user,
-                           event_role: 'submitter')
-    @event.event_users.new(user: current_user,
-                           event_role: 'speaker')
-    authorize! :new, @event
+    authorize! :create, @event
 
     if @event.save
       ahoy.track 'Event submission', title: 'New submission'
@@ -152,7 +141,7 @@ class ProposalController < ApplicationController
   def event_params
     params.require(:event).permit(:event_type_id, :track_id, :difficulty_level_id,
                                   :title, :subtitle, :abstract, :description,
-                                  :require_registration, :max_attendees)
+                                  :require_registration, :max_attendees, event_users_attributes: [:user_id, :event_role])
   end
 
   def user_params
--- a/app/views/conference/_conference_details.html.haml
+++ b/app/views/conference/_conference_details.html.haml
@@ -32,7 +32,7 @@
                   Sorry, no places left
             - if !current_user.nil? && current_user.proposal_count(conference) > 0
               = link_to "My Proposals", conference_program_proposal_index_path(conference.short_title), :class =>"btn btn-default"
-            - elsif conference.program.cfp_open?
+            - elsif can? :new, conference.program.events.new
               = link_to "Submit Proposal", new_conference_program_proposal_path(conference.short_title), :class =>"btn btn-default"
             - if current_user.nil? || !current_user.subscribed?(conference)
               = link_to 'Subscribe', conference_subscriptions_path(conference.short_title), method: :post, class: 'btn btn-default'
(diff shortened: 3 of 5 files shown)


ChrisBr:
@differentreality was there a reason why the cfp and new record check was in the in the ichain conditional?

app/models/ability.rb
```diff
@@ -45,6 +45,11 @@ def not_signed_in
     # can view Commercials of confirmed Events
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can [:show, :create], User
+
+    can [:new, :create], Event do |event|
+      event.program.cfp_open? && event.new_record?
```
differentreality: For the not_signed_in user that includes user creation, which we don't want to have in case of ichain (right?).

app/models/ability.rb
```diff
@@ -45,6 +45,11 @@ def not_signed_in
     # can view Commercials of confirmed Events
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can [:show, :create], User
+
+    can [:new, :create], Event do |event|
```
differentreality: For the not_signed_in user this includes user creation, which we don't want to have with ichain (right?).

app/models/ability.rb
```diff
@@ -87,9 +88,8 @@ def signed_in(user)
     end
 
     # cannot create an event if program does not have open cfp
-    cannot [:new, :create], Event do |event|
-      user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
-      !event.program.cfp_open? || !event.new_record? || !user_inclusion
+    can [:new, :create], Event do |event|
```
differentreality: This is included in https://github.com/ChrisBr/osem/blob/fix_ability/app/models/ability.rb#L86
```
can :manage, Event do |event|
  event.users.include?(user)
end
```

The extra thing we want is to supersede that `can :manage` and disallow access (cannot) if the cfp is closed.

app/models/ability.rb
ChrisBr: yes, but these two lines are valid for iChain on or off, aren't they?

app/models/ability.rb
ChrisBr: So what do you suggest?

ChrisBr:
```
!event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id 
```

app/models/ability.rb
differentreality: But with ichain we do **not** create users from within OSEM.
So these lines are only valid when iChain is **off**. Because those lines refer to not_signed_in user, in which case the new action of proposal form includes fields for user account creation.

app/models/ability.rb
differentreality: I think we need
```
cannot [:new, :create], Event do |event|
  !event.program.cfp_open?
end
```
But I haven't tested that.

differentreality:
The error you are referring to is on https://events.opensuse.org/conference/oSC16/program/proposal/new when you are **NOT** signed in?

differentreality:
:+1: events.o.o seems to be liking it too!

ChrisBr:
Just for documentation:
This is necessary because the manage block does not cover it. Usually :manage also includes :new and :create. However we check: 

```
    can :manage, Event do |event|
      event.users.include?(user)
    end
```
``      event.users.include?(user)`` is always false because the event_users record is not saved when we enter the ability. So that means that we need to explicitly cover the new and create actions if the cfp is open!

ChrisBr:
Yeah, it's already deployed to events.o.o as a quick fix!

ChrisBr:
@differentreality also mentioned that it shouldn't be possible to create an event with someone else as a submitter, we have to think about it (probably atm possible)!

sonalkr132:
> Submitting proposal wasn't possible anymore for unprivileged users.

Can you please elaborate on that? Especially *unprivileged users* part. I am able to do following on master:
* create a conference `goto`, and assign cfp dates (range should include current date).
* sign out from admin user and sign in with a different user (has no `role` in `goto`)
* go to new proposal page: `http://localhost:3000/conference/dsd/program/proposal/new`
* fill in fields and submit proposal

___

from 381e8ec :
> This is necessary because the manage block does not cover it. event.users.include?(user) is always false for :new and :create because the event_users are not saved yet.

`manage` block is not applicable for `:new` and `:create` action, the follow up `cannot` block overrode the `manage` block.

app/models/ability.rb
```diff
@@ -86,10 +86,8 @@ def signed_in(user)
       event.users.include?(user)
     end
 
-    # cannot create an event if program does not have open cfp
-    cannot [:new, :create], Event do |event|
-      user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
-      !event.program.cfp_open? || !event.new_record? || !user_inclusion
+    can [:new, :create], Event do |event|
+      event.program.cfp_open? && event.new_record?
     end
```
sonalkr132: > it shouldn't be possible to create an event with someone else as a submitter, we have to think about it (probably atm possible)! 

yes, it will be possible to create an event with someone else as a submitter on your branch. `can` followed by `can` rule are `ORed`. Equivalent rule for what you have used is:
```Ruby
event.users.include?(user) ||  event.program.cfp_open? && event.new_record?
```

ChrisBr:
We use iChain on production! Set iChain to true and try your workflow again ...

app/models/ability.rb
ChrisBr: Nope

sonalkr132:
@ChrisBr please check #943 :confused:

app/models/ability.rb
sonalkr132: From cancancan wiki on [Ability Precedence](https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence):
> Adding can rules do not override prior rules, but instead are logically or'ed.

ChrisBr: I know, I've read the readme ;)

Anyway, creating an event with someone else as submitter was possible before, too! And checking user.include doesn't work for new and create!

ChrisBr: @sonalkr132 no, this is wrong
``
event.users.include?(user) ||  event.program.cfp_open? && event.new_record?
``
is not equivalent to what I used! Because I only use 
``
event.program.cfp_open? && event.new_record?
``
for :create and :new authorization.
But to be more clear, I changed manage to :update, :read, :destroy and the other one to :create!

ChrisBr:
``event.users.include?`` doesn't work, however, we can use event.event_users like @sonalkr132 did it before. 

The reason why it worked for iChain disabled was that it was explicitly allowed for not signed in users (see https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L61). We don't want to allow that when iChain is enabled because User creation is handled by the iChain proxy! So for iChain enabled it was never allowed (the condition in :manage event was ALWAYS wrong). I splitted the condition now in two conditions: the first one for update, destroy, read if user is included and the second one for create if cfp is open and event_users includes the user. Hope that clarifies a little bit!

ChrisBr:
Ok, so I've run out of ideas ... I don't see any possibility to check that event_users only includes the current_user. The problem with ``event.event_users`` is that the ability instances a new event object and it uses the event_params hash, which does not include the event_users (we override it in the create action). Also using load_resource and dropping the overriding does not work because of the noted reasons.

Atm it's anyway not possible to add someone else as event_user because we filter it in the strong parameters. 

@hennevogel @differentreality any idea how to continue?

hennevogel:
But `event_users` empty is also okay. What's the problem with this? Events can be created with either yourself in or empty `event_users`

ChrisBr:
@hennevogel Example

proposals_controller:
``
@event.event_users.new(user: 'intruder', role: 'submitter')
@event.event_user.new(user: 'me', role: 'submitter')
authorize! :create, @event
``

ability.rb
``
can :create, Event do |event|
  user_inclusion = event.event_users.map { |event_user| event_user.user.id }.compact.include? user.id
  event.event_users.empty || user_inclusion
end
``

-> you would expect that @event.event_users has two users assigned and the ability check fails (because there is the user 'intruder'), but event_users is empty and the ability check succeeds! And in the end you have these two event_users!

hennevogel:
Alright, time to make this a has_many/belongs_to between User and Event

ChrisBr:
@hennevogel can you please elaborate? There is already a assocation between user -> event_users <- event

ChrisBr:
@hennevogel we have to fix this before 1.0 because it's currently broken for iChain, so it's not possible to submit a proposal when iChain is enabled!

differentreality:
@ChrisBr what if we just do
`@event.users << current_user`
in the proposals controller? Without changing ability? 
That will add a user to the event, with role participant. (Create action adds submitter/speaker, we can just delete the event_user record for participant.)

hennevogel:
I mean switching from an Event has_many Users to Event belongs to user.

ChrisBr:
@differentreality as discussed with @hennevogel: We will change the relation in the future but not for the 1.0 release as we will most likely introduce some regressions! Instead we will stick with this implementation. I'll fix up the test failures on the weekend!

ChrisBr:
Superseded by https://github.com/openSUSE/osem/pull/1001

Issue #903, opened 2016-03-23
Roles authorization


sonalkr132:
Steps to reproduce:
* create conference (say: `goto`)
* sign up two user `cfp_member` and `another_organizer`
* sign in with admin and assign `cfp_member` to `cfp` team and `another_organizer` to `organizer` team.
* sign in with `cfp_member` user
* go to roles page of goto conf: `http://localhost:3000/admin/conference/goto/roles`
* click on `users` button; page with go to organizers show page
* toggle `another_organizer` (it is crucial that you have two organizers)

You just removed an organizer from his own conference being in cfp team :sweat_smile:
You can also add any user to any team, edit any role (description).

differentreality:
I will look into it.

Rolence:
I did what you outlined above but didn't remove my organizer. Can you elaborate more on the bug!!

sonalkr132:
![ezgif-1316420179](https://cloud.githubusercontent.com/assets/7680662/14077602/cbff026e-f50a-11e5-8b24-b55e143e56c9.gif)

sonalkr132:
It is happening cause `authorize! @role, :toggle_user` is missing from `toggle_action` of `Admin::RolesControlle`

differentreality:
Nice gif ;) 
I can confirm the issue.

nishanthvijayan:
The issue seems to be caused by [this line](https://github.com/openSUSE/osem/blob/master/app%2Fmodels%2Fability.rb#L201)
:manage includes all actions on the object and so includes :toggle_user as well
and if there are multiple can rules on same action and object they are logically or'ed
So the [rule on toggle_user](https://github.com/openSUSE/osem/blob/master/app%2Fmodels%2Fability.rb#L207) becomes useless
and like @sonalkr132 mentioned there is no authorization check in both admin/roles_controller#toggle_user and admin/roles_controller#update

PR #942, opened 2016-04-02
Fix authorization issues with Roles

nishanthvijayan:
Only organizers should be able to add/remove users from all roles
Cfp/info_desk/voluters coordinators should be to able to add/remove users only to their own teams

Also organizers of a contest should be able to edit/update the description of any role of that contest.
Cfp/info_desk/voluters coordinators should be to able to edit/update the description of their role only.

Solves #903

### DIFF
--- a/app/controllers/admin/roles_controller.rb
+++ b/app/controllers/admin/roles_controller.rb
@@ -2,11 +2,13 @@ module Admin
   class RolesController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     before_action :set_selection
+    authorize_resource :role, except: :index
     # Show flash message with ajax calls
     after_action :prepare_unobtrusive_flash, only: :toggle_user
 
     def index
       @roles = Role.where(resource: @conference)
+      authorize! :index, @role
     end
 
     def show
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -123,9 +123,7 @@ class Ability
 
   def signed_in_with_organizer_role(user)
     # ids of all the conferences for which the user has the 'organizer' role
-    conf_ids_for_organizer = []
-    conf_ids_for_organizer =
-        Conference.with_role(:organizer, user).pluck(:id) if user.has_role? :organizer, :any
+    conf_ids_for_organizer = Conference.with_role(:organizer, user).pluck(:id)
 
     can [:new, :create], Conference if user.has_role?(:organizer, :any)
     can :manage, Conference, id: conf_ids_for_organizer
@@ -162,24 +160,15 @@ class Ability
                          commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids_for_organizer).pluck(:id)).pluck(:id)
 
     # Abilities for Role (Conference resource)
-    can :index, Role
-    can :manage, Role do |role|
+    can [:index, :show], Role
+    can [:edit, :update, :toggle_user], Role do |role|
       role.resource_type == 'Conference' && (conf_ids_for_organizer.include? role.resource_id)
-    end
-
-    # Can add or remove users from role, when user has that same role for the conference
-    # Eg. If you are member of the CfP team, you can add more CfP team members (add users to the role 'CfP')
-    can :toggle_user, Role do |role|
-      role.resource_type == 'Conference' &&
-      (Conference.with_role(role.name.parameterize.underscore.to_sym, user).pluck(:id).include? role.resource_id)
     end
   end
 
   def signed_in_with_cfp_role(user)
     # ids of all the conferences for which the user has the 'cfp' role
-    conf_ids_for_cfp = []
-    conf_ids_for_cfp =
-      Conference.with_role(:cfp, user).pluck(:id) if user.has_role? :cfp, :any
+    conf_ids_for_cfp = Conference.with_role(:cfp, user).pluck(:id)
 
     can :manage, Event, program: { conference_id: conf_ids_for_cfp }
     can :manage, EventType, program: { conference_id: conf_ids_for_cfp }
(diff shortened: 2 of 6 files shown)


app/controllers/admin/roles_controller.rb
```diff
@@ -2,11 +2,13 @@ module Admin
   class RolesController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     before_action :set_selection
+    authorize_resource :role
     # Show flash message with ajax calls
     after_action :prepare_unobtrusive_flash, only: :toggle_user
 
     def index
       @roles = Role.where(resource: @conference)
+      authorize! :index, @roles
```
sonalkr132: Doesn't `authorize_resource :role` already cover `index` action? Why do we need it here too?

sonalkr132:
It would really help while reviewing it, if you could add [ability tests](https://github.com/openSUSE/osem/blob/master/spec/models/ability_spec.rb) for changes you have made.

app/models/ability.rb
```diff
@@ -162,17 +162,9 @@ def signed_in_with_organizer_role(user)
                          commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids_for_organizer).pluck(:id)).pluck(:id)
 
     # Abilities for Role (Conference resource)
-    can :index, Role
```
differentreality: That is actually on purpose here.

app/models/ability.rb
```diff
@@ -196,15 +188,13 @@ def signed_in_with_cfp_role(user)
                          commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids_for_cfp).pluck(:id)).pluck(:id)
 
     # Abilities for Role (Conference resource)
-    can :index, Role
-
-    can :manage, Role do |role|
+    can :read, Role do |role|
```
differentreality: Even though what you are doing is technically correct, I prefer not using read but explicit actions instead. 
* It will make the code more readable (not everyone knows what read corresponds to)
* And it will protect us, in the future, from changes made to read in cancancan gem, which we might miss when updating it

app/models/ability.rb
```diff
@@ -196,15 +188,13 @@ def signed_in_with_cfp_role(user)
...
       role.resource_type == 'Conference' && (conf_ids_for_cfp.include? role.resource_id)
     end
 
-    # Can add or remove users from role, when user has that same role for the conference
+    # Can add or remove users from role and edit role description, when user has that same role for the conference
     # Eg. If you are member of the CfP team, you can add more CfP team members (add users to the role 'CfP')
-    can :toggle_user, Role do |role|
+    can [:toggle_user, :update, :edit], Role do |role|
       role.resource_type == 'Conference' &&
```
differentreality: I would only keep toggle user here. This is not from a technical perspective, but from an essential one. I'd rather allow only the organizer to change the description of roles in his/her conference.

nishanthvijayan: Yea, that makes sense.

app/models/ability.rb
```diff
@@ -197,14 +190,13 @@ def signed_in_with_cfp_role(user)
...
       (Conference.with_role(role.name.parameterize.underscore.to_sym, user).pluck(:id).include? role.resource_id)
```
nishanthvijayan: `parameterize.underscore`  are they really required?
None of the 4 role names have any special characters in the.

app/models/ability.rb
```diff
@@ -240,20 +232,20 @@ def signed_in_with_volunteers_coordinator_role(user)
     # ids of all the conferences for which the user has the 'volunteers_coordinator' role
     conf_ids_for_volunteers_coordinator = []
     conf_ids_for_volunteers_coordinator =
-        Conference.with_role(:volunteer_coordinator, user).pluck(:id) if user.has_role? :volunteer_coordinator, :any
+        Conference.with_role(:volunteers_coordinator, user).pluck(:id) if user.has_role? :volunteers_coordinator, :any
```
nishanthvijayan: @differentreality  this `if user.has_role? :volunteers_coordinator, :any` seems redundant.Is there a reason why its here?

differentreality: There is not a reason any more, no. Now we already check that in https://github.com/nishanthvijayan/osem/blob/fix-roles-authorization/app/models/ability.rb#L100

It is important to check somewhere so that we can reduce the code executed for authorization. We only need to execute the relative lines, not everything for everyone. As we are still having a lot of calculations being made for authorization.

spec/models/ability_spec.rb
```diff
@@ -190,12 +191,18 @@
       it{ should_not be_able_to(:manage, other_event.commercials.first) }
       it{ should be_able_to(:index, my_event.comment_threads.first) }
       it{ should_not be_able_to(:index, other_event.comment_threads.first) }
+
+      it{ should be_able_to(:toggle_user, cfp_role) }
+      it{ should be_able_to(:update, cfp_role) }
```
differentreality: If we check for cfp role, perhaps it makes sense to check for the other roles too? In a shared example?

app/models/ability.rb
```diff
@@ -240,20 +232,20 @@ def signed_in_with_volunteers_coordinator_role(user)
...
+    can :show, Role do |role|
       role.resource_type == 'Conference' && (conf_ids_for_volunteers_coordinator.include? role.resource_id)
     end
 
-    # Can add or remove users from role, when user has that same role for the conference
+    # Can add or remove users from role and edit role description, when user has that same role for the conference
     # Eg. If you are member of the CfP team, you can add more CfP team members (add users to the role 'CfP')
-    can :toggle_user, Role do |role|
+    can [:toggle_user, :update, :edit], Role do |role|
       role.resource_type == 'Conference' &&
```
nishanthvijayan: Why are we checking if `role.resource_type == 'Conference'` ?

differentreality: Because we could use roles on another resource (even though we are not currently). This will keep our code correct no matter what we do in the future.

spec/models/ability_spec.rb
nishanthvijayan: We could check ftggle and update of all roles,but why in a shared example?

differentreality: To reduce the lines of code as we will be testing the same thing for multiple roles. Choose whichever way you prefer.

nishanthvijayan:
Updated.

nishanthvijayan:
Signed in user has only cfp role and so the toggle switch is kept hidden in organizers page

toggle_hidden
toggle_visible
info_desk_toggle_hidden
vol_coord_info_hidden

differentreality:
@nishanthvijayan I think we are missing some test casee. We should be testing can and cannot for all available actions of the controller, for all roles of the conference for which the user has a role, and for the roles of other conferences:
Eg. 
  * All roles can index Role for all conferences (conference where they have a role, conference where they don't have a role).
  * CfP role `can, :toggle_user, cfp_role`, but not other roles of that conference. `Cannot :toggle_user, other_conference`, `can :show, role` for all roles of all conferences.

nishanthvijayan:
@differentreality  updated

differentreality:
:+1:

PR #944, opened 2016-04-02
Serializers test

sonalkr132:
So far, 0% coverage gain. Yay!

### DIFF
--- a/app/serializers/event_serializer.rb
+++ b/app/serializers/event_serializer.rb
@@ -5,7 +5,7 @@ class EventSerializer < ActiveModel::Serializer
 
   def date
     t = object.start_time
-    t.blank? ? '' : %{ #{l t, format: :short}#{t.formatted_offset(false)} }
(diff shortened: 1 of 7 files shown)


app/serializers/event_serializer.rb
```diff
@@ -5,7 +5,7 @@ class EventSerializer < ActiveModel::Serializer
 
   def date
     t = object.start_time
-    t.blank? ? '' : %{ #{l t, format: :short}#{t.formatted_offset(false)} }
+    t.blank? ? '' : %{ #{I18n.l t, format: :short}#{t.formatted_offset(false)} }
```
nishanthvijayan: I think you can remove t.formatted_offset(false) and add it to :short  by adding ":z" .I'm not sure, so please do verify they give the same output.

kalabiyau: @nishanthvijayan I would recommend to look at the very last comment in #904 about that case

hennevogel: This is out of scope for this PR....

hennevogel:
Awesome work. Thanks

PR #946, opened 2016-04-04
Defined Venue has_one Commercial nice new relationship

Shriyanshagro:
#777 
Defined an association of Venue has_one Commercial nice new relationship.

### DIFF
--- a/app/controllers/admin/venue_commercials_controller.rb
+++ b/app/controllers/admin/venue_commercials_controller.rb
@@ -9,10 +9,10 @@ module Admin
       authorize! :create, @commercial
 
       if @commercial.save
-        redirect_to edit_admin_conference_venue_path,
+        redirect_to admin_conference_venue_path,
                     notice: 'Commercial was successfully created.'
       else
-        redirect_to edit_admin_conference_venue_path,
+        redirect_to admin_conference_venue_path,
                     error: 'An error prohibited this Commercial from being saved: '\
                     "#{@commercial.errors.full_messages.join('. ')}."
 
@@ -21,10 +21,10 @@ module Admin
 
     def update
       if @commercial.update(commercial_params)
-        redirect_to edit_admin_conference_venue_path,
+        redirect_to admin_conference_venue_path,
                     notice: 'Commercial was successfully updated.'
       else
-        redirect_to edit_admin_conference_venue_path,
+        redirect_to admin_conference_venue_path,
                     error: 'An error prohibited this Commercial from being saved: '\
                     "#{@commercial.errors.full_messages.join('. ')}."
       end
@@ -32,7 +32,7 @@ module Admin
 
     def destroy
       @commercial.destroy
-      redirect_to edit_admin_conference_venue_path, notice: 'Commercial was successfully destroyed.'
+      redirect_to admin_conference_venue_path, notice: 'Commercial was successfully destroyed.'
     end
 
     def render_commercial
--- a/app/controllers/admin/venue_commercials_controller.rb
+++ b/app/controllers/admin/venue_commercials_controller.rb
@@ -10,10 +10,10 @@ module Admin
 
       #FIXME redirect to tab#commercial
       if @commercial.save
-        redirect_to admin_conference_venue_path,
+        redirect_to edit_admin_conference_venue_path,
                     notice: 'Commercial was successfully created.'
       else
-        redirect_to admin_conference_venue_path,
+        redirect_to edit_admin_conference_venue_path,
                     error: 'An error prohibited this Commercial from being saved: '\
                     "#{@commercial.errors.full_messages.join('. ')}."
 
@@ -22,10 +22,10 @@ module Admin
 
     def update
       if @commercial.update(commercial_params)
-        redirect_to admin_conference_venue_path,
+        redirect_to edit_admin_conference_venue_path,
                     notice: 'Commercial was successfully updated.'
       else
-        redirect_to admin_conference_venue_path,
+        redirect_to edit_admin_conference_venue_path,
                     error: 'An error prohibited this Commercial from being saved: '\
                     "#{@commercial.errors.full_messages.join('. ')}."
       end
@@ -33,7 +33,7 @@ module Admin
 
     def destroy
       @commercial.destroy
-      redirect_to admin_conference_venue_path, notice: 'Commercial was successfully destroyed.'
+      redirect_to edit_admin_conference_venue_path, notice: 'Commercial was successfully destroyed.'
     end
 
     def render_commercial
--- a/app/controllers/admin/venue_commercials_controller.rb
+++ b/app/controllers/admin/venue_commercials_controller.rb
@@ -3,6 +3,7 @@ module Admin
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :venue ,through: :conference, singelton: true
     before_action :set_venue
+    before_action :set_commercial , only: [:update , :destroy]
 
     def create
       @commercial = @venue.build_commercial(commercial_params)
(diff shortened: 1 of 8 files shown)


hennevogel:
Please provide screenshots and fix the rubocop issues

giphy

Shriyanshagro:
ezgif com-crop
Here is a gif to show new added association (Venue-Commercial) with implied Views.
Also By now I have not removed :photos from venue attributes , will replace  it once mentor approves of this association.

Shriyanshagro:
update
Here is a gif to show update and delete functionality of commercials under venue.

Shriyanshagro:
Rubocop has three pending test , and I think those pending test have nothing to do with the changes I have made. Can maintainers please look into the failures ?

Here is the attached screenshot of failures.  
pending

hennevogel:
The problems are described [at the top](https://travis-ci.org/openSUSE/osem/builds/120742881#L1737) of the logfile. You can also run rubocop locally with 

`bundle exec rubocop`

app/controllers/admin/venue_commercials_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+    end
+
+    private
+
+    def commercial_params
+      params.require(:commercial).permit(:url)
+    end
+
+    def set_venue
```
ChrisBr: You don't need that, load_and_authorize resource already does that for your!
And you can also remove the before filter!

app/controllers/admin/venue_commercials_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      @venue = @conference.venue
+    end
+
+    def set_commercial
```
ChrisBr: You don't need that, load_and_authorize resource already does that for your!
And you can also remove the before filter!

ChrisBr: Maybe you need an load_and_authorize commercial filter for it. Make sure that you test it as a non admin user (e.g. organizer)

ChrisBr:
@Shriyanshagro you should also display the venue in conference show, otherwise only the admin sees the commercial which doesn't makes sense! Maybe we can even display the commercial on the event site if there is no event or conference commercial!

Futhermore we could drop the photo if we have the commercial? @hennevogel what do you think? We would need to think about a way to migrate it but we could at least remove it from the ui so that admins can't add it in the future? (Which is another step to remove dependencies of storing images on our server)

So, in general :+1: @Shriyanshagro! Good idea!

app/controllers/admin/venue_commercials_controller.rb
Shriyanshagro: Yeah you are correct.
I guess only 
load_and_authorize_resource :commercial, find_by: :id
would also work ??

app/controllers/admin/venue_commercials_controller.rb
```diff
@@ -0,0 +1,62 @@
+module Admin
+  class VenueCommercialsController < Admin::BaseController
+    load_and_authorize_resource :conference, find_by: :short_title
+    load_and_authorize_resource :venue, through: :conference, singelton: true
```
Shriyanshagro: Yeah , I guess this line is not needed here.

ChrisBr: The other way around! Leave this line and remove your self coded set_venue

app/controllers/admin/venue_commercials_controller.rb
Shriyanshagro: No , it isn't loading resources while updating and deleting, without this.
And even if add this -
It's not working

app/controllers/admin/venue_commercials_controller.rb
Shriyanshagro: Yeah, that would also work. Instead of this I was using @conference.venue, but ratherless, I guess this is better idea.

ChrisBr: It's ``singleton: true`` and not ``singelton: true``

app/controllers/admin/venue_commercials_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    before_action :set_venue
+    before_action :set_commercial, only: [:update, :destroy]
```
ChrisBr: Before actions not necessary, should be handled by cancancan

app/controllers/admin/venue_commercials_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def create
+      @commercial = @venue.build_commercial(commercial_params)
```
ChrisBr: Should also not necessary, cancancan will do that for you!

app/controllers/admin/venue_commercials_controller.rb
Shriyanshagro: Done :+1:

Shriyanshagro:
venue
Gif of new view for venue page

Shriyanshagro:
splashpage
Gif of new view of SplashPage, with added Commercial for venue.

sonalkr132:
@Shriyanshagro Please [rebase](http://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master) to fix hakiri warnings and [squash](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits) commits c3dab64 and fed48b3

Shriyanshagro:
@sonalkr132, what about Hakiri issues here??

sonalkr132:
I am not sure. We have them marked false positive on hakiri.
Your branch seems to be 2 commits behind the current head of our canonical master (openSUSE/osem's master), make sure you have them include them as well (ie fetch/pull the changes before you rebase).

Shriyanshagro:
@sonalkr132, it still didn't work, I guess we need to mark this false positive too.

sonalkr132:
> I guess we need to mark this false positive too.

Yes. Lines with warning are introduced in this PR and we don't have them marked false positive.

Shriyanshagro:
Can someone mark hakiri warnings false positive. Or either guide me how to do so?

app/views/conference/_venue.html.haml
```diff
@@ -6,12 +6,21 @@
 -else
   .container
     .row
-      .col-md-12
+      - if @conference.venue.commercial.nil?
+        .col-md-6
+          %img{ "data-src" => "holder.js/500x300?text=No Commercial Set", class: 'img-responsive img-rounded' }
+
+      -else @conference.venue.commercial.persisted?
```
hennevogel: else takes no condition. do you mean elsif?

app/views/conference/_venue.html.haml
```diff
@@ -6,12 +6,21 @@
 -else
   .container
     .row
-      .col-md-12
+      - if @conference.venue.commercial.nil?
+        .col-md-6
+          %img{ "data-src" => "holder.js/500x300?text=No Commercial Set", class: 'img-responsive img-rounded' }
```
hennevogel: Please do not show placeholder images to visitors. If there is no commercial, just show the address

hennevogel:
Otherwise great work @Shriyanshagro

app/views/conference/_venue.html.haml
```diff
@@ -6,11 +6,7 @@
 -else
   .container
     .row
-      - if @conference.venue.commercial.nil?
-        .col-md-6
-          %img{ "data-src" => "holder.js/500x300?text=No Commercial Set", class: 'img-responsive img-rounded' }
-
-      -else @conference.venue.commercial.persisted?
```
hennevogel: Why `.persisted?`

PR #947, opened 2016-04-05
Switch to trusty as base OS

hennevogel:
Maybe this get's us more reliable travis workers. Especially with phantomjs...


ChrisBr:
Why do you need to adapt the tests?

Otherwise lgtm ...

PR #950, opened 2016-04-07
Add devise confirmation controller test

sonalkr132:
It would be awesome if we could test path here but response after confirmation is not redirect[1] but 200. If anyone knows how to test response `path` here, please let me know. 

[1] `expect(response).to redirect_to path` won't work


ChrisBr:
LGTM

PR #951, opened 2016-04-07
Sponsors

ChrisBr:
The sponsors section is really ugly... At least some small improvement
Before
osem3

After:
osem1
osem2

### DIFF
--- a/app/views/conference/_sponsors.html.haml
+++ b/app/views/conference/_sponsors.html.haml
@@ -6,28 +6,17 @@
   .row
     .col-md-12
       %h2 Sponsors
-      %p.lead
-        whom without,
-        %span.notranslate
-          = @conference.short_title
-        wouldn't be possible! 
   - @conference.sponsorship_levels.each do |sponsorship_level|
     -if sponsorship_level.sponsors.any?
+      %h4
+        = sponsorship_level.title
       - sponsorship_level.sponsors.each_slice(3) do |slice|
-        .row.row-centered
+        .row
           - slice.each do |sponsor|
-            .col-md-4.col-sm-4.col-centered.col-top
-              .thumbnail
-                - if sponsor.logo.blank?
-                  %h3.text-center
-                    = link_to(sponsor.website_url, title: "#{sponsor.name}") do
-                      = sponsor.name
-                - else
-                  = link_to(sponsor.website_url, class: 'thumbnail') do
-                    = image_tag sponsor.logo, class: "img-responsive img-sponsor img-sponsor-#{sponsorship_level.position}", title: "#{sponsor.name}"
-                .caption
-                  -if sponsor.description.present?
-                    = markdown(sponsor.description)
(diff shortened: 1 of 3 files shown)


sonalkr132:
It is definitely better than what we have right now. However, what do you think of following:
screenshot from 2016-04-12 12-35-52

It doesn't require much lines of code, it requires that we add a dependency of [imagemagic](http://www.imagemagick.org/script/index.php) tho.
````
If uploaded image is smaller than a x*y px
    # take the image.jpg and places it on a white background
    convert image.jpg -background white -gravity center -extent 340x120 output.jpg
else
    # it resizes with aspect ratio and ensures that page is not slowed by huge logo images
    convert image.gif    -resize 340x120  resize_image.gif
end
```

Also, instead of description on hover it would be nicer if we do pop-up model (like 101.opensuse.org). If you are okay with it, I can take it up from here?

hennevogel:
Please do not expose the sponsorship levels to the users without explanation...

app/views/conference/_sponsors.html.haml
```diff
@@ -6,28 +6,17 @@
...
-                    = link_to(sponsor.website_url, title: "#{sponsor.name}") do
-                      = sponsor.name
-                - else
-                  = link_to(sponsor.website_url, class: 'thumbnail') do
-                    = image_tag sponsor.logo, class: "img-responsive img-sponsor img-sponsor-#{sponsorship_level.position}", title: "#{sponsor.name}"
-                .caption
-                  -if sponsor.description.present?
-                    = markdown(sponsor.description)
+            .col-md-4.col-sm-4.col-top
+              - if sponsor.logo.present?
```
hennevogel: And if not?

ChrisBr:
@sonalkr132 I like the idea!

@hennevogel @sonalkr132 I'm on vacation so don't expect any response from me for the next few weeks!

app/views/conference/_sponsors.html.haml
sonalkr132: validation is [added in this PR](https://github.com/openSUSE/osem/pull/951/files#diff-c50c4ba5add3c326d0f0646377bbbcd4R10) which ensures the presence of logo. (`- if ..` is not needed)

sonalkr132:
@ChrisBr I am taking this up.

sonalkr132:
@hennevogel 
> Please do not expose the sponsorship levels to the users without explanation...

It would be nice we had some explanation, however is it must to have? Checkout following websites, which just show the level of sponsorship:

* http://rubyconfindia.org/#sponsor
* http://rubyconf.org/sponsors
* http://railsconf.com/sponsors
* https://www.ng-conf.org/#/sponsors

hennevogel:
@sonalkr132 because it has no meaning without an explanation and you usually don't want to explain it to the public.

hennevogel:
Done differently in #967

PR #952, opened 2016-04-09
Remove program model from ahoy

sonalkr132:
We are not using this anywhere in our codebase.

From #728 
> ahoy/program.rb model authomatically created

@differentreality, can you please explain that?


differentreality:
Not really.  I think it is created automatically, so I either missed it on commit or on rebase. Good catch!

PR #953, opened 2016-04-10
Datatable cleanup

sonalkr132:
Things I have removed here were introduced in 152e8a3. In d21e19b97 we changed the implementation. ie we are using `jquery-datatables-rails` gem, leaving the previous code from 152e8a3 dead.

I have removed the model as well cause it is needed only if we are doing [server side processing](https://datatables.net/examples/server_side/simple.html) for records fetching, however we are not. Following is what needs to set for server side processing:
```Js
    $('#example').DataTable( {
        "processing": true,
        "serverSide": true,
        "ajax": "scripts/server_processing.php"
    } );
} );
```

When ours is:
```Js
    $('.datatable').DataTable({
      // ajax: ...,
      stateSave: true,
      autoWidth: false,
      pagingType: 'full_numbers',
      "lengthMenu": [[25, 50, 100, -1], [25, 50, 100, "All"]]
    });
  });
```

I have tested following pages (which are using dataTable) manually to make sure that things are working fine:

* Questions show: http://localhost:3000/admin/conference/conf_short_title/questions/1
* Tracks show: http://localhost:3000/admin/conference/conf_short_title/program/tracks/1
* Role users index: http://localhost:3000/admin/conference/conf_short_title/roles/organizer
* Tickets purchase show: http://localhost:3000/admin/conference/conf_short_title/tickets/1
* Registered users index: http://localhost:3000/admin/conference/conf_short_title/registrations
* Users index: http://localhost:3000/admin/users

cc: @ChrisBr


hennevogel:
https://www.youtube.com/watch?v=OM7zRfHG0no&feature=youtu.be&t=10

That's the way I like it :+1:

PR #959, opened 2016-04-14
Show registrations size in admin pages

davidsf:
Show the registrations number of events that need registration in admin/event/index and admin/event/show

### DIFF
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -25,6 +25,8 @@
         %th
           %b Pre-registration
         %th
+          %b Registrations
+        %th
           %b Highlight
         %th
           %b Type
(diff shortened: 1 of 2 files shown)


app/views/admin/events/index.html.haml
```diff
@@ -25,6 +25,8 @@
         %th
           %b Pre-registration
         %th
+          %b Registrations
+        %th
```
sonalkr132: @davidsf This change is making [following assertion](https://github.com/openSUSE/osem/blob/master/spec/views/admin/events/index.html.haml_spec.rb#L24) fail:
```Ruby
expect(rendered).to have_selector('table thead th:nth-of-type(6)', text: 'Highlight')
```
Please fix it.

hennevogel:
Please fix the tests and reopen

PR #962, opened 2016-04-15
Show registrations size in admin pages

### LINKED ISSUE #958: View number of registrations in admin / events
If the event needs registrations, the admin can't view how many people has already registered.

davidsf:
Show registrations size in admin pages. Corrected test. fix #958


hennevogel:
This is already implemented in #923

PR #963, opened 2016-04-17
Remove unused methods of user model and tests

sonalkr132:
* Remove unused functions from user model (check log message for details)
* Fix get_roles method of User model. 'Addie' should show as organizer of ng-conf :
screenshot from 2016-04-16 13-11-25

* Add tests for user model


hennevogel:
LGTM

Issue #965, opened 2016-04-18
Word limit on user biography


sonalkr132:
User can update there biography with word count exceeding the limit cause we don't have relevant validation in place:
![screenshot from 2016-04-18 11-22-16](https://cloud.githubusercontent.com/assets/7680662/14594928/a3dd6416-0558-11e6-8858-523261a8457b.png)

![screenshot from 2016-04-18 11-29-05](https://cloud.githubusercontent.com/assets/7680662/14594965/f96d652a-0558-11e6-9eb1-dc206fdc5090.png)

Ana06:
I'll do it! :smile:

differentreality:
Upon visiting edit_user_path, before typing anything in the bio box, it says you have used 0 words (even if there is text as biography).

Ana06:
This issue can be closed :wink:

hennevogel:
Yes boss :-)

PR #968, opened 2016-04-24
Word limit on user biography

Ana06:
Issue #965

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -54,14 +54,8 @@ class User < ActiveRecord::Base
             },
             presence: true
 
-  ##
-  # Checkes if the user attended the event
-  # This is used for events that require registration
-  # The user must have registered to attend the event
-  # Gets an event
-  # === Returns
-  # * +true+ if the user attended the event
-  # * +false+ if the user did not attend the event
+  validate :check_biography_length
+  
   def attended_event? event
     event_registration = event.events_registrations.find_by(registration: self.registrations)
 
@@ -190,6 +184,15 @@ class User < ActiveRecord::Base
   def setup_role
     if User.count == 1 && User.first.email == 'deleted@localhost.osem'
       self.is_admin = true
+    end
+  end
+
+  ##
+  # Check if biography has an allowed number of words. Used as validation.
+  #
+  def check_biography_length
+    if self.biography.present?
+      errors.add(:biography, 'is limited to 150 words.') unless self.biography.split.length <= 150
     end
   end
 end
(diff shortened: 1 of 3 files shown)


Ana06:
Sorry that I closed and reopen the pull request. I had read it wrong at the beginning and I checked that the biography hasn't got more than 150 characters instead of 150 words.

spec/models/user_spec.rb
```diff
@@ -66,6 +66,31 @@
...
+          lobortis metus id, tristique scelerisque nisl. Suspendisse potenti.
+          Suspendisse ac metus magna. Integer lobortis pharetra eros euismod
+          fringilla. Phasellus vitae orci vel magna laoreet mattis non eu neque.
+          Mauris ac dictum leo. Nullam dapibus convallis molestie. Integer
+          dignissim massa at odio feugiat tempus. Pellentesque ultrices rutrum
+          eros, a pellentesque lorem auctor in. Suspendisse sollicitudin dolor
+          vitae justo dignissim, a condimentum turpis molestie. Aenean
+          scelerisque, arcu eu congue mollis, nibh nulla finibus.
+        EOS
+        expect{create(:user, biography: long_text)}.to_not change { User.count }
```
sonalkr132: I think we can make it more apparant that it is a validation test if it was something like:
```Ruby
expect(build(:user, biography: long_text)).to_not be_valid
```
And please put it along with other [validation tests](https://github.com/Ana06/osem/blob/biography/spec/models/user_spec.rb#L14). It is a method indeed, but I think it is a validation method.

Ana06:
Done :wink:

differentreality:
We had removed that validation in a previous PR, not sure why.
https://github.com/openSUSE/osem/pull/963/files#diff-4676c008b11a5480d73d4a6de01e45b9L202

sonalkr132:
> We had removed that validation in a previous PR, not sure why.

Cause the method just existed there and it was not used (for validation).

app/models/user.rb
```diff
@@ -168,4 +170,13 @@ def setup_role
       self.is_admin = true
     end
   end
+
+  ##
+  # Check if biography has an allowed number of words. Used as validation.
+  #
+  def check_biography_length
+    if self.biography.present?
+      errors.add(:biography, 'is limited to 150 words.') unless self.biography.split.length <= 150
```
differentreality: if > 150  should also do.

differentreality:
@sonalkr132 right, I see what happened. At some point we fixed the js that wasn't working, but apparently the validation wasn't in place already for some reason.
@Ana06 I would introduce the same function name back (biography_limit).

Ana06:
I have changed the function name and I have solved the problem @differentreality pointed out in #965.

hennevogel:
LGTM! Thanks Ana

PR #970, opened 2016-04-27
Switch to carrierwave for for upload management

hennevogel:
It's way more flexible and brings simple support to use cloud storage like AWS or cloudinary. Additionally it fits perfect for our needs to host on PaaS services like heroku.


config/cloudinary.yml.example
```diff
@@ -0,0 +1,4 @@
+production:
+  cloud_name: "sample"
+  api_key: "874837483274837"
```
differentreality: Are those keys meant to be public?

hennevogel: They are examples (taken from the cloudinary docu)

differentreality:
Photos for conference, sponsor and logding look good! :+1: 

I don't think we are showing a photo for Venue anymore.. Not sure if that was intended in https://github.com/openSUSE/osem/pull/946

hennevogel:
Yeah I've seen this. That Venue doesn't degrade anymore is a bug (#974) :-)

hennevogel:
@differentreality so this can be merged?

differentreality:
Yeap, let's do this :+1:

nishanthvijayan:
@hennevogel  Now that they are not being used shouldn't we remove fields like logo_file_size, logo_content_type, logo_updated_at from the tables conferences sponsors and events?

PR #976, opened 2016-04-30
More options for events registrations

differentreality:
* Adds email notifications
  1. To user, when user registers to attend an event
  2. To user, when user is unregistered from an event automatically from the app (during scheduling)
  3. To event submitter, when max_attendees attributes is changed automatically (also informs submitter of user emails that were automatically unregistered)

* Adds switch-checkboxes
  1. For registering to events
  2. For submitter/organizer to unregister people manually (the switch box offers the option to undo your change and register the person back, as long as you haven't reloaded the page)

* Creates EventsRegistrations controller (that enables us to authorize actions)

**New workflow:**
To solve the issue of max_attendees being larger than the room size and no one knowing about it, when the event gets scheduled, if the room.size is smaller than max_attendees value:
  * max_attendees is automatically reduced to the room.size
  * exceeding registrations are deleted


differentreality:
screenshot
screenshot-1
screenshot-2
screenshot-3
screenshot-4

differentreality:
I am gonna close this, as it has gone unreviewed for a very long time and rebase is needed anyway. I will probably open a new PR, possibly with less changes (just moving events that require registration to their own page), until we can decide on the workflow we want to follow about max_attendees

Issue #956, opened 2016-04-13
Registration update by organizers


sonalkr132:
Registration update section in **admin area** has following section:
![screenshot from 2016-04-13 09-00-44](https://cloud.githubusercontent.com/assets/7680662/14482107/b5e0e486-0156-11e6-97f1-9b534a31f649.png)

It breaks any attempt to update the registration with following error message in flash:
> An error prohibited the Registration for : User email can't be blank. User password can't be blank. User username has already been taken. User username can't be blank.

Why do organizers need to update user related info(Name, Nickname and Affiliation) anyway?

differentreality:
I think having the option to change some basic user information should be given to people than cannot otherwise access user information (eg. Info Desk volunteers). Name and affiliation might be important to be fixed in case of typo or something for attendees like sponsors, keynote speakers.

PR #977, opened 2016-05-01
fix admin/registrations#update

differentreality:
#956

### DIFF
--- a/app/controllers/admin/registrations_controller.rb
+++ b/app/controllers/admin/registrations_controller.rb
@@ -14,8 +14,7 @@ module Admin
     def edit; end
 
     def update
-      @user.update_attributes(registration_params[:user_attributes])
-      params[:registration].delete :user_attributes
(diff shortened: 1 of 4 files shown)


app/controllers/admin/registrations_controller.rb
```diff
@@ -14,6 +14,9 @@ def index
     def edit; end
 
     def update
+      @user.update_attributes(registration_params[:user_attributes])
```
ChrisBr: What about strong_parameters?

app/controllers/admin/registrations_controller.rb
```diff
@@ -14,6 +14,9 @@ def index
...
+      params[:registration].delete :user_attributes
```
ChrisBr: With strong parameters you also don't need this line...

ChrisBr:
LGTM

PR #982, opened 2016-05-02
Demo Data

hennevogel:
Add a rake task to load the :full_conference factory into the database. We're going to use this for the demo instance and the review apps.


spec/factories/question.rb
```diff
@@ -2,11 +2,9 @@
...
 FactoryGirl.define do
   factory :question do
-    title 'blah'
+    title { Faker::Lorem.sentence }
     question_type
-    after(:build) do |question|
-      question.answers << build(:answer)
-      question.conferences << build(:conference)
-    end
+    conferences { [create(:conference)] }
```
ChrisBr: create_list ?

spec/factories/question.rb
```diff
@@ -2,11 +2,9 @@
...
+    answers { [create(:answer)] }
```
ChrisBr: same here

spec/factories/question.rb
hennevogel: What is the difference between create_list with 1 and create? :-)

ChrisBr:
LGTM

spec/factories/question.rb
ChrisBr: I guess you don't have to wrap it into the array anymore?

hennevogel: True. But I find this not very obvious...

ChrisBr: Ok, then leave it as it is

PR #986, opened 2016-05-02
Fix proposal#update so that it shows errors when trying to create user

differentreality:
Creating a propsal without a user, when the user details are not correct, raises an error and does not render 'new' action with @user.errors

### DIFF
--- a/app/controllers/proposal_controller.rb
+++ b/app/controllers/proposal_controller.rb
@@ -29,6 +29,8 @@ class ProposalController < ApplicationController
 
   def create
     @url = conference_program_proposal_index_path(@conference.short_title)
+    @event = Event.new(event_params)
+    @event.program = @program
 
     # We allow proposal submission and sign up on same page.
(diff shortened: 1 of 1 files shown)


hennevogel:
I don't get this change. Can you please explain why moving event instantiation up fixes this?

differentreality:
Creating a user fails (due to user validations):
https://github.com/openSUSE/osem/blob/master/app/controllers/proposal_controller.rb#L38
So we render :new action (that is the proposal#new action):
https://github.com/openSUSE/osem/blob/master/app/controllers/proposal_controller.rb#L42

At this point @event is not set for the form:
https://github.com/openSUSE/osem/blob/master/app/views/proposal/new.html.haml#L23

hennevogel:
Please rebase and check @ChrisBr's changes

ChrisBr:
I guess we can close this one as we don't need to manually initialize the @event object anymore!

Issue #955, opened 2016-04-12
Move registration_limit check to ability file


sonalkr132:
[Source of suggestion](https://github.com/openSUSE/osem/pull/795#discussion_r55738013)

We are checking for registration limit of conference in `new` action of [ConferenceRegistrationsController](https://github.com/openSUSE/osem/blob/master/app%2Fcontrollers%2Fconference_registrations_controller.rb#L26-L29):

```Ruby
def new
  # code ommited
  if @conference.registration_limit_exceeded?
      redirect_to root_path, error: "Sorry, registration limit exceeded for #{@conference.title}"
      return
    end
end
```
We should be handling this with authorization of the `new` action.

Ana06:
I'll do it :wink:

Ana06:
This issue can be closed :cactus:

PR #989, opened 2016-05-04
registration_limit check moved to ability file

Ana06:
Issue #955

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -6,7 +6,7 @@ class ConferenceRegistrationsController < ApplicationController
 
   def new
     @registration = Registration.new(conference_id: @conference.id)
-    authorize! :new, @registration
+    authorize! :new, @registration, message: "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
 
     # Redirect to registration edit when user is already registered
     if @conference.user_registered?(current_user)
@@ -21,11 +21,6 @@ class ConferenceRegistrationsController < ApplicationController
     # avoid openid sign_in to redirect to register/new when the sign_in user had already a registration
     if current_user && @conference.user_registered?(current_user)
       redirect_to edit_conference_conference_registrations_path(@conference.short_title)
-    end
-
-    if @conference.registration_limit_exceeded?
-      redirect_to root_path, error: "Sorry, registration limit exceeded for #{@conference.title}"
-      return
     end
 
     # @user variable needs to be set so that _sign_up_form_embedded works properly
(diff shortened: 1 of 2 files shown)


hennevogel:
How is the user ever going to know that this is denied because the registration limit is exceeded?

Ana06:
In fact, when it is denied because the registration is not opened it is also said: `You are not authorized to access this page.` and it is also quite confusing.

I have added an error for both cases. I think that we don't need to say what of the two possibilities is causing the error as the user is redirected to root where information about registration is shown (it shows `Sorry, no places left.` or hides the register button).

I have also defined a different `can` for create, as there is a validation for that in `registration.rb`.

Also, when the user is not signed in, why is it checked that the registration is a new record?

And what happens with users with roles? I have only checked the registration limit for signed_in and not_signed_in users. Am not sure what we want to do with other users.

app/models/ability.rb
```diff
@@ -50,7 +50,12 @@ def not_signed_in
         registration.new_record?
       end
 
-      can [:new, :create], Registration do |registration|
+      can [:new], Registration do |registration|
+        conf = registration.conference
+        conf.registration_open? && registration.new_record? && !conf.registration_limit_exceeded?
+      end
+
+      can [:create], Registration do |registration|
```
differentreality: We do not want to allow `create` action either, when the registration_limit is reached.

Ana06: But there is a validation in Registration that checks it. It won't have any effect if I add it in the ability file.

hennevogel:
Please rebase and check @ChrisBr's changes

Ana06:
I've rebased, but there were not merge conflicts and I am not sure of what @ChrisBr's changes I should look at.

app/models/ability.rb
differentreality: ability.rb should still fully reflect what is happening, ie if the user has access to a specific action, or not.

And please change variable `conf` into `conference`.

app/controllers/conference_registrations_controller.rb
```diff
@@ -6,7 +6,7 @@ class ConferenceRegistrationsController < ApplicationController
 
   def new
     @registration = Registration.new(conference_id: @conference.id)
-    authorize! :new, @registration
+    authorize! :new, @registration, message: "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not opened."
```
differentreality: The registration is not **open** (instead of opened) sounds better in this case.

hennevogel:
@Ana06 what is the status of this PR?

Ana06:
@hennevogel you told me to check @ChrisBr's changes but I didn't find those changes. I just rebased and I don't know if I should do anything else.

differentreality:
:+1:

Issue #991, opened 2016-05-09
[production][conference_registrations#new] NoMethodError: undefined method `>' for nil:NilClass


hennevogel:
conference_registrations#new threw this 1 times
<pre>
NoMethodError: undefined method `&gt;&#39; for nil:NilClass
</pre>
[https://events.opensuse.org/conference/oSC16/register/new](https://events.opensuse.org/conference/oSC16/register/new)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca99dc71fa00b1000002/problems/57308ae9f12576000c000003)

hennevogel:
This happens because e.max_attendees is nil by default. hotfixed in production by removing this condition.

differentreality:
It is cause because events already had require_registration set, before we introduced max_attendees. Now a user cannot set require_registration attribute without putting a value for max_attendees too.

I will send a task to fix this for the existing events with require_registration.

hennevogel:
But whats wrong with an Event that requires registration but does not have a max_attendees limit set?

differentreality:
Since we dropped 0 value for max_attendees, I haven't provisioned for this scenario, where registrations are limitless. 
Usually registration is required because there is a limit to the amount of people that can be accommodated.

ChrisBr:
@differentreality is this still valid with #994 merged?

differentreality:
@ChrisBr should be closed via #994

PR #993, opened 2016-05-09
Task for max_attendees to have a value when require_registration is set

differentreality:
Fixes https://github.com/openSUSE/osem/issues/991
`bundle exec rake events_registrations:set_max_attendees`

### DIFF
--- a/lib/tasks/events_registrations.rake
+++ b/lib/tasks/events_registrations.rake
@@ -1,4 +1,22 @@
 namespace :events_registrations do
+  desc 'Sets max_attendees to 1 or number of registrations, if require_registration is set'
+  task set_max_attendees: :environment do
+    @events = Event.all.where(require_registration: true, max_attendees: nil)
+    puts "Fixing max_attendees attribute for #{@events.length} events."
+    puts "The IDs of those events are: #{@events.pluck(:id)}"
+
+    @events.each do |event|
+      if event.registrations.any?
+        event.max_attendees = event.registrations.count
+      else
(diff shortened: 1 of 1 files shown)


differentreality:
Note: Specially for oSC16, the submitter of the event must go and edit the proposal to add the actual max_attendees value.

lib/tasks/events_registrations.rake
```diff
@@ -1,4 +1,22 @@
...
+  task set_max_attendees: :environment do
+    @events = Event.all.where(require_registration: true, max_attendees: nil)
+    puts "Fixing max_attendees attribute for #{@events.length} events."
+    puts "The IDs of those events are: #{@events.pluck(:id)}"
+
+    @events.each do |event|
+      if event.registrations.any?
+        event.max_attendees = event.registrations.count
+      else
+        event.max_attendees = 1
```
hennevogel: Erm this is not really what we want right?

differentreality: Not really, but I am not sure what the alternative is.
We have no way of knowing the value of max_attendees. OK it is probably not going to be 1, but we don't know if it is 10 or 30 or...

hennevogel: Then I don't think we should rely on it

hennevogel:
:-1:  I think instead of adding a rather random number to `max_attendees` we should get rid of the `max_attendees_and_require_registration` validation because it really should be possible to require registration for the Event without having a limit

PR #994, opened 2016-05-10
allow max_attendees to be nil

differentreality:
#991

### DIFF
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -13,8 +13,7 @@ class Program < ActiveRecord::Base
     end
 
     def with_registration_open
-      where(require_registration: true, state: :confirmed).
-      map { |e| e if e.max_attendees > e.registrations.count }.compact
+      select { |e| e if e.registration_possible? }
     end
 
     # All confirmed events of the conference with attribute require_registration
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -615,6 +615,3 @@ DEPENDENCIES
   web-console (~> 2.0)
   webmock
   whenever
-
-BUNDLED WITH
-   1.11.2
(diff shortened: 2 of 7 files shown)


app/models/program.rb
```diff
@@ -14,7 +14,7 @@ def require_registration
 
     def with_registration_open
       where(require_registration: true, state: :confirmed).
-      map { |e| e if e.max_attendees > e.registrations.count }.compact
+      map { |e| e if e.max_attendees.nil? || (e.max_attendees && e.max_attendees > e.registrations.count) }.compact
```
hennevogel: Can you make this into a method for Event or use `Event.registration_possible?`

spec/controllers/admin/events_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+
+describe Admin::EventsController, type: :controller do
+  let(:conference) { create(:full_conference) }
+  let(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
+  let(:organizer) { create(:user, role_ids: organizer_role.id, last_sign_in_at: Time.now - 1.day) }
+  let(:event) { create(:event, program: conference.program) }
+
+  shared_examples 'scheduling' do
+    scenario 'event has a room' do
```
hennevogel: What is this testing?

spec/controllers/admin/events_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      room = create(:room, venue: event.program.conference.venue)
+      event.room = room
+      event.save!
+      expect(event.room).to eq room
+    end
+
+    scenario 'event has a start_time' do
```
hennevogel: Same here? What is this testing? That ActiveRecord attribute assignment works?

hennevogel:
Awesome

PR #1000, opened 2016-05-15
Fix ability model spec

nishanthvijayan:
Cancancan does not support testing multiple permissions in one be_able_to clause. [See issue ](https://github.com/CanCanCommunity/cancancan/issues/300)  
Replaces all occurrences of such examples with separate examples for each permission.


ChrisBr:
@nishanthvijayan cool :+1: Nice catch!

Merging ...

PR #1004, opened 2016-05-19
Make proposal show view more robust

### LINKED ISSUE #718: [production][proposal#show] NoMethodError: undefined method `color' for nil:NilClass
proposal#show threw this 2 times
<pre>
NoMethodError: undefined method `color&#39; for nil:NilClass
</pre>
[https://events.opensuse.org/conference/osc13/proposal/124](https://events.opensuse.org/conference/osc13/proposal/124)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca99dc71fa00b1000002/problems/5620db69764b230006000002)

ChrisBr:
Partially fixes #718. Some objects have an foreign_key set on track, difficulty level or room but the
referenced object does not exists. This is caused by introducing dependent: nullify after deleting
some of the objects. This commit will make the show page more robust, however, we need to
fix the corrupt objects in an additional commit.

### DIFF
--- a/app/views/proposal/show.html.haml
+++ b/app/views/proposal/show.html.haml
@@ -58,7 +58,7 @@
           .col-md-12
             %dt Room:
             %dd
-              - if @event.room_id
+              - if @event.room
                 = @event.room.name
           .col-md-12
             %dt Conference:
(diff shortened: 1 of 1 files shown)


differentreality:
:+1:

PR #1005, opened 2016-05-19
Hide Submit proposal if iChain is enabled and user is not logged in

ChrisBr:
because it's not possible to create a user while submitting a proposal when iChain is enabled. User would get not authorized error message.


hennevogel:
:+1:

Issue #966, opened 2016-04-18
Sponsors without sponsor level


sonalkr132:
Sponsor level is mandatory field of sponsors model. If a user tries to create sponsor without first creating a sponsor level, their validation is always going to fail(cause they have nothing to choose for sponsor level). We should show sponsor's page only if sponsor levels are defined (similar to showing `rooms` only if `venue` is defined).
![screenshot from 2016-04-18 12-32-05](https://cloud.githubusercontent.com/assets/7680662/14596109/c9852a88-0561-11e6-885a-c68ac51dfa9b.png)

PR #1006, opened 2016-05-19
Display sponsor pages only if sponsorship levels are defined

nishanthvijayan:
Solves #966

### DIFF
--- a/app/views/layouts/_admin_sidebar.html.haml
+++ b/app/views/layouts/_admin_sidebar.html.haml
@@ -107,7 +107,7 @@
         - if can? :update, @conference.sponsorship_levels.build
           %li{:class=> "#{active_nav_li(admin_conference_sponsorship_levels_path(@conference.short_title))}" }
             = link_to 'Sponsorship Levels', admin_conference_sponsorship_levels_path(@conference.short_title)
-        - if can? :update, @conference.sponsors.build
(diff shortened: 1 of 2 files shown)


app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -107,7 +107,7 @@
         - if can? :update, @conference.sponsorship_levels.build
           %li{:class=> "#{active_nav_li(admin_conference_sponsorship_levels_path(@conference.short_title))}" }
             = link_to 'Sponsorship Levels', admin_conference_sponsorship_levels_path(@conference.short_title)
-        - if can? :update, @conference.sponsors.build
+        - if !@conference.sponsorship_levels.empty?  && @conference.sponsorship_levels.first.persisted? && (can? :update, @conference.sponsors.build)
```
hennevogel: Why `.persisted`?

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -107,7 +107,7 @@
         - if can? :update, @conference.sponsorship_levels.build
```
nishanthvijayan: This builds a sponsorship_level hence the `.persisted?` check.A better alternative is to check `@conference.sponsorship_levels.count > 0` instead of `!@conference.sponsorship_levels.empty?  && @conference.sponsorship_levels.first.persisted?`. Should I change it count?

hennevogel: No that's just semantics...

PR #1009, opened 2016-05-23
Fix program routing outside of Admin

hennevogel:
We don't have a ProgramsController so we don't need routes for it.

https://errbit-opensuse.herokuapp.com/apps/5620ca99dc71fa00b1000002/problems/57424461216ae10009000000


ChrisBr:
LGTM

PR #1015, opened 2016-05-26
Wrong test: proposal cannot be rated

Ana06:
This test was not doing what it was expected to do, as it was a copy of _returns true if proposals can be rated_


differentreality:
:+1:

PR #1016, opened 2016-05-29
Allow updating currency of ticket if its the only ticket of that conference

nishanthvijayan:
Currently if a new ticket is added to a conference its currency should be the same as that of the existing type of tickets.But the validation that checks this also prevents updating a ticket's currency even if its the only type  of ticket.

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -59,6 +59,7 @@ class Ticket < ActiveRecord::Base
   private
 
   def tickets_of_conference_have_same_currency
+    return if self.persisted? && Ticket.where(conference_id: conference_id).count == 1
     unless Ticket.where(conference_id: conference_id).all?{|t| t.price_currency == self.price_currency }
       errors.add(:price_currency, 'is different from the existing tickets of this conference.')
     end
--- a/spec/models/ticket_spec.rb
+++ b/spec/models/ticket_spec.rb
@@ -45,6 +45,11 @@ describe Ticket do
     end
   end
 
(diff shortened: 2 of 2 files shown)


ChrisBr:
@nishanthvijayan please discuss this with @rishabhs95 and his mentors as they plan to work on the ticketing system ...

differentreality:
@ChrisBr do you think we really need to have the same currency for all tickets? It might be a bit unorthodox to have one ticket in euro and one in USD, but why not? Then, for example, people from different parts of the world can pay in the currency they prefer.

@lagartoflojo is this something that would interfere with the planning you ve made so far?

In any case, updating the currency of 1 single ticket should be possible. @lagartoflojo can we merge this?

rishabhs95:
I think all the tickets should have the same currency, which, the conference's organiser prefers to receive payments in. Also, if user picks to pay for multiple tickets, both in Euro and in USD, then it could lead to a conflict while payment.

Still, I think we should be able to update currency of a single ticket.

It could be a problem only in a case, when the organiser updates the currency of a ticket(USD -> EUR) which has been already purchased by attendees (as already purchased tickets would have transferred amount in USD while after update our database will reflect the tickets in EUR), which doesn't happen I think(organisers don't change currencies in between).

If that's right, we should be able to merge it.

hennevogel:
People who committed to pay €10 have not committed to pay $11.26 (10€)

If you want to you can make it possible to change the ticket currency if there are no purchases...

PR #1018, opened 2016-05-30
Add payment mode selection for organisers

rishabhs95:
add capability for organisers to select payment modes for their tickets as Online/Offline.

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -9,7 +9,7 @@ class Ticket < ActiveRecord::Base
   # If we would allow different currencies per conference we also have to handle convertions between currencies!
   validate :tickets_of_conference_have_same_currency
 
-  validates :price_cents, :price_currency, :title, presence: true
+  validates :price_cents, :price_currency, :payment_mode, :title, presence: true
 
   validates_numericality_of :price_cents, greater_than: 0
 
--- a/db/migrate/20160526142456_add_payment_mode_to_tickets.rb
+++ b/db/migrate/20160526142456_add_payment_mode_to_tickets.rb
@@ -1,0 +1,5 @@
+class AddPaymentModeToTickets < ActiveRecord::Migration
+  def change
+    add_column :tickets, :payment_mode, :string
+  end
+end
--- a/spec/models/ticket_spec.rb
+++ b/spec/models/ticket_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe Ticket do
   let(:conference) { create(:conference) }
-  let(:ticket) { create(:ticket, price: 50, price_currency: 'USD', conference: conference) }
+  let(:ticket) { create(:ticket, price: 50, price_currency: 'USD', payment_mode: 'Online', conference: conference) }
   let(:user) { create(:user) }
 
   describe 'validation' do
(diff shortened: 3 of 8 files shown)


hennevogel:
You really think people want to mix online/offline payment methods?

rishabhs95:
i thought this will provide freedom to organiser to select which payment mode they will prefer for each ticket.

they can choose whether they want to receive payments for it online/offline.

rishabhs95:
please suggest if it should be flexible with payment mode associated to each tickets or on a larger level(like conference)?

db/migrate/20160526142456_add_payment_mode_to_tickets.rb
```diff
@@ -0,0 +1,5 @@
+class AddPaymentModeToTickets < ActiveRecord::Migration
+  def change
+    add_column :tickets, :payment_mode, :string
```
lagartoflojo: By adding the column like this, all old tickets will have a NULL `payment_method`. Is that valid? Should we do something about that?

app/models/ticket.rb
```diff
@@ -9,7 +9,7 @@ class Ticket < ActiveRecord::Base
   # If we would allow different currencies per conference we also have to handle convertions between currencies!
   validate :tickets_of_conference_have_same_currency
 
-  validates :price_cents, :price_currency, :title, presence: true
+  validates :price_cents, :price_currency, :payment_mode, :title, presence: true
```
ChrisBr: Please also add an inclusion validation as I see you only allow the values Online/Offline in the view.
http://guides.rubyonrails.org/active_record_validations.html#inclusion

spec/models/ticket_spec.rb
```diff
@@ -22,6 +22,10 @@
       should validate_presence_of(:price_currency)
     end
 
+    it 'is not valid without a payment_mode' do
```
ChrisBr: Also add a test for the inclusion validation

db/migrate/20160526142456_add_payment_mode_to_tickets.rb
rishabhs95: yes... i will edit it keeping "Offline" as default payment_mode as the existing tickets required offline payments

spec/models/ticket_spec.rb
rishabhs95: ok :)

rishabhs95:
Just a clarification for the use case for associating payment_mode with tickets: 

According to my understanding, a conference is an overall event and various talks and other events are organised inside it. These events can be of two types: paid and free. The payment method for the paid events can be further classified into online and offline payment.

One example of event which requires online payment would be provision of accommodation or some other special paid talks. Online payment signifies commitment on attendees behalf and the organisers can focus on arrangements instead of last minute 'no-shows'.

However, there are other events like tours or parties which attendees are encouraged to decide on the spot. The organisers require a rough estimate of number of attendees. Offline payments is suitable in such scenario.

hennevogel:
@rishabhs95 tickets are for a conference as a whole. Not for individual happenings during it

rishabhs95:
Okay @hennevogel, I will make the changes accordingly

hennevogel:
Please send a new pull request then :-)

PR #1019, opened 2016-05-30
Changelog/Revision History page

nishanthvijayan:
Sets up a Revision History page.
 - The page displays changes to conferences and associated resources.
 - Only organizers can view and revert the revision history of their conferences
 -  Admins can view and revert the revision history for all conferences
 - When describing a change, attributes whose values changed from one type of blank value to another (eg: nil to '  ')   will be ignored
 - The previous and current values of each modified attribute can be viewed by clicking the 'View Changes' button of the corresponding change
 - Revert feature:
  - Reverting the creation of an object deletes it
  - Reverting the deletion of an object re-creates it
  - Reverting the creation of a conference and user is not permitted
  - If a change include changes to multiple attributes, each of these attributes can be reverted individually. But this is allowed only for changes where the object was updated and not created/deleted and if the object is alive (not deleted).
 - Few models have not been tracked as they are expected to be reworked soon (TicketPurchase, Question related stuff)
 - This feature adds a new field named conference_id to the versions table.This is used to fetch versions belonging to a conference/ models belonging to a conference.A rake task has been provided to fill this new column for pre-existing versions (Conference and Event model were already being tracked)
The take task can be invoked by `bundle exec rake data:set_conference_in_versions`

### DIFF
--- a/app/assets/javascripts/osem-datatables.js
+++ b/app/assets/javascripts/osem-datatables.js
@@ -7,6 +7,11 @@ $(function () {
       pagingType: 'full_numbers',
       "lengthMenu": [[25, 50, 100, -1], [25, 50, 100, "All"]]
     });
+
+    $('#versionstable').DataTable({
+      pagingType: 'full_numbers',
+      order: [[ 0, 'desc' ]]
+    });
   });
 });
 
--- a/app/assets/javascripts/osem-revisionhistory.js
+++ b/app/assets/javascripts/osem-revisionhistory.js
@@ -1,0 +1,10 @@
+$(document).ready(function() {
+    $('.show-changeset').click(function(){
+        if ($(this).text() == 'View Changes'){
+            $(this).text('Hide Changes');
+        }else {
+            $(this).text('View Changes');
+        }
+        $('#changeset-' + this.id).toggle();
+    });
+});
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,0 +1,58 @@
+module Admin
+  class VersionsController < Admin::BaseController
+    skip_authorization_check
+
+    def index
+      authorize! :index, PaperTrail::Version.new(item_type: 'User')
+      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
+      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+    end
+
+    def revert_attribute
+      @version = PaperTrail::Version.find(params[:id])
+      authorize! :revert_attribute, @version
+
+      if params[:attribute] && @version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.keys.include?(params[:attribute])
+        if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
+          flash[:error] = 'The item is already in the state that you are trying to revert it back to'
+
+        else
+          @version.item[params[:attribute]] = @version.changeset[params[:attribute]][0]
+          if @version.item.save
+            flash[:notice] = 'The selected change was successfully reverted'
+          else
+            flash[:error] = "An error prohibited this change from being reverted: #{@version.item.errors.full_messages.join('. ')}."
+          end
+        end
+
+      else
+        flash[:error] = 'Revert failed. Attribute missing or invalid'
+      end
+
+      redirect_to admin_revision_history_path
+    end
+
+    def revert_object
+      @version = PaperTrail::Version.find(params[:id])
+      authorize! :revert_object, @version
+
+      if @version.event != 'create'
+        if @version.reify.save
+          flash[:notice] = 'The selected change was successfully reverted'
+        else
+          flash[:error] = "An error prohibited this change from being reverted: #{@version.reify.errors.full_messages.join('. ')}."
+        end
+
+      elsif @version.event == 'create' && @version.item
+        # if @version represets a create event and is not currently deleted
+        @version.item.destroy
+        flash[:notice] = 'The selected change was successfully reverted'
+
+      else
+        flash[:error] = 'The item is already in the state that you are trying to revert it back to'
+      end
+
+      redirect_to admin_revision_history_path
+    end
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -370,4 +370,125 @@ module ApplicationHelper
   def selected_scheduled?(schedule)
     (schedule == @selected_schedule) ? 'Yes' : 'No'
   end
+
+  # Recieves a PaperTrail::Version object
+  # Outputs the list of attributes that were changed in the version (ignoring changes from one blank value to another)
+  # Eg: If version.changeset = '{"title"=>[nil, "Premium"], "description"=>[nil, "Premium = Super cool"], "conference_id"=>[nil, 3]}'
+  # Output will be 'title, description and conference'
+  def updated_attributes(version)
+    version.changeset.
+      reject{ |_, values| values[0].blank? && values[1].blank? }.
+      keys.map{ |key| key.gsub('_id', '').tr('_', ' ')}.join(', ').
+      reverse.sub(',', ' dna ').reverse
+  end
+
+  def change_creator_link(user_id)
+    user = User.find_by(id: user_id)
+    if user
+      link_to user.name, admin_user_path(id: user_id)
+    else
+      'Someone (probably via the console)'
+    end
+  end
+
+  # Recieves a PaperTrail::Version object
+  # Returns object in its current state if its alive
+  # Returns object as it was before version's change(unless its a create event's version)
+  # Else Returns object as it was after version's change
+  def get_version_object(version)
+    version.item || version.reify || version.next.reify
+  end
+
+  def event_change_description(version)
+    case
+    when version.event == 'create' then 'submitted new'
+
+    when version.changeset['state']
+      case version.changeset['state'][1]
+      when 'unconfirmed' then 'accepted'
+      when 'withdrawn' then 'withdrew'
+      when 'canceled', 'rejected', 'confirmed' then version.changeset['state'][1]
+      when 'new' then 'resubmitted'
+      end
+
+    when version.changeset['start_time'] && version.changeset['start_time'][0].nil?
+      'scheduled'
+
+    when version.changeset['start_time'] && version.changeset['start_time'][1].nil?
+      'unscheduled'
+
+    else
+      "updated #{updated_attributes(version)} of"
+    end
+  end
+
+  def users_role_change_description(version)
+    version.event == 'create' ? 'added' : 'removed'
+  end
+
+  def subscription_change_description(version)
+    user = get_version_object(version).user
+    user_name = user.name unless user.id.to_s == version.whodunnit
+    version.event == 'create' ? "subscribed #{user_name} to" : "unsubscribed #{user_name} from"
+  end
+
+  def registration_change_description(version)
+    if version.item_type == 'Registration'
+      user = get_version_object(version).user
+    else
+      registration_id = get_version_object(version).registration_id
+      registration_last_version = PaperTrail::Version.where(item_type: 'Registration', item_id: registration_id).last
+      user = get_version_object(registration_last_version).user
+    end
+
+    if user.id.to_s == version.whodunnit
+      case version.event
+      when 'create' then 'registered to'
+      when 'update' then "updated #{updated_attributes(version)} of the registration for"
+      when 'destroy' then 'unregistered  from'
+      end
+    else
+      case version.event
+      when 'create' then "registered #{user.name} to"
+      when 'update' then "updated #{updated_attributes(version)} of  #{user.name}'s registration for"
+      when 'destroy' then "unregistered #{user.name} from"
+      end
+    end
+  end
+
+  def comment_change_description(version)
+    user = get_version_object(version).user
+    if version.event == 'create'
+      version.previous.nil? ? 'commented on' : "re-added #{user.name}'s comment on"
+    else
+      "deleted #{user.name}'s comment on"
+    end
+  end
+
+  def vote_change_description(version)
+    user = get_version_object(version).user
+    if version.event == 'create'
+      version.previous.nil? ? 'voted on' : "re-added #{user.name}'s vote on"
+    elsif version.event == 'update'
+      "updated #{user.name}'s vote on"
+    else
+      "deleted #{user.name}'s vote on"
+    end
+  end
+
+  def user_change_description(version)
+    if version.event == 'create'
+     change_creator_link(version.item_id) + ' signed up'
+    elsif version.event == 'update'
+      if version.changeset.keys.include?('reset_password_sent_at')
+        'Someone requested password reset of'
+      elsif version.changeset.keys.include?('confirmed_at') && version.changeset['confirmed_at'][0].nil?
+        (version.whodunnit.nil? ? change_creator_link(version.item_id) : change_creator_link(version.whodunnit)) + ' confirmed account of'
+      elsif version.changeset.keys.include?('confirmed_at') && version.changeset['confirmed_at'][1].nil?
+        change_creator_link(version.whodunnit) + ' unconfirmed account of'
+      else
+        change_creator_link(version.whodunnit) + " updated #{updated_attributes(version)} of"
+      end
+    end
+  end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -114,6 +114,14 @@ class Ability
     # for admins
     can :manage, :all if user.is_admin
 
+    cannot :revert_object, PaperTrail::Version do |version|
+      (version.event == 'create' && %w(Conference User Event).include?(version.item_type))
+    end
+
+    cannot :revert_attribute, PaperTrail::Version do |version|
+      version.event != 'update' || version.item.nil?
+    end
+
     cannot :destroy, Program
     # Do not delete venue, when there are rooms being used
     cannot :destroy, Venue do |venue|
@@ -167,6 +175,10 @@ class Ability
     can [:index, :show], Role
     can [:edit, :update, :toggle_user], Role do |role|
       role.resource_type == 'Conference' && (conf_ids_for_organizer.include? role.resource_id)
+    end
+
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
+      version.item_type == 'User' || (conf_ids_for_organizer.include? version.conference_id)
     end
   end
 
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,6 +1,6 @@
 class Event < ActiveRecord::Base
   include ActiveRecord::Transitions
-  has_paper_trail
+  has_paper_trail on: [:create, :update], ignore: [:updated_at, :guid, :week], meta: { conference_id: :conference_id }
 
   acts_as_commentable
 
@@ -280,12 +280,18 @@ class Event < ActiveRecord::Base
 
   def set_week
     self.week = created_at.strftime('%W')
-    save!
+    self.without_versioning do
+      self.save!
+    end
   end
 
   def before_end_of_conference
     errors.
         add(:created_at, "can't be after the conference end date!") if program.conference && program.conference.end_date &&
         (Date.today > program.conference.end_date)
+  end
+
+  def conference_id
+    program.conference_id
   end
 end
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -1,6 +1,8 @@
 # cannot delete program if there are events submitted
 
 class Program < ActiveRecord::Base
+  has_paper_trail on: [:update], ignore: [:updated_at], meta: { conference_id: :conference_id }
+
   belongs_to :conference
 
   has_one :cfp, dependent: :destroy
@@ -53,8 +55,8 @@ class Program < ActiveRecord::Base
   validate :voting_start_date_before_end_date
   validate :voting_dates_exist
 
-  before_create :create_event_types
-  before_create :create_difficulty_levels
+  after_create :create_event_types
+  after_create :create_difficulty_levels
   validate :check_languages_format
 
   # Returns all event_schedules for the selected schedule ordered by start_time
@@ -157,12 +159,12 @@ class Program < ActiveRecord::Base
   # Creates default EventTypes for this Conference. Used as before_create.
   #
   def create_event_types
-    event_types << EventType.create(title: 'Talk', length: 30, color: '#FF0000', description: 'Presentation in lecture format',
-                                    minimum_abstract_length: 0,
-                                    maximum_abstract_length: 500)
-    event_types << EventType.create(title: 'Workshop', length: 60, color: '#0000FF', description: 'Interactive hands-on practice',
-                                    minimum_abstract_length: 0,
-                                    maximum_abstract_length: 500)
+    EventType.create(title: 'Talk', length: 30, color: '#FF0000', description: 'Presentation in lecture format',
+                     minimum_abstract_length: 0,
+                     maximum_abstract_length: 500, program_id: self.id)
+    EventType.create(title: 'Workshop', length: 60, color: '#0000FF', description: 'Interactive hands-on practice',
+                     minimum_abstract_length: 0,
+                     maximum_abstract_length: 500, program_id: self.id)
     true
   end
 
@@ -170,15 +172,15 @@ class Program < ActiveRecord::Base
   # Creates default DifficultyLevels for this Conference. Used as before_create.
   #
   def create_difficulty_levels
-    difficulty_levels << DifficultyLevel.create(title: 'Easy',
-                                                description: 'Events are understandable for everyone without knowledge of the topic.',
-                                                color: '#70EF69')
-    difficulty_levels << DifficultyLevel.create(title: 'Medium',
-                                                description: 'Events require a basic understanding of the topic.',
-                                                color: '#EEEF69')
-    difficulty_levels << DifficultyLevel.create(title: 'Hard',
-                                                description: 'Events require expert knowledge of the topic.',
-                                                color: '#EF6E69')
+    DifficultyLevel.create(title: 'Easy',
+                           description: 'Events are understandable for everyone without knowledge of the topic.',
+                           color: '#70EF69', program_id: self.id)
+    DifficultyLevel.create(title: 'Medium',
+                           description: 'Events require a basic understanding of the topic.',
+                           color: '#EEEF69', program_id: self.id)
+    DifficultyLevel.create(title: 'Hard',
+                           description: 'Events require expert knowledge of the topic.',
+                           color: '#EF6E69', program_id: self.id)
     true
   end
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -1,6 +1,10 @@
 class Role < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
-  has_and_belongs_to_many :users, join_table: :users_roles
+  has_many :users_roles
+  has_many :users, through: :users_roles
+
+  has_paper_trail on: [:update], only: [:name, :description], meta: { conference_id: :resource_id }
+
   before_destroy :cancel
   scopify
 
--- a/app/views/admin/versions/_object_changes.html.haml
+++ b/app/views/admin/versions/_object_changes.html.haml
@@ -1,0 +1,28 @@
+.col-md-10.col-md-offset-1.changeset{id: "changeset-#{version.id}"}
+  %br
+  %br
+  %table.table.table-bordered.table-hover
+    %thead
+      %tr
+        %th Updated Attribute
+        - if version.event != 'create'
+          %th Previous Value
+        - if version.event != 'destroy'
+          %th New Value
+        - if can? :revert_attribute, version
+          %th Action
+    %tbody
+      - if version.event!= 'destroy'
+        - version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.each do |attribute, values|
+          %tr
+            %td= attribute
+            - if version.event != 'create'
+              %td= values[0].blank? ?  '-' : values[0]
+            %td= values[1].blank? ?  '-' : values[1]
+            - if can? :revert_attribute, version
+              %td= link_to 'Revert', admin_revision_history_revert_attribute_path(id: version.id, attribute: attribute), class: 'btn btn-sm btn-primary', data: { confirm: "Are you sure you want to revert #{attribute}?" }
+      - else
+        - version.reify.attributes.each do |attribute, value|
+          %tr
+            %td= attribute
+            %td= value.blank? ?  '-' : value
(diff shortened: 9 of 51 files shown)


nishanthvijayan:
screen

differentreality:
Can you upload a screenshot with pagination please?

app/models/program.rb
```diff
@@ -49,7 +51,7 @@ def confirmed
 #   validates :conference_id, presence: true, uniqueness: true
   validates :rating, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 10 }
 
-  before_create :create_event_types
+  after_create :create_event_types
```
differentreality: This change is not directly relevant to this PR. Could you extract it (along with the changes in factories) into a different PR?

nishanthvijayan: Because of versioning, on trying to create any factory that needs program, it'll throw up an error saying the factory is invalid.
The reason is that on creating a program this before_create creates  2 event_types without a program id.This triggers paper_trail to record the create event of these EventTypes  resulting in a call to the EventType method conference_id which throws up `undefined method conference_id for nil:NilClass (NoMethodError)` as these 2 event types do not have a program at that point.
I included these fixes as the errors are  direct consequences of the changes introduced in this PR.

nishanthvijayan:
pagination

app/models/program.rb
nishanthvijayan: @differentreality On creating a conference, it creates these 2 event types as well.Do we need to show that in the changelog page.If not paper_trail tracking  of Event Type should be turned off inside `create_event_types`.What do you think?

differentreality: I think the creation of an event type is an action that happens, so it should be logged, even though it is supposed to happen automatically for every conference.

differentreality: What about the next line? Don't we have the same issue with difficulty levels?
https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L53

nishanthvijayan: I wasn't tracking DifficultyLevel.I tried it now locally and it has the same issue.So if were to similarly track DifficultyLevel similar change should be made for [https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L53]( https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L53)

app/models/role.rb
```diff
@@ -1,6 +1,9 @@
 class Role < ActiveRecord::Base
   belongs_to :resource, polymorphic: true
   has_and_belongs_to_many :users, join_table: :users_roles
+
+  has_paper_trail on: [:update], only: [:name, :description], meta: {conference_id: :resource_id}
```
hennevogel: There is no interface to create roles, why track it?

app/views/admin/versions/_event_change_description.html.haml
```diff
@@ -0,0 +1,35 @@
+- if version.event == 'create'
```
hennevogel: Please use [case](http://www.blackbytes.info/2015/10/ruby-case/)

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -0,0 +1,72 @@
+- if !version.item_type.constantize.exists?(version.item_id)
```
hennevogel: Please use [case](http://www.blackbytes.info/2015/10/ruby-case/)

app/models/role.rb
differentreality: It is only tracking **update** action. Action **toggle_user** will also be tracked.

nishanthvijayan: Actually this only tracks change of description of role.
Tracking adding/removing users (toggle_user) is a bit complicated and I'll add that in another PR.

app/views/admin/versions/_event_change_description.html.haml
```diff
@@ -0,0 +1,35 @@
...
+  = 'submitted new'
+
+- elsif version.event == 'destroy'
+  = 'deleted'
```
nishanthvijayan: @differentreality  There is no interface to delete an event.Should I remove this?

differentreality: Yeap you should

app/views/admin/versions/_event_change_description.html.haml
nishanthvijayan: Which form should I use?
```
case 
  when boolean condition
    statements
  when boolean condition
     statements
```
form?
Or using procs/lambdas?

hennevogel: I use the simple form and would move this into a helper

nishanthvijayan: Okay I'm a bit confused.How do I use the simple form here? the conditions are complex and is not just a simple variable comparison

hennevogel: Use a Boolean case. 

```ruby
irb(main):012:0> result = case
irb(main):013:1* when false && false then 'false'
irb(main):014:1> when false && true then 'half false'
irb(main):015:1> when true && true then 'truthy'
irb(main):016:1> end
=> "truthy"
```

nishanthvijayan: RuboCop offence [EmptyCaseCondition](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyCaseCondition)

hennevogel: Please use complete sentences when communicating. It helps understanding what you are trying to say very much :-)

nishanthvijayan: Sorry :) .
Trying to use case like you suggested threw up this [RuboCop offence](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyCaseCondition)

nishanthvijayan:
I've added one more commit.
Whenever an event is created, the `after_create :set_week` was causing one more version to be created.

event_evil_twin


The latest commit fixes this issue.

differentreality:
Aren't we ignoring changes to attribute **week**?

app/models/event.rb
```diff
@@ -255,13 +255,16 @@ def generate_guid
   end
 
   def set_week
-    self.week = created_at.strftime('%W')
-    save!
+    self.week = Time.zone.now.strftime('%W')
```
differentreality: Why do you need this change in this PR?

nishanthvijayan:
@differentreality yes we are ignoring week attribute.
I think what's happening is because of timing.Paper_trail determines object changes based on the previous version of that item.But here, when the save! inside set_week is called the create event's version has not been saved into versions table yet.So when saving the update event's version it determines that all the attributes have changes.I think this is what's happening

app/models/event.rb
nishanthvijayan: The idea is to remove the save! call and set the week attribute before create

nishanthvijayan:
I've changed the fix for the  extra event version issue.
I've also implemented revert and view changes feature.When displaying changes to attributes, it'll ignore attributes where the change is from one blank type value to another Eg: if an attribute's value changes from nil to "", its ignored

ezgif-985856866

nishanthvijayan:
I've changed the revert button to allow reverting single attributes of each change by means of a dropdown
ezgif-461106150

spec/controllers/admin/versions_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+          expect(flash[:error]).to match('The item is already in the state that you are trying to revert it back to')
+        end
+
+        it 'reverting deletion of object  creates it again' do
+          conference.program.event_types.first.destroy
+          get :revert, id: PaperTrail::Version.last.id
+          conference.reload
+          expect(PaperTrail::Version.last.event).to eq 'create'
+          expect(conference.program.event_types.count).to eq 2
+        end
```
differentreality: I would like to know the count of event_types *prior* to the action we are testing, so that this tests makes more sense when reading it.

spec/controllers/admin/versions_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        it 'revert fails when attribute is invalid' do
+          conference.update_attributes(short_title: 'testtitle', description: 'Some random text')
+          # Note: even though title is a valid attribute of conference, it was not  updated in the change we are trying to revert
+          get :revert, id: PaperTrail::Version.last.id, attribute: 'title'
+          conference.reload
+          expect(conference.short_title).to eq 'testtitle'
+          expect(conference.description).to eq 'Some random text'
+          expect(flash[:error]).to match('Revert failed.Invalid attribute')
```
differentreality: We should also check the value of title.

app/controllers/admin/versions_controller.rb
```diff
@@ -9,14 +9,35 @@ def index
...
+          if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
+            flash[:error] = 'The item is already in the state that you are trying to revert it back to'
+          else
+            @version.item[params[:attribute]] = @version.changeset[params[:attribute]][0]
+            @version.item.save!
+            flash[:notice] = 'The selected change was successfully reverted'
+          end
+
+        else
+          flash[:error] = 'Revert failed.Invalid attribute'
```
differentreality: Leave a space after the dot. There a few more places where you haven't.

app/views/admin/versions/index.html.haml
```diff
@@ -0,0 +1,48 @@
...
+      %h1 Revision History
+      %p.text-muted
+        Log of changes made to conferences and associated resources
+
+- @versions.each do |version|
+  .row
+    .col-md-9
+      %p
+        = version.whodunnit.nil? ? 'Someone (probably via the console)' : User.find(version.whodunnit).name
+
```
differentreality: Name is not a mandatory field. We should show something when there is no user.name

app/views/admin/versions/index.html.haml
```diff
@@ -0,0 +1,48 @@
...
+          - else
+            = 'deleted'
+
+        =  render partial: 'object_desc_and_link', locals: { version: version }
+
+        %span.text-muted
+          = distance_of_time_in_words(Time.now,version.created_at) + ' ago'
+
+    .col-md-3
+      .btn-group{role: "group"}
```
differentreality: single quotes

app/views/admin/versions/index.html.haml
```diff
@@ -0,0 +1,48 @@
...
+        %a.btn.btn-success.btn-sm.show-changeset{id: "#{version.id}"} View Changes
+        %button.btn.btn-default.dropdown-toggle.btn-sm.btn-primary{"data-toggle" => "dropdown", :type => "button"}
```
differentreality: single quotes, and new ruby syntax style.
You can have
`'type' => 'button' `or `type: 'button'`

differentreality:
Please update your 1st comment to give a short description of what you are actually implementing, what is not included, how things are supposed to work.
Eg info like:
* Only organizers can view and revert the revision history of their conferences
* Admins can view and revert the revision history for all conferences
* what can /cannot be reverted

app/views/layouts/_user_menu.html.haml
```diff
@@ -44,3 +44,7 @@
     = link_to(admin_users_path) do
       %span.fa.fa-user
       Users
+%li
+  = link_to(admin_revision_history_path) do
+    %span.fa.fa-history
+    Revision History
```
differentreality: You should only show this if a user has the appropriate access to see the page. Otherwise there will be a link which won't be accessible by the user.

differentreality:
The review app is erroring but locally it works (with bundle install and server restart).

app/controllers/admin/versions_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        flash[:error] = 'The item is already in the state that you are trying to revert it back to'
+      end
+      redirect_to admin_revision_history_path
+    end
+
+    private
+
+    def authorize_version
+      @version = PaperTrail::Version.find(params[:id])
+      return if current_user.has_role?(:organizer, Conference.find(@version.conference_id)) || current_user.is_admin
```
differentreality: You should also change that into an ability, so that you can check if the user can do something instead of checking if the user has a role.

db/migrate/20160524050538_add_conference_id_to_versions.rb
```diff
@@ -0,0 +1,5 @@
+class AddConferenceIdToVersions < ActiveRecord::Migration
+  def change
+    add_column :versions, :conference_id, :integer
```
differentreality: What about the old verion records that have nil for confernece_id? You should create a rake task to assign the conference_id for the old records too, when possible.

Gemfile
```diff
@@ -183,6 +183,10 @@ gem 'faker'
 # for seeds
 gem 'factory_girl_rails'
 
+# for pagination
+gem 'will_paginate', '~> 3.1.0'
```
differentreality: What about we just use datatable?

differentreality:
For adding/removing roles I would rephrase the description to sth like this:

UserX added role {organizer} to user {UserY} for conference {ConferenceX}

And the user name should also be a link.

app/assets/javascripts/osem-revisionhistory.js
```diff
@@ -0,0 +1,10 @@
+$(document).ready(function() {
+    $('.show-changeset').click(function(){
+        if ($(this).text() == 'View Changes'){
+            $(this).text('Hide Changes');
+        }else {
+            $(this).text('View Changes');
+        }
+        $('#changeset-' + this.id).toggle();
+    });
+});
```
differentreality: Please add an empty line at the end of the file (this is the styling we are following)

app/controllers/admin/versions_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        else
+          flash[:error] = 'Revert failed. Invalid attribute'
+        end
+
+      elsif @version.event != 'create'
+        @version.reify.save!
+        flash[:notice] = 'The selected change was successfully reverted'
+
+      elsif @version.event == 'create' && %w(Conference User).include?(@version.item_type)
+        flash[:error] = 'Reverting creation of a conference is not permitted'
```
differentreality: of a conference and a user?

app/helpers/application_helper.rb
```diff
@@ -276,4 +276,148 @@ def sign_in_path
...
+    # FIXME
+    version.event == 'create' ? "subscribed #{name_or_username(user_id)} to" : "unsubscribed #{name_or_username(user_id)} from"
+  end
+
+  def registration_change_description(version)
+    user_id = get_version_object(version).user_id
+    if user_id.to_s == version.whodunnit
+      case version.event
+      when 'create' then 'registered to'
+      when 'update' then "updated #{updated_attributes(version)} of  #{name_or_username(user_id)}'s registration for"
```
differentreality: let's make name_or_username(user_id) a link to that user

differentreality: In this case that the registration.user is making the change we can rephrase this, as in create and destroy actions.

app/helpers/application_helper.rb
```diff
@@ -276,4 +276,148 @@ def sign_in_path
...
+  # Eg: If version.changeset = '{"title"=>[nil, "Premium"], "description"=>[nil, "Premium = Super cool"], "conference_id"=>[nil, 3]}'
+  # Output will be 'title, description and conference'
+  def updated_attributes(version)
+    version.changeset.
+      reject{ |_, values| values[0].blank? && values[1].blank? }.
+      keys.map{ |key| key.gsub('_id', '').tr('_', ' ')}.join(', ').
+      reverse.sub(',', ' dna ').reverse
+  end
+
+  def name_or_username(user_id)
```
differentreality: Does this not cover us?
https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L75

app/helpers/application_helper.rb
nishanthvijayan: I didn't understand.How should it be rephrased?

app/helpers/application_helper.rb
nishanthvijayan: It doesn't check if its blank.
[https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L75](https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L75) only checks if user.name is nil

differentreality: Then perhaps we should fix this directly in the model and drop the helper all together.

app/helpers/application_helper.rb
differentreality: Like the other 2 cases. Line 358 and 360.
Now it reads UserX updated attributes of UserX's registration.

differentreality:
Datatable looks nice, let s put it in!

lib/tasks/version.rake
```diff
@@ -0,0 +1,23 @@
...
+  desc 'Sets conference_id in all pre-existing PaperTrail::Version objects'
+  task set_conference_id: :environment do
+
+    PaperTrail::Version.all.each do |version|
+      # All pre-existing versions are either of Conference or Event
+      if version.item_type == 'Conference'
+        version.conference_id = version.item_id
+        version.save
+      elsif version.item_type == 'Event'
+        event = (version.item || version.reify || version.next.reify)
```
differentreality: Even if the actual Event does not exist for some reason, the object attribute of changes has kept a record of either the conference_id (old records) or the program_id (through which we can go to conference).

That way we can ensure more (or all?) our records will get a conference_id.
So if event does not exist, search for what we want in the more hidden data ;)

nishanthvijayan: The task will set the conference_id even if the event has been deleted.It fails only when if the program of that event has been deleted(which is unlikely as there is no interface to do this) as in that case there is no way to find the conference_id using the program_id

app/helpers/application_helper.rb
nishanthvijayan: Done

app/assets/javascripts/osem-revisionhistory.js
nishanthvijayan: Done

lib/tasks/version.rake
```diff
@@ -0,0 +1,21 @@
+namespace :versions do
```
hennevogel: Please put this into the data namespace, not to your own

nishanthvijayan: Done

app/helpers/application_helper.rb
```diff
@@ -286,4 +286,122 @@ def sign_in_path
...
+  def change_creator_link(user_id)
+    if user_id
+      link_to User.find(user_id).name, admin_user_path(id: user_id)
+    else
+      'Someone (probably via the console)'
+    end
+  end
+
+  # Recieves a PaperTrail::Version object
+  # Returns the object in its current state if its alive else the state before or after this version (after if its a create event's version)
```
differentreality: Please rephrase this in a little more comprehensive way.

app/helpers/application_helper.rb
```diff
@@ -286,4 +286,122 @@ def sign_in_path
...
+  def vote_change_description(version)
+    user = get_version_object(version).user
+    if user.id.to_s == version.whodunnit && version.event == 'create'
+      'voted on'
+    elsif user.id.to_s != version.whodunnit && version.event == 'create'
+      "re-added #{user.name}'s vote on"
+    elsif version.event == 'update'
+      "updated #{user.name}'s vote on"
+    else
+      "deleted #{user.name}'s vote on"
```
differentreality: How do we enter this case?

app/helpers/application_helper.rb
```diff
@@ -286,4 +286,122 @@ def sign_in_path
...
+      "deleted #{user.name}'s comment on"
+    end
+  end
+
+  def vote_change_description(version)
+    user = get_version_object(version).user
+    if user.id.to_s == version.whodunnit && version.event == 'create'
+      'voted on'
+    elsif user.id.to_s != version.whodunnit && version.event == 'create'
+      "re-added #{user.name}'s vote on"
```
differentreality: When do we enter this case?

app/helpers/application_helper.rb
nishanthvijayan: I've updated the comment.

hennevogel:
Crashes for me in `admin/versions/index.html.haml` with 

```
ActionView::Template::Error (undefined method `title' for #<Campaign:0x007fbdc13c6ff0>)
```

nishanthvijayan:
@hennevogel Is it crashing now?

app/helpers/application_helper.rb
```diff
@@ -335,4 +335,124 @@ def speaker_width(rooms)
...
+  # Output will be 'title, description and conference'
+  def updated_attributes(version)
+    version.changeset.
+      reject{ |_, values| values[0].blank? && values[1].blank? }.
+      keys.map{ |key| key.gsub('_id', '').tr('_', ' ')}.join(', ').
+      reverse.sub(',', ' dna ').reverse
+  end
+
+  def change_creator_link(user_id)
+    if user_id
```
differentreality: It is not enough to check if a user_id was given, we need to double check the user exists before displaying a link which will otherwise fail.

app/models/event.rb
```diff
@@ -1,6 +1,6 @@
 class Event < ActiveRecord::Base
   include ActiveRecord::Transitions
-  has_paper_trail
+  has_paper_trail ignore: [:updated_at, :guid, :week], meta: { conference_id: :conference_id }
```
differentreality: Should that **not** show me week in changes? (I am seeing it on updating an event)

differentreality:
There is a migration error when trying to create the review app, but locally it works.

differentreality:
It would make sense to add a revert button on the table of changes, on every line of a changed attribute, so that if you are seeing all the changes, you can just click the button (instead of going through the dropdown).

lib/tasks/version.rake
```diff
@@ -0,0 +1,21 @@
+namespace :data do
+  desc 'Sets conference_id in all pre-existing PaperTrail::Version objects'
+  task set_conference_id: :environment do
```
differentreality: The name of this task is too general. This could refer to anything. 
The name of the task should tell us a bit more,eg, `set_conference_in_versions`, and it should be less generic. What if we want to set the conference_id attribute in other tables in the future... It will be confusing.

differentreality:
Reverting some changes is not possible (eg. user sign up). How about we don't show the revert button all together?

nishanthvijayan:
@differentreality So then how do we revert create/destroy of other objects? Or is that not necessary?

differentreality:
I mean we don't show the button only when it cannot be used!

nishanthvijayan:
Oh :) .I'll look into that

RichiH:
Instead of, or additionally to, reverting singular changes, it might also be nice to have a "make this version current" button which simply reinstates old state.

spec/features/versions_spec.rb
```diff
@@ -0,0 +1,373 @@
...
+
+    visit admin_revision_history_path
+    expect(page).to have_text("Someone (probably via the console) created new campaign Test Campaign in conference #{conference.short_title}")
+    expect(page).to have_text("Someone (probably via the console) updated utm source, utm medium, utm term and utm content of campaign Test Campaign in conference #{conference.short_title}")
+    expect(page).to have_text("Someone (probably via the console) deleted campaign Test Campaign in conference #{conference.short_title}")
+  end
+
+  scenario 'display changes in user', feature: true, js: true, versioning: true do
+    sign_out
+    visit new_user_password_path
```
differentreality: You have 3 different sub-scenarios to test here, you should split them up :)

spec/controllers/admin/versions_controller_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  let(:admin) { create(:admin) }
+
+  with_versioning do
+    describe 'GET #revert' do
+      before :each do
+        sign_in admin
+      end
+
+      context 'when no attribute is set' do
+        it 'reverts all changes for update events' do
```
differentreality: update events is confusing, it makes me think of updating events, which is not what you are testing for here.
`update action` perhaps?

differentreality:
We can probably do better with our tests. I can see more than a couple of things we should rewrite. But I think it would be easier for everyone to focus on that in a different PR, after this one is merged.

differentreality:
I'd rather show the timestamp of the change in the beginning of the line, so that date and time don't get split up in 2 lines

app/controllers/admin/versions_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+        elsif @version.item.nil?
+          flash[:error] = 'Reverting a specific attribute is not allowed for deleted objects'
+
+        elsif @version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.keys.include?(params[:attribute])
+          if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
+            flash[:error] = 'The item is already in the state that you are trying to revert it back to'
+          else
+            @version.item[params[:attribute]] = @version.changeset[params[:attribute]][0]
+            @version.item.save!
```
differentreality: What if that save does not succeed?

differentreality:
I have some older version records that for Event creation it shows a `create` record (photo 1) and an `update` record (photo 2). The update record holds the same info as the create one, same timestamps, except it adds the week. Problem is that in update record we enable the revert option, which will try to revert all attributes in nil (eg. abstract, title, event_type_id). That is of course not allowed, as we cannot have an event without a title or an abstract or an event_type_id, and the revert action throws an exception.
1) We should check if revert action is possible before showing that button. And we should authorize this a bit more in ability, so that we avoid cases like the user hitting revert and trying to make event_type_id nil.
2) You should be catching that exception to begin with and showing an error flash message in https://github.com/openSUSE/osem/pull/1019/files#diff-bc6a21dd66d0444573ff114349a614cbR27

app/models/ability.rb
```diff
@@ -166,6 +166,10 @@ def signed_in_with_organizer_role(user)
     can [:edit, :update, :toggle_user], Role do |role|
       role.resource_type == 'Conference' && (conf_ids_for_organizer.include? role.resource_id)
     end
+
+    can [:index, :revert], PaperTrail::Version do |version|
```
differentreality: We should properly authorize `revert` action. Please read my latest comments on the PR about this. Then you disable the revert button based on authorization and not custom checks. Same for the revert buttons inside the table with the changes.

app/controllers/admin/versions_controller.rb
nishanthvijayan: I've added a check to see if save fails. But adding those conditions is raising this rubocop [offence](https://travis-ci.org/openSUSE/osem/builds/145311507#L1796).

I was thinking of splitting up this controller action.One for restoring the object to the version and once action for revert a single attribute.This would make it easier to define ability and will fix the rubocop offence mentioned above

app/views/admin/versions/_object_changes.html.haml
```diff
@@ -9,6 +9,8 @@
           %th Previous Value
         - if version.event != 'destroy'
           %th New Value
+        - if version.event == 'update' && !version.item.nil?
```
nishanthvijayan: @differentreality I want to move this check to ability.But this check should be invoked only when reverting a single attribute change.So this cannot be added to the ability with the current setup.

differentreality:
I think we should not be reverting the creation of an Event.
* It is not possible to perform the `destroy` action anywhere in the application, events can only be marked as `withdrawn`
* Reverting the deletion of an Event is more complicated than other models (we need to revert event.users too, which is currently not being done and things secretly break).

app/assets/javascripts/osem-datatables.js
```diff
@@ -1,6 +1,6 @@
 $(function () {
   $(document).ready(function() {
-    $('.datatable').DataTable({
+    $('.data-table').DataTable({
```
differentreality: Why?

nishanthvijayan: Yeah that was unnecessary.I've changed it.

app/helpers/application_helper.rb
differentreality: This does not seem to appear when  reverting the action of deleting a vote. It would be nice to distinguish between a user voting and an admin deleting a vote and then reverting that deletion.
Take a look at https://osem-demo-pr-1019.herokuapp.com/admin/revision_history ID 541

app/assets/javascripts/osem-datatables.js
nishanthvijayan: Here, using 'datatable' as the classname to identify and initialize datatables is trouble in the making.
The datatable library when initializing a table as a datatable adds 'dataTable' class to the table.So if someone were to initialize a table as datatable from someplace else, like done [here](https://github.com/openSUSE/osem/blob/master/app/views/admin/volunteers/_volunteers_table.html.haml#L24) , this would cause dataTable library to raise reinitialization error alert
reinitialize-warning

differentreality: There is  a solution for handling that case with datatables described in the link provided by the error.

app/helpers/application_helper.rb
nishanthvijayan: Everything seems to be working correctly with Vote.
 - If User A revert the deletion of User A's vote, it'll show
   User A voted on event QWERTY
 - If User A reverts the deletion of User B's vote, it'll show:
   User A re-added User B's vote on event QWERTY

vote

nishanthvijayan:
@differentreality The huge loading time is because of how we use datatable.
Describing changes requires 2-4 additional queries for each change(version).The pagination of the table is client-side and there is no server-side pagination.All versions are fetched in one go and datatable renders all of them.
So if there are 1000 versions accessible by the user, then all 1000 would be fetched and all the associated queries foir describing them will be made on loading the page instead of just fetching 10 records (& associated stuff).
Datatable has an option for [server-side processing](https://datatables.net/release-datatables/examples/data_sources/server_side.html)

> With server-side processing enabled, all paging, searching, ordering actions that DataTables performs are handed off to the server...As such, each draw of the table will result in a new Ajax request being made to get the required data.

But in the exmaple provided in their docs, the data to displayed is simple, but in our case its complex html (buttons and such).Also another problem with this approach is that we'll have to implement pagination (and sorting and searching if needed) in the server-side.I don't think its worth the effort.My suggestion is that we go back to using will-paginate and drop Datatables.

differentreality:
@nishanthvijayan we should investigate on how we can keep sorting and search (which should be applied on the whole set of objects, not just the ones visible on the page). But I would leave this for a new PR.

app/helpers/application_helper.rb
differentreality: I think it is best if we use the phrasing 're-added User's X vote' in both case. Because then we can tell the difference between a user going to the proposal and voting, and a user reverting a change from the changelog.

nishanthvijayan:
Sorting already works for the whole set of version objects.

differentreality:
@nishanthvijayan this has been a lot and very good work :) 
You can go ahead and squash commits and make sure you rebase. I have added a comment about some phrasing, please also fix this.

I will make a final review during the next few days, and hopefully we can get this in.

differentreality:
@nishanthvijayan please rebase again to fix travis failure

nishanthvijayan:
@differentreality In the last commit, I've reintroduced datatable and have attempted to implement serverside processing similar to #1117

nishanthvijayan:
@differentreality I've removed the last commit and all the will_paginate stuff.It uses datatables now.

differentreality:
@nishanthvijayan please rebase all your PRs so that we can start merging.

nishanthvijayan:
@differentreality Done

differentreality:
:+1:

PR #1023, opened 2016-05-31
public schedule turned

Ana06:
timeline-smaller

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -82,6 +82,16 @@ class Conference < ActiveRecord::Base
   end
 
   ##
+  # Check if there is any event in the program scheduled.
+  #
+  # ====Returns
+  # * +false+ ->  If there aren't schedules events
+  # * +true+ -> If there are schedules events
+  def scheduled?
+    self.program.events.where.not(start_time: nil).any?
+  end
+
+  ##
   # Checks if the user is registered to the conference
   #
   # ====Args
--- a/app/views/conference/_schedule.html.erb
+++ b/app/views/conference/_schedule.html.erb
@@ -13,51 +13,39 @@
         <div class="tab-pane <%= 'active' if @dates.first == date %>" id ="<%= date %>" >
           <table class="table table-bordered" id="schedule" width="100%">
             <tr>
-              <th>Time</th>
-              <% @rooms.each do |room| %>
-                <th><%= room.name %></th>
+              <th>Rooms</th>
+              <%# The conference starts at 9 and ends at 18:45 %>
+              <% intervals = (19 - 9) * 4 %>
+              <% (9..18).each do |i| %>
+                  <th colspan="4"><%= "#{i}:00" %></th>
               <% end %>
             </tr>
-
-                <% span = {} %>
-                <% @rooms.each do |room| %>
-                  <% span[room.id] = 1 %>
-                <% end %>
-
-            <% conf_start = DateTime.parse("#{date} 09:00") %>
-            <% conf_end = DateTime.parse("#{date} 19:00") %>
-
-            <% while conf_start < conf_end %>
-              <%-
-                start_hour = conf_start.hour
-                start_min = conf_start.min %>
-
+            <% @rooms.each do |room| %>
+              <% span = 1 %>
               <tr>
-                <td style="width:5%">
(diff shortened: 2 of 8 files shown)


app/views/conference/_schedule.html.erb
```diff
@@ -13,51 +13,39 @@
         <div class="tab-pane <%= 'active' if @dates.first == date %>" id ="<%= date %>" >
           <table class="table table-bordered" id="schedule" width="100%">
             <tr>
-              <th>Time</th>
-              <% @rooms.each do |room| %>
-                <th><%= room.name %></th>
+              <th>Rooms</th>
+              <%# The conference starts at 9 and ends at 18:45 %>
+              <% intervals = (19 - 9) * 4 %>
```
ChrisBr: Is it really necessary to write this explicit?

(but we want to change that anyway later ...)

ChrisBr: Just hardcode it, it doesn't matter if it's 10 or 100 hours, one hour is always 4 * 15 minutes!

Ana06: But I need to do `(end_hour - start_hour) * 4` when we allow to change the start and end time, that's why I keep it like that.

ChrisBr: You could just iterate over the number of hours and then you have 4 boxes for each hour!

hennevogel: And you should never do this in a view...

Ana06: I am iterating over the number of hours. I need the intervals to set the width of each column, although I could calculate the width directly. I am changing that and moving it to the controller.

Ana06:
What about that?

schedule-less-info

I suppose that we don't want to add the star now (until the user can select his own schedule), but I added to see how it looked like.

app/models/conference.rb
```diff
@@ -82,6 +82,16 @@ def date_range_string
   end
 
   ##
+  # Check if there is any event in the program scheduled.
+  #
+  # ====Returns
+  # * +false+ ->  If there aren't schedules events
+  # * +true+ -> If there are schedules events
+  def scheduled?
```
ChrisBr: Why do you move this to conference?

app/views/conference/_schedule.html.erb
```diff
@@ -13,51 +13,36 @@
         <div class="tab-pane <%= 'active' if @dates.first == date %>" id ="<%= date %>" >
           <table class="table table-bordered" id="schedule" width="100%">
             <tr>
-              <th>Time</th>
-              <% @rooms.each do |room| %>
-                <th><%= room.name %></th>
+              <th>Rooms</th>
+              <% (@start_hour..@end_hour-1).each do |i| %>
```
ChrisBr: What is ``i`` ? I assume the hour? Please use a self-explanatory variable name

app/views/conference/_schedule.html.erb
```diff
@@ -13,51 +13,36 @@
...
-                  <% if !event.empty? %>
-                    <!-- There is an event, calculate the span and show it -->
-                    <% span[room.id] = event[0].event_type.length / 15 %>
-                    <%= render partial: 'schedule_item', locals: {event: event, room: room, span: span} %>
-                    <% span[room.id] = span[room.id] + 1 %>
-                    <% elsif  span[room.id] > 1 %>
-                      <!-- span is <%= span[room.id] %> and bigger than 1, substract 1 from span -->
-                      <% span[room.id] = ( span[room.id] - 1 ) if span[room.id] > 1 %>
+                <td><%= room.name %></td>
+                  <% events_for_date = room.events.find_all{ |e| e.start_time.to_date == date } %>
```
ChrisBr: To be honest, I don't like assigning variables in the view :(
@hennevogel ^^

app/models/conference.rb
Ana06: Because it was used only once (in the schedule view) like that:

`@conference.program.events.scheduled.any?`

So, moving it to conference, in the schedule view I can write only:

`@conference.scheduled?`

which I think is more readable.

hennevogel:
As we have discussed to do it differently I close this.

PR #1024, opened 2016-05-31
Limit event types length

Ana06:
I've limit the event types length to be a multiple of 15, adding validation and tests.

I've created a variable `LENGTH_STEP` in `EventType` for if we want to change the step (10 minutes instead of 15 for example). I plan to use this variable in the schedule also.

### DIFF
--- a/spec/models/event_type_spec.rb
+++ b/spec/models/event_type_spec.rb
@@ -1,0 +1,32 @@
+require 'spec_helper'
+
+describe EventType do
+  let(:conference) { create(:conference) }
+  let(:event_type) { create(:event_type, program: conference.program) }
+
+  describe 'association' do
+    it { is_expected.to belong_to :program }
+    it { is_expected.to have_many :events }
+  end
+
+  describe 'validation' do
+    it 'has a valid factory' do
+      expect(build(:event_type)).to be_valid
+    end
+
+    it { is_expected.to validate_presence_of(:title) }
+    it { is_expected.to validate_presence_of(:minimum_abstract_length) }
+    it { is_expected.to validate_presence_of(:maximum_abstract_length) }
+
+    describe 'has a valid length' do
+      it 'length is greater than 0' do
+        should validate_numericality_of(:length).
+        is_greater_than(0)
+      end
+
+      it 'length is multiple of LENGTH_STEP' do
+        expect(build(:event_type, program: conference.program, length: 37)).not_to be_valid
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


Ana06:
There is a test failing in my computer: 

> Failures:
>   1) conference/show.html.haml renders banner component

But it doesn't fail here... :confused:

ChrisBr:
@Ana06 please provide some stacktrace / error message then!

spec/models/event_type_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      expect(build(:event_type)).to be_valid
+    end
+
+    it { is_expected.to validate_presence_of(:title) }
+    it { is_expected.to validate_presence_of(:minimum_abstract_length) }
+    it { is_expected.to validate_presence_of(:maximum_abstract_length) }
+
+    describe 'has a valid length' do
+      it 'length is greater than 0' do
+        should validate_numericality_of(:length).
```
ChrisBr: You can also split this test!

ChrisBr: http://betterspecs.org/#single

Ana06:
I closed the console and when running the tests again I realised that if I run only the file where the error is supposed to be, then there is no error. But If I run all tests, the error appears. It is very strange :confused: Here is the start and end of the error (is is a really long error):

image

image

spec/models/event_type_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        is_greater_than(0)
+      end
+
+      it 'length is multiple of LENGTH_STEP' do
```
ChrisBr: The test is not testing what it says

ChrisBr: I'd group it into some valid and invalid context

spec/models/event_type_spec.rb
ChrisBr: And please use the new expect syntax
http://betterspecs.org/#expect

spec/models/event_type_spec.rb
ChrisBr: Also the positive case is missing

ChrisBr: Something like this:
```
 describe 'length' do
     it 'is valid when length is multiple of LENGTH_STEP' do
      ...
    end 
     it 'is not valid when length is multiple of LENGTH_STEP' do
      ...
    end
end
```

ChrisBr:
LGTM, @hennevogel what do you think?

hennevogel:
If it looks good for you let's go :-)

PR #1025, opened 2016-06-01
Add payment mode selection with conference for organisers

rishabhs95:
add capability for organisers to select payment modes for their conferences as Online/Offline.

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -57,6 +57,9 @@ class Conference < ActiveRecord::Base
   validates_format_of :short_title, with: /\A[a-zA-Z0-9_-]*\z/
   validates :registration_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
 
+  validates :payment_method, presence: true, null: false, inclusion: {in: %w(online offline),
+                                                                      message: "'%{value}' is not a valid payment method"}
+
   # This validation is needed since a conference with a start date greater than the end date is not possible
   validate :valid_date_range?
   before_create :generate_guid
--- a/app/views/admin/conference/edit.html.haml
+++ b/app/views/admin/conference/edit.html.haml
@@ -23,4 +23,13 @@
         = f.input :end_date, :as => :string, :input_html => { :id => "conference-end-datepicker", :readonly => "readonly"  }
       = f.inputs name: "Registrations" do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+      = f.inputs name: "Tickets" do
+        = f.label :payment_method
+        .btn-group{"data-toggle" => :payment_method, class: "pull-right"}
+          %label.btn.btn-default
+            = f.radio_button :payment_method, 'online'
+            Pay Online
+          %label.btn.btn-default
+            = f.radio_button :payment_method, 'offline'
+            Pay at the venue
       = f.action :submit, :as => :button, :button_html => {:class => "btn btn-primary"}
--- a/spec/models/conference_spec.rb
+++ b/spec/models/conference_spec.rb
@@ -1545,6 +1545,20 @@ describe Conference do
       should validate_presence_of(:end_date)
     end
 
+    it 'is not valid without payment mode' do
+      should validate_presence_of(:payment_method)
+    end
+
+    it 'is valid with payment mode as Online' do
+      should allow_value('online').for(:payment_method).
+        with_message("'%{value}' is not a valid payment method")
+    end
+
+    it 'is valid with payment mode as Offline' do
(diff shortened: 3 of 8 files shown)


rishabhs95:
will replace dropdown with better UI alternative as suggested by hernan

db/migrate/20160531152256_add_payment_mode_to_conferences.rb
```diff
@@ -0,0 +1,5 @@
+class AddPaymentModeToConferences < ActiveRecord::Migration
+  def change
+    add_column :conferences, :payment_mode, :string, default: 'Offline'
```
lagartoflojo: Do we want the `payment_mode` column to ever be `NULL`?

lagartoflojo: Also, how about using Enums?
I think it fits this use case pretty nicely.
http://api.rubyonrails.org/v4.1.0/classes/ActiveRecord/Enum.html

app/models/conference.rb
```diff
@@ -57,6 +57,9 @@ class Conference < ActiveRecord::Base
   validates_format_of :short_title, with: /\A[a-zA-Z0-9_-]*\z/
   validates :registration_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
 
+  validates :payment_mode, presence: true, inclusion: {in: %w(Online Offline),
```
lagartoflojo: I would go for lower-case `online` and `offline`

db/migrate/20160531152256_add_payment_mode_to_conferences.rb
```diff
@@ -0,0 +1,5 @@
+class AddPaymentModeToConferences < ActiveRecord::Migration
+  def change
+    add_column :conferences, :payment_mode, :string, default: 'offline'
```
lagartoflojo: Hmm now that I think about it, "payment method" is better than payment mode :-\
http://www.google.com/trends/explore#q=%22payment%20method%22%2C%20%22payment%20mode%22&cmpt=q&tz=Etc%2FGMT-2

rishabhs95: Ok, will change the name to `payment_method`

app/models/conference.rb
rishabhs95: Yes, I've made the changes for it.

db/migrate/20160531152256_add_payment_mode_to_conferences.rb
rishabhs95: I am not using `payment_mode` to ever be `NULL`, so I've added `null: false`  on `payment_method` as well.

spec/models/conference_spec.rb
```diff
@@ -1545,6 +1545,18 @@
...
+    end
+
+    it 'is valid with payment mode as Online' do
+      should allow_value('online').for(:payment_method)
+    end
+
+    it 'is valid with payment mode as Offline' do
+      should allow_value('offline').for(:payment_method)
+    end
+
```
kalabiyau: I would also test the message of the validator to be consistent.

rishabhs95: I have updated the validator tests for custom messages as you suggested

app/models/conference.rb
```diff
@@ -57,6 +57,9 @@ class Conference < ActiveRecord::Base
   validates_format_of :short_title, with: /\A[a-zA-Z0-9_-]*\z/
   validates :registration_limit, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
 
+  validates :payment_method, presence: true, null: false, inclusion: {in: %w(online offline),
```
hennevogel: This type of data is called [enum](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html)

app/views/admin/conference/edit.html.haml
```diff
@@ -23,4 +23,13 @@
         = f.input :end_date, :as => :string, :input_html => { :id => "conference-end-datepicker", :readonly => "readonly"  }
       = f.inputs name: "Registrations" do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+      = f.inputs name: "Tickets" do
```
hennevogel: We use [Bootstrap-Switch-Checkbox](https://github.com/openSUSE/osem/wiki/Bootstrap-Switch-Checkbox) for this kind of input

lagartoflojo:
@hennevogel I think we can close this PR, as we are not going to offer the "offline" payment option, correct?

hennevogel:
true!

PR #1030, opened 2016-06-07
New schedule

Ana06:
image

I still have to fix few things for devices with bigger screens.


hennevogel:
* You need to make sure that the height of the Talks do not vary. Otherwise there is too much flickering when you turn the carousl
* I don't like that the carousel wraps. Days _never_ wrap in reality.
* No interval for the carousel please

app/models/event.rb
```diff
@@ -223,6 +223,13 @@ def calculate_progress
     (100 * result.values.count(true) / result.values.compact.count).to_s
   end
 
+  ##
+  # Returns the number of columns this event should span in the schedule
+  #
+  def span
```
ChrisBr: The function does not what the name states!

app/models/event_type.rb
```diff
@@ -10,6 +10,7 @@ class EventType < ActiveRecord::Base
 
   alias_attribute :name, :title
 
+  # If LENGTH_STEP > 60 the public schedule must be changed
```
ChrisBr: Can we somehow force that the step is between 15 and 60 minutes? Changes the schedule is not that easy!

app/models/conference.rb
```diff
@@ -82,6 +82,16 @@ def date_range_string
   end
 
   ##
+  # Check if there is any event in the program scheduled.
+  #
+  # ====Returns
+  # * +false+ ->  If there aren't schedules events
+  # * +true+ -> If there are schedules events
+  def scheduled?
```
ChrisBr: Why did you move this function to conference?

hennevogel:
Another thing that I noticed: Talks on the fringe (e.g. between 12:00 and 13:00) get pushed outside. You will never see them. I guess the carousel items should overlap.

app/models/event.rb
Ana06: no? :confused:

app/models/conference.rb
Ana06: because it was called only once as `conference.program.schedule?` and moving it I can write `conference.schedule?` I think is better.

Ana06:
I am limiting the number of lines to 3 instead of using truncate.

I am using `word-break: break-word;` to split the title in several lines without making the table's size change, but it doesn't work in Firefox. I use `word-break: break-all;` for this case, it doesn't look as good but I think it is better than allowing that change the table size. Also, the '...' doesn't apear in Firefox because the css elipses is not allowed. 

I have a doubt. There are some css things that if I put them inside `style=""` work fine but if I move them to `osem-schedule.css.scss` then stop working. That happens with the `text-font` for example, why does this happen?

app/models/event_type.rb
Ana06: In fact, we need the step to be divisor of 60 to display the schedule properly. How can I force that the step is a divisor of 60? I can add a test, but I am not pretty sure if that is what you mean.

hennevogel: This is a constant. Whoever changes this must deal with the outcome. They have been warned. All is good..

Ana06:
I've added the swipe support in the carousel.

This is how it looks now:

image

I am changing it to adjust the height, but I am not sure if it is better:

image

The problem is that there is a lot of space in the event cells where there is not anything but I can not add more number of lines as I am using percentages for the height and I can not determine the number of lines with that.

hennevogel:
@Ana06 the swipe works nicely. cool

Ana06:
Now the schedule height fits a 600px screen. This how it looks with 3 rooms:

image

And when there are many rooms:

image


I think it looks good. The only "problem" is that the line height has to be the same for different screen sizes. I am changing the speaker picture size taking into account the height I have free instead of the device. Its minimum height is `17px` and the maximum `29px`.

Ana06:
If the speaker picture is too small, I can make it a little bit bigger. I mean that the minimum height could be bigger.

The maximum is `29px` for 280px screen width, but It could be bigger if I hide it in devices with less than 300px for example. In fact, I should hide it for devices with less than 280px anyway.

The reason why I have two heights is because the tittle height is multiple of 17, but not the cell height, just to compensate it. With a difference of `12px` is enough.

Ana06:
Is there anything else I should change?

hennevogel:
Awesome stuff :-) :+1:

app/views/conference/schedule.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="container">
-  <% if @conference.program.events.scheduled.any? %>
+  <% if @rooms.length > 1 and @conference.program.events.scheduled.any? %>
```
nishanthvijayan: @Ana06  Is there any reason why its > 1 and not >= 1 ?
Right now if there is only 1 room, the schedule won't be displayed, even if there are events scheduled in that one room.

Ana06: @nishanthvijayan yes, if there is only one room the all events approach is better, as it lists all the events by date and time and in that case the room is not important.

PR #1031, opened 2016-06-08
Some cherry-picked fixes for 1.1

hennevogel:


app/controllers/admin/questions_controller.rb
```diff
@@ -5,7 +5,7 @@ class QuestionsController < Admin::BaseController
 
     def index
       authorize! :index, Question.new(conference_id: @conference.id)
-      @questions = Question.where(global: true).all | Question.where(conference_id: @conference.id)
```
differentreality: `Question.where(conference_id: @conference.id)` was meant to show questions that were added to the conference but are not (yet) enabled. But adding a question now, enables it automatically. So as long as it is not disabled, we would be OK.

differentreality:
:+1:

PR #1032, opened 2016-06-10
All events grouped by date and time

Ana06:
Change All Events to grouped by day and time. It looks like:

image

And in the mobile phone:

image


Ana06:
I added more information (room, track and speaker picture) in the event boxes than there was before because there is place for it and it may be useful.

I've just added unscheduled events, as they where here before, so I suppose we want to keep them:


image

Ana06:
Do we want to add tabs with the dates at the top as in the schedule to be able to see the events that take place a concrete day easily?

differentreality:
I personally, and got used to the tabs. It also gives a nice overview of which are the total days of the conf.

Ana06:
I've added a dropdown instead of tabs:

image

When you click a date it takes you to the selected date. It also works fine in a mobile phone:

image

I think I should also add a button to go up in every date title, it may be useful if there are many events.

Why are all dates in the format `year-month-day`? I understand what it means but it is a strange format for me, `day/month/year` would be more natural, although maybe that is something that depends on the country.

ChrisBr:
> Why are all dates in the format year-month-day? I understand what it means but it is a strange format for me, day/month/year would be more natural, although maybe that is something that depends on the country.

https://en.wikipedia.org/wiki/ISO_8601

app/controllers/conference_controller.rb
```diff
@@ -15,6 +15,9 @@ def schedule
     @rooms = @conference.venue.rooms if @conference.venue
     @events = @conference.program.events
     @dates = @conference.start_date..@conference.end_date
+    confirmed_events = @conference.program.events.where(state: 'confirmed')
```
ChrisBr: @conference.program.events.confirmed

http://guides.rubyonrails.org/active_record_querying.html#scopes

app/controllers/conference_controller.rb
```diff
@@ -15,6 +15,9 @@ def schedule
...
+    @scheduled_events = confirmed_events.where('start_time IS NOT NULL').order(:start_time)
```
ChrisBr: Maybe it's worth to introduce also a scope for this.
(Which you can also use in the next line combined with a ``not``)

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,67 @@
...
+  <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown">
+    Dates
+    <span class="caret"></span>
+  </button>
+  <ul class="dropdown-menu">
+    <% @dates.each do |date| %>
+      <!-- we check if there is any event that day to indicate it appropiately -->
+      <% date_ini =  DateTime.parse("#{date} 00:00") %>
+      <% date_fin =  date_ini + 1.day %>
+      <% events_date = @scheduled_events.where('start_time >= ? AND start_time <= ?', date_ini, date_fin) %>
```
ChrisBr: Maybe you can do that in a helper?
I don't like all that logic in the view ...

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,67 @@
...
+  </ul>
+</div>
+
+<div class="row">
+
+  <!-- scheduled events -->
+  <% date = nil %>
+  <% time = nil %>
+  <% @scheduled_events.each do |event| %>
+    <% event_date = event.start_time.strftime("%Y-%m-%d") %>
```
ChrisBr: Please move that to the model

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,67 @@
...
+    <% unless event_date.eql?(date) %>
+      <div class='col-xs-12 col-md-12'  style='margin-top: 30px; border-bottom: 1px solid #848484;'>
+        <h3 id='program-<%= event_date %>'><%= date = event_date %></h3>
+      </div>
+    <% end %>
+    <% event_time = event.start_time.strftime("%H:%M") %>
```
ChrisBr: Same for this one!

app/views/conference/_event.html.haml
```diff
@@ -0,0 +1,27 @@
+%a.unstyled-link{href: url_for(conference_program_proposal_path(@conference.short_title, event.id))}
+  %div{ class: 'event-item', style: 'border: 1px solid #E6E6E6; width: 100%; padding: 0px 10px 10px 15px; margin: 2px;cursor:pointer"'}
```
ChrisBr: Please don't use inline CSS!

app/controllers/conference_controller.rb
differentreality: We already have btw the following scopes:
https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L29
https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L25

Ana06: The scheduled scope should not check if the event is confirmed? I think so. It depends if we want to consider the unconfirmed event with start_time scheduled or not and consequently showing them in the schedule. That scope is already used in the schedule, so I think I should use it here and maybe modified what the scope does.

Ana06: Also in the event model, an event is scheduled if it has not only start_time, but also room. So I think that if the room is necessary in the schedule model it should also be necessary in the program scope. In this case I would introduce that it is confirmed too.

differentreality: Or you can just combine the two scopes so that you can do what you want.

Ana06: Yes, but do it makes sense in any case to consider an event schedule without room or without being confirmed?

ChrisBr: I agree with @Ana06 !

app/controllers/conference_controller.rb
```diff
@@ -16,6 +16,9 @@ def schedule
     @events = @conference.program.events
     @dates = @conference.start_date..@conference.end_date
 
+    @scheduled_events = @conference.program.events.scheduled.order(:start_time)
+    @unscheduled_events = @conference.program.events.unscheduled
```
hennevogel: Unscheduled events don't happen...

app/helpers/application_helper.rb
```diff
@@ -276,4 +276,10 @@ def sign_in_path
   def unread_notifications(user)
     Comment.accessible_by(current_ability).find_since_last_login(user)
   end
+
+  def any_event_date?(events, date)
```
hennevogel: I don't understand the methods name. Please use a sentence if you're not sure

app/helpers/application_helper.rb
```diff
@@ -276,4 +276,10 @@ def sign_in_path
...
+    date_ini =  DateTime.parse("#{date} 00:00")
+    date_fin =  date_ini + 1.day
```
hennevogel: Why assign them to variables first?

app/models/event.rb
```diff
@@ -69,11 +69,19 @@ class Event < ActiveRecord::Base
...
-  # Checkes if the event has a start_time and a room
+  # Checkes if the event is confirmed and has a start_time and a room
   # ====Returns
   # * +true+ or +false+
   def scheduled?
-    room && start_time ? true : false
+    room && start_time && (state == 'confirmed') ? true : false
+  end
+
+  def date_string
```
hennevogel: This is a view helper. Not a model method...

app/models/event.rb
```diff
@@ -69,11 +69,19 @@ class Event < ActiveRecord::Base
...
+    start_time.strftime('%Y-%m-%d')
+  end
+
+  def time_string
```
hennevogel: This is a view helper. Not a model method...

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,73 @@
 <h1 class="text-center">Program for <%= @conference.title %></h1>
```
hennevogel: When you work on this please hamlfy it.

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,73 @@
...
+      <div class='col-xs-12 col-md-12 all-events-title'>
+        <span class='date-title' id='program-<%= event.date_string %>'>
+          <%= date = event.date_string %>
+        </span>
+        <a title="Go up" class="pull-right" href="#program">
+          <i class="fa fa-angle-double-up fa-lg" aria-hidden="true"></i>
+        </a>
+      </div>
+    <% end %>
+    <% if !event.time_string.eql?(time) %>
```
hennevogel: What's the difference between 
```
unless event.date_string.eql?(date)
```
and
```
if !event.date_string.eql?(date)
```

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,73 @@
...
+      <div class='col-xs-12 col-md-1 start-time'>
```
hennevogel: Please do not mix bootstrap col's with other CSS especially if you manipulate the [box-model](https://www.w3.org/TR/css3-box/) as this can break the grid

app/controllers/conference_controller.rb
ChrisBr: @hennevogel wrong! That was the point why we implemented the "All Events" tab in the first place. Because we wanted to show the accepted and confirmed talks even if they aren't scheduled yet!

app/assets/stylesheets/osem-schedule.css.scss
```diff
@@ -75,3 +75,67 @@ a.unstyled-link {
   background-image: linear-gradient(to top, #f7faf2 24%, #53ab4a 97%,
 #53ab4a 100%);
 }
+
+.event-item{
```
hennevogel: Please use bootstrap components and not your own styling. Makes it way easier to maintain in a group.

app/views/conference/schedule.html.erb
```diff
@@ -25,3 +25,19 @@
       <%= render 'all_events' %>
   <% end %>
 </div>
+
+<%= javascript_tag do %>
```
hennevogel: I think we should rather put this onto a different controller method now. Not in a tab. This view is getting complicated!

app/views/conference/_all_events.html.erb
```diff
@@ -1,19 +1,73 @@
...
-          </small>
-        </h3>
-        <h4>
-          presented by <%= event.speaker_names %>
-        </h4>
-        <p>
-          <%= truncate(event.abstract, :length => 400) -%>
-          <%= link_to 'more', conference_program_proposal_path(@conference.short_title, @conference.program.id, event.id) if event.abstract.length > 400 %>
-        </p>
+<div class="dropdown program-dropdown">
```
hennevogel: I think this is not needed. Just do simple anchors for the days and events like this:

ezgif com-video-to-gif 3

app/views/conference/_event.html.haml
```diff
@@ -0,0 +1,26 @@
+%a.unstyled-link{href: url_for(conference_program_proposal_path(@conference.short_title, event.id))}
+  %div{ class: 'event-item'}
```
hennevogel: Like I said above. Please use bootstrap components.

app/helpers/application_helper.rb
Ana06: In the case of `date_ini` to be able to do `date_ini + 1.day` instead of parsing the date again. In the case of `date_fin` I thought it was more clear. I'll delete `date_fin.`

app/views/conference/_all_events.html.erb
Ana06: I don't know what you mean.

app/views/conference/_all_events.html.erb
Ana06: Why not? I think it may be useful if there are a lot of events and as @differentreality said it gives you an overview of which are the total days of the conference.

app/assets/stylesheets/osem-schedule.css.scss
Ana06: I am not sure of what bootstrap component I should use here. :confused:

app/views/conference/_all_events.html.erb
hennevogel: We are using [.haml](http://haml.info/) as view format. Not .erb

Ana06: okey. I'll change it! :wink:

app/assets/stylesheets/osem-schedule.css.scss
hennevogel: Looks like a job for [panels](http://getbootstrap.com/components/#panels)

Ana06: It looks different with panels as they don't have shadow but it is good too.

app/views/conference/schedule.html.erb
Ana06: Do you mean moving it to a file in the javascript folder?

Ana06:
I've added the start and end time in the boxes, that is how it looks now:

image

app/assets/stylesheets/osem-schedule.css.scss
hennevogel: You might have noticed that nothing has shadows... ;-)

app/controllers/conference_controller.rb
hennevogel: Jupp. Noticed, deleted my comment :-)

app/helpers/application_helper.rb
```diff
@@ -276,4 +276,23 @@ def sign_in_path
   def unread_notifications(user)
     Comment.accessible_by(current_ability).find_since_last_login(user)
   end
+
+  def date_s(datetime)
```
hennevogel: In two months I'm going to have to look up what `date_s` is. I don't have to look up what `strftime` is, I know that. Please do not obfuscate...

app/views/conference/schedule.html.erb
hennevogel: No I mean not rendering everything in `ConferenceController#schedule` and switching between views in `schedule.html.erb` but having an separate method and view like `ConferenceController#all`.

Ana06: okey, but I think we should merge #1030 before, to avoid a lot of merge conflicts.

app/assets/stylesheets/osem-schedule.css.scss
```diff
@@ -214,6 +214,15 @@ td.no-padding{
   display: inline-block;
 }
 
+.title{
```
hennevogel: Why invent your own? http://getbootstrap.com/css/#type-headings

app/assets/stylesheets/osem-schedule.css.scss
```diff
@@ -214,6 +214,15 @@ td.no-padding{
...
+  font-size: 23px;
+}
+
+.subtitle{
```
hennevogel: Why invent your own? http://getbootstrap.com/css/#helper-classes-colors

app/views/conference/_event.html.haml
```diff
@@ -1,32 +1,32 @@
...
-  %div.event-item
-    - if speaker = event.speakers.first
-      = image_tag speaker.gravatar_url, :class => "img-circle pull-right all-speaker-pic", |
-                                        :alt => speaker.name, |
-                                        :title => speaker.name |
-    %h3
-      = event.title
-      %br
-      %small
+  %div.panel.panel-default
```
hennevogel: `%div` is default.

```
.panel == <div class="panel">
```

app/assets/stylesheets/osem-schedule.css.scss
Ana06: Because the `h3` has margin at the top and the `panel` padding too, so it is too much margin. They were `h3` and `small` before but it looks horrible:

image

hennevogel: If you read a little bit further the docu tells you there are also heading classes. So you can do `p.h3` :-)

hennevogel:
This crashes with 

```
app/controllers/conference_controller.rb:41:in `allevents'
NoMethodError (undefined method `start_time' for nil:NilClass):
```

app/assets/stylesheets/osem-schedule.css.scss
Ana06: `%p.h3` looks the same than `%h3`, but `%span.h3` looks great! :smile:

Ana06:
@hennevogel It shouldn't crash now. :relieved:

Ana06:
Apart from grouping the events by date and time, I have moved the all event section to a new controller, I have changed to haml all the views related to the schedule and I have made that both the schedule and the all event section open in the current date and time during the conference.

So, I think there is no everyting else to do in the public schedule. Well, I have to add the individual schedule, but I think that would be better to do it after I have change the relations to allow multiple schedules in the backend.

app/models/event.rb
```diff
@@ -69,11 +69,11 @@ class Event < ActiveRecord::Base
   end
 
   ##
-  # Checkes if the event has a start_time and a room
+  # Checkes if the event is confirmed and has a start_time and a room
   # ====Returns
   # * +true+ or +false+
   def scheduled?
-    room && start_time ? true : false
+    room && start_time && (state == 'confirmed') ? true : false
```
ChrisBr: Do you really need the explicit `` ? true : false`` ?

app/controllers/conference_controller.rb
```diff
@@ -21,10 +25,40 @@ def schedule
     conf_end = 20
     @conf_period = conf_end - @conf_start
 
-    if @dates == Date.current
-      @today = Date.current.strftime('%Y-%m-%d')
+    @day = Time.find_zone(@conference.timezone).today
+    # the schedule takes you to today if it is a date of the schedule
+    if @dates.include? @day
```
ChrisBr: Do we really need to calculate that in the controller?

app/controllers/conference_controller.rb
```diff
@@ -21,10 +25,40 @@ def schedule
...
     else
-      @today = @conference.start_date.strftime('%Y-%m-%d')
+      @day = @dates.first
+    end
+  end
+
+  def allevents
+    @dates = @conference.start_date..@conference.end_date
+
+    @scheduled_events = @conference.program.events.scheduled.order(:start_time)
```
ChrisBr: Hm, maybe we should add the ordering by start_time as default ordering for scheduled events?

app/models/event.rb
Ana06: I think so, because if I eliminate it and for example the `room` is `nil` it will returns `nil` instead of `false`.

I could write that:

`room.present? && start_time.present? && (state == 'confirmed')`

But I don't know if it is an improvement.

app/controllers/conference_controller.rb
```diff
@@ -21,10 +25,40 @@ def schedule
...
+    @unscheduled_events = @conference.program.events.unscheduled
+
+    day = Time.find_zone(@conference.timezone).today
```
ChrisBr: What is the difference between:

```day = Time.find_zone(@conference.timezone).today```

and

```today = DateTime.now.in_time_zone(@conference.timezone)```

File: app/controllers/conference_controller.rb [hunk 27]
```diff
@@ -21,10 +25,40 @@ def schedule
...
+    if @dates.include? day
```
ChrisBr: Do we need to calculate that in the controller?

app/models/program.rb
```diff
@@ -30,6 +30,10 @@ def scheduled
       where.not(start_time: nil).where.not(room: nil)
     end
 
+    def unscheduled
+      confirmed.where('start_time IS NULL OR room_id IS NULL')
```
ChrisBr: Maybe just 
```events - scheduled```

(However, please check if the result is the same!)

File: app/controllers/conference_controller.rb (hunk 26 above)
Comment (Ana06) [inline app/controllers/conference_controller.rb]: The difference is that `today` has the time and `day` doesn't have it.

I think I should change the names of the variables to make it more clear (`day_with_time` instead of `today`) and also `today` is equivalent to that:

`today = Time.find_zone(@conference.timezone).now`

which is similar to how `day` is get so I think I should change that too.

File: app/views/conference/allevents.html.haml [hunk 29]
```diff
@@ -0,0 +1,66 @@
...
+            Unscheduled events
+          %a{ title: "Go up", class: "pull-right", href: "#program" }
+            %i{ class: "fa fa-angle-double-up fa-lg", 'aria-hidden' => true }
+      - @unscheduled_events.each do |event|
+        .col-xs-12.col-md-12
+          .unscheduled-event
+            = render partial: 'event', locals: {event: event}
+
+:javascript
+  $('.program-selector').on('click', function(e) {
```
ChrisBr: Is it necessary to do that with JS? Does ```data-toggle="tab"``` not work?

app/views/conference/schedule.html.haml
```diff
@@ -0,0 +1,81 @@
...
+          .visible-sm-inline
+            = render partial: 'carousel', locals: { date: date, number_columns: 2 }
+
+          .visible-md-inline.visible-lg-inline
+            = render partial: 'carousel', locals: { date: date, number_columns: 3 }
+
+:javascript
+  // change of active tab and the button title when a date is clicked
+  $(function() {
+    $('.date-tab').on('click', function(e) {
```
ChrisBr: Is it necessary to do that with JS? Does ```data-toggle="tab"``` not work?

app/models/program.rb
ChrisBr: Please also take https://github.com/openSUSE/osem/pull/1082 into account! Seems like @nishanthvijayan is also working on sth in this direction!

app/controllers/conference_controller.rb
Ana06: But some times we don't care it they are ordered or not, for example here:

 `@conference.program.events.scheduled.any?`

app/models/program.rb
Ana06: `all - scheduled` works  :smile:

Ana06: But it is not what we want. As I am using unscheduled in the all events to show then, so unscheduled events should be confirmed

app/views/conference/schedule.html.haml
Ana06: I am not sure of where I should add `data-toggle="tab"`

app/controllers/conference_controller.rb
Ana06: where do you want to move it?

app/controllers/conference_controller.rb
```diff
@@ -21,10 +25,40 @@ def schedule
     conf_end = 20
     @conf_period = conf_end - @conf_start
 
-    if @dates == Date.current
-      @today = Date.current.strftime('%Y-%m-%d')
+    @day = Time.find_zone(@conference.timezone).today
+    # the schedule takes you to today if it is a date of the schedule
+    if @dates.include? @day
+      @has_date = true
+      current_hour = DateTime.now.in_time_zone(@conference.timezone).strftime('%H').to_i
```
hennevogel: Classical Over-engineering. Don't make it more complicated then it needs to be.

app/controllers/conference_controller.rb
hennevogel: If we don't care it doesn't matter if they are sorted or not right? So sort them... :-)

app/helpers/application_helper.rb
```diff
@@ -335,4 +335,25 @@ def speaker_width(rooms)
...
     speaker_height(rooms) - 4
   end
+
+  ##
+  # Checks if there is any event in the collection events that starts in the given date
+  #
+  # ====Returns
+  # * +True+ -> If there is any event for the given date
+  # * +False+ -> If there is not any event for the given date
+  def any_event_for_this_date?(events, date)
```
hennevogel: [Range conditions](http://guides.rubyonrails.org/active_record_querying.html#range-conditions) are usually done differently

app/models/event.rb
hennevogel: It is an improvement

hennevogel:
A not logged in user is not allowed to visit `/allevents`. You need to adopt the [ability](https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L37)

app/helpers/application_helper.rb
Ana06: Because I don't like to add it in the view.

app/controllers/conference_controller.rb
Ana06: If the conference has finished and someone enter to see the schedule I think that we should render the next day.

I have to consider that case even if I do other thing. :confused:

app/views/conference/allevents.html.haml
Ana06: I am not sure of where I should add `data-toggle="tab"`

Ana06:
I've made all the changes you have told me expect few of them where I have added a comment. 

I've also reorder and squash the commits because I think it was quite confusing. I think it is better now. :smile:

app/controllers/conference_controller.rb
hennevogel: > I have to consider that case even if I do other thing.

Why? Don't introduce too much magic, displaying the current day first is enough.

app/helpers/application_helper.rb
hennevogel: I do know why helpers exist, thank you ;-) Why is this not a `Program` instance method?

differentreality:
@hennevogel why should we not allow not logged in users to see /allevents? We allow them to see the schedule, and the whole idea behind /allevents was to show them all accepted talks prior to scheduling them.

app/controllers/conference_controller.rb
Ana06: But I think that when you are in a conference you want to see what is taking place now or what is going to take place and you usually use your mobile phone for that. That means that if it is 20:00 you will take the user to 9:00 and he would have to click next 11 times. I really think that it is useful.

Ana06:
@differentreality what @hennevogel meant was exactly what you are saying, that not logged users should be able to see /all events. I had forgotten to add it in the ability file, but I have already fixed it.

differentreality:
Ah right :) Sorry..

app/models/program.rb
```diff
@@ -87,6 +91,17 @@ def languages_list
...
 
+  ##
+  # Checks if there is any event in the program that starts in the given date
+  #
+  # ====Returns
+  # * +True+ -> If there is any event for the given date
+  # * +False+ -> If there is not any event for the given date
+  def any_event_for_this_date?(date)
+    parsed_date = DateTime.parse("#{date} 00:00").utc
+    events.where(start_time: parsed_date..(parsed_date + 1)).any?
```
ChrisBr: You know that ``Event#start_time`` could be in a different timezone than UTC?

app/views/conference/allevents.html.haml
ChrisBr: getbootstrap.com/javascript/#markup

config/routes.rb
```diff
@@ -116,6 +116,10 @@
     member do
       get :schedule
     end
+
+    member do
+      get :allevents
```
ChrisBr: To be honest, I don't like the name allevents ... 
Do we already have a route ``/conference/<slug>/events`` ?
Any other suggestion?

app/models/program.rb
Ana06: Can they? The `start_time` is given in the admin schedule in the backend and all event are saved without timezone, that is with the default one which is UTC. :confused:

app/views/conference/allevents.html.haml
Ana06: But that is for tabs and pill and these is a dropdown. I think I can not use it in a dropdown, at least I've tried it and I didn't manage to make it works.

config/routes.rb
Ana06: I am not sure of what you mean with `conference/<slug>/events`.

In some places the all events is called program. Maybe program is a better name.

differentreality: By slug he means the short_title of a conference.
And no, we don't have that route, because all routes for events go through program.
(Exception:  api_v1_conference_events_path	GET	/api/v1/conferences/:conference_id/events(.:format)	api/v1/events#index {:format=>"json"} )

app/models/program.rb
ChrisBr: From my perspective, the event timezone is the same timezone as the conference timezonve.
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L86

config/routes.rb
ChrisBr: @Ana06 https://en.wikipedia.org/wiki/Semantic_URL#Slug

Anyway, I still don't like the URL ```allevents```

app/controllers/conference_controller.rb
ChrisBr: Model function with a good name?

These two lines are really hard to understand:

```
current_hour = Time.find_zone(@conference.timezone).now.strftime('%H').to_i
@hour_column = (current_hour > @conf_start && current_hour < conf_end) ? current_hour - @conf_start : 0
```

app/models/program.rb
Ana06: But we  are always saving them as UTC without taking into account timezone:
https://github.com/openSUSE/osem/blob/master/public/javascripts/schedule/schedule.js#L53

I think that this is the correct way to do it, as saving everything related to the schedule in UTC we can forget about time zone changes. The only thing for what we have to take it into account is for opening the schedule in the current day and time, as it may be different to UTC.

If we are always going to save event in UTC maybe we should validate that the events timezone is UTC.

ChrisBr: Ok

> If we are always going to save event in UTC maybe we should validate that the events timezone is UTC.

What do you mean? How would you validate it as events have no timezone?

app/views/conference/schedule.html.haml
ChrisBr: What about this fiddle:
http://jsfiddle.net/KyleMit/pp9e2hkd/0/
?

app/models/program.rb
Ana06: Something like:

``` Ruby
validate :start_time_utc

def start_time_utc
  errors.add(:start_time, 'must be in UTC time zone') unless (start_time.strftime('%z') == '+0000')
end
```

That way we can use that events are saving in UTC safely. :bowtie:

app/views/conference/schedule.html.haml
Ana06: But that is for nav tabs and I am using a dropdown, so I think I can not use it :confused:

config/routes.rb
```diff
@@ -116,6 +116,10 @@
     member do
       get :schedule
     end
+
+    member do
+      get :events
```
differentreality: Any reason why this is in a new `member do` block and not added on the one above?

Ana06:
Is there anything else to change here?

app/controllers/conference_controller.rb
```diff
@@ -13,19 +13,34 @@ def show; end
 
   def schedule
     @rooms = @conference.venue.rooms if @conference.venue
+    unless @conference.program.events.scheduled.any?
+      redirect_to events_conference_path(@conference.short_title)
+    end
+
     @events = @conference.program.events
-    @events_xml = @events.scheduled.order(start_time: :asc).group_by{ |event| event.start_time.to_date }
+    @events_xml = @events.scheduled.group_by{ |event| event.start_time.to_date }
```
differentreality: I'd rather we did not change that. I saw you included it in the definition of scheduled, but if for some reason that definition changes in the future, the XML export will be buggy. For the XML export the order is of high importance.

app/models/program.rb
```diff
@@ -27,7 +27,11 @@ def confirmed
     end
 
     def scheduled
-      where.not(start_time: nil).where.not(room: nil)
+      where.not(start_time: nil).where.not(room: nil).order(:start_time)
```
differentreality: I would make the ordering here explicit. Eg. `.order(start_time: :asc)`

hennevogel:
Okay let's go :-)

Ana06:
:tada:

Issue #973, opened 2016-04-29
Statistics on top of Events#index & Registrations#index


hennevogel:
In the admin interface both are basically 2 huge lists containing all submitted events and all registered people. We should show some statistics on top like:

![dashboard](https://cloud.githubusercontent.com/assets/514785/14912456/b973d790-0dfc-11e6-919b-baf99fae31f1.png)

For Events it could also include this (from the general dashboard)

![screenshot from 2016-04-29 11-23-47](https://cloud.githubusercontent.com/assets/514785/14912489/e8e32b48-0dfc-11e6-8254-8ab7a4e8a0e5.png)

Ana06:
I'll do it. 
The statistic on the top are the same as the ones in the dashboard, aren't they?
The event types, difficulty levels and tracks should be also at the top?

hennevogel:
For Registrations I would say those 3 are important:

* Number of overall registrations
* Registration attended vs. Registration not attended
* Distribution of affiliation (if that is possible)

For Events I would say those 3 are important:

* Confirmed vs. Unconfirmed Events
* Scheduled (start_time set) vs. Unscheduled events
* Track Distribution

Ana06:
This issue can be closed :smile:

PR #1034, opened 2016-06-11
Statistics

Ana06:
Statistics on top of Events#index. Issue #973.

I added:

- Events state
- Scheduled (start_time set) vs. Unscheduled events of the confirmed events.
- Track Distribution of the confirmed events.

image

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -325,9 +325,56 @@ class Conference < ActiveRecord::Base
   # * +hash+ -> hash
   def scheduled_event_distribution
     confirmed_events = program.events.where(state: 'confirmed')
-    scheduled_value =  { 'value' => confirmed_events.where('start_time IS NOT NULL').count, 'color' => 'green' }
-    unscheduled_value =  { 'value' => confirmed_events.where('start_time IS NULL').count, 'color' => 'red' }
+    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
+    unscheduled_value =  { 'value' => confirmed_events.where(start_time: nil).count, 'color' => 'red' }
     { 'Scheduled' => scheduled_value, 'Unscheduled' => unscheduled_value }
+  end
+
+  ##
+  # Returns a hash with Registration attended vs. Registration not attended
+  # { "Attended" => { value: number of registration attended, color: color },
+  #   "Not attended" => { value: number of registration not attended, color: color }
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def registration_distribution
+    reg = registrations.includes(:user)
+    attended_value =  { 'value' => reg.where(attended: true).count, 'color' => 'magenta' }
+    not_attended_value =  { 'value' => reg.where.not(attended: true).count, 'color' => 'blue' }
+    { 'Attended' => attended_value, 'Not attended' => not_attended_value }
+  end
+
+  ##
+  # Returns a hash with affiliation =>
+  # {value: count of registration whose user has that affilation, color: color}
+  # In case that the affiliation is blank, it groups them in None and
+  # if the number of persons that have an affiliation are less than the 2% of
+  # the total number of registered people, they are grouped in others.
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def affiliation_distribution
+    counted_affiliations = registrations.joins(:user).group(:affiliation).count
+    result = {}
+    i=1
+    others = 0
+    none = 0
+    counted_affiliations.each do |key, value|
+      if value < 0.02 * registrations.length
+        others += value
+      elsif key.blank?
+        none += value
+      else
+        result[key.capitalize] = { 'value' => value, 'color' => next_color(i) }
+        i += 1
+      end
+    end
+    if others > 0
+      result['Others'] = { 'value' => others, 'color' => next_color(i) }
+      i += 1
+    end
+    result['None'] = { 'value' => none, 'color' => next_color(i) } if none > 0
+    result
   end
 
   ##
@@ -549,6 +596,16 @@ class Conference < ActiveRecord::Base
   end
 
   private
+
+  # Returns a different html colour for every i. We make use of big prime numbers
+  # to avoid repetition and to make consecutive colors clearly different.
+  def next_color(i)
+    color = '#000000'
+    color[1..2] = ((i*113)%239 + 16).to_s(16)
+    color[3..4] = ((i*67)%239 + 16).to_s(16)
+    color[5..6] = ((i*151)%239 + 16).to_s(16)
+    color
+  end
 
   after_create do
     self.create_contact
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -317,6 +317,20 @@ class Conference < ActiveRecord::Base
   end
 
   ##
+  # Returns a hash with scheduled vs unscheduled events
+  # { "Scheduled" => { value: number of confirmed and scheduled events, color: color },
+  #   "Unscheduled" => { value: number of confirmed and unscheduled events, color: color }
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def scheduled_event_distribution
+    confirmed_events = program.events.where(state: 'confirmed')
+    scheduled_value =  { 'value' => confirmed_events.where('start_time IS NOT NULL').count, 'color' => 'green' }
+    unscheduled_value =  { 'value' => confirmed_events.where('start_time IS NULL').count, 'color' => 'red' }
+    { 'Scheduled' => scheduled_value, 'Unscheduled' => unscheduled_value }
+  end
+
+  ##
   # Returns a hash with user distribution => {value: count of user state, color: color}
   # active: signed in during the last 3 months
   # unconfirmed: registered but not confirmed
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -597,14 +597,18 @@ class Conference < ActiveRecord::Base
 
   private
 
-  # Returns a different html colour for every i. We make use of big prime numbers
-  # to avoid repetition and to make consecutive colors clearly different.
+  # Returns a different html colour for every i and consecutive colors are
+  # clearly different.
   def next_color(i)
-    color = '#000000'
-    color[1..2] = ((i*113)%239 + 16).to_s(16)
-    color[3..4] = ((i*67)%239 + 16).to_s(16)
-    color[5..6] = ((i*151)%239 + 16).to_s(16)
-    color
+    '#' + next_color_component(:r, i) + next_color_component(:g, i) + next_color_component(:b, i)
+  end
+
+  # Auxiliar function which is used in next_color and returns each component of
+  # the color. We make use of big prime numbers to avoid repetition and to make
+  # consecutive colors clearly different.
+  def next_color_component(component, i)
+    big_prime_numbers = {r: 113, g: 67, b: 151}
+    ((i*big_prime_numbers[component])%239 + 16).to_s(16)
   end
 
   after_create do
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -14,161 +14,162 @@
     = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Confirmed events scheduled', data: @scheduled_event_distribution}
   .col-md-4
     = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Tracks of confirmed events', data: @tracks_distribution_confirmed}
-.row{ style: "margin-top: 40px;"}
+.row
   .col-md-12
-    %table.table.table-striped.table-bordered.table-hover.datatable
-      %thead
-        %th
-          %b ID
-        %th
-          %b Title
-        - if @program.rating_enabled?
+    %div.margin-event-table
+      %table.table.table-striped.table-bordered.table-hover.datatable
+        %thead
           %th
-            %b Rating
-        %th
-          %b Submitter
-        %th
-          %b Speaker
-        -if @program.languages.present?
+            %b ID
           %th
-            %b Language
-        %th
-          %b Requires Registration
-        %th
-          %b Highlight
-        %th
-          %b Type
-        %th
-          %b Track
-        %th
-          %b Difficulty
-        %th
-          %b State
-      - @events.each do |event|
-        %tr
-          %td
-            = event.id
-          %td
-            =link_to event.title, admin_conference_program_event_path(@conference.short_title, event)
-
+            %b Title
           - if @program.rating_enabled?
-            %td.col-md-1{'data-order' => "#{event.average_rating}"}
-              - if event.average_rating.to_f > 0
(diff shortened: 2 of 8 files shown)


Ana06:
I've just added 	statistics on top of Registrations#index (also in issue #973):

- Number of registrations (including the new ones since you last logged in)
- Registration attended vs. Registration not attended
- Distribution of affiliation

image

For the affiliation, as it has no color, I created a function to generate different colours that are different specially for consecutive ones and that are always the same ones (to avoid confusion if they change every time the page is reloaded). I think it works really well:

image

I show _None_, for the ones that have no affiliation, but maybe we want that the ones that do not have affiliation to not appear in the statistics to have more space for the rest.

Also, the registration, submissions and hours statistic in the dashboard are very similar. Maybe we could move it to a partial. I am using the first one so it would use the partial to avoid copying the code as I have done now. I am not sure if I should do it in this PR, as maybe it will make more difficult to review the rest.

I have a doubt. Why this two things don't do the same?

` reg.where.not(attended: true)`
 `reg.where(attended: false)`

The first one works as I expected, but the second doesn't and I don't understand why.

ChrisBr:
@Ana06 

> For the affiliation, as it has no color, I created a function to generate different colours that are different specially for consecutive ones and that are always the same ones (to avoid confusion if they change every time the page is reloaded). 

How do you make sure that it's always the same color?

ChrisBr:
> I show None, for the ones that have no affiliation, but maybe we want that the ones that do not have affiliation to not appear in the statistics to have more space for the rest.

I'd include it for now! We will see how useful it is in production and can adapt if necessary.

ChrisBr:
> I have a doubt. Why this two things don't do the same?

reg.where.not(attended: true)
reg.where(attended: false)

What I do in this case is to try it in the rails console (```rails c```). While executing the statements, rails will show you the generated and executed SQL statements. They're probably not the same as you get different results. Maybe this will already give you some insights to understand what's going wrong. If now, please paste the output of the rails console and we will try to debug with you.

ChrisBr:
> Registration attended vs. Registration not attended

This stastic makes only sense to display AFTER (maybe after the conference started) the conference...

app/models/conference.rb
```diff
@@ -317,6 +317,60 @@ def event_distribution
...
   ##
+  # Returns a hash with scheduled vs unscheduled events
+  # { "Scheduled" => { value: number of confirmed and scheduled events, color: color },
+  #   "Unscheduled" => { value: number of confirmed and unscheduled events, color: color }
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def scheduled_event_distribution
+    confirmed_events = program.events.where(state: 'confirmed')
+    scheduled_value =  { 'value' => confirmed_events.where('start_time IS NOT NULL').count, 'color' => 'green' }
```
ChrisBr: I'd prefer to use here also ``where.not``

```
confirmed_events.where.not(start_time: nil)
```

app/models/conference.rb
```diff
@@ -317,6 +317,60 @@ def event_distribution
...
+    unscheduled_value =  { 'value' => confirmed_events.where('start_time IS NULL').count, 'color' => 'red' }
```
ChrisBr: Same here!

app/models/conference.rb
hennevogel: scope?

app/models/conference.rb
```diff
@@ -317,6 +317,60 @@ def event_distribution
...
+  # In case that the affiliation is nil, it groups it in not_attended_value
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def affiliation_distribution
+    affiliations = []
+    registrations.each do |r|
+      affiliations << (r.user.affiliation ? r.user.affiliation : 'None')
+    end
+    counted_affiliations = affiliations.group_by(&:capitalize).map {|k, v| [k, v.length]}
```
ChrisBr: Hm, I'd prefer to do this in some ActiveRecord grouping? Because using database features is way faster than doing it in Ruby code.

Ana06:
@ChrisBr 

> How do you make sure that it's always the same color?

Because they are not aleatory, that is why I needed a function that given  `i >= 1` gives you a different color for every `i` (**but always the same one for the same `i`**), and consecutive/close values of `i` gives highly different colors. The function name is `next_color`, I've used big prime numbers. II printed the first generated colors to show you that they are different and also because depending on the primes colors chosen the colours are more or less beautiful, I like that combination.

Ana06:
@ChrisBr 

> What I do in this case is to try it in the rails console (rails c). While executing the statements, rails will show you the generated and executed SQL statements. They're probably not the same as you get different results. Maybe this will already give you some insights to understand what's going wrong. If now, please paste the output of the rails console and we will try to debug with you.

They are not the same, but I think they should be equivalent:

image

The only logical explanation would be that the attended value of the user that apear in the fist query but not in the second was `nil`, but it is not, it is `false`:

image

Ana06:
I'll change the _Registration attended vs. Registration not attended_ to be displayed only after the conference has started. I'll make the changes you added in the code too.

app/views/admin/registrations/index.html.haml
```diff
@@ -3,14 +3,34 @@
     .page-header
       %h1
         Registrations
-        = "(#{@registrations.length})" if @registrations
+        = "(#{@total_reg})" if @registrations
```
differentreality: Would be awesome if that variable name wasn't abbreviated, so that I don't have to think twice when I am reading it.
`@total_reg`  also just holds `@registrations.count`, so do we really need to define a new variable for it?

Either way, since we have `@registrations` already loaded, we can use `.length` instead of `.count`, which triggers a new count query on our DB.

app/models/conference.rb
```diff
@@ -317,6 +317,57 @@ def event_distribution
...
+  # In case that the affiliation is nil, it groups it in not_attended_value
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def affiliation_distribution
+    counted_affiliations = registrations.joins(:user).group(:affiliation).count
+    result = {}
+    i=1
+    counted_affiliations.each do |key, value|
+      capitalized_key = (key.nil? ? 'None' : key.capitalize)
```
differentreality: Affiliation, except for nil, might also just be an empty string. We still want to show 'None' for that. Perhaps check for `.blank?`

ChrisBr:
> Because they are not random, that is why I needed a function that given i >= 1 gives you a different color for every i (but always the same one for the same i), and consecutive/close values of i gives highly different colors. The function name is next_color, I've used big prime numbers. II printed the first generated colors to show you that they are different and also because depending on the primes colors chosen the colours are more or less beautiful, I like that combination.

> The colors could change if there are changes in the affiliation, but what I meant was that without having changes in the affiliation if you reload the page they'll be the sames (otherwise it would be confusing). They would have changed if they were random and also we may probably have different colors but maybe ugly ones.

-> That's what I meant. You can't be sure that the affiliation has always the same i. As long as you have thought about that issue it should be fine.

However, what we should maybe consider is grouping affiliations with less than a certain number together to other (e.g. less than 5 people are in that affiliation). Otherwise we will have A LOT affiliations only with one person which will make the pie diagram not very useful.

ChrisBr:
> I have a doubt. Why this two things don't do the same?
> reg.where.not(attended: true)
> reg.where(attended: false)

@Ana06 played a little bit around with that and it's indeed true that ``reg.where(attended: false)`` does not return a value. However, if you explicitely set the value to false

```
r = reg.where.not(attended: true).first
r.attended = false
r.save
```
and execute the query again, it works. Also if you select ``reg.where(attended: 0)`` if works. From the SQLite docu

> SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true)

So I assume, something goes wrong while setting the default value of attended.

app/models/conference.rb
ChrisBr: Isn't the nil case anyway covered by you query?

app/views/admin/registrations/index.html.haml
```diff
@@ -3,14 +3,34 @@
...
+      %p
+        %small
+          #{'Registration'.pluralize(@total_reg)}
+        - if @new_reg
+          %span.label.label-success{title: "+#{@new_reg} since you last logged in!"}
+            +
+            = @new_reg
+  .col-md-4
+    = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Affiliation', data: @affiliation_distribution}
+  = unless @conference.pending?
```
ChrisBr: Shouldn't that be:
```
- unless @conference.pending?
```
?

Ana06: True, but it also works with `=`, because I tried it. Anyway, I'll change it.

app/models/conference.rb
Ana06: The problem is that empty string  is grouped separated from nil. I can group them together

Ana06:
I've included the empty string in None, I've also group those affiliations that are less than the 2% of the registrations in others.

There is a problem, rubocop doesn't like that the Conference class has more that 500 lines.

differentreality:
Of course it does :)
Should we raise its limit again?

Ana06:
@ChrisBr 

> However, if you explicitely set the value to false and execute the query again, it works. 

Are you sure that it works? Look:

image

It's true that `reg.where(attended: 0)` works. Another reason why `reg.where(attended: false)` should work.

ChrisBr:
> Are you sure that it works? Look:
> 
> It's true that reg.where(attended: 0) works. Another reason why reg.where(attended: false) should work.

@Ana06 see this answer on stackoverflow: 

http://stackoverflow.com/questions/21066345/regression-in-activerecord-4-sqlite3-boolean-query-support

hennevogel:
@ChrisBr 
> This stastic makes only sense to display AFTER (maybe after the conference started) the conference...

I also want to know how I'm doing while the conference runs IMHO

app/models/conference.rb
```diff
@@ -536,6 +597,16 @@ def registration_limit_exceeded?
 
   private
 
+  # Returns a different html colour for every i. We make use of big prime numbers
+  # to avoid repetition and to make consecutive colors clearly different.
```
hennevogel: You math studies are showing through ;-)

ChrisBr:
> >This stastic makes only sense to display AFTER (maybe after the conference started) the conference...

> I also want to know how I'm doing while the conference runs IMHO

That's why I wrote after the conference started!

app/views/admin/events/index.html.haml
```diff
@@ -8,6 +8,13 @@
       %p.text-muted
         All the submissions of your speakers
 .row
+  .col-md-4
+    = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Events state', data: @event_distribution}
+  .col-md-4
+    = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Confirmed events scheduled', data: @scheduled_event_distribution}
+  .col-md-4
+    = render partial: 'admin/conference/doughnut_chart', locals: {title: 'Tracks of confirmed events', data: @tracks_distribution_confirmed}
+.row{ style: "margin-top: 40px;"}
```
hennevogel: Do not manipulate the box-model of bootstrap grid components.

Ana06:
@hennevogel 

> I also want to know how I'm doing while the conference runs IMHO

Do you mean when the conference has started? It is shown if the start day of the conference is earlier or equal than today.

app/models/conference.rb
```diff
@@ -317,6 +317,67 @@ def event_distribution
...
   ##
+  # Returns a hash with scheduled vs unscheduled events
+  # { "Scheduled" => { value: number of confirmed and scheduled events, color: color },
+  #   "Unscheduled" => { value: number of confirmed and unscheduled events, color: color }
+  #
+  # ====Returns
+  # * +hash+ -> hash
+  def scheduled_event_distribution
+    confirmed_events = program.events.where(state: 'confirmed')
+    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
```
hennevogel: I don't like that it includes the color. This has clearly nothing to do with the model...

app/views/admin/registrations/index.html.haml
```diff
@@ -11,6 +11,26 @@
             = link_to 'Export XLS', {format: :xlsx}, class: 'btn btn-success'
       %p.text-muted
         All the people who registered to your event
+.row
+  .col-sm-4.col-xs-4
+    .dashbox.text-center
```
hennevogel: Please do not mix `dashbox` and `doughnut_chart`. It looks horrible

app/models/conference.rb
Ana06: and where do I put the color?

hennevogel: Where you render this data. In the view :-)

app/views/admin/registrations/index.html.haml
Ana06: But I think it has no sense to show the registration with a `doughnut_chart`. Maybe we can remove the number of registrations from the statistics, as it is in the title.

app/models/conference.rb
Ana06: The rest doughnut charts set the color in the model too and doing it in the view would be complicated.

Ana06:
I eliminated the registrations dashbox, that is how it looks now:

image

app/models/conference.rb
hennevogel: OK

differentreality:
What is left to do here? Except of course for increasing the lines number for Conference class.

Would be very interesting to see the registration statistics any time now for oSC16.

Ana06:
I think it is finished.

app/models/conference.rb
```diff
@@ -536,6 +597,16 @@ def registration_limit_exceeded?
...
+  # Returns a different html colour for every i. We make use of big prime numbers
+  # to avoid repetition and to make consecutive colors clearly different.
+  def next_color(i)
+    color = '#000000'
+    color[1..2] = ((i*113)%239 + 16).to_s(16)
+    color[3..4] = ((i*67)%239 + 16).to_s(16)
+    color[5..6] = ((i*151)%239 + 16).to_s(16)
+    color
+  end
+
```
mdeniz: Can be that way? For non math people :smile: 

```ruby
def next_color(i)
  "#" 
   + big_prime_color_component(:r, i) 
   + big_prime_color_component(:g, i) 
   + big_prime_color_component(:b, i)
end

def big_prime_color_component(component, i)
  big_prime_numbers = {r: 113, g: 67, b: 151}
  ((i*big_prime_numbers[component.to_sym])%239 + 16).to_s(16)
end
```

Ana06:
Following @mdeniz  suggestion I made the next_color function more readable. Also I think I hadn't push the _Registrations dashbox eliminated from registrations_ commit, although I had added the screenshot. I've just added it.

hennevogel:
giphy

differentreality:
Let's add 10 extra lines for conference class so that rubocop doesn't complain.

Ana06:
Rubocop doesn't complain now :relieved:

hennevogel:
Okay let's get this in and let's think about some other stats we can show in #973

Issue #733, opened 2015-11-12
Support FRAB XML output


danimo:
It would be nice if there was a way to export a conference to FRAB (http://frab.github.io/frab/) XML. The rationale is that a lot of conference recording tools developed by c3voc.de which are actively used consume FRAB output. This allows for same day publishing & streaming, e.g. https://github.com/voc/intro-outro-generator.

Examples can be found here:

http://programm.froscon.de/2015/schedule.xml
https://events.ccc.de/congress/2014/Fahrplan/schedule.xml

The XSD is located at https://raw.githubusercontent.com/voc/schedule/master/validator/xsd/schedule.xml.xsd

Online-validator (also accepts URLs): https://c3voc.de/schedulexml/

ChrisBr:
@danimo nice idea! Maybe we can include that into #741

jospoortvliet:
just out of curiosity, was this done? I saw in #741 "transferring" but not sure what that means...

differentreality:
Nop, it was not done. Yet..

differentreality:
@ChrisBr @hennevogel are you going to include this in Ana's GSOC project?

hennevogel:
Nope. I'm not even convinced that this is a good feature in general

afranke:
If it helps convincing you, note that many mobile conference apps consumes this format, which gives you schedule apps for your event for free.

PR #1035, opened 2016-06-13
Initial XML export support (FRAB compatible)

danimo:
https://github.com/openSUSE/osem/issues/733
This enables basic support for FRAB XML exports, which can be read by numerous conference planning tools for Android and iOS, and by the C3VOC toolchain.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -223,6 +223,12 @@ class Event < ActiveRecord::Base
     (100 * result.values.count(true) / result.values.compact.count).to_s
   end
 
+  ##
+  # Returns the slug of an event
+  def slugify
+    "#{self.id}-#{self.title.downcase.gsub(/[^a-z1-9]+/, '-').chomp('-')}"
+  end
+
   private
 
   ##
--- a/app/views/conference/_all_days.xml.haml
+++ b/app/views/conference/_all_days.xml.haml
@@ -1,0 +1,3 @@
+- @conference.program.events.group_by{ |event| event.start_time.to_date }.each_with_index do | day, index |
+  %day{ :date => day[0], :index => index+1 }
+    =render 'all_rooms', :events => day[1]
--- a/app/views/conference/_event.xml.haml
+++ b/app/views/conference/_event.xml.haml
@@ -1,0 +1,20 @@
+- event.public?
+  %event{ :guid => event.guid, :id => event.id }
+    %date=event.start_time.iso8601
+    %start=event.start_time.strftime("%H:%M")
+    %duration=event.event_type.length_timestamp
+    %room=event.room.name
+    %type=event.event_type.name
+    %language="en" ## fix me, add language property to Event/Conference?
+    %slug=event.slugify
+    %title=event.title
+    %subtitle=event.subtitle
+    %track=event.track.name
+    %abstract=event.abstract
+    %recording
+      %license/
+      %optout=false #fixme
+    %persons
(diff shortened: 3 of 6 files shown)


danimo:
todo items:

* Dummy entries: Did I miss their representation in Frab or do they not exist?
  *  If they not exist, do should they be added in subsequent requests?

* Error handling: If only one talk misses certain properties, the export will explode. More error checking is due.

app/models/event.rb
```diff
@@ -223,6 +223,12 @@ def calculate_progress
     (100 * result.values.count(true) / result.values.compact.count).to_s
   end
 
+  ##
+  # Returns the slug of an event
+  def slugify
+    "#{self.id}-#{self.title.downcase.gsub(/[^a-z1-9]+/, '-').chomp('-')}"
```
danimo: Maybe better use an existing gem for this? Any ideas?

app/views/conference/_event.xml.haml
```diff
@@ -0,0 +1,20 @@
+- event.public?
+  %event{ :guid => event.guid, :id => event.id }
+    %date=event.start_time.iso8601
+    %start=event.start_time.strftime("%H:%M")
+    %duration=event.event_type.length_timestamp
+    %room=event.room.name
+    %type=event.event_type.name
+    %language="en" ## fix me, add language property to Event/Conference?
```
danimo: Can we pull this information from somewhere?

app/views/conference/_event.xml.haml
```diff
@@ -0,0 +1,20 @@
...
+    %slug=event.slugify
+    %title=event.title
+    %subtitle=event.subtitle
+    %track=event.track.name
+    %abstract=event.abstract
+    %recording
```
danimo: Recording metadata would probably require implementation in OSEM.

app/views/conference/_event.xml.haml
differentreality: Very soon, yes. https://github.com/openSUSE/osem/pull/905

app/views/conference/_event.xml.haml
differentreality: What recording metadata are we talking about?

danimo: Currently only license (could inherit from the project, but sometimes it makes sense that individual talks have different licenses) and opt-out (i.e. when the speaker does not want to be recorded).

app/views/conference/_all_days.xml.haml
```diff
@@ -0,0 +1,3 @@
+- @conference.program.events.group_by{ |event| event.start_time.to_date }.each_with_index do | day, index |
```
differentreality: `@conference.program.events.scheduled` assuming that we only want to show scheduled events in here.

app/views/conference/_event.xml.haml
```diff
@@ -0,0 +1,20 @@
...
+    %date=event.start_time.iso8601
+    %start=event.start_time.strftime("%H:%M")
+    %duration=event.event_type.length_timestamp
+    %room=event.room.name
+    %type=event.event_type.name
+    %language="en" ## fix me, add language property to Event/Conference?
+    %slug=event.slugify
+    %title=event.title
+    %subtitle=event.subtitle
+    %track=event.track.name
```
differentreality: It might not have a track, and that will fail.

danimo:
Closing in favor of #1038.

PR #1038, opened 2016-06-13
XML export

### LINKED ISSUE #733: Support FRAB XML output

differentreality:
Small rework of #1035 
Fixes #733

### DIFF
--- a/app/controllers/conference_controller.rb
+++ b/app/controllers/conference_controller.rb
@@ -14,6 +14,7 @@ class ConferenceController < ApplicationController
   def schedule
     @rooms = @conference.venue.rooms if @conference.venue
     @events = @conference.program.events
+    @events_xml = @conference.program.events.scheduled.group_by{ |event| event.start_time.to_date }
     @dates = @conference.start_date..@conference.end_date
 
     if @dates == Date.current
--- a/app/controllers/conference_controller.rb
+++ b/app/controllers/conference_controller.rb
@@ -14,7 +14,7 @@ class ConferenceController < ApplicationController
   def schedule
     @rooms = @conference.venue.rooms if @conference.venue
     @events = @conference.program.events
-    @events_xml = @conference.program.events.scheduled.group_by{ |event| event.start_time.to_date }
+    @events_xml = @events.scheduled.group_by{ |event| event.start_time.to_date }
     @dates = @conference.start_date..@conference.end_date
 
     if @dates == Date.current
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -12,6 +12,13 @@ class EventType < ActiveRecord::Base
 
   LENGTH_STEP = 15
 
+  ##
+  # Return the length in timestamp format (HH:MM)
+  #
+  def length_timestamp
+    [length / 60, length % 60].map { |t| t.to_s.rjust(2, '0') }.join(':')
+  end
+
   private
 
   ##
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -12,13 +12,6 @@ class EventType < ActiveRecord::Base
 
   LENGTH_STEP = 15
 
-  ##
-  # Return the length in timestamp format (HH:MM)
-  #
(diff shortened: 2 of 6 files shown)


danimo:
LGTM. Thanks for the improvements.

app/models/event_type.rb
```diff
@@ -12,6 +12,13 @@ class EventType < ActiveRecord::Base
 
   LENGTH_STEP = 15
 
+  ##
+  # Return the length in timestamp format (HH:MM)
+  #
+  def length_timestamp
```
hennevogel: This is a helper, not a model method.

app/controllers/conference_controller.rb
```diff
@@ -14,6 +14,7 @@ def show; end
   def schedule
     @rooms = @conference.venue.rooms if @conference.venue
     @events = @conference.program.events
+    @events_xml = @conference.program.events.scheduled.group_by{ |event| event.start_time.to_date }
```
hennevogel: This will probably be cached by the query caching but why do you load everything twice?

app/views/conference/schedule.xml.haml
```diff
@@ -0,0 +1,35 @@
+%schedule
+  %version= @conference.revision
+  %conference
+    %acronym= @conference.short_title
+    %title= @conference.title
+    %start= @conference.start_date
+    %end= @conference.end_date
+    %days= (@conference.end_date - @conference.start_date).to_i + 1
+    %timeslot_duration= '00:15' #FIXME
```
hennevogel: FIXME why?

danimo: this shouldn't be hard coded. I've seen it in other places as well, so if it's not supposed to be configurable, it should probably be defined as a constant somewhere.

hennevogel: Please let people know in the comment :-)

danimo: Fair enough :)

hennevogel:
Can you add a test that this renders?

danimo:
Perfect. I've test deployed this here: https://conference.owncloud.org/conference/oCC2015/schedule.xml

It validates using https://c3voc.de/schedulexml/ and imports in the C3VOC tracker.
See https://events.ccc.de/congress/2014/wiki/Apps for apps that can deal with the format.

differentreality:
I ll squash the commits after review is over.

spec/controllers/conference_controller_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe ConferenceController do
-  let(:conference) { create(:conference,  splashpage: create(:splashpage, public: true)) }
+  let(:conference) { create(:full_conference,  splashpage: create(:splashpage, public: true)) }
```
hennevogel: I know this seems convenient but do you really need all of this to test the schedule? I doubt it..

spec/controllers/conference_controller_spec.rb
```diff
@@ -24,6 +24,29 @@
     end
   end
 
+  describe 'GET #schedule' do
+    context 'XML' do
+      before :each do
+        conference.program.schedule_public = true
```
hennevogel: Please do this with the factory instead

hennevogel:
@differentreality please send another PR that adopts the tests. I'll merge this to be able to deploy it before oSC16...

PR #1039, opened 2016-06-14
Bundle update

hennevogel:
Supersedes #1037


differentreality:
Hakiri issue still not fixed.

#1037 introduced nokogiri version 1.6.8

Do you think we should leave rubocop update for another PR (possibly along with more enabled cops)?

hennevogel:
Yeah this needs more work obviously

PR #1041, opened 2016-06-15
Add languages_list in controller admin/events#edit

differentreality:
The variable exists in proposal#edit and proposal#new but not in admin/events#edit, as a result when editing an event from the admin panel, the drop down does not actually show the available languages.


hennevogel:
LGTM

PR #1042, opened 2016-06-18
Online payment feature - integrate active merchant

rishabhs95:
an overview of feature is as follows :
- Purchase Tickets : 
start
- Make Payment : 
form
- Card Test Case handling : 
test1
test2
test3
test4
- Paid ticket purchases : 
purchased
purchaseagain


app/controllers/conference_registrations_controller.rb
```diff
@@ -33,8 +33,10 @@ def new
   end
 
   def show
-    @total_price = Ticket.total_price(@conference, current_user)
+    TicketPurchase.destroy_all(user_id: current_user.id, conference_id: @conference.id, paid: 'f')
```
hennevogel: Why?

app/models/ability.rb
```diff
@@ -79,6 +79,7 @@ def signed_in(user)
 
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
+    can [:index, :new, :create], Payment
```
hennevogel: So a user can create a Payment for some other user?

app/models/payment.rb
```diff
@@ -0,0 +1,75 @@
...
+      month:               expiration_month,
+      year:                expiration_year,
+      verification_value:  card_verification_value
+    )
+  end
+
+  def validate_card
+    unless credit_card.validate.empty?
+      credit_card.errors.full_messages.each do |message|
+        errors.add(:base, message)
```
hennevogel: Why do you mess with the error collection? The validations already add errors to it? I don't get it, can you explain?

app/models/payment.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+    end
+    credit_card.validate.empty?
+  end
+
+  def price_in_cents
+    (amount * 100).round
+  end
+
+  def purchase(user, conference)
```
hennevogel: Hm, shouldn't this be PaymentsController#create code?

app/models/ticket_purchase.rb
```diff
@@ -48,9 +44,21 @@ def self.purchase_ticket(conference, quantity, ticket, user)
...
                                     conference_id: conference.id,
-                                    user_id: user.id).first
+                                    user_id: user.id,
+                                    paid: 'f').first
 
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  def self.update_paid_ticket_purchases(conference, user, payment)
```
hennevogel: `TicketPurchase` already belongs to a conference a user, why do you pass this in here?

app/models/ticket_purchase.rb
```diff
@@ -48,9 +44,21 @@ def self.purchase_ticket(conference, quantity, ticket, user)
...
+    paid_ticket_purchases = TicketPurchase.where(conference_id: conference.id,
+                                                 user_id: user.id,
+                                                 paid: 'f')
+    begin
+      paid_ticket_purchases.each do |ticket|
+        ticket.update_columns(paid: 't', payment_id: payment.id)
```
hennevogel: `update_columns` bypasses all Validations and Callbacks etc. Why do you use this?

config/environments/development.rb
```diff
@@ -97,3 +97,9 @@
   end
 
 end
+
+
+ActiveMerchant::Billing::Base.mode = :test
+
+  ::GATEWAY = ActiveMerchant::Billing::StripeGateway.new(
+    :login => ENV['SECRET_KEY'])
```
hennevogel: Why do you use `ENV['SECRET_KEY']`?

app/controllers/conference_registrations_controller.rb
rishabhs95: Actually as we want to keep just the online payment option I think that only those `ticket_purchases` should be present who have been paid, all other should be deleted. I have placed it in this controller so as immediately after `ticket_purchase` then `purchase`, you see the `conference_registration` and if somehow the payment failed, we should delete the `ticket_purchase`

app/models/ability.rb
rishabhs95: No, a user can only pay his tickets. I will repair the abilities to make it correct.

app/models/payment.rb
rishabhs95: Ok I will improve the error collection.

app/models/ticket_purchase.rb
rishabhs95: I am just updating the `paid` value in the `ticket_purchase` to 'paid=true' here

app/models/ticket_purchase.rb
rishabhs95: Yes, I am finding a better alternative for that. I will try to directly assign the value, then.

config/environments/development.rb
rishabhs95: It provides the Gateway login for the application. I will try to use a better alternative for this as well.

app/controllers/conference_registrations_controller.rb
hennevogel: So you want to manipulate existing data that is there before your code goes live? We use rake tasks for this, not code in the app

app/models/ticket_purchase.rb
hennevogel: Yes. I'm interested why you pass `conference` and `user` into this function.

app/models/ticket_purchase.rb
hennevogel: Okay :-)

app/models/ticket_purchase.rb
rishabhs95: I am passing the user and conference here, just to find the relevant `ticket_purchase` records

app/controllers/conference_registrations_controller.rb
rishabhs95: I am not modifying the existing data in the application with this code, rather , I am handling a case in which, if the user creates a ticket_purchase and somehow fails to complete his payment, (for example, they click the back button or close the webpage), then the `ticket_purchase` record should be deleted as well.

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,34 @@
+class PaymentsController < ApplicationController
+  before_action :authenticate_user!
+  load_and_authorize_resource
+  load_resource :conference, find_by: :short_title
+  authorize_resource :conference_registrations, class: Registration
+
+  def index
+    @payments = Payment.where(user_id: current_user.id)
```
lagartoflojo: If you add `has_many :payments` to User, then here you can just say `current_user.payments`

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, 'f')
```
lagartoflojo: This reads strange. What is `'f`'? (That's a rhetorical question, I know it means unpaid tickets, but just by glancing at this line, I cannot figure that out).

Something like this would read much more nicely:

```ruby
current_users.tickets.by_conference(@conference).unpaid.total_price
```

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def create
+    @payment = Payment.new(payment_params)
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, 'f')
+
+    if @payment.valid?
+      if @payment.purchase(current_user, @conference)
```
lagartoflojo: Why a nested `if`?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+        @payment.save
+        @update_ticket_purchases = TicketPurchase.update_paid_ticket_purchases(@conference, current_user, @payment)
+        return redirect_to conference_conference_registrations_path(@conference.short_title), flash: { success: 'Thanks! You have purchased your tickets successfully.' }
```
lagartoflojo: `return redirect_to ...` looks strange, it's not very standard.
I guess you need it because you don't want to run the `redirect_to` plus the `render 'new'` line below, correct?
There's another way you can do that...

app/controllers/payments_controller.rb
lagartoflojo: As a side not, using `'f'` or `'F'` or `'no'` or `0` or anything like that instead of a real `false` is generally frowned upon :)

app/models/payment.rb
```diff
@@ -0,0 +1,75 @@
...
+    unless response.success?
+      errors.add(:base, response.message)
+      self.status = 2
+      return false
+    end
+    self.user_id = user.id
+    self.conference_id = conference.id
+    self.last4 = credit_card.display_number
+    self.authorization_code = response.authorization
+    self.status = 1
```
lagartoflojo: What does it mean when `status` is `1`?

app/controllers/conference_registrations_controller.rb
lagartoflojo: What's strange is that you are modifying data in the `show` method.
Could you do this somewhere else?

app/controllers/payments_controller.rb
rishabhs95: Yeah, that looks much better and easy to understand, I will modify the code accordingly.

app/models/payment.rb
rishabhs95: I will add a comment with the code to clarify that

lagartoflojo: More than a comment, how can you make the *code* be more explicit about this?
Hint: take a look at the Rails documentation for enum :)

differentreality:
Is this not going to support Maestro cards?

rishabhs95:
It will support all cards according to what Gateway does the Organiser use... which generally includes all cards like Visa, MasterCard, Maestro, etc.

app/models/payment.rb
```diff
@@ -0,0 +1,64 @@
...
+  belongs_to :user
+  belongs_to :conference
+
+  attr_accessor :credit_card_number
+  attr_accessor :credit_card_type
+  attr_accessor :card_verification_value
+  attr_accessor :expiration_month
+  attr_accessor :expiration_year
+
+  validates :first_name, presence: true
```
ChrisBr: Do we need to store all that data in our database? AFAIK we need to implement the [PCI](https://en.wikipedia.org/wiki/Payment_Card_Industry_Data_Security_Standard) standard for events.o.o ...

lagartoflojo: It's not being stored in the db.. at least there's no columns for this in the corresponding migration.

rishabhs95: Yes, the corresponding data is not saved into the database rather they are directly accessed.

ChrisBr: Ah sorry, was confused! Thanks for the clarification ...

app/controllers/conference_registrations_controller.rb
rishabhs95: I have repaired this and added it in ticketPurchases controller.

app/models/ticket_purchase.rb
rishabhs95: This is because the models were not getting access to `current_user` and current `conference` as they are specific to sessions.

app/controllers/payments_controller.rb
lagartoflojo: I'm not sure why my previous comments don't show up anymore, but this line still looks the same to me.

Also, please remove all usage of `'f'` as a parameter. As I commented above, it's very hard to know what it means. If I show you this line:

```ruby
Ticket.total_price(@conference, current_user, 'f')
# or
Ticket.total_price(@conference, current_user, false)
```
I can kind of deduce that we are getting "the total price of tickets for this user in this conference"... but what do `'f'` or `false` mean?? I have no idea by just looking at this line. We need a clearer parameter here.

And again, please don't use `'f'` but rather `false`. You have a lot of `'f'` usage throughout this PR that you can improve :)

rishabhs95: I don't know how I missed that comment and why it doesn't show up on GH, but yes, I will change that

lagartoflojo: Are you force-pushing? Maybe that's why?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def create
+    @payment = Payment.new(payment_params)
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, 'f')
+
+    if @payment.valid? && @payment.purchase(current_user, @conference, price_in_cents)
+      @payment.save
+      @update_ticket_purchases = TicketPurchase.update_paid_ticket_purchases(@conference, current_user, @payment)
+    end
+
+    if @payment.save
```
lagartoflojo: `@payment.save` again?

app/controllers/payments_controller.rb
rishabhs95: Yes, I used to squash the commits with the earlier commit and force-pushed the changes... so I will not squash the commits while repairing changes from now onwards

spec/models/ticket_spec.rb
```diff
@@ -88,11 +88,11 @@
              user: user,
              ticket: ticket,
              quantity: 20)
-      expect(ticket.quantity_bought_by(user)).to eq(20)
+      expect(ticket.quantity_bought_by(user, 'f')).to eq(20)
```
lagartoflojo: you still need to fix the `'f'`s here in the specs :)

Question: i don't see any tests for when `unpaid` is `true` instead of false...

app/models/payment.rb
```diff
@@ -0,0 +1,60 @@
+class Payment < ActiveRecord::Base
```
lagartoflojo: We're missing unit tests for this class

app/models/ticket.rb
```diff
@@ -21,21 +21,31 @@ def paid?(user)
...
+    quantity = 0
+    if result
+      result.each do |ticket|
+        quantity += ticket.quantity
+      end
+    end
+    quantity
+  end
+
+  def unpaid?(user)
```
lagartoflojo: this method is not tested..

config/environments/development.rb
rishabhs95: well, as the `SECRET_KEY` is the gateway API key, I think that using environment is the most secure way of providing it into the application... please suggest if I should change this or not... but I think I can definitely make the name more clear to something like `PAYMENT_GAYEWAY_SECRET_KEY`

app/controllers/payments_controller.rb
rishabhs95: I have repaired this in the latest commit.

app/models/ticket.rb
rishabhs95: Yes, i will add tests for that

app/models/payment.rb
rishabhs95: Okay, I am adding tests for this class

spec/models/ticket_spec.rb
rishabhs95: As I was just creating ticket_purchases in this controller `paid` still remained `false`, I think i will add tests for making payments while testing `payments.rb` for values with `paid` as `true`.

app/models/ticket_purchase.rb
hennevogel: Then this is controller code, not model code :-)

config/environments/development.rb
hennevogel: Please have a look at how [OSEM is configued](https://github.com/openSUSE/osem/blob/master/INSTALL.md#configure) with environment variables and `config/secrets.yml.example`

lagartoflojo: @hennevogel I'm confused, is the problem that we're not using secrets.yml here, or the name of the env variable?

hennevogel: Both :)

app/models/ticket_purchase.rb
rishabhs95: okay :)

app/controllers/conference_registrations_controller.rb
```diff
@@ -28,8 +28,9 @@ def new
   end
 
   def show
-    @total_price = Ticket.total_price(@conference, current_user)
+    @total_price = Ticket.total_price(@conference, current_user, true)
```
lagartoflojo: what is `true` referring to?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, false)
+  end
+
+  def create
+    @payment = Payment.new(payment_params)
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, false)
+
+    if @payment.valid? && @payment.purchase(current_user, @conference, price_in_cents)
+      if @payment.save
```
lagartoflojo: Do we need the nested `if`?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  load_and_authorize_resource
+  load_resource :conference, find_by: :short_title
+  authorize_resource :conference_registrations, class: Registration
+
+  def index
+    @payments = current_user.payments
+  end
+
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, false)
```
lagartoflojo: What is `false` referring to?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def price_in_cents
+    (@payment.amount * 100).round
+  end
+
+  def update_paid_ticket_purchases(conference, user, payment)
+    paid_ticket_purchases = TicketPurchase.where(conference_id: conference.id,
+                                                 user_id: user.id,
+                                                 paid: false)
+    begin
```
lagartoflojo: what's this `begin` block for?

app/controllers/ticket_purchases_controller.rb
```diff
@@ -4,11 +4,11 @@ class TicketPurchasesController < ApplicationController
   authorize_resource :conference_registrations, class: Registration
 
   def create
+    TicketPurchase.destroy_all(user_id: current_user.id, conference_id: @conference.id, paid: false)
     message = TicketPurchase.purchase(@conference, current_user, params[:tickets][0])
     if message.blank?
-      if current_user.ticket_purchases.any?
-        redirect_to conference_conference_registrations_path(@conference.short_title),
-                    notice: "Thank you for supporting #{@conference.title} by purchasing a ticket."
+      if current_user.ticket_purchases.where(paid: false).any?
```
lagartoflojo: maybe here you could say `current_user.ticket_purchases.unpaid.any?`

app/models/ticket.rb
```diff
@@ -21,21 +21,31 @@ def paid?(user)
     ticket_purchases.find_by(user: user, paid: true).present?
   end
 
-  def quantity_bought_by(user)
-    result = ticket_purchases.where(user_id: user.id).first
-    result ? result.quantity : 0
+  def quantity_bought_by(user, paid)
+    result = ticket_purchases.where(user_id: user.id, paid: paid)
```
lagartoflojo: `result` is an awfully generic variable name.. could you name it better?

app/views/conference_registrations/show.html.haml
```diff
@@ -89,29 +89,30 @@
...
+                  = ticket.title
+                  = word_pluralize(ticket.quantity, 'Ticket')
+                  for
+                  = ticket.price.symbol
+                  = humanized_money ticket.price
+              %hr
+          - if @tickets.any?
+            = link_to 'Buy more tickets', conference_tickets_path(@conference.short_title), class: "btn btn-default"
+          - else
+            You haven't bought any tickets.
```
lagartoflojo: Will we ever get to the case where we haven't bought any tickets?
Or is that part not yet implemented in this PR?

app/views/payments/index.html.haml
```diff
@@ -0,0 +1,32 @@
+.container
+  .row
+    .col-md-12
+      .page-header
+        %h1
+          Payment Attempts for
+          = @conference.title
+      - if !flash[:notice].blank?
```
lagartoflojo: can't you just say `if flash[:notice]` ?

lagartoflojo: Or even: `if flash[:notice].present?` and then you don't need the `!` bang

app/views/payments/new.html.haml
```diff
@@ -0,0 +1,14 @@
+.container
+  .row
+    .col-xs-6.col-xs-offset-3
+      %h1
+        Buy tickets for
+        = @total_amount_to_pay.symbol
```
lagartoflojo: what does this `.symbol` method do?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      should_not allow_value(0).for(:amount)
+    end
+
+    it 'is not valid with a amount smaller than zero' do
+      should_not allow_value(-1).for(:amount)
+    end
+
+    it 'is valid with a amount greater than zero' do
+      should allow_value(1).for(:amount)
+    end
```
lagartoflojo: from line 10 until here, you can write the specs like:

```ruby
it { is_expected.to validate_presence_of(:last_name) }`
```

That way, you don't need so much repetition.

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  end
+
+  describe 'self#purchase' do
```
lagartoflojo: you don't need to put `self` here, `describe '#purchase` is enough.

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      purchase = TicketPurchase.where(conference_id: conference.id,
+                                      user_id: participant.id,
+                                      ticket_id: ticket_1.id).first
+
+      expect(TicketPurchase.count).to eq(1)
+      expect(purchase.quantity).to eq(1)
+      expect(message.blank?).to be true
+
+      response = Payment.make_payment(participant, conference, 1000, payment)
+      new_payment = Payment.first
```
lagartoflojo: why do you need to do this?

spec/models/ticket_spec.rb
```diff
@@ -82,17 +82,37 @@
...
+  end
+
   describe '#quantity_bought_by' do
     it 'returns the correct value if the user has bought this ticket' do
       create(:ticket_purchase,
              user: user,
              ticket: ticket,
              quantity: 20)
-      expect(ticket.quantity_bought_by(user)).to eq(20)
+      expect(ticket.quantity_bought_by(user, false)).to eq(20)
```
lagartoflojo: what does `false` refer to?

spec/models/ticket_spec.rb
```diff
@@ -82,17 +82,37 @@
...
+    context 'user has paid' do
+      before { ticket_purchase.update_attributes(paid: true) }
+
+      it 'returns false' do
+        expect(ticket.unpaid?(user)).to eq(false)
+      end
+    end
+  end
+
   describe '#quantity_bought_by' do
```
lagartoflojo: there's no tests for when the second parameter of this method is `true` (whatever `true` means in this context)

spec/models/ticket_spec.rb
```diff
@@ -102,11 +122,11 @@
              user: user,
              ticket: ticket,
              quantity: 20)
-      expect(ticket.total_price(user)).to eq(Money.new(100000, 'USD'))
+      expect(ticket.total_price(user, false)).to eq(Money.new(100000, 'USD'))
```
lagartoflojo: there's no tests for when the second parameter is `true`

spec/models/ticket_spec.rb
```diff
@@ -116,7 +136,7 @@
     describe 'user has bought' do
       context 'no tickets' do
         it 'returns zero' do
-          expect(Ticket.total_price(conference, user)).to eq(Money.new(0, 'USD'))
+          expect(Ticket.total_price(conference, user, false)).to eq(Money.new(0, 'USD'))
```
lagartoflojo: there's no tests for when the second parameter is `true`

app/controllers/ticket_purchases_controller.rb
rishabhs95: I can't reuse the unpaid function here because it is a ticket-model specific function... so I'll have to either define the function again or just use a where clause.

app/views/conference_registrations/show.html.haml
rishabhs95: for the moment, there can be a case where a user purchased no tickets, if the user cancels payment after ticket_purchase but they will be informed that their registration will not be considered as complete until they buy the tickets.

app/views/payments/new.html.haml
rishabhs95: symbol method provides the currency symbol for variables with type `Money`

spec/models/payment_spec.rb
rishabhs95: I created the variable to check if the record is present in `Payment` table or not.

app/controllers/ticket_purchases_controller.rb
lagartoflojo: What do you mean? I don't see any `.unpaid` method or scope in TicketPurchase.

app/views/conference_registrations/show.html.haml
lagartoflojo: > they will be informed that their registration will not be considered as complete until they buy the tickets.

Is that in this PR?

app/controllers/ticket_purchases_controller.rb
rishabhs95: Yes, I meant it is a method accessible only in `Ticket` model

app/views/conference_registrations/show.html.haml
rishabhs95: Yes, users are informed at their registration page with a message to buy tickets

app/views/conference_registrations/show.html.haml
```diff
@@ -89,29 +89,30 @@
...
+            = link_to 'Please buy some tickets to support us!', conference_tickets_path(@conference.short_title)
+            %p
+              (Your registration won't be complete without buying a ticket)
```
rishabhs95: This message informs user that their registration is incomplete without purchasing tickets

app/views/payments/new.html.haml
lagartoflojo: Okay, would it make more sense to use `Money#format` instead of `.symbol` and `humanized_money` together?

app/controllers/ticket_purchases_controller.rb
lagartoflojo: Ah, right, you DO have to define a new `unpaid` method... or rather, a scope :)

spec/models/payment_spec.rb
lagartoflojo: but there's definitely a Payment: in line 56 you create one

rishabhs95: Okay then, I will reuse the created payment

app/models/payment.rb
rishabhs95: I have added tests for this class.

app/models/ticket.rb
rishabhs95: Added tests for testing this method.

app/views/payments/new.html.haml
rishabhs95: I think `humanised_money` and `#symbol` are used at a lot of places, is it okay if I replace by `Money#format` method that in an another PR?

spec/models/payment_spec.rb
rishabhs95: I have changed this as you suggested

spec/models/ticket_spec.rb
rishabhs95: actually, the second value(paid value) doesn't matter in this case... I will remove the second parameter in this function call

lagartoflojo: @rishabhs95, if a method accepts a parameter, it must be tested with that parameter, regardless whether it has a default. So you have to test both cases, when `paid` is `true` and when it's `false`.

app/views/payments/new.html.haml
lagartoflojo: Ok, but please make a Trello card for it.

app/controllers/ticket_purchases_controller.rb
```diff
@@ -4,11 +4,11 @@ class TicketPurchasesController < ApplicationController
...
 
   def create
+    TicketPurchase.destroy_all(user_id: current_user.id, conference_id: @conference.id, paid: false)
     message = TicketPurchase.purchase(@conference, current_user, params[:tickets][0])
     if message.blank?
-      if current_user.ticket_purchases.any?
-        redirect_to conference_conference_registration_path(@conference.short_title),
-                    notice: "Thank you for supporting #{@conference.title} by purchasing a ticket."
+      if current_user.ticket_purchases.unpaid.any?
+        redirect_to new_conference_payment_path, notice: 'Please pay here to purchase tickets.'
```
lagartoflojo: On line 7 you destroy all of the user's unpaid tickets, and then on line 10 you check if she has any unpaid tickets... I don't think line 10 is will ever return `true`. What's the desired flow here?

lagartoflojo: Ah, or can an unpaid ticket be created on line 8?
The flow is not very clear..

spec/features/ticket_purchases_spec.rb
```diff
@@ -26,26 +26,20 @@
         fill_in "tickets__#{ticket.id}", with: '2'
         expect(current_path).to eq(conference_tickets_path(conference.short_title))
 
-        click_button 'Support'
+        click_button 'Continue'
```
lagartoflojo: Are there feature tests for failed cases? Like when the payment doesn't go through, for example?

spec/models/payment_spec.rb
lagartoflojo: This still needs a fix :)

app/controllers/payments_controller.rb
lagartoflojo: Have you looked at how you can re-write this as:

```
current_user.tickets.by_conference(@conference).unpaid.total_price
```

app/models/ticket.rb
```diff
@@ -21,21 +21,31 @@ def paid?(user)
     ticket_purchases.find_by(user: user, paid: true).present?
   end
 
-  def quantity_bought_by(user)
-    result = ticket_purchases.where(user_id: user.id).first
-    result ? result.quantity : 0
+  def quantity_bought_by(user, paid: false)
+    purchased_tickets = ticket_purchases.where(user_id: user.id, paid: paid)
```
lagartoflojo: how about:
```ruby
purchased_tickets = user.ticket_purchases.paid
```

lagartoflojo: Actually, you can also much easily add up the quantities.

Look at the [sum method](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/Calculations.html#method-i-sum).

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+  end
+
+  def create
+    @payment = Payment.new(payment_params)
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+
+    if @payment.valid? && @payment.purchase(current_user, @conference, price_in_cents) && @payment.save
```
lagartoflojo: `@payment.save` will return `false` if the payment is not valid. Do we need to have `@payment.valid?` here as well?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  private
+
+  def price_in_cents
+    (@payment.amount * 100).round
+  end
+
+  def update_paid_ticket_purchases(conference, user, payment)
+    paid_ticket_purchases = TicketPurchase.where(conference_id: conference.id,
+                                                 user_id: user.id,
+                                                 paid: false)
```
lagartoflojo: How about:
```
current_user.ticket_purchases.by_conference(conference).unpaid
```

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  private
+
+  def price_in_cents
+    (@payment.amount * 100).round
+  end
+
+  def update_paid_ticket_purchases(conference, user, payment)
+    paid_ticket_purchases = TicketPurchase.where(conference_id: conference.id,
```
lagartoflojo: why is the name of this variable **paid**_ticket_purchases if the where has `paid: false`?

app/models/payment.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def purchase(user, conference, price_in_cents)
+    begin
+      response = GATEWAY.purchase(price_in_cents, credit_card, currency: conference.tickets.first.price_currency)
+    rescue
+      false
+    end
+
+    unless response
+      errors.add(:base, 'Unable to recieve any response')
```
lagartoflojo: should say: receive

app/models/payment.rb
```diff
@@ -0,0 +1,65 @@
...
+    self.user_id = user.id
+    self.conference_id = conference.id
+    self.last4 = credit_card.display_number
+    self.authorization_code = response.authorization
+    self.status = 'success'
+    response.success?
+  end
+
+  # method to test `purchase` method
+  def self.make_payment(user, conference, price_in_cents, payment)
```
lagartoflojo: why do you need this additional method which is used only in testing?

app/models/ticket.rb
```diff
@@ -21,21 +21,31 @@ def paid?(user)
...
+    if purchased_tickets
+      purchased_tickets.each do |ticket|
+        quantity += ticket.quantity
+      end
+    end
+    quantity
+  end
+
+  def unpaid?(user)
+    ticket_purchases.find_by(user: user, paid: false).present?
```
lagartoflojo: how about:
```
ticket_purchases.unpaid.find_by(user: user).any?
```

app/models/payment.rb
```diff
@@ -0,0 +1,65 @@
...
+      number:              credit_card_number,
+      month:               expiration_month,
+      year:                expiration_year,
+      verification_value:  card_verification_value
+    )
+  end
+
+  def purchase(user, conference, price_in_cents)
+    begin
+      response = GATEWAY.purchase(price_in_cents, credit_card, currency: conference.tickets.first.price_currency)
```
lagartoflojo: all tickets use the same currency?

app/controllers/conference_registrations_controller.rb
```diff
@@ -28,8 +28,9 @@ def new
   end
 
   def show
-    @total_price = Ticket.total_price(@conference, current_user)
-    @tickets = current_user.ticket_purchases.where(conference_id: @conference.id)
+    @total_price = Ticket.total_price(@conference, current_user, paid: true)
+    @tickets = current_user.ticket_purchases.where(conference_id: @conference.id, paid: true)
```
lagartoflojo: How about:
```
current_user.ticket_purchases.by_conference(@conference).paid
```

app/views/payments/new.html.haml
rishabhs95: Done.

app/controllers/ticket_purchases_controller.rb
rishabhs95: Actually, line 7 deletes the unpaid ticket which failed their payments, then, new `ticket_purchases` are created on line 8 and then payment is attempted for new purchased_tickets on line 11

app/controllers/payments_controller.rb
rishabhs95: it is named as such as it updates the tickets whose payments are completed recently but had `paid: false`. it updates `paid` to `true`.

app/models/payment.rb
rishabhs95: Yes, all tickets in a conference always have the same currency.

app/controllers/conference_registrations_controller.rb
rishabhs95: Okay, will use a scope for this.

app/models/payment.rb
rishabhs95: I used this, as for testing this instance method through model, it is unaccessible and I want to use it as an instance method in code rather than class method which is needed for tests

app/models/payment.rb
rishabhs95: Okay, will change that

app/controllers/payments_controller.rb
rishabhs95: Okay, will change that

app/controllers/payments_controller.rb
rishabhs95: Yes, I guess as we clubbed both conditions... we won't require this now

app/controllers/ticket_purchases_controller.rb
```diff
@@ -4,11 +4,11 @@ class TicketPurchasesController < ApplicationController
   authorize_resource :conference_registrations, class: Registration
 
   def create
+    TicketPurchase.destroy_all(user_id: current_user.id, conference_id: @conference.id, paid: false)
     message = TicketPurchase.purchase(@conference, current_user, params[:tickets][0])
     if message.blank?
-      if current_user.ticket_purchases.any?
-        redirect_to conference_conference_registration_path(@conference.short_title),
-                    notice: "Thank you for supporting #{@conference.title} by purchasing a ticket."
+      if current_user.ticket_purchases.unpaid.any?
```
lagartoflojo: should this line be limited to only THIS conference?

app/controllers/payments_controller.rb
lagartoflojo: Also, checking the code for `TicketPurchase.purchase`, it seems to always return a truthy value, doesn't it?
Does it belong in the conditional then, or do you want to somehow refactor that method?

app/controllers/payments_controller.rb
lagartoflojo: Ok, still a bit confusing.. could you come up with a clearer variable name?

rishabhs95: Okay, maybe something like `update_purchased_ticket_purchases`

app/controllers/ticket_purchases_controller.rb
rishabhs95: Yes, it should be

app/helpers/payments_helper.rb
```diff
@@ -0,0 +1,13 @@
+module PaymentsHelper
```
rishabhs95: Added tests for testing this helper

spec/models/ticket_spec.rb
rishabhs95: Okay, adding tests for both

app/models/ticket.rb
rishabhs95: Used this method to calculate sum now

spec/models/payment_spec.rb
lagartoflojo: > Rishabh Saxena, [13.07.16 08:57]
actually, in the part where you said to reuse already created payment rather than creating new_payment variable... but creating the variable new_payment also lets me check the record and its values like new_purchase.amount
> Rishabh Saxena, [13.07.16 08:59]
whereas the variable which stores the result of created payment (i.e, response) doesn't

I don't get it --
You have a Payment instance already, called `payment`.
Why can't you say

```ruby
response = payment.purchase(participant, conference, 1000)
# (...)
expect(payment.amount).to eq(10)
```

Why do you need that `make_payment` method at all, and why do you need to create a new payment?

app/models/ticket.rb
lagartoflojo: I don't see the change.. maybe you didn't push?

app/controllers/payments_controller.rb
lagartoflojo: Yup, something like that could work.

app/models/payment.rb
lagartoflojo: >  I want to use it as an instance method in code rather than class method which is needed for tests

Why are class methods needed for tests? You can perfectly write tests using instance methods.

lagartoflojo:
In this screen:

screenshot from 2016-07-13 17-12-23

Would it maybe make more sense not to group by purchase but to see an overview of all the tickets you bought together?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      expect(TicketPurchase.count).to eq(0)
+    end
+
+    it 'creates no ticket purchase or payment if amount is 0' do
+      tickets = { ticket_1.id.to_s => '0' }
+      TicketPurchase.purchase(conference, participant, tickets)
+
+      expect(TicketPurchase.count).to eq(0)
+    end
```
lagartoflojo: The tests on line 63 and 70 confuse me. This is `payment_spec.rb`, so that means we are testing the Payment class. But in these tests, you never call a method from Payment, only from TicketPurchase. Sounds like these tests should go in ticket_purchase_spec.rb, or am I missing something?

spec/helpers/payments_helper_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+
+describe PaymentsHelper, type: :helper do
+  let(:conference) { create(:conference) }
+  let(:event) { create(:event, program: conference.program) }
```
lagartoflojo: where are these variables used?

spec/helpers/payments_helper_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  describe '#years' do
+    it 'returns the correct set of options' do
+      expect(years).to match_array(Array(Date.current.year..Date.current.year + 15))
+    end
+  end
+
+  describe '#card_types' do
+    it 'returns the correct set of card_types array' do
+      expect(card_types).to match_array(Array([["American Express", "american_express"], ["Discover", "discover"],
```
lagartoflojo: Do the card types depend on which payment gateway we're using? If so, can we get the card types from the gateway somehow?

spec/models/payment_spec.rb
```diff
@@ -39,39 +39,40 @@
...
+      expect(Payment.count).to eq(0)
     end
 
     it 'creates no ticket purchase or payment if amount is 0' do
       tickets = { ticket_1.id.to_s => '0' }
       TicketPurchase.purchase(conference, participant, tickets)
 
       expect(TicketPurchase.count).to eq(0)
+      expect(Payment.count).to eq(0)
+    end
```
lagartoflojo: the `TicketPurchase.purchase` method (actually, the entire TicketPurchase class) does not mention Payment anywhere. Why are we testing it here? What do these tests achieve?

spec/models/payment_spec.rb
```diff
@@ -39,39 +39,40 @@
...
+
+    let(:payment) { create(:payment) }
```
lagartoflojo: inside the test, you say `payment = Payment.new`, so apparently you never end up using the `payment` variable defined here.. or is it here for a different reason?

spec/models/payment_spec.rb
```diff
@@ -39,39 +39,40 @@
...
+
+      expect(TicketPurchase.count).to eq(1)
+      # expect(purchase.quantity).to eq(1)
+      expect(message.blank?).to be true
+
+     payment = Payment.new
+     payment.purchase(participant, conference, 1000)
+
+      expect(Payment.count).to eq(1)
+      expect(payment.blank?).to be true
```
lagartoflojo: Why should `payment.blank?` equal true?

spec/helpers/payments_helper_spec.rb
rishabhs95: Card types depend upon the gateway but actually, I have not used this component so I should rather drop this as gateways automatically detect card gateways

lagartoflojo: That makes ssnese :)

spec/features/ticket_purchases_spec.rb
rishabhs95: Added tests for failing payments

spec/helpers/payments_helper_spec.rb
rishabhs95: removed unused variables

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(Payment.count).to eq(0)
+    end
+
+    it 'creates no ticket purchase or payment if amount is 0' do
+      tickets = { ticket_1.id.to_s => '0' }
+      TicketPurchase.purchase(conference, participant, tickets)
+
+      expect(TicketPurchase.count).to eq(0)
+      expect(Payment.count).to eq(0)
+    end
```
lagartoflojo: the `TicketPurchase.purchase` method (actually, the entire TicketPurchase class) does not mention Payment anywhere. Why are we testing it here? What do these tests achieve?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    let(:payment) { create(:payment) }
```
lagartoflojo: inside the test, you say `payment = Payment.new`, so apparently you never end up using the `payment` variable defined here.. or is it here for a different reason?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      purchase = TicketPurchase.where(conference_id: conference.id,
+                                      user_id: participant.id,
+                                      ticket_id: ticket_1.id).first
+
+      expect(TicketPurchase.count).to eq(1)
+      expect(purchase.quantity).to eq(1)
+      expect(message.blank?).to be true
+
+      payment = Payment.new
+      payment.purchase(participant, conference, 1000)
```
lagartoflojo: OK so you called `payment.purchase`, but what do we expect to happen now? We are missing some... "expectations" :)

spec/models/payment_spec.rb
rishabhs95: Actually `payments` are made only after a ticket is purchased, i.e., `TicketPurchase` record is made.

app/controllers/ticket_purchases_controller.rb
```diff
@@ -4,11 +4,11 @@ class TicketPurchasesController < ApplicationController
   authorize_resource :conference_registrations, class: Registration
 
   def create
+    TicketPurchase.destroy_all(user_id: current_user.id, conference_id: @conference.id, paid: false)
```
lagartoflojo: `TicketPurchase.by_user(user).by_conference(conference).unpaid.destroy_all` :)

lagartoflojo: Or even:
```
current_user.ticket_purchases.by_conference(@conference).unpaid
```
which is the call used in line 10

config/environments/development.rb
```diff
@@ -96,4 +96,10 @@
     end
   end
 
+  #Initialize Payment Gateway with valid credentials
+  ActiveMerchant::Billing::Base.mode = :test
+  ::GATEWAY = ActiveMerchant::Billing::StripeGateway.new(:login => ENV['SECRET_KEY'])
```
lagartoflojo: This env variable could have a more specific name. `OSEM_STRIPE_SECRET_KEY` ?
Also, maybe it should be added to `dotenv.example`

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @payments = current_user.payments
+  end
+
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+  end
+
+  def create
+    @payment = Payment.new(payment_params)
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
```
lagartoflojo: I think this variable is only used when the purchase fails and we `render :new`, right?
If so, I would move it to the `else` block in line 22, so that it's clear it's only used there.

app/views/conference_registrations/show.html.haml
lagartoflojo: Where the admins see the list of registrations, do they also see if the user has bought a ticket?

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,34 @@
+= semantic_form_for(@payment, url: conference_payments_path) do |f|
+  .form-group
+    = f.label :first_name, 'First name(as on card)'
```
lagartoflojo: I think we could just ask for full name, instead of first & last, and call the text field `:name`, which should be allowed by the credit card class:
https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/billing/credit_card.rb#L251
That would simplify this form, because really there's no reason to ask for both separately. What do you think>

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,34 @@
...
+    = f.text_field :first_name, class: "form-control", placeholder: "John", id: "first_name"
+  .form-group
+    = f.label :last_name, 'Last name(as on card)'
+    = f.text_field :last_name,class: "form-control", placeholder: "Doe", id: "last_name"
+  %span.pull-right
+    %img.img-responsive{:src => image_url('credit_card.png')}
+  .form-group
+    = f.label :credit_card_number, "Credit Card number(without spaces)"
```
lagartoflojo: need a space before the parenthesis: `number (without`

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,34 @@
...
+  .form-group.col-md-10
+    = f.label :card_verification_value, 'Security Code(3 on back, AmEx: 4 on front)'
+    = f.text_field :card_verification_value, class: "form-control", placeholder: "XXX", id: "card_verification_value"
+    = f.number_field :amount, value: @total_amount_to_pay, class: "form-control", type: 'hidden'
+  %ul.price-tags.pull-right
+    %li.text-muted
+      you will pay
+    %li
+      %a
+        = number_to_currency @total_amount_to_pay
```
lagartoflojo: I like this "you will pay" ribbon, but I think it should be AFTER the "charge card" button..

screenshot from 2016-07-21 13-23-23

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,34 @@
...
+      %span.input-group-addon
+        %i.fa.fa-credit-card
+  .form-group.col-md-6
+    = f.label :expiration_month
+    = f.select :expiration_month, months, {}, class: "form-control", id: "expiration_month"
+  .form-group.col-md-6
+    = f.label :expiration_year
+    = f.select :expiration_year, years, {}, class: "form-control", id: "expiration_year"
+  .form-group.col-md-10
+    = f.label :card_verification_value, 'Security Code(3 on back, AmEx: 4 on front)'
```
lagartoflojo: space before parenthesis `Code (3`

app/views/payments/index.html.haml
```diff
@@ -0,0 +1,32 @@
+.container
```
lagartoflojo: Is this page used anywhere in the flow?

app/views/tickets/_ticket.html.haml
```diff
@@ -9,7 +9,7 @@
           = markdown(ticket.description)
   %td.col-sm-1.col-md-1
     - if ticket.bought?(current_user)
-      = text_field_tag("tickets[][#{ticket.id}]", ticket.quantity_bought_by(current_user),
+      = text_field_tag("tickets[][#{ticket.id}]", 0,
```
lagartoflojo: this `text_field_tag` looks exactly like the one on line 15.. why do we need the `if`?

db/migrate/20160606040848_create_payments.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePayments < ActiveRecord::Migration
+  def change
+    create_table :payments do |t|
+      t.string :first_name
+      t.string :last_name
+      t.string :last4
+      t.decimal :amount, precision: 12, scale: 3
+      t.string :authorization_code
+      t.integer :status, default: 0
```
lagartoflojo: I would also add `null: false` for the other fields: first name, last name, last4, amount, status.
Not for authorization code, since that one only gets set whe n the payment goes through, i guess?

db/migrate/20160606040848_create_payments.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.integer :user_id, null: false
+      t.integer :conference_id, null: false
+      t.datetime :created_at
+      t.datetime :updated_at
```
lagartoflojo: the `t.timestamps` call on line 15 adds the `created_at` and `updated_at` columns. We don't need to add them again.

db/migrate/20160606040848_create_payments.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePayments < ActiveRecord::Migration
+  def change
+    create_table :payments do |t|
+      t.string :first_name
+      t.string :last_name
+      t.string :last4
+      t.decimal :amount, precision: 12, scale: 3
```
lagartoflojo: ActiveMerchant expects an integer amount. Do we really need a decimal here?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      redirect_to conference_conference_registration_path(@conference.short_title), flash: { success: 'Thanks! You have purchased your tickets successfully.' }
+    else
+      render 'new'
+    end
+  end
+
+  private
+
+  def update_purchased_ticket_purchases
+    paid_ticket_purchases = current_user.ticket_purchases.by_conference(@conference).unpaid
```
lagartoflojo: we can do this in one line:
```
current_user.ticket_purchases.by_conference(@conference).unpaid.update_all(paid: true, payment: payment)
```

lagartoflojo: For that to work though, you have to add `belongs_to :payment` to TicketPurchase.

app/controllers/payments_controller.rb
rishabhs95: Okay

app/views/conference_registrations/show.html.haml
rishabhs95: Yes, it shows the payment status for purchased tickets

kalabiyau:
@rishabhs95, @lagartoflojo from my side you get really big LGTM. It is a great piece of work!

app/views/conference_registrations/show.html.haml
differentreality: I don't think that message is very accurate.
1) Tickets are not always mandatory as this line implies
2) The registration is actually already successful, because we only redirect to the tickets page **if** the registration saves successfully
https://github.com/openSUSE/osem/blob/master/app/controllers/conference_registrations_controller.rb#L50

differentreality:
@rishabhs95  Your tests are failing because you have not authorized all the controller actions you are using. I am happy to help if you have problems with cancancan authorization.

lagartoflojo:
@rishabhs95 should we close this?

lagartoflojo:
After the discussion in https://github.com/openSUSE/osem/issues/1105, we decided to close this PR. It has been superseded by https://github.com/openSUSE/osem/issues/1108.

Issue #1033, opened 2016-06-10
Changing the text color of tracks taking the background color into account


Ana06:
The text color of the track name is always white without taking into account the background color, so it is imposible/difficult to read it:

![image](https://cloud.githubusercontent.com/assets/16052290/15981029/94bd9662-2f70-11e6-9348-aa65b0c55fa0.png)

I think we should use black or white text depending on the background color chosen. Similarly to Github with labels:

![image](https://cloud.githubusercontent.com/assets/16052290/15981070/045bb35a-2f71-11e6-8976-94a7c3d1713f.png)

Can I do it?

Ana06:
This can be closed. :smile_cat:

PR #1044, opened 2016-06-20
Changing the text color of tracks taking the background color into account

Ana06:
Issue #1033 

Before:

image

Now:

image

I didn't change it in the schedule as it is being changed and there are two pull-request open. I'll change afterwards.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -286,4 +286,16 @@ module ApplicationHelper
   def unread_notifications(user)
     Comment.accessible_by(current_ability).find_since_last_login(user)
   end
+
+  # Returns black or white deppending on what of them contrast more with the
+  # given color. Useful to print text in a coloured background.
+  # hexcolor is a hex color of 7 characters, being the first one '#'.
+  # Reference: https://24ways.org/2010/calculating-color-contrast
+  def contrast_color(hexcolor)
+    r = hexcolor[1..2].to_i(16)
+    g = hexcolor[3..4].to_i(16)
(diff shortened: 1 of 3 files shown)


hennevogel:
We are using [tinycolor](https://github.com/bgrins/TinyColor) already in the splash page to calculate colors. Can't we use this instead of re-inventing it ourselves?

Ana06:
We could, but tinycolor `mostReadable` function selects black or white depending on what of them has more readability, and readability is given by:

``` Javascript
tinycolor.readability=function(color1,color2){
    var c1=tinycolor(color1),c2=tinycolor(color2);
    return(Math.max(c1.getLuminance(),c2.getLuminance())+.05)/(Math.min(c1.getLuminance(),c2.getLuminance())+.05)
}
```

So, it is not exactly the same and will select a different color for some cases. In the following image you can see the difference for the colors I used before. The first column is my option and the second is tinycolor option. As you can see, they are different for two colors. In my opinion, and as we are talking about color it is probably a matter of taste, my option looks better.

image


Also, I don't like the javascript needed to use tinycolor for that. That the javascript I wrote for this:

``` Javascript
$(function () {
    var elements = document.getElementsByClassName("track");
    for(var i=0; i < elements.length; i++) {
        var background = elements[i].style.backgroundColor
        var color = tinycolor.mostReadable(background,["#000", "#fff"]).toHexString();
        elements[i].style.color = color;
    }
  });
});
```

And we will have a dependency of a library to use one function that is easily to implement.

In short, I'll use my first implementation. :wink:

hennevogel:
Her we go :-)

PR #1047, opened 2016-06-20
XML export fix tests. Add not nil room for scheduled events.

differentreality:


differentreality:
@hennevogel  I have trouble defining this in a factory. Due to program being created automatically on conference creation?
https://github.com/differentreality/osem/blob/15076b11f14701d697a5e691d32332348da6b7a9/spec/controllers/conference_controller_spec.rb#L32

hennevogel:
@differentreality do it in an `after(:create)` block?

app/models/program.rb
```diff
@@ -27,7 +27,7 @@ def confirmed
     end
 
     def scheduled
-      where.not(start_time: nil)
+      where.not(start_time: nil).where.not(room: nil)
```
hennevogel: Hm. How can this happen?

differentreality: I guess the same way we currently have the tours scheduled with time, but without room. (which is, I think, what made XML export fail)

hennevogel: okay

PR #1048, opened 2016-06-20
Show most recent conference first in conferences#index

differentreality:


app/controllers/conference_controller.rb
```diff
@@ -5,7 +5,7 @@ class ConferenceController < ApplicationController
   load_resource :program, through: :conference, singleton: true, except: :index
 
   def index
-    @current = Conference.where('end_date >= ?', Date.current).order('start_date ASC')
+    @current = Conference.where('end_date >= ?', Date.current).reorder('start_date ASC')
```
hennevogel: `'start_date DSC'`

differentreality: Why DESC? Don't we want to show the earliest conference first? (ie. first show oSC16 and then show summit asia in events.o.o)

hennevogel: Yes I think so too. Isn't that what you are trying to do? Now it's `ASC` and the wrong order right? :-)

differentreality: Yes, but it is the wrong order because sth goes wrong with ordering. Which is why I used reorder. It is still wrong with DESC.

hennevogel: By default we order by `start_date DESC`.

https://github.com/openSUSE/osem/blob/master/app/models/conference.rb#L10

That's why. I guess you should change it there...

differentreality: Turns out it might not be such a good idea to change that in the model. 
Because that would affect other parts of the application, eg. dropdwon with list of conferences in admin panel, and there we don't want to see the oldest conference first. It makes sense seeing the most recent one.

Here in the controller we are only selecting the most recent conferences, and then it makes sense to reorder these as start_date: :asc.

Either way we need to touch something.

hennevogel:
Okay LGTM

PR #1059, opened 2016-06-27
Update API key section in TRANSLATION.md

Ana06:
The API key was moved to the OSEM_TRANSIFEX_APIKEY variable in .env.development

### DIFF
--- a/TRANSLATION.md
+++ b/TRANSLATION.md
@@ -41,19 +41,12 @@ Eg. Assuming you have signed up yourself as a translator for OSEM, to translate 
 You can view OSEM in other languages by selecting the language you want from the language selector icon (bottom right corner).
 
 ## 4. API key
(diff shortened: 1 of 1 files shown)


TRANSLATION.md
```diff
@@ -41,15 +41,9 @@ Eg. Assuming you have signed up yourself as a translator for OSEM, to translate
 You can view OSEM in other languages by selecting the language you want from the language selector icon (bottom right corner).
 
 ## 4. API key
-  When you run your own instance of OSEM, if you want to have the translations available you need to add the necessary [code] (http://docs.transifex.com/live/webmasters/#api) from Transifex:
+  When you run your own instance of OSEM, if you want to have the translations available you need to add the necessary [code] (http://docs.transifex.com/live/webmasters/#api) from Transifex. Specifically, you have to add the api key for transifex.com in the `.env.development` file:
```
differentreality: I would remove the whole sentence about adding code from transifex, since we are not doing that anymore. 
So:
When you run your own instance of OSEM, if you want to have the translations available you need to add the api key in the .env.development file:

Issue #1064, opened 2016-06-29
Use computer's local timezone when creating conference


shyukri:
Currently we use American Samoa timezone by default (GMT -11)

shyukri:
Tracked in PR#1065

PR #1065, opened 2016-06-29
Scheduling new conference: use default Time.zone.now instead of the 1…

shyukri:
…st available option which was American Samoa. Should fix issue#1064.


differentreality:
Fixes #1064

hennevogel:
LGTM. Thanks @shyukri

Issue #1062, opened 2016-06-28
Adding Add buttons at the top


Ana06:
Currently the _Add_  button in the rooms page in the backend is at the end:

![image](https://cloud.githubusercontent.com/assets/16052290/16415091/bfb3041e-3d3b-11e6-9d2a-371e03eff80b.png)

So if there are many rooms to have to scroll a lot to create a new room. We should also add a new button at the top:

![good_solution](https://cloud.githubusercontent.com/assets/16052290/16415584/4d691008-3d3e-11e6-9fb0-5d859cf4e005.png)

That should be done for rooms, tracks, events types, ... and in short in everywhere to be consistent.

shyukri:
What should be the reasonable limit to display 2 btns? 5 or more items?

differentreality:
What do you mean?

shlok007:
Hi! Is this issue still open? If that's an issue, I suggest instead of adding two buttons, we could just have a single button at the top.

Ana06:
I think that it is not open, because @shyukri implemented it in #1066. :no_mouth:

shlok007:
I saw a closed PR. Sorry :).

hennevogel:
I don't like it. Too many buttons are way more confusing than learning to press END on your keyboard.

PR #1066, opened 2016-06-29
Add default buttons on top if the items are > 5 on tracks, event_type…

shyukri:
…s, lodgings, rooms. This should prevent scrolling down to find the button.
This is for #1062


hennevogel:
I don't like it. Too many buttons are way more confusing than learning to press END on your keyboard

Ana06:
@hennevogel I think that this is issue #1062 so if you don't like it maybe that issue should be closed. :wink:

PR #1073, opened 2016-07-01
Use protocol relative URL when loading OSM tiles

danimo:
This avoids mixed content warnings and gives us our padlock back on Chromium
and Firefox.

This also switches the domain for the tile servers from osm.org to 
openstreetmap.org because the certificates do not mention osm.org and hence
using this domain the tiles fail to load.


danimo:
This should be a trivial one and makes SSL lovers happy. //cc @differentreality @hennevogel

hennevogel:
LGTM. Thank you Daniel

Issue #1057, opened 2016-06-27
open the schedule for the current day by default


muelli:
I'd appreciate it if I could open the schedule and see the current day, rather than the first day of the conference.

PR #1076, opened 2016-07-02
Open schedule for the current day by default

nishanthvijayan:
Solves #1057

### DIFF
--- a/app/views/conference/_schedule.html.erb
+++ b/app/views/conference/_schedule.html.erb
@@ -55,6 +55,12 @@ $(function() {
     }
 });
 
+// Open current day's schedule by default
+$(function() {
+    today = new Date().toISOString().split('T')[0];
+    $('.date-tab[href="#' + today +'"]').trigger( "click" );
+});
+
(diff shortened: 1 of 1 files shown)


app/views/conference/_schedule.html.erb
```diff
@@ -55,6 +55,12 @@ $(function() {
     }
 });
 
+// Open current day's schedule by default
+$(function() {
+    today = new Date().toISOString().split('T')[0];
+    $('.date-tab[href="#' + today +'"]').trigger( "click" );
+});
+
```
Ana06: This doesn't take the conference time zone into account so I doubt that it works properly. 

I was already working on that as I was working in the public schedule already, I have just realised that you too. :confounded:

I have just added a commit to https://github.com/openSUSE/osem/pull/1032 in which the schedule is open in the current day and **time**, taking into account the **time zone**.

nishanthvijayan: :+1:  I'll close this

Issue #1077, opened 2016-07-02
Different colors when creating new tracks


Ana06:
When creating a track the default color for it is always black:

![image](https://cloud.githubusercontent.com/assets/16052290/16540381/3e32541e-4063-11e6-8cb6-d86705840200.png)

It would be nice if it suggest you a different colour for every track, so you don't have to change it if you don't care about the color.

PR #1079, opened 2016-07-02
Different colors for new tracks, types and levels

Ana06:
Every time a new track, event type or difficulty level is created, the color picker has a different color by default. :smile: 

Closes https://github.com/openSUSE/osem/issues/1077

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -595,6 +595,20 @@ class Conference < ActiveRecord::Base
     registration_limit > 0 && registrations.count >= registration_limit
   end
 
+  # Returns an hexadecimal color given a collection. The returned color changed
+  # when the number of element in the collection changes and for consecutive
+  # number of elements it returns highly different colors.
+  def next_color_for_collection(collection)
+    # we have different start indices for every collection to generate a
+    # different color for every of them.
+    start_index = {
+      tracks: (program.tracks.count + 1),
+      levels: (program.difficulty_levels.count + 51),
+      types: (program.event_types.count + 101)
+    }
+    next_color(start_index[collection])
+  end
+
   private
 
   # Returns a different html colour for every i and consecutive colors are
--- a/app/models/difficulty_level.rb
+++ b/app/models/difficulty_level.rb
@@ -3,4 +3,5 @@ class DifficultyLevel < ActiveRecord::Base
   has_many :events, dependent: :nullify
 
   validates :title, presence: true
+  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
 end
--- a/app/models/difficulty_level.rb
+++ b/app/models/difficulty_level.rb
@@ -3,5 +3,13 @@ class DifficultyLevel < ActiveRecord::Base
   has_many :events, dependent: :nullify
 
   validates :title, presence: true
-  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
+  validates :color, format: /\A#[0-9A-F]{6}\z/
+
+  before_validation :capitalize_color
+
+  private
+
+  def capitalize_color
+    self.color = color.upcase if color.present?
+  end
 end
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -7,6 +7,7 @@ class EventType < ActiveRecord::Base
   validates :minimum_abstract_length, presence: true
   validates :maximum_abstract_length, presence: true
   validate :length_step
+  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
 
   alias_attribute :name, :title
 
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -7,7 +7,9 @@ class EventType < ActiveRecord::Base
   validates :minimum_abstract_length, presence: true
   validates :maximum_abstract_length, presence: true
   validate :length_step
-  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
+  validates :color, format: /\A#[0-9A-F]{6}\z/
+
+  before_validation :capitalize_color
 
   alias_attribute :name, :title
 
@@ -21,5 +23,9 @@ class EventType < ActiveRecord::Base
   #
   def length_step
     errors.add(:length, "must be multiple of #{LENGTH_STEP}") if length % LENGTH_STEP != 0
+  end
+
+  def capitalize_color
+    self.color = color.upcase if color.present?
   end
 end
--- a/spec/features/event_types_spec.rb
+++ b/spec/features/event_types_spec.rb
@@ -32,7 +32,7 @@ feature EventType do
       within('table#event_types') do
         expect(page.has_content?('Party')).to be true
         expect(page.has_content?('13042')).to be true
-        expect(page.has_content?('#e4e4e4')).to be true
+        expect(page.has_content?('#E4E4E4')).to be true
         expect(page.assert_selector('tr', count: 3)).to be true
       end
 
--- a/spec/models/event_type_spec.rb
+++ b/spec/models/event_type_spec.rb
@@ -18,6 +18,22 @@ describe EventType do
     it { is_expected.to validate_presence_of(:minimum_abstract_length) }
     it { is_expected.to validate_presence_of(:maximum_abstract_length) }
 
+    it 'is valid if its color is a correct hexadecimal color with 7 characters' do
+      should allow_value('#FF0000').for(:color)
+    end
+
+    it 'is not valid if its color is not a correct hexadecimal color' do
+      should_not allow_value('#AB1H7G').for(:color)
+    end
+
+    it 'is not valid if its color has less than 7 characters' do
+      should_not allow_value('#fff').for(:color)
+    end
+
+    it 'is not valid if its color has more than 7 characters' do
+      should_not allow_value('#123A4567').for(:color)
+    end
+
     describe 'length' do
       it 'validates numericality and greater than 0' do
         is_expected.to validate_numericality_of(:length).is_greater_than(0)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -95,7 +95,7 @@ Metrics/BlockNesting:
   Max: 4
 
 Metrics/ClassLength:
-  Max: 560
+  Max: 570
 
 # avoid redundunt curly braces when it is obvious that hash is used
 Style/BracesAroundHashParameters:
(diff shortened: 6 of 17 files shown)


Ana06:
The default colors in event_types and difficulty levels are in capital letters and the ones that comes from the color picker were not. Now all colors are turned to capital letters when showing them. We could also put the default colors in lower case letters but I think It looks better in capital letters.

Ana06:
I've just realised that there are test that I should change :anguished:

Ana06:
The same test that is failing in https://github.com/openSUSE/osem/pull/1080, it is failing here too because of the same reason. I think we should add validations for colors in the models.

app/models/conference.rb
```diff
@@ -595,8 +595,38 @@ def registration_limit_exceeded?
     registration_limit > 0 && registrations.count >= registration_limit
   end
 
+  # Return a different html color for every new track. Consecutive talks have
+  # highly different colors.
+  def next_track_color
```
ChrisBr: Is it really necessary to have 3 different methods for the (almost) the same functionality?
I propose to either 

- just use next_color (I don't really care if a track and difficulty level have the same color)
- just use next_color and hardcode the start_index when you call the next_color

Three different methods are just too much for my taste ...

ChrisBr:
LGTM if you fix the failing test.
And please squash your commits and the .rubocop change should be in a separate commit!

app/models/conference.rb
Ana06: But if we put then together we lose the fact that following tracks has highly different colors as we could create other things in the middle. It is not that important that a track and a difficult level have the same color but it is important that we create all of them with the same index).

I can call `next_color_for_collection(collection, start_index)` directly in the controller and eliminate the other 3 methods, would that be fine?

Ana06:
When you say that I squash my commits, do you mean turning the three commits into one?

app/models/conference.rb
ChrisBr: I don't think that this is really a big problem!

ChrisBr:
You should group in your commits what logically belongs together, e.g.

Tests fixed & Hexadecimal color changed to capital letters belongs together

Also it is good practice to have only one commit / functionality in one PR as it makes reviewing a lot easier!

Ana06:
I added the _tests fixed_ separately because I think that some of the tests that are doing so because of the first commit. But I can put it together with _Hexadecimal color changed to capital letters_ if you think it is better that way.

I'll try to have only one functionality per PR from now on. :see_no_evil:

Ana06:
@ChrisBr:

> Is it really necessary to have 3 different methods for the (almost) the same functionality?
> I propose to either
> 
> just use next_color (I don't really care if a track and difficulty level have the same color)
> just use next_color and hardcode the start_index when you call the next_color
> Three different methods are just too much for my taste ...

`next_color` needs an index, so to use it for all collection and I would have to use the maximum of the number of elements in all collections or something like that which will complicate it instead of doing it simpler.

I've simplified and now I am using only one function. I could eliminate the start_index if we don't care if the colors are the same for different collections, but I think it is something simple and I would keep it. If you don't like that solution, I think that the best alternative would be using random numbers like that:

```
  def generate_color
    next_color(random_number_beetween_0_and_238)
  end
```

This way we keep the fact that next_color generate beautiful colors and as they are random numbers we can expect that the generated colours are different. This would be the simplest I think.

Ana06:
Instead of doing `upcase` all the time in the view I capitalize the colors before validation. I've also create a rake task to change the old colors.

Is there a way to avoid having the same function `capitalize_color` in the three models?

app/models/conference.rb
```diff
@@ -595,6 +595,20 @@ def registration_limit_exceeded?
...
+  # number of elements it returns highly different colors.
+  def next_color_for_collection(collection)
+    # we have different start indices for every collection to generate a
+    # different color for every of them.
+    collections = {
+      tracks: { collection: program.tracks, start_index: 1 },
+      levels: { collection: program.difficulty_levels, start_index: 51 },
+      types: { collection: program.event_types, start_index: 101 }
+    }
+    next_color(collections[collection][:collection].count + collections[collection][:start_index])
```
ChrisBr: Why don't you store the count in the hash?

```
tracks: { collection: program.tracks.count, start_index: 1 },
```

app/models/difficulty_level.rb
```diff
@@ -3,4 +3,5 @@ class DifficultyLevel < ActiveRecord::Base
   has_many :events, dependent: :nullify
 
   validates :title, presence: true
+  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
```
ChrisBr: I'd really like to see that the validations are in a separate commit! Do you know how to do that?

app/views/admin/event_types/index.html.haml
```diff
@@ -39,4 +39,3 @@
 .row
   .col-md-12.text-right
     = link_to 'Add Event Type', new_admin_conference_program_event_type_path(@conference.short_title), class: 'btn btn-primary'
-
```
ChrisBr: Why is there a whitespace change?

spec/features/event_types_spec.rb
```diff
@@ -32,7 +32,7 @@
       within('table#event_types') do
         expect(page.has_content?('Party')).to be true
         expect(page.has_content?('13042')).to be true
-        expect(page.has_content?('#e4e4e4')).to be true
+        expect(page.has_content?('#E4E4E4')).to be true
```
ChrisBr: Does that belong to this commit? I guess it belongs to 89662894b130c855d8e468e465f4f89add7bd89f, doesn't it?

app/models/difficulty_level.rb
```diff
@@ -3,5 +3,13 @@ class DifficultyLevel < ActiveRecord::Base
   has_many :events, dependent: :nullify
 
   validates :title, presence: true
-  validates :color, format: /\A#[0-9a-fA-F]{6}\z/
+  validates :color, format: /\A#[0-9A-F]{6}\z/
```
ChrisBr: You already have the validation added in 1ba59c2352925a1816a9f1113cf3749e8f869ac6 ?
I'd really like to see it in a separate commit! However, in this commit it makes more sense than in 1ba59c2352925a1816a9f1113cf3749e8f869ac6 ...

spec/models/event_type_spec.rb
```diff
@@ -18,6 +18,18 @@
     it { is_expected.to validate_presence_of(:minimum_abstract_length) }
     it { is_expected.to validate_presence_of(:maximum_abstract_length) }
 
+    it 'is valid if its color is a correct hexadecimal color with 7 characters' do
```
ChrisBr: Also validate with more than 7 characters!

app/models/event_type.rb
```diff
@@ -22,4 +24,8 @@ class EventType < ActiveRecord::Base
   def length_step
     errors.add(:length, "must be multiple of #{LENGTH_STEP}") if length % LENGTH_STEP != 0
   end
+
+  def capitalize_color
+    self.color = color.upcase if color.present?
```
ChrisBr: Maybe we should introduce [normalize attributes gem](https://github.com/fnando/normalize_attributes)

^^ @hennevogel

spec/features/event_types_spec.rb
Ana06: It should be in 8966289, is there a way to move it to the other commit?

ChrisBr: You could do an [interactive rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i):

```
git rebase -i HEAD~3
```

app/views/admin/event_types/index.html.haml
Ana06: When I save a file in atom if there is a empty line at the end it removes it. I'll exclude that from the PR.

hennevogel:
@ChrisBr maybe but not in the scope of this PR :-)

app/models/conference.rb
Ana06: why that would be better?

I would have `.count` three times instead of one.

app/models/difficulty_level.rb
Ana06: The reason why  I put that validation here is that if there is not color or it doesn't have the right format what I have done do not always work properly, and the tests fail.

So, I am not sure if I have to leave it here, move to the next commit or move it to a new commit.

app/models/difficulty_level.rb
Ana06: I changed it here as only capital letters are now allowed.

app/models/conference.rb
ChrisBr: Why not calculate 

``
collection.count + start_index
`` 

in the hash? Doing it that way you don't need a nested hash. Plus you could name the hash ``start_index`` and just call 

```
next_color(start_index[:collection])
```

which makes it more easy to understand?

app/models/difficulty_level.rb
ChrisBr: Ok

ChrisBr:
If answer/change my last comment, it LGTM :+1:

Ana06:
@ChrisBr  done! :wink:

app/models/conference.rb
```diff
@@ -595,6 +595,20 @@ def registration_limit_exceeded?
...
+  # Returns an hexadecimal color given a collection. The returned color changed
+  # when the number of element in the collection changes and for consecutive
+  # number of elements it returns highly different colors.
+  def next_color_for_collection(collection)
+    # we have different start indices for every collection to generate a
+    # different color for every of them.
+    start_index = {
+      tracks: (program.tracks.count + 1),
+      levels: (program.difficulty_levels.count + 51),
+      types: (program.event_types.count +  101)
```
ChrisBr: Double whitespace?
(wondering why rubocop does not complain ...)

app/models/event_type.rb
ChrisBr: Agreed! @Ana06 could you please open an issue after we merged this PR. Could be a junior job....

Ana06: Sure! :smile:

app/models/conference.rb
Ana06: It seems that if you don't add any white space after or before the + it complains. But if add more add one it doesn't complain.

ChrisBr:
LGTM let's :ship: it ...

Issue #1078, opened 2016-07-02
Changing the text color of event types and difficulty levels 


Ana06:
Changing the text color of event types and difficulty levels taking the background color into account. Similar to what it is done with tracks. See issue #1033.

PR #1080, opened 2016-07-02
Text color of event types and difficulty levels

Ana06:
Changing the text color of event types and difficulty levels taking the background color into account. Similar to what it is done with tracks. See issue https://github.com/openSUSE/osem/issues/1033.

Closes https://github.com/openSUSE/osem/issues/1078

### DIFF
--- a/app/views/admin/difficulty_levels/index.html.haml
+++ b/app/views/admin/difficulty_levels/index.html.haml
@@ -20,7 +20,7 @@
             %td
               = difficulty_level.description
             %td
-              %span.label{style: "background-color: #{difficulty_level.color};"}
+              %span.label{style: "background-color: #{difficulty_level.color}; color: #{ contrast_color(difficulty_level.color) };"}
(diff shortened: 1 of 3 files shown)


Ana06:
There is a test failing because the factory creates the `event_type` without color. I could take into account that the color can be nil or have an incorrect format when rendering, but I think that maybe it is better to validate it, I mean doing the color compulsory. In fact, it is not possible to create an `event_type` from the user side without color.

The tests are not failing for `difficulty_levels` and `tracks` because the factory has the color, but there is not validation in the model. So, I think It should be added here too.

ChrisBr:
LGTM

ChrisBr:
If you fix the failing test :rocket:

Ana06:
This text should be fixed with the validation I've added in https://github.com/openSUSE/osem/pull/1079. So when we merge that, I'll rebase here. :smile:

ChrisBr:
LGTM

PR #1082, opened 2016-07-04
Handle cancellation of scheduled events

nishanthvijayan:
Currently if a scheduled event is cancelled/withdrawn,
  - It remains scheduled.The public schedule page will still display the event without any change
  - The event disappears from the admin scheduling interface.
  - If a new event is scheduled in the same room and time of the cancelled event, it won't be shown in the public schedule until the cancelled event is unscheduled (which is difficult as it won't be accessible from the scheduling interface)  
  - There is no way to let attendees know that the event has been cancelled other than manually by  changing the title/abstract.

With these changes, if a scheduled event is cancelled/withdrawn,
  - A red label with text 'CANCELED' will be displayed before its title in the schedule
  - If a new event is scheduled in the same room and time as that of the canceled event
    - The cancelled event will no longer be displayed in the schedule
    - A label with text 'REPLACEMENT' will be displayed before its title in the schedule
    -  'Please note that this talk replaces #{cancelled event's title}' will be shown in the replacement event's proposal page

cancelled_event
replacement

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -38,6 +38,8 @@ class Event < ActiveRecord::Base
   validate :max_attendees_no_more_than_room_size
 
   scope :confirmed, -> { where(state: 'confirmed') }
+  scope :canceled, -> { where(state: 'canceled') }
+  scope :withdrawn, -> { where(state: 'withdrawn') }
   scope :highlighted, -> { where(is_highlight: true) }
 
   state_machine initial: :new do
@@ -228,6 +230,13 @@ class Event < ActiveRecord::Base
   #
   def end_time
     self.start_time + self.event_type.length.minutes
+  end
+
+  ##
+  # Returns events that are scheduled in the same room and start_time as event
+  #
+  def intersecting_events
+    room.events.where(start_time: start_time).where.not(id: id)
   end
 
   private
--- a/app/views/conference/_schedule_item.html.haml
+++ b/app/views/conference/_schedule_item.html.haml
@@ -4,7 +4,14 @@
   %a.unstyled-link{href: url_for(conference_program_proposal_path(@conference.short_title, event.id))}
     %div{ class: "elipsis break-words event-title", |
           style: "-webkit-line-clamp: #{ event_lines(@rooms) }; height: #{ event_height(@rooms) }px;"} |
+
+      - if event.state == 'canceled' || event.state == 'withdrawn'
+        %span.label.label-danger.schedule-label CANCELED
+      - elsif event.state == 'confirmed' && (!event.intersecting_events.canceled.empty? || !event.intersecting_events.withdrawn.empty?)
+        %span.label.label-info.schedule-label REPLACEMENT
+
       = event.title
+
     - if speaker = event.speakers.first
       = image_tag speaker.gravatar_url, :class => "img-circle pull-right speaker-pic", |
                                         :alt => speaker.name, |
--- a/app/views/proposal/show.html.haml
+++ b/app/views/proposal/show.html.haml
@@ -14,6 +14,12 @@
             = link_to "Edit", edit_conference_program_proposal_path(@conference.short_title, @event), :class => "btn btn-mini btn-primary"
           - if can? :schedule, @conference
             = link_to "Schedule", schedule_conference_path(@conference.short_title), :class =>"btn btn-success"
+
+      - if @event.state == 'canceled' || @event.state == 'withdrawn'
+        %span.label.label-danger CANCELED
+      - elsif @event.state == 'confirmed' && (!@event.intersecting_events.canceled.empty? || !@event.intersecting_events.withdrawn.empty?)
+        %span.label.label-info REPLACEMENT
+
   .row
     .col-md-3
       .speakerinfo
@@ -34,6 +40,16 @@
     .col-md-9
       .row
         .col-md-12
+          .lead
+            - if @event.state == 'confirmed' && !@event.intersecting_events.withdrawn.empty?
+              = "Please note that this talk replaces"
+              = link_to @event.intersecting_events.withdrawn.first.title,
+                    conference_program_proposal_path(@conference.short_title, @event.intersecting_events.withdrawn.first.id)
+            - elsif @event.state == 'confirmed' && !@event.intersecting_events.canceled.empty?
+              = "Please note that this talk replaces"
+              = link_to @event.intersecting_events.canceled.first.title,
+                    conference_program_proposal_path(@conference.short_title, @event.intersecting_events.canceled.first.id)
+
           - if @event.commercials.empty?
             %h5.text-warning
               No video of the event yet, sorry!
--- a/app/assets/stylesheets/osem-schedule.css.scss
+++ b/app/assets/stylesheets/osem-schedule.css.scss
@@ -193,6 +193,14 @@ td.no-padding{
   color:  #D8D8D8 !important;
 }
 
+.schedule-label{
+  height: 14px;
+  line-height: 11px;
+  overflow: hidden;
+  display: inline-block;
+  white-space: normal;
+}
+
 /* Small devices (tablets, 768px and up) */
 @media (min-width: 768px) {
   .room, .event-title{
@@ -226,6 +234,11 @@ td.no-padding{
 
   td.event{
     padding: 3px 5px !important;
+  }
+
+  .schedule-label{
+    height: 15px;
+    line-height: 12px;
   }
 }
 
--- a/app/views/conference/_carousel.html.haml
+++ b/app/views/conference/_carousel.html.haml
@@ -33,6 +33,11 @@
                     - span -= 1
                   - else
                     - event = events.find{|e| e.start_time <= start_room_time and e.end_time > start_room_time}
+
+                    - if event && (event.state == 'canceled' || event.state == 'withdrawn') && !event.intersecting_events.confirmed.empty?
+                      - replacement_event = event.intersecting_events.confirmed.first
+                      - event = (replacement_event.start_time <= start_room_time && replacement_event.end_time > start_room_time) ? replacement_event : nil
+
                     - if event
                       / There is an event, calculate the span and show it
                       - event_span =  (event.end_time.to_i - start_room_time.to_i) / 60 / EventType::LENGTH_STEP


app/models/event.rb
```diff
@@ -153,6 +153,25 @@ def process_rejection(options)
     end
   end
 
+  def process_cancellation
+    return unless scheduled?
```
ChrisBr: I've to admit that I don't like to change the data here!
What about just displaying this different in the view?
@hennevogel @differentreality ^^

app/models/event.rb
```diff
@@ -153,6 +153,25 @@ def process_rejection(options)
...
+    abstract << "\nPlease note that this talk has been cancelled"
+    title.prepend('[CANCELLED] ')
+    self.without_versioning do
+      self.save!
+    end
+  end
+
+  def process_resubmit
```
ChrisBr: Same here!

ChrisBr:
Please also talk with @Ana06 as it seems your PR interfere a little bit!

Ana06:
I have changed the schedule definition, as I only consider that an event is scheduled if its state is confirmed and it has room and start_time. So, what @nishanthvijayan says:

> If a new event is scheduled in the same room and time as that of the cancelled event the cancelled event will then be unscheduled automatically

won't be true with that definition, but I think that it still make sense. An event can be unscheduled because it is cancelled, if it is confirmed again it will be scheduled in its previous time and room. What happens if  a new event is scheduled in the same room and time as that of the cancelled event the cancelled event will then lose its room and time.

So, it doesn't interfere really as for me everything works well. The only thing we may change is that users are allowed to see the event page of a cancelled event and also to add the cancelled events in another section of the all events. I don't like showing cancelled events in the schedule even if they have the _CANCELLED_ text prepended, but I would show then in the all events section, to allow users to know what events have been cancelled.

If @nishanthvijayan agree I can add the new section in the all events and change the ability file and I think we can finish that as there is not anything that really interfere.

nishanthvijayan:
@Ana06 I think it would be less confusing for users if we just show the cancelled event in schedule with [CANCELLED].Its straightforward. 
@differentreality?

Ana06:
@nishanthvijayan I don't agree. An event can be cancelled because of many reason and even much before that the conference start. If it doesn't appear you assume that it has been cancelled. And also if a lot of events has been cancelled the schedule would be really crowded although there are not a lot of events. I think that when you more open the schedule is when you are at the conference and that it is better not to show what it is not useful. 

I think that they should appear somewhere, but not in the schedule. I'll show then in the all events section, as we do with unscheduled events.

hennevogel:
I agree with Christian. Do no change data, do not replace events automatically. Just check the status of the Event when rendering the schedule and tell users about it. BTW there are more obvious ways than changing the title/description, a [label](http://getbootstrap.com/components/#labels) or [the title in strike-through](http://getbootstrap.com/css/#strikethrough-text) for instance...

Ana06:
but @hennevogel do we want to show cancelled events in the schedule? even if we inform that they are cancelled, if there are cancelled they are not scheduled any more.

Currently, I am not showing then in #1032. I can show then if you think that they are important.

hennevogel:
@Ana06 people can cancel/withdraw their events in the middle of everything when organizers can't handle this. Visitors then show up to the talk and don't know whats going on...

Ana06:
okey @hennevogel then I'll change the `scheduled` definition to have a room and a start_time. So @nishanthvijayan can continue working on that, although I haven't finished yet with the schedule as #1032 is still open, so maybe there will be some merge conflicts if he needs to change things there.

I have a doubt, what do we want to show in the unscheduled events section of the all events then? We can not say that unscheduled is the opposite of scheduled because that would include events that have never been confirmed. The current definition is events which are confirmed but which do not have start_time or room. What happens if an unscheduled event is cancelled? Do we want to include cancelled events here?

differentreality:
I don't think we need to care about unscheduled events being cancelled that much, because they were never assigned a room and a time. Sure we are showing a list of confirmed events, just to give people an idea of what to expect prior we are done with scheduling, but unless an event makes it into the schedule there won't be such big confusion to attendees (eg. go to a room at a certain time and find no one).
The issue is when an event is assigned room and time and in the middle of the conference it is suddenly cancelled. This is an issue that we should consider from the conference organizer's point of view. Is not strictly a programming issue but a rather meaningful one. 
Having events disappear from the schedule in the middle of the conference does not convey the right information to attendees. They are most likely to think our schedule has broken, rather than the event was cancelled.

As far as definitions of scheduled/unscheduled events are concerned, let's remember this is just variables that can hold the values we want them to have :) 
So you show what makes sense to show. If needed we will define scheduled_confirmed or something else that makes sense for us.

Labels for cancelled and substitute talk sound like a good idea, even more visual.

@Ana06 your PR is bigger, would it be helpful for you if we wait until #1032 is merged before merging this one? Difficult rebases can be painful.

Ana06:
okey, then I'll keep the cancelled events in the schedule ans show only the confirmed ones as unscheduled in the all events.

@differentreality currently I think there wouldn't be any merge problem (as the modified files are not the one of the schedule), but if your are going to add labels in the schedules there would be them. But as the schedule is completely changed in #1032 I do not know if it make sense to do it before merging it, as I would have to be redone it when rebasing. Also, I need to have #1032 merge as soon as possible, as I need it when changing the relations for allowing multiples schedules. Although If you prefer me to wait I could work on other things before.

Ana06:
I've just seen in IRC that @nishanthvijayan is already working on labels, so I'll wait until he finish that and then I'll solve merge conflicts in #1032.

app/controllers/admin/schedules_controller.rb
```diff
@@ -58,6 +58,15 @@ def update
       # start_time = DateTime.strptime(time + zone.formatted_offset, "%Y-%m-%d %k:%M %Z")
       start_time = DateTime.strptime(time, '%Y-%m-%d %k:%M')
       event.start_time = start_time
+
+      # Check if there are any cancelled events scheduled that event intesects
+      blocking_event = room.events.where(start_time: event.start_time..(event.end_time - 1)).first
+      if blocking_event && blocking_event.state == 'canceled'
+        event.title.prepend('[AMENDMENT] ')
+        event.abstract << "\nPlease note that this talk replaces #{blocking_event.title}"
```
ChrisBr: As discussed, please do not alter title and abstract! Use labels instead!

ChrisBr:
Thought we already discussed that we don't want to alter title & abstract?

> The new (replacement) event's title will be prepended with '[AMENDMENT]'
> 'Please note that this talk replaces #{cancelled event's title}' will be added to its abstract

Please also use labels here!

app/views/conference/_schedule_item.html.haml
```diff
@@ -5,6 +5,10 @@
     %div{ class: "elipsis break-words event-title", |
           style: "-webkit-line-clamp: #{ event_lines(@rooms) }; height: #{ event_height(@rooms) }px;"} |
       = event.title
+
+      - if event.state == 'canceled'
+        %span.label.label-danger CANCELLED
```
differentreality: Let's stick with the American spelling `canceled` for uniformity reasons with our event state name.

RichiH:
@differentreality I agree with your distinction on if an event has been assigned a time and location or not. One thing that might clutter the overview is when the same slot is re-used for something else. In this case, replacing the initial assignment might be better.

Also, something I never saw at any conference, but which might be useful, is a list of cancelled, or rescheduled, events. This would allow looking for changes in one canonical places; especially useful during busy conferences.

differentreality:
@RichiH what do you mean about the same slot being used for something else? If there was supposed to be Presentation X at 9:30 am and now there is sth else, that's what we want to address and make clear in the schedule, so that people that saw the previous version of the schedule don't show up at 9:30 am for Presentation X. 

List of canceled events would be useful indeed, but not very relevant to this PR. It should be part of reports that can be exported. Have added it to the list :)

differentreality:
I wonder if we should label the newly scheduled events as `AMENDEMENT` or `REPLACEMENT`.

differentreality:
The events can be either canceled by organizers, or withdrawn by submitters.
I'd also suggest we put the label in front of the event title.

RichiH:
@differentreality labelling them is a good idea, but it should probably always be the same external reason, no matter why something was canceled.

Also, you are right, I misinterpreted what @nishanthvijayan said in the initial comment; sorry for the confusion.

nishanthvijayan:
@differentreality I've changed it to consider withdrawn events as well.
The labels are now CANCELED (for both canceled & withdrawn) and REPLACEMENT.
Is there anything else to be done here?

app/views/proposal/show.html.haml
```diff
@@ -1,3 +1,8 @@
+- if @event.state == 'confirmed' && !@event.intersecting_events.withdrawn.empty?
+  - @event.abstract.prepend("Please note that this talk replaces #{@event.intersecting_events.withdrawn.first.title}\n")
```
ChrisBr: Why do you still alter the abstract? 
Just hardcode it in the html?

app/views/proposal/show.html.haml
```diff
@@ -1,3 +1,8 @@
...
+- elsif @event.state == 'confirmed' && !@event.intersecting_events.canceled.empty?
+  - @event.abstract.prepend("Please note that this talk replaces #{@event.intersecting_events.canceled.first.title}\n")
```
ChrisBr: Same here!

Ana06:
I've tried what happens with the REPLACEMENT tag in a 300x600 screen, and it breaks the table:

image

It also has problems when the title starts in the next line independently of the screen size:

image

nishanthvijayan:
@Ana06 breaks the table? I didn't understand.Can you be more specific as to what's wrong in the above images?

Ana06:
@nishanthvijayan all the columns with the time at the top should have the same width. That is what happens without the tag. 

If you have four consecutive events with the tag the table won't enter in the screen.

I think it is not too difficult to solve. In the cells I used css to hide and short the title if It didn't fit the cell size. So, something similar can be done for the tags (short them if there is not enough space). Here there is what I used in case that you want to use it:

https://github.com/openSUSE/osem/blob/master/app/assets/stylesheets/osem-schedule.css.scss#L97

And I also used `-webkit-line-clamp` and `height` css attributes, but they are dynamically calculated and that is why they do not appear in the `.css`, but I think that for the tag we only want to show one line.


In the second image the first event at 9:15 it is not displayed correctly, as the height has changed and you can see a half of the next line of the title that used to be hidden.

Ana06:
Now the labels don't break the table. :tada:

There is still something that I don't really like. For 15 minutes events in a screen with a width less than 450px (specific case, probably very rare), the tag uses two lines:

replacement

The problem is that the title has only one line, so it is difficult to know what the talk is about. Also, splitting the tag in two lines doesn't look very nice. Personally I would prefer two lines of title and a shorted tag that only uses one line. But I am not sure that users will understand it and taking into account that it is a very rare case, maybe it is better to keep it as it is now.

nishanthvijayan:
@Ana06 I've changed it to use ellipsis to truncate the label text when needed.Can you take a look at it?

Ana06:
@nishanthvijayan the only problem of what you have done is that you need to use `table-layout: fixed;` and that makes all columns of the same width, and now the room column doesn't have the same width (only the ones with the time). So it seems that it is not that easy to use the ellipsis... 

Okey, I have been trying other things and I like this one, hidden the text we do not want to show:

image

image


What do you think @nishanthvijayan? If you like it, here is the `css` I used:

``` css
.schedule-label{
  height: 14px;
  line-height: 11px;
  overflow: hidden;
  display: inline-block;
  white-space:normal;
}
```

And I needed to adjust the height and line height for devices with at least 992px width:

``` css
@media (min-width: 992px) {
  .schedule-label{
    height: 15px;
    line-height: 12px;
  }
```

nishanthvijayan:
Yeah that looks great.Thanks for the help!

differentreality:
In smaller size I see the whole label CANCELED, but then in larger size I don't. That doesn't seem right. @Ana06 is this due to your schedule design?

I wouldn't like to do any more front-end design for the schedule in this PR. If there are things to be improved, they can be introduced on a new PR, or perhaps @Ana06 might have some valuable input on it.

differentreality:
I believe that the same logic we have for the schedule, we also need to introduce in other places, such as the proposal#show page, as that page also provides details about time and room and can be confusing for the attendees when looking at it. 
@nishanthvijayan please check if we should make the information available in other places.

differentreality:
We can see what shows on the schedule in 
https://osem-demo-pr-1082.herokuapp.com/conference/osemdemo/schedule

@nishanthvijayan a screenshot of latest view is always helpful in the PR, in my opinion.

app/views/proposal/show.html.haml
```diff
@@ -47,6 +47,15 @@
       .row
         .col-md-12
           %p
+            - if @event.state == 'confirmed' && !@event.intersecting_events.withdrawn.empty?
```
differentreality: This is not at all visible here. I wonder how we could make it more obvious to visitors.

Ana06:
@differentreality Yes, what I suggested is to short the tag to allow it to use only one line. As the font size is bigger for big screens that's why it short it for big devices and not for small ones. The alternative is to allow the tag to use more than one line and then the title get too short. I think people will understand it because of the colors. And if you add the tag it in the proposal#show page it would be even more clear.

Ana06:
It will be helpful if we add the tag in the all events section of the schedule. But that section has been completely changed in #1032, which hasn't been merged because this PR has to be merged before, so I wouldn't do it now. If you do it I'll have to redone it when merging.

app/views/proposal/show.html.haml
nishanthvijayan: - This could be moved to the [top of the page](https://github.com/nishanthvijayan/osem/blob/5585eafbcd6e6db3679552c54101e1ea3bf17dec/app/views/proposal/show.html.haml#L37)
 - Like Ana suggested, the coloured label could be displayed here as well (before the title)

differentreality: Let's see it :)

differentreality:
@Ana06 we will wait until #1032 is merged and then finalize this PR.

nishanthvijayan:
proposal_show_cancaled

proposal_show_replacment

hennevogel:
@nishanthvijayan  Please rebase

hennevogel:
Okay let's get this in! Thanks everyone :-)

PR #1084, opened 2016-07-04
Modify conference registration to a singular resource

### LINKED ISSUE #954: conference_registration should be singular resource
[Source of previous discussion.](https://github.com/openSUSE/osem/pull/795#discussion_r54597070)

Corresponding to a conference and a user, there can be only one registration, however our [routes](https://github.com/openSUSE/osem/blob/master/config%2Froutes.rb#L110) define it as plural resource:
```ruby
 resource :conference_registrations, path: 'register'
```
This should be: ` resource :conference_registration, path: 'register'`
We would also want to change the name of our controller from [ConferenceRegistration**s**Controller ](https://github.com/openSUSE/osem/blob/master/app%2Fcontrollers%2Fconference_registrations_controller.rb)to `ConferenceRegistrationController`

arun1595:
Fixes #954

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -10,7 +10,7 @@ class ConferenceRegistrationsController < ApplicationController
 
     # Redirect to registration edit when user is already registered
     if @conference.user_registered?(current_user)
-      redirect_to edit_conference_conference_registrations_path(@conference.short_title)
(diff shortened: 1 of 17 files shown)


differentreality:
:+1:

ChrisBr:
@differentreality :+1: means we can merge?

differentreality:
Oops yes, I should have clicked merge!

PR #1085, opened 2016-07-05
Name validation repeated in Track

Ana06:
The validation to check that the name is present in track.rb is twice.


differentreality:
We really  wanted to validate! 👍

PR #1088, opened 2016-07-06
fix #1054: modified redirect path during registration

### LINKED ISSUE #1054: Register with Ichain: redirect to login page

shlok007:
Solution to issue #1054


hennevogel:
new_user_session_path doesn't exist with https://github.com/openSUSE/devise_ichain_authenticatable

PR #1089, opened 2016-07-06
Redirect user to the login page while registering with ichain

### LINKED ISSUE #1054: Register with Ichain: redirect to login page

arun1595:
Fixes #1054

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -14,7 +14,7 @@ class ConferenceRegistrationsController < ApplicationController
       return
     # ichain does not allow us to create users during registration
     elsif (ENV['OSEM_ICHAIN_ENABLED'] == 'true') && !current_user
-      redirect_to root_path, alert: 'You need to sign in or sign up before continuing.'
+      redirect_to new_user_ichain_session_path, alert: 'You need to sign in or sign up before continuing.'
       return
     end
 


differentreality:
@arun1595 have you tried this locally?

arun1595:
I tried to test it locally by creating a conference and setting up a registration period. But when I try to register(as a regular user) with the url `/conference/conference_name/register/new`, I get a flash message that registration limit is exceeded or the registration isn't open yet. I am sure the registration period is not over but how do we track if the registration limit is exceeded?

I would like to know the workflow on how to register for a conference and to track registration limit.

arun1595:
I didn't set the `ENV[`OSEM_ICHAIN_ENABLED`]`

arun1595:
I did some digging and I found that `CanCan::AccessDenied` exception is raised when a guest user tries to perform a new action on `@registration`. But however, when I looked in the ability.rb file, I don't see any abilities defined for actions on `Registration` when `ENV['OSEM_ICHAIN_ENABLED']` is set to true.

In the `not_signed_in` method in `ability.rb`, every ability on `Registration` is defined when `ENV['OSEM_ICHAIN_ENABLED']` is set to false.

As far as my understanding, I am not sure how we can check whether the user is redirected to `new_user_ichain_session_path` because the user's not authorized in the first place.

Correct me if any of my assumptions is wrong.

differentreality:
Now you are on the right track of what is going on and why your code is not working.
When ICHAIN is enabled we cannot create a new user during registration. The ability is correctly not allowing that. Which is precisely why we want to make the interface a bit more user friendly and direct the user towards the right direction (ie. to sign in).

You have to think about the workflow you want here. What do you want to happen? Once you clarify what you want to implement, then you can go ahead and code it.

differentreality:
Ping @arun1595

arun1595:
I'll work on this and update the PR soon.

differentreality:
I am closing this due to inactivity. Feel free to reopen a new PR when you are ready to work on it!

PR #1093, opened 2016-07-08
Make abstract optional

### LINKED ISSUE #1092: proposal#new with empty abstract is not possible
Even if the abstract is set to 0 words minimum, the user still cannot submit the form because we have marked the field as required (the box telling the user to fill it out appears at all times).

arun1595:
Fixes #1092
Removed client-side and server-side validation for `abstract`


differentreality:
But that is **not** what we want!! No we don't validate, even if we should! And most of the times we should. 9 out of 10 times the abstract will be required.

PR #1094, opened 2016-07-08
Schedule versioning

Ana06:
**Organizers can create multiple proposals for schedules in the backend, only 1 schedule can be selected and that's the conference schedule.**

I have created the new associations and models needed for the schedule versioning, illustrated in the following image:

nuevodocumento 5_1

The `Schedule` and `EventSchedule` models are new. `EventSchedule` is where the time and room is saved. So, for every `Schedule` and `Event` we can have one `EventSchedule `allowing that an Event is scheduled in different rooms and times for every schedule.

Also, the `selected_schedule` in program indicates what of the schedules that belongs to that program is the public one.


Ana06:
I've added the integration of the schedule together with the necessary changes to make use of the introduced associations and models (in a separate commit to make reviewing it easier):

image

image


I will continue the rest of the schedule integration (adding a save button, having several schedules, etc) in another branch (maybe one branch for each part) using this branch as a starting point . But we need that part here, otherwise the admin schedule wouldn't work until I finished the whole integration. That way we can merge that before.

I still have to adapt the public schedule to make use of the new associations and models and to fix some test after that too. I was waiting to merge #1032 but I think I am going to do it now to avoid having to wait, and I'll solve merge conflicts afterwards.

Ana06:
Why the commits in the PR are not in the same order than in my computer? :confounded: 

image

Ana06:
I think that everything is working now and that the test are fixed. :wink:

lib/tasks/several_schedules.rake
```diff
@@ -0,0 +1,24 @@
+namespace :data do
+  desc 'Move the start_time and romm atributes from Event to EventSchedule'
```
ChrisBr: Typo: room attributes

lib/tasks/several_schedules.rake
```diff
@@ -0,0 +1,24 @@
...
+                               schedule: schedule,
+                               start_time: event.start_time,
+                               room_id: event.room_id)
+          event.start_time = nil
+          event.room_id = nil
+          event.save
+        end
+      end
+    end
+    puts 'The start_time and room atributes has been moved from Event to EventSchedule'
```
ChrisBr: Typo/spelling: attributes

app/assets/stylesheets/osem.css.scss
```diff
@@ -86,3 +86,51 @@ p.comment-body {
 #proposal-info div dt, #proposal-info div dd {
   display: inline-block;
 }
+
+.room-name{
```
ChrisBr: Do you need these CSS in several views or why is it not e.g. in schedule.css ?

app/controllers/admin/schedules_controller.rb
```diff
@@ -31,14 +31,17 @@ def update
...
       if params[:date] == 'none'
-        event.start_time = nil
-        event.room = nil
-        event.save!
+        event_schedule.destroy if event_schedule.present?
         render json: { 'status' => 'ok' }
         return
       end
+
+      event_schedule = EventSchedule.create(event: event, schedule_id: params[:schedule]) unless event_schedule.present?
```
ChrisBr: Why do you create a new EventSchedule in the #update action?

ChrisBr: Furthermore, if you do:

``
event.event_schedules.create(schedule_id: params[:schedule])
``
you don't need to pass event as a parameter and it's more explicit!

ChrisBr: But in general, I don't like that you do #create and #update in one action, would prefer to split that.

app/assets/stylesheets/osem.css.scss
Ana06: Because that is for the admin schedule and I thought that osem-schedule.css was only for the public schedule. Do I move it?

app/controllers/conference_controller.rb
```diff
@@ -13,8 +13,10 @@ def show; end
 
   def schedule
     @rooms = @conference.venue.rooms if @conference.venue
-    @events = @conference.program.events
-    @events_xml = @events.scheduled.order(start_time: :asc).group_by{ |event| event.start_time.to_date }
+    program = @conference.program
+    @events = program.events
+    @events_xml = program.schedules.find(program.selected_schedule).event_schedules.where('start_time IS NOT NULL AND room_id IS NOT NULL').order(start_time: :asc)
```
ChrisBr: Uff, that's a very long statement ...
What about a scope or function?

app/models/event.rb
```diff
@@ -16,9 +16,9 @@ class Event < ActiveRecord::Base
 
   has_many :events_registrations
   has_many :registrations, through: :events_registrations
+  has_many :event_schedules, dependent: :destroy
```
ChrisBr: What about a 
```
has_many :schedule, through: event_schedules
```

association?

app/controllers/admin/schedules_controller.rb
Ana06: But now `room` and `start_time` attributes are in `EventSchedule `and not in `Event`, so if the `event_schedule` is not created for this `schedule` I can not assign the attributes, so I think I have to create it here, when it is scheduled.

app/models/program.rb
```diff
@@ -26,8 +27,12 @@ def confirmed
       where(state: :confirmed)
     end
 
-    def scheduled
-      where.not(start_time: nil).where.not(room: nil)
+    def scheduled(schedule_id)
```
ChrisBr: Why not a scope?

ChrisBr: http://guides.rubyonrails.org/active_record_querying.html#scopes

app/models/event.rb
```diff
@@ -224,19 +233,33 @@ def calculate_progress
   end
 
   ##
-  # Returns end of the event
+  # Returns the room in which the event is scheduled
   #
-  def end_time
-    self.start_time + self.event_type.length.minutes
+  def scheduled_room
+    event_schedule = event_schedules.find_by(schedule_id: program.selected_schedule) if program.present? && program.selected_schedule.present?
```
ChrisBr: This schouldn't be necessary:
``
``

ChrisBr: What about:

``
event_schedules.find_by(schedule_id: program.try(:selected_schedule)).try(:room)
``

app/models/event.rb
```diff
@@ -224,19 +233,33 @@ def calculate_progress
...
+    event_schedule.room if event_schedule.present?
+  end
+
+  ##
+  # Returns the start time at which this event is scheduled
+  #
+  def scheduled_start_time
+    event_schedule = event_schedules.find_by(schedule_id: program.selected_schedule) if program.present? && program.selected_schedule.present?
```
ChrisBr: You do several times
``
event_schedules.find_by(schedule_id: program.selected_schedule)
``
What about introducing a scope for that?

app/views/admin/schedules/show.html.haml
```diff
@@ -1,48 +1,166 @@
...
+          %a{ href: "##{date}" }
+            = date
+    .tab-content
+      - @dates.each do |date|
+        .tab-pane{ class: "#{ (@dates.first == date) ? 'active' : '' }", id: "#{date}" }
+          = render partial: 'day_tab', locals: { date: date }
+
 
 :javascript
+
```
ChrisBr: You should move the JS in a dedicated .js file. Inline JS is bad style and hard to read / review!

db/migrate/20160705110711_add_start_time_to_event_schedule.rb
```diff
@@ -0,0 +1,5 @@
+class AddStartTimeToEventSchedule < ActiveRecord::Migration
```
ChrisBr: Why is that in a separate migration?

spec/factories/event_schedule.rb
```diff
@@ -0,0 +1,22 @@
+# Read about factories at https://github.com/thoughtbot/factory_girl
+
+FactoryGirl.define do
+  factory :event_schedule do
+    event
+    after(:build) do |event_schedule|
```
ChrisBr: Uff, there is a lot of logic in the factory. Is that really all necessary to get a valid factory?

app/models/event.rb
Ana06: Do I also change it in `Room`?

Ana06: Why do we want that association?

db/migrate/20160705110711_add_start_time_to_event_schedule.rb
Ana06: Because I forgot it in the previous one. Should I rollback the previous one and add it in that one?

spec/factories/event_schedule.rb
Ana06: I think so. :pensive:
In some tests we only care about the room and in some other about the start_time, so we do not have to create it here. It has no sense creating a room without venue. And then if we are creating an event_schedule in the test is always because we are scheduling it, so it must belong to the selected_schedule, and if it doesn't exist it is created.

db/migrate/20160705110711_add_start_time_to_event_schedule.rb
ChrisBr: Yes, please!

app/models/event.rb
```diff
@@ -39,6 +39,7 @@ class Event < ActiveRecord::Base
 
   scope :confirmed, -> { where(state: 'confirmed') }
   scope :highlighted, -> { where(is_highlight: true) }
+  scope :schedule, ->(schedule_id) {  event_schedules.find_by(schedule_id: (schedule_id.present? ? schedule_id : program.try(:selected_schedule))) }
```
ChrisBr: I would even split that in two different scopes!

ChrisBr: And it is not a schedule, it's the event schedule.

app/models/event.rb
```diff
@@ -224,19 +233,26 @@ def calculate_progress
   end
 
   ##
-  # Returns end of the event
+  # Returns the room in which the event is scheduled
+  #
+  def scheduled_room
```
ChrisBr: Why do we need to wrap a simple attribute call in a function?
Same for ``scheduled_start_time``

db/migrate/20160705110711_add_start_time_to_event_schedule.rb
ChrisBr: Yes please

app/assets/stylesheets/osem.css.scss
hennevogel: All of the files get merged into one large CSS file so all CSS is for everything. It's just a matter of better naming for humans :-)

ChrisBr: And where I have to look when I want to change sth ;)

app/models/event.rb
Ana06: This is wrong. I think it can not be an scope, because scopes are for calls on models or association object, but the event_schedules is of an specific event. I'll move that to a private function in Event.

spec/factories/event_schedule.rb
hennevogel: There is a simple rule to follow when you create a factory. Only create what is really necessary and do not try to cover all circumstances at once. If some test cares about some specific aspects they should take care of setting those.

app/models/event.rb
Ana06: Because there are many places where we did `event.room` so doing `event.scheduled_room` we keep everything the same. And at the beginning the function `event_schedule` was private as it was only used for the `scheduled_room` and `scheduled_start_time `but now I am using it in more places and it is public. So, I can eliminate `scheduled_room` and `scheduled_start_ time` functions and do everywhere `event. event_schedule.try(:room)`. Do I change it then?

app/models/event.rb
```diff
@@ -242,26 +250,30 @@ def calculate_progress
...
+    event_schedule.try(:room)
   end
 
   ##
-  # Returns events that are scheduled in the same room and start_time as event
+  # Returns the start time at which this event is scheduled
   #
-  def intersecting_events
-    room.events.where(start_time: start_time).where.not(id: id)
+  def scheduled_start_time
```
ChrisBr: Why do you wrap the attributes in functions?

app/models/event.rb
```diff
@@ -242,26 +250,30 @@ def calculate_progress
   end
 
   ##
-  # Returns end of the event
+  # Returns the room in which the event is scheduled
   #
-  def end_time
-    self.start_time + self.event_type.length.minutes
+  def scheduled_room
+    event_schedule.try(:room)
```
ChrisBr: Maybe you should add some validations to EventSchedule ;) Then you don't need the .try everywhere!

app/models/event_schedule.rb
```diff
@@ -0,0 +1,27 @@
+class EventSchedule < ActiveRecord::Base
+  belongs_to :schedule
+  belongs_to :event
+  belongs_to :room
```
ChrisBr: Maybe you should add some validations for schedule, room and event. Then you don't need the .try everywhere and I'm not sure if an EventSchedule without these associations makes sense!

db/migrate/20160704092814_add_selected_schedule_to_program.rb
```diff
@@ -0,0 +1,5 @@
+class AddSelectedScheduleToProgram < ActiveRecord::Migration
+  def change
+    add_column :programs, :selected_schedule, :integer # Selected schedule ID
```
ChrisBr: Please move that to 
```
 db/migrate/20160704091928_create_schedules.rb 
```

app/controllers/admin/schedules_controller.rb
```diff
@@ -7,10 +7,11 @@ class SchedulesController < Admin::BaseController
     load_resource :venue, through: :conference, singleton: true
 
     skip_before_filter :verify_authenticity_token, only: [:update]
-    layout 'schedule'
 
     def show
-      authorize! :update, @program.events.new
+      event = @program.events.new
```
ChrisBr: Can we split the :show action in a :show, :create, :update and :delete action?

app/models/event.rb
Ana06: Because there are many places where we did event.room so doing event.scheduled_room we keep everything the same. And at the beginning the function event_schedule was private as it was only used for the scheduled_room and scheduled_start_timebut now I am using it in more places and it is public. So, I can eliminate scheduled_room and scheduled_start_ time functions and do everywhere event. event_schedule.try(:room). Do I change it then?

app/models/event.rb
Ana06: but `event_schedule` is only present if the event is scheduled

app/controllers/admin/schedules_controller.rb
Ana06: There is already an :update action. I think I don't understand what you mean :confused:

app/controllers/admin/schedules_controller.rb
```diff
@@ -7,10 +7,11 @@ class SchedulesController < Admin::BaseController
...
+      authorize! :update, event
+      event.destroy
```
ChrisBr: Huh? In line 12 you instantiate the event and in 14 you destroy it?

app/controllers/admin/schedules_controller.rb
ChrisBr: Sorry, missed to expand the code ...

However, this looks to me like it belongs to a create action:

```
unless @program.selected_schedule.present?
  schedule = @program.schedules.create
  @program.selected_schedule = schedule.id
  @program.save!
end
@schedule_id = @program.selected_schedule
```

ChrisBr: And why not authoring also in a before filter?

app/controllers/admin/schedules_controller.rb
```diff
@@ -31,14 +39,17 @@ def update
         error_message = "Could not find event GUID: #{params[:event]}"
```
ChrisBr: Maybe we should change the param to ```#{params[:guid]}```

app/controllers/admin/schedules_controller.rb
```diff
@@ -31,14 +39,17 @@ def update
...
       end
 
+      event_schedule = event.event_schedules.find_by(schedule_id: params[:schedule])
```
ChrisBr: For me this action looks more like it would belong to a EventSchedules Controller!
Also the :update action covers now :create, :update and :destroy ...

app/controllers/conference_controller.rb
ChrisBr: This looks like a Schedule#index action. 
@hennevogel ^^

app/models/event.rb
ChrisBr: Hmm, @hennevogel what do you think?

app/models/event.rb
```diff
@@ -242,26 +250,30 @@ def calculate_progress
...
+  # Returns the start time at which this event is scheduled
   #
-  def intersecting_events
-    room.events.where(start_time: start_time).where.not(id: id)
+  def scheduled_start_time
+    event_schedule.start_time
+  end
+
+  def event_schedule(schedule_id=nil)
+    event_schedules.find_by(schedule_id: (schedule_id.present? ? schedule_id : program.try(:selected_schedule)))
```
ChrisBr: I would split that in 

```selected_event_schedule``` and ```event_schedule```

and as yesterday discussed, I would try to do it with an association!

File: app/controllers/admin/schedules_controller.rb (hunk 20 above)
Comment (Ana06) [inline app/controllers/admin/schedules_controller.rb]: Yes, I instantiate the event to keep the authorization as it was, but then I destroy it because otherwise when I do `@program.save!` in line 29 I save the new event. Well, the validations fail, so I do not save it but it doesn't work.

Comment (ChrisBr) [inline app/controllers/admin/schedules_controller.rb]: Why not just 
``` authorize! :update, @program.events.new ``` ?

File: app/controllers/admin/schedules_controller.rb (hunk 21 above)
Comment (Ana06) [inline app/controllers/admin/schedules_controller.rb]: Both event and room use the guid, so I think it is clearer in the view to call it event, to avoid confusion

https://github.com/openSUSE/osem/pull/1094/files#diff-4db956c0d81325278cf026c6f02ae666R122

File: app/controllers/admin/schedules_controller.rb (hunk 20 above)
Comment (Ana06) [inline app/controllers/admin/schedules_controller.rb]: Because then I'll save that new event when doing `@program.save!`
But if I move that to the create action as you suggested then I can have just  `authorize! :update, @program.events.new`

Comment (ChrisBr) [inline app/controllers/admin/schedules_controller.rb]: Yes, because I think it is bad practice to have a create call in a :show action

Comment (ChrisBr) [inline app/controllers/admin/schedules_controller.rb]: And you could even move the authorize in a before_filter!

File: spec/factories/event_schedule.rb (hunk 12 above)
Comment (Ana06) [inline spec/factories/event_schedule.rb]: `room` and `start_time` are compulsory, and I have to create a `venue` before creating rooms. I also need to check if a room or date has been created already before creating it.
And always I create an `event_schedule` is because I want to schedule the event so I need that it belongs to the selected schedule, which must be created if it doesn't exit.

I could eliminate the part of the selected_schedule, but then I will have to do it every time I create an `event_schedule`, as in everywhere I use this factory this has to be done. I think that would be worst.

The rest was already on event and I just moved it (I kept the creation of the venue in both places but maybe it is not needed in event):

https://github.com/openSUSE/osem/pull/1094/files#diff-9c3fa59ac2be2f7c567e2bef4ec1367dL29

https://github.com/openSUSE/osem/pull/1094/files#diff-9c3fa59ac2be2f7c567e2bef4ec1367dL36

File: app/controllers/conference_controller.rb (hunk 5 above)
Comment (Ana06) [inline app/controllers/conference_controller.rb]: schedule controller is in the admin folder. Everything related with the schedule is now in conference controller. Do you want to change that?

**Comment by Ana06** [comment]
@ChrisBr I've made the corrections you sent me by email.

I am not pretty sure how we want the schedules index to look like:

image


Also, the authorization for the schedule used to be like that and I am not sure why:

`authorize! :update, @program.events.new`

I have changed to only allow to manage the schedule and event_schedules  to users with organizer role. Is that okey?

File: app/assets/javascripts/osem-schedule.js [hunk 24]
```diff
@@ -0,0 +1,109 @@
...
+var Schedule = {
+  initialize: function(conference_param, schedule_id_param) {
+    conference = conference_param;
+    schedule_id = schedule_id_param;
+  },
+  remove: function(element) {
+    var e =  $("#" + element);
+    var event_schedule_id = e.attr("event_schedule_id");
+    if(event_schedule_id != null){
+      var url = '/admin/conference/' + conference + '/event_schedule/' + event_schedule_id;
```
ChrisBr: Don't do that! Use instead the Rails url helper eg in a data attribute and get the whole url. This way you also don't need to initialize the schedule and conference ids.

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,109 @@
...
+        success: callback,
+        dataType : 'json'
+      });
+    }
+    var unscheduled = $(".unscheduled-events");
+    e.appendTo(unscheduled);
+    e.find(".schedule-event-delete-button").hide();
+  },
+  add: function (event_id, room_id, date, time, event_schedule_id) {
+    var url = '/admin/conference/' + conference + '/event_schedule';
```
ChrisBr: Same here!

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,109 @@
...
+      type: type,
+      data: params,
+      success: callback,
+      dataType : 'json'
+    });
+  }
+};
+
+$(document).ready( function() {
+  // hide the remove button for unshceduled events
```
ChrisBr: Typo!

app/controllers/admin/event_schedule_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      event = Event.where(guid: params[:event]).first
+      error_message = "Could not find event GUID: #{params[:event]}" if event.nil?
+
+      schedule = Schedule.where(id: params[:schedule]).first
+      error_message = "Could not find schedule: #{params[:schedule]}" if schedule.nil?
+
+      room = Room.where(guid: params[:room]).first
+      error_message = "Could not find room GUID: #{params[:room]}" if room.nil?
+
+      error_message = 'Date and time must be present' if params[:date].eql?('') || params[:time].eql?('')
```
ChrisBr: All this looks like validations for me?!

app/controllers/admin/event_schedule_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      unless error_message.nil?
```
ChrisBr: This clearly doesn't belong in a function called ``get_event_schedule_params(params)``

app/controllers/admin/event_schedule_controller.rb
```diff
@@ -0,0 +1,47 @@
+module Admin
+  class EventScheduleController < Admin::BaseController
+    load_and_authorize_resource :event_schedule
+
+    def create
+      event_schedule = EventSchedule.create(get_event_schedule_params(params))
```
ChrisBr: Shouldn't be necessary,  load_and_authorize_resource :event_schedule should do that!

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
+module Admin
+  class ScheduleController < Admin::BaseController
+    # By authorizing 'conference' resource, we can ensure there will be no unauthorized access to
+    # the schedule of a conference, which should not be accessed in the first place
+    load_and_authorize_resource :schedule
```
ChrisBr: Does it work? I think it should something like 
``
load_and_authorize_resource :schedule, through: :program
``

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    load_and_authorize_resource :conference, find_by: :short_title
+    load_and_authorize_resource :program, through: :conference, singleton: true
+    load_resource :venue, through: :conference, singleton: true
+
+    skip_before_action :verify_authenticity_token, only: [:update]
+
+    def index
+      @schedules = @conference.program.schedules
```
ChrisBr: If  `` load_and_authorize_resource :schedule `` would work, this shouldn't be necessary!
Maybe you have another look at https://github.com/CanCanCommunity/cancancan ?

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def create
+      new_schedule = @program.schedules.create
+      redirect_to action: 'show', id: new_schedule.id
+    end
+
+    def show
+      @schedule_id = params[:id].to_i
+      schedule = Schedule.find(@schedule_id)
```
ChrisBr: Load and authorize should do that

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      @event_schedules = schedule.event_schedules
```
ChrisBr: You could cover this also with a before filter

``
 load_and_authorize_resource :event_schedules, through: :schedule 
``

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      @unscheduled_events = @program.events.confirmed - schedule.events
+      @selected_schedule_id = @conference.program.selected_schedule.try(:id)
+      @dates = @conference.start_date..@conference.end_date
+      @rooms = (@venue && @venue.rooms.any?) ? @venue.rooms : [Room.new(name: 'No Rooms!', size: 0)]
+    end
+
+    def update
+      if params[:selected_schedule].present?
```
ChrisBr: Why first the present check?

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    skip_before_action :verify_authenticity_token, only: [:update]
+
+    def index
+      @schedules = @conference.program.schedules
+      @selected_schedule = @conference.program.selected_schedule
+    end
+
+    def create
+      new_schedule = @program.schedules.create
```
ChrisBr: Can sth go wrong? What happens then?

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def update
+      if params[:selected_schedule].present?
+        if params[:selected_schedule] == 'true'
+          @program.selected_schedule_id = params[:id].to_i
+        elsif params[:selected_schedule] == 'false' && (@program.selected_schedule_id == params[:id].to_i)
+          @program.selected_schedule_id = nil
+        end
+        @program.save!
```
ChrisBr: What happens if sth goes wrong? Why the bang operator?

app/controllers/admin/schedule_controller.rb
```diff
@@ -0,0 +1,54 @@
+module Admin
+  class ScheduleController < Admin::BaseController
```
ChrisBr: Why did you rename 

```
 app/controllers/admin/schedules_controller.rb 
```
to
```
app/controllers/admin/schedule_controller.rb 
```

http://alexander-clark.com/blog/rails-conventions-singular-or-plural/

app/models/event.rb
```diff
@@ -242,26 +232,28 @@ def calculate_progress
   end
 
   ##
-  # Returns end of the event
+  # Returns the room in which the event is scheduled
   #
-  def end_time
-    self.start_time + self.event_type.length.minutes
+  def selected_schedule_room
```
ChrisBr: just room and time! If it is not scheduled there is no room or time, period!

app/views/admin/schedule/index.html.haml
```diff
@@ -0,0 +1,32 @@
...
+        %th Selected
+        %th Actions
+      %tbody
+        - @schedules.each do |schedule|
+          %tr
+            %td
+              Schedule
+              = schedule.id
+            %td
+              = (schedule == @selected_schedule) ? 'Yes' : 'No'
```
ChrisBr: Why not a helper?

app/views/admin/schedule/show.html.haml
```diff
@@ -0,0 +1,36 @@
...
+    .page-header
+      %h1 Schedule
+      %p.text-muted
+        Create the schedules for the conference
+
+.row
+  .col-md-2
+    Selected schedule
+    = check_box_tag @conference.short_title, @schedule_id, (@schedule_id == @selected_schedule_id),
+      method: :put, url: "/admin/conference/#{@conference.short_title}/schedule/#{@schedule_id}?selected_schedule=",
```
ChrisBr: Please use the rails url_helper and don't use string interpolation. If the route changes, your view is broken!

Ana06:
@ChrisBr: 
> Don't do that! Use instead the Rails url helper eg in a data attribute and get the whole url. This way you also don't need to initialize the schedule and conference ids.

I would need the `schedule_id` even if I had the url as the url is using `event_schedule_id`, which is needed to create new event_schedules.

Also all unscheduled events use the same url (to create a new schedule), so I think it is not worthwhile adding it in a data attribute in every event. The schedule events use the same url + the event_schedule_id for both editing and deleting. I use the schedule_id attribute to know if an event is already scheduled or not to know if I should edit or create the event_schedule, if both has an url attribute it would be more difficult to do that. This is here:
https://github.com/openSUSE/osem/pull/1094/files#diff-96573ccecfe6b86eaba8cd21511262bbR35

I have initialize the Schedule with the url instead of the conference id, so I am using the helper but it is not exactly what you suggested. Is it okey like that?

app/controllers/admin/event_schedule_controller.rb
Ana06: Yes, and all that validations already exists. I'll eliminate this. :wink:

Ana06:
@ChrisBr:

> Why did you rename
>  `app/controllers/admin/schedules_controller.rb `
> to
> `app/controllers/admin/schedule_controller.rb `
> http://alexander-clark.com/blog/rails-conventions-singular-or-plural/

If controller should be plural, why conference controller is singular? :confused: 

I suppose I should also pluralize event schedule controller.

Ana06:
I've just realized that `max_attendees_no_more_than_room_size` should be moved to `EventSchedule`, shouldn't it?

Otherwise we are only checking that for the `event_schedules` in the selected schedule.

I think I should render the room size in the admin schedule (it used to be rendered but I eliminated it when integrating the view). Also, it should be allowed to drop events in a room whose size is not enough for the number of max attendees. Do I add that in this PR or in a different one?

ChrisBr:
> If controller should be plural, why conference controller is singular? :confused:

It should also be plural! Even for singular routes you should name the controller in plural.

> Because you might want to use the same controller for a singular route (/account) and a plural route (/accounts/45), singular resources map to plural controllers. So that, for example, resource :photo and resources :photos creates both singular and plural routes that map to the same controller (PhotosController).

from http://edgeguides.rubyonrails.org/routing.html#singular-resources

> I suppose I should also pluralize event schedule controller.

Yes, it should be event_schedules_controller!

ChrisBr:
> I've just realized that max_attendees_no_more_than_room_size should be moved to EventSchedule, shouldn't it?
> 
> Otherwise we are only checking that for the event_schedules in the selected schedule.

Makes sense

> I think I should render the room size in the admin schedule (it used to be rendered but I eliminated it when integrating the view). Also, it should be allowed to drop events in a room whose size is not enough for the number of max attendees. Do I add that in this PR or in a different one?

Let's finish this one first, it's already quite big and hard to review!

Ana06:
I've pluralized event schedule controller name and moved the `max_attendees_no_more_than_room_size` validation. I've also open an issue to pluralize conference and proposal controllers https://github.com/openSUSE/osem/issues/1120

Is there anything else to change?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Admin
+  class EventSchedulesController < Admin::BaseController
+    load_and_authorize_resource :event_schedule
+
+    def create
+      @event_schedule.save
```
ChrisBr: What happens when save goes wrong?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+    end
+
+    def update
+      @event_schedule.update(event_schedule_params)
```
ChrisBr: What happens when update goes wrong?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+    end
+
+    def destroy
+      @event_schedule.destroy
```
ChrisBr: Same here!

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,50 @@ class SchedulesController < Admin::BaseController
...
-        redirect_to admin_conference_index_path
-        return
-      end
-      @dates = @conference.start_date..@conference.end_date
-      if @venue && @venue.rooms.any?
-        @rooms = @venue.rooms
-      else
-        @rooms = [ Room.new(name: 'No Rooms!', size: 0) ]
-      end
+    def index
```
ChrisBr: def index ; end

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,50 @@ class SchedulesController < Admin::BaseController
...
-      authorize! :update, @program.events.new
-      event = Event.where(guid: params[:event]).first
-      error_message = nil
-      if event.nil?
-        error_message = "Could not find event GUID: #{params[:event]}"
+    def create
+      if @schedule.save
+        redirect_to action: 'show', id: @schedule.id
+      else
+        flash[:error] = 'Could not create schedule'
```
ChrisBr: What about showing the error message in the flash?

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,50 @@ class SchedulesController < Admin::BaseController
...
-
-      Rails.logger.debug("Loading #{time}")
-      # FIXME: Same here as in events_controller.rb. Event timezone should be applied
-      # only on output
-      # zone = ActiveSupport::TimeZone::new(@conference.timezone)
-      # start_time = DateTime.strptime(time + zone.formatted_offset, "%Y-%m-%d %k:%M %Z")
-      start_time = DateTime.strptime(time, '%Y-%m-%d %k:%M')
-      event.start_time = start_time
-      event.save!
+      @program.save!
```
ChrisBr: Why the bang ``!`` ? What happens if this goes wrong? We will get a 500...

spec/models/event_spec.rb
```diff
@@ -25,27 +25,6 @@
     it { is_expected.to validate_presence_of(:program) }
     it { is_expected.to validate_presence_of(:event_type) }
 
-    describe 'max_attendees_no_more_than_room_size' do
```
ChrisBr: Where is this test now?

ChrisBr: Got it!

app/controllers/admin/schedules_controller.rb
ChrisBr: And what I predicted happens in our heroku app ;)
Because the new migration didn't run, there is no @program.selected_schedule_id field and you get a 500.

Please don't use the bang and render an error if sth goes wrong!

ChrisBr: Also, you update a program attribute here, that means you should move this action to the programs_controller.

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,107 @@
...
+        data: params,
+        success: callback,
+        dataType : 'json'
+      });
+    }
+    var unscheduled = $(".unscheduled-events");
+    e.appendTo(unscheduled);
+    e.find(".schedule-event-delete-button").hide();
+  },
+  add: function (event_id, room_id, date, time, event_schedule_id) {
```
ChrisBr: Why not two different functions for #create and #update ?

ChrisBr:
Because of the 

`   Selected schedule
    = check_box_tag @conference.short_title, @schedule.id, (@schedule.id == @selected_schedule.try(:id)),
      method: :put, url: (admin_conference_schedule_path(@conference.short_title, @schedule) + '?selected_schedule='),
      class: 'switch-checkbox', data: { size: 'small',
                                        off_color: 'warning',
                                        on_text: 'Yes',
                                        off_text: 'No' }`

I get a message from my browser every time I reload the page...

Ana06:
@ChrisBr:
`Why not two different functions for #create and #update ?`

Because they will be identical except by the type and the url in the params, but I will still need the `if(event_schedule_id != null)` condition to know if I have to call create or update. I think it will complicate the code. Every time we want to change something in the way we schedule events (for example to add a save button) it will be necessary to change both functions.

Ana06:
> Because of the
> 
> `Selected schedule = check_box_tag @conference.short_title, @schedule.id, (@schedule.id == @selected_schedule.try(:id)), method: :put, url: (admin_conference_schedule_path(@conference.short_title, @schedule) + '?selected_schedule='), class: 'switch-checkbox', data: { size: 'small', off_color: 'warning', on_text: 'Yes', off_text: 'No' }`
> 
> I get a message from my browser every time I reload the page...

What message? I think I don't get any message :confused:

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,39 @@ class SchedulesController < Admin::BaseController
...
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :program, through: :conference, singleton: true
+    load_and_authorize_resource :schedule, through: :program
+    load_resource :event_schedules, through: :schedule
+    load_resource :selected_schedule, through: :program, singleton: true
     load_resource :venue, through: :conference, singleton: true
 
-    skip_before_filter :verify_authenticity_token, only: [:update]
-    layout 'schedule'
+    skip_before_action :verify_authenticity_token, only: [:update]
```
differentreality: There is no `update` action in this controller

app/views/admin/schedules/show.html.haml
```diff
@@ -1,48 +1,36 @@
...
+    .page-header
+      %h1 Schedule
+      %p.text-muted
+        Create the schedules for the conference
+
+.row
+  .col-md-2
+    Selected schedule
+    = check_box_tag @conference.short_title, @schedule.id, (@schedule.id == @selected_schedule.try(:id)),
+      method: :patch, url: (admin_conference_program_path(@conference.short_title) + '?[program][selected_schedule_id]='),
```
differentreality: This fails
2016-08-04T21:35:06.495866+00:00 app[web.1]: Started PATCH "/admin/conference/osemdemo/program?[program][selected_schedule_id]=1" for 94.66.223.71 at 2016-08-04 21:35:06 +0000
2016-08-04T21:35:06.506358+00:00 app[web.1]: Processing by Admin::ProgramsController#update as JS
2016-08-04T21:35:06.556467+00:00 app[web.1]: ActiveRecord::UnknownAttributeError (unknown attribute 'selected_schedule_id' for Program.):

app/models/event.rb
```diff
@@ -35,8 +35,6 @@ class Event < ActiveRecord::Base
   validates :program, presence: true
   validates :max_attendees, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_nil: true }
 
-  validate :max_attendees_no_more_than_room_size
```
differentreality: The problem with removing this from here is that when a proposal is scheduled in a room that can have up to 10 people, when the submitter edits the proposal, can set `max_attendees` to 15, and enable require_registration. Which will allow 15 people to register to a session that takes place in a room that can only hold only 10 people.

Now of course the problem can work the other way around, which is what I assume you are trying to fix, ie. what if we try to schedule a talk with 15 registered attendees in a room that can only have 10 seats. That's indeed an issue, one we are aware of and trying to resolve. But in my opinion:
1) we should not break scheduling because of this. An organizer wants to schedule a specific session into a specific room at a specific time, a warning would be excellent, but disallowing it all together, I don't know. Mainly because the organizer cannot properly remedy this. Once 15 people have registered to attend, the organizer cannot delete the registrations.

2) the workflow of max_attendees is a subject for the proposal registration process to solve, not this PR. Unless there is something really failing with your changes. Is there?

app/assets/javascripts/osem-switch.js
```diff
@@ -11,4 +11,27 @@ $(function () {
...
+    var url = $(this).attr('url');
+    var method = $(this).attr('method');
+
+    if(state){
+      url += $(this).attr('value');
+    }
+
+    var callback = function(data) {
+      if(data.status != 'ok'){
+        alert("The schedule couldn't been updated");
```
differentreality: Should we be handling error messages through pop-up windows (instead of flash messages)?

differentreality:
@Ana06 could you please add a comment with the overall workflow you are implementing? Like a few bullet points of what you are doing, what is no longer there, what is new.

app/views/admin/schedules/show.html.haml
Ana06: Yes, because `selected_schedule_id` was introduced here: https://github.com/openSUSE/osem/pull/1094/files#diff-7e170cb4cd18afa0c7d53e4ba331c660R7

app/models/event.rb
Ana06: > The problem with removing this from here is that when a proposal is scheduled in a room that can have up to 10 people, when the submitter edits the proposal, can set max_attendees to 15, and enable require_registration. Which will allow 15 people to register to a session that takes place in a room that can only hold only 10 people.

Maybe I should keep the validation in both event and event schedule.

> we should not break scheduling because of this. An organizer wants to schedule a specific session into a specific room at a specific time, a warning would be excellent, but disallowing it all together, I don't know. Mainly because the organizer cannot properly remedy this. Once 15 people have registered to attend, the organizer cannot delete the registrations.

For me allowing the organizer to schedule a talk with can have 15 attendees if there is only place for 10 has no sense. If it is possible that this talk take place in that room then the size of the room should be increased, and the organizer can do that before scheduling the event.

> the workflow of max_attendees is a subject for the proposal registration process to solve, not this PR. Unless there is something really failing with your changes. Is there?

No, nothing is failing. I moved it because I thought that we need to validate all event schedules, I didn't take into account that the max attendees can be change after scheduling. I can keep validations in both models and we can discuss afterward (outside this PR) if we want to change the way it works (validate less things or in a different way).

app/assets/javascripts/osem-switch.js
Ana06: I was not sure of that. Maybe it is not even necessary to show a message as the event will go back to its original position if it couldn't be scheduled.

Ana06:
@differentreality 
- Integrate the schedule in the backend
- Allow having several schedules in the backend: 
  - Add the relations and models illustrated in the image of the first comment
  - Add an index with all schedules were more schedules can be created
  - Add a checkbox inside show to be able to publish a schedule
- Adapt the public schedule to work with the new relations and models
- Simplify the javascript needed for the drag and drop (events are not loaded with javascript anymore) and the javascript code has been moved
- To schedule an event it is used event schedules controller instead of schedule controller
- I changed the routes
- Change the ability file and the tests to work with the previous changes

And I think that's all. :smiley:

app/assets/javascripts/osem-switch.js
differentreality: Yes, but the switch-checkbox won't show that until the page is refreshed.
You can use unobtrusive flash (we already have the gem in)
https://github.com/leonid-shevtsov/unobtrusive_flash

Ana06: That's true. I'll change it then. :wink:

app/models/event.rb
differentreality: Increasing the room.size is not realistic, rooms have a maximum size (consider fire regulations, or amount of workstations needed for a workshop to take place).

As far as I am concerned, you could leave everything related to this validation back to its original state, and not deal with this in this PR at all.

Ana06: > Increasing the room.size is not realistic, rooms have a maximum size (consider fire regulations, or amount of workstations needed for a workshop to take place).
Then we shouldn't allow that an event with more attendees is scheduled in that room.

app/views/admin/schedules/show.html.haml
ChrisBr: Yes, and it doesn't run the migration twice ... 
We would need to restart the heroku server or you run it locally!

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,111 @@
...
+      var my_url = url + '/' + event_schedule_id;
+      var callback = function(data) {
+        console.log(data);
+        if(data.status == 'ok'){
+          e.attr("event_schedule_id", null);
+          e.appendTo($(".unscheduled-events"));
+          e.find(".schedule-event-delete-button").hide();
+        }
+        else{
+          alert(error_remove);
```
ChrisBr: No, please no alerts. Use a flash instead!

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+      $.ajax({
+        url: my_url,
+        type: 'DELETE',
+        success: callback,
+        dataType : 'json'
+      });
+    }
+    else{
+        alert(error_remove);
```
ChrisBr: Same here

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,111 @@
...
+    var callback = function(data) {
+      console.log(data);
+      if(data.status == 'ok'){
+        event.appendTo(new_parent);
+        event.attr("event_schedule_id", data.event_schedule_id);
+        event.find(".schedule-event-delete-button").show();
+      }
+      else{
+        event.appendTo(previous_parent);
+        alert("The event couldn't been scheduled");
```
ChrisBr: Same here

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
+module Admin
+  class EventSchedulesController < Admin::BaseController
+    load_and_authorize_resource :event_schedule
+
+    def create
+      if @event_schedule.save
+        render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+      else
+        render json: { 'status' => 'error' }
```
ChrisBr: Why not sending the error message?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    def update
+      if @event_schedule.update(event_schedule_params)
+        render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+      else
+        render json: { 'status' => 'error' }
```
ChrisBr: Same here

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    def destroy
+      if @event_schedule.destroy
+        render json: { 'status' => 'ok' }
+      else
+        render json: { 'status' => 'error' }
```
ChrisBr: Same here

app/controllers/admin/programs_controller.rb
```diff
@@ -15,8 +15,14 @@ def update
 
       if @program.update_attributes(program_params)
         ConferenceScheduleUpdateMailJob.perform_later(@conference) if send_mail_on_schedule_public
-        redirect_to admin_conference_program_path(@conference.short_title),
-                    notice: 'The program was successfully updated.'
+        if program_params[:selected_schedule_id].nil?
```
ChrisBr: Please don't do that, instead use respond_to

```
respond_to do |format|
  format.html { redirect_to(...) }
  format.js
end
```

If you do it like this, you can use it only with js when ``selected_schedule_id`` is present. With respond_to it is more general.

http://apidock.com/rails/ActionController/MimeResponds/InstanceMethods/respond_to

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,37 @@ class SchedulesController < Admin::BaseController
...
-        event.room = nil
-        event.save!
-        render json: { 'status' => 'ok' }
-        return
-      end
-      room = Room.where(guid: room_params).first
-      if room.nil?
-        error_message = "Could not find room GUID: #{params[:room]}"
+        redirect_to admin_conference_schedules_path(conference_id: @conference.short_title),
+                    error: 'Could not create schedule'
```
ChrisBr: Why not showing the error message?

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,37 @@ class SchedulesController < Admin::BaseController
...
-      if @conference.nil?
-        redirect_to admin_conference_index_path
-        return
-      end
-      @dates = @conference.start_date..@conference.end_date
-      if @venue && @venue.rooms.any?
-        @rooms = @venue.rooms
+    def create
+      if @schedule.save
+        redirect_to action: 'show', id: @schedule.id
```
ChrisBr: Why no flash message?

app/controllers/admin/schedules_controller.rb
```diff
@@ -4,72 +4,37 @@ class SchedulesController < Admin::BaseController
...
+    load_resource :event_schedules, through: :schedule
+    load_resource :selected_schedule, through: :program, singleton: true
     load_resource :venue, through: :conference, singleton: true
 
-    skip_before_filter :verify_authenticity_token, only: [:update]
-    layout 'schedule'
+    def index; end
 
-    def show
-      authorize! :update, @program.events.new
```
ChrisBr: Have you checked that there is a policy for creating a new schedule?

ChrisBr: Found it!

app/models/ability.rb
```diff
@@ -143,6 +143,8 @@ def signed_in_with_organizer_role(user)
     can :manage, Vposition, conference_id: conf_ids_for_organizer
     can :manage, Vday, conference_id: conf_ids_for_organizer
     can :manage, Program, conference_id: conf_ids_for_organizer
+    can :manage, Schedule, program: { conference_id: conf_ids_for_organizer }
```
ChrisBr: Please add tests for this!

Ana06: I've already added tests for that
https://github.com/openSUSE/osem/pull/1094/files#diff-e9501dda6c23fd3e1076e2dff66d0f68R205

app/controllers/admin/schedules_controller.rb
Ana06: It is shown

image


Unless you mean something else... :confused:

app/controllers/admin/schedules_controller.rb
Ana06: To inform that it was successfully created?

Ana06:
@ChrisBr I made the last corrections you added.

I also moved the `max_attendees_no_more_than_room_size` to event again, as @differentreality suggested. I am not sure that we don't need to validate it in event_schedule too, but as @differentreality says:

> the workflow of max_attendees is a subject for the proposal registration process to solve, not this PR.

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
+module Admin
+  class EventSchedulesController < Admin::BaseController
+    load_and_authorize_resource :event_schedule
+
+    def create
+      if @event_schedule.save
+        render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+      else
+        render json: { 'status' => "The event couldn't be scheduled" }
```
ChrisBr: Why not displaying the error message ``@event_schedule.errors.full_messages``?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    def update
+      if @event_schedule.update(event_schedule_params)
+        render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+      else
+        render json: { 'status' => "The event couldn't be scheduled" }
```
ChrisBr: Why not displaying the error message ``@event_schedule.errors.full_messages``?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+    end
+
+    def destroy
+      if @event_schedule.destroy
+        render json: { 'status' => 'ok' }
+      else
+        render json: { 'status' => "The event couldn't be unscheduled" }
```
ChrisBr: Why not displaying the error message ``@event_schedule.errors.full_messages``?

app/controllers/admin/programs_controller.rb
```diff
@@ -13,20 +13,28 @@ def update
...
-        ConferenceScheduleUpdateMailJob.perform_later(@conference) if send_mail_on_schedule_public
-        redirect_to admin_conference_program_path(@conference.short_title),
-                    notice: 'The program was successfully updated.'
-      else
-        flash[:error] = "Updating program failed. #{@program.errors.to_a.join('. ')}."
-        render :new
+      respond_to do |format|
+        if @program.update_attributes(program_params)
+          ConferenceScheduleUpdateMailJob.perform_later(@conference) if send_mail_on_schedule_public
+          format.html do
```
ChrisBr: I would prefer to have two respond_to blocks to be more explicit and to not wrap the whole update statement inside it!

app/controllers/admin/schedules_controller.rb
ChrisBr: I mean ``@schedule.errors.full_messages`` ?

Ana06:
@ChrisBr I made the last correction you pointed out and created `event_schedules_controller_spec.rb` and `schedules_controller_spec.rb` :smiley:

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,116 @@
...
+    if(event_schedule_id != null){
+      var my_url = url + '/' + event_schedule_id;
+      var callback = function(data) {
+        console.log(data);
+        if(data.status == 'ok'){
+          e.attr("event_schedule_id", null);
+          e.appendTo($(".unscheduled-events"));
+          e.find(".schedule-event-delete-button").hide();
+        }
+        else{
```
ChrisBr: Why do you handle that in the success callback? I'd expect sth like

```
$.ajax({
  url: my_url,
  type: 'DELETE',
  success: success_callback,
  error: error_callback (-> eg show error),
  dataType : 'json'
+      });
```
Then you don't need the conditional in the success callback. And what I'm wondering, why does the success callback fire if it is an error?

app/assets/javascripts/osem-schedule.js
```diff
@@ -0,0 +1,116 @@
...
+      else{
+        event.appendTo(previous_parent);
+        showError("The event couldn't been scheduled");
+      }
+    }
+    $.ajax({
+      url: my_url,
+      type: type,
+      data: params,
+      success: callback,
```
ChrisBr: Same here!

app/assets/javascripts/osem-switch.js
```diff
@@ -11,4 +11,27 @@ $(function () {
...
+
+    var callback = function(data) {
+      if(data.status != 'ok'){
+        showError(data.status);
+      }
+    }
+    $.ajax({
+      url: url,
+      type: method,
+      success: callback,
```
ChrisBr: Same here!

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -0,0 +1,35 @@
+module Admin
+  class EventSchedulesController < Admin::BaseController
+    load_and_authorize_resource :event_schedule
+
+    def create
+      if @event_schedule.save
+        render json: { 'status' => 'ok', event_schedule_id: @event_schedule.id }
+      else
+        render json: { 'status' => "The event couldn't be scheduled. #{@event_schedule.errors.full_messages.join('. ')}" }
```
ChrisBr: The problem here is that you don't use HTTP status to specify that an error happened! See this blog article:
http://travisjeffery.com/b/2012/04/rendering-errors-in-json-with-rails/

This will also point out, that you should use different callbacks for error and success
(what I already commented in the JS files...)

ChrisBr: Please adapt this for all actions and also please use the new ruby has syntax!

app/controllers/admin/programs_controller.rb
```diff
@@ -15,18 +15,28 @@ def update
...
+        end
       else
-        flash[:error] = "Updating program failed. #{@program.errors.to_a.join('. ')}."
-        render :new
+        respond_to do |format|
+          format.html do
+            flash[:error] = "Updating program failed. #{@program.errors.to_a.join('. ')}."
+            render :new
+          end
+          format.js { render json: { 'status' => "The selected schedule couldn't been updated #{@program.errors.to_a.join('. ')}" } }
```
ChrisBr: Same here!

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        it 'saves the event schedule to the database' do
+          expected = expect do
+            post :create, conference_id: conference.short_title, event_schedule:
+                 attributes_for(:event_schedule,
+                                schedule_id: schedule.id,
+                                event_id: create(:event, program: conference.program).id,
+                                room_id: create(:room, venue: venue).id,
+                                start_time: conference.start_date)
+          end
+          expected.to change { EventSchedule.count }.by 1
```
ChrisBr: Why not just wrapping the post into a ``{ }``?

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        it 'renders JSON without errors' do
+          post :create, conference_id: conference.short_title, event_schedule:
+               attributes_for(:event_schedule,
+                              schedule_id: schedule.id,
+                              event_id: create(:event, program: conference.program).id,
+                              room_id: create(:room, venue: venue).id,
+                              start_time: conference.start_date)
+
+          expect(response).to be_success
+          expect(JSON.parse(response.body)['status']).to eq('ok')
```
ChrisBr: Just test the status code!

ChrisBr: To DRY, please wrap the post into a before filter!

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        end
+      end
+
+      context 'with invalid attributes' do
+        it 'does not save the event schedule to the database' do
+          expected = expect do
+            post :create, conference_id: conference.short_title, event_schedule:
```
ChrisBr: Please wrap the post into a before filter!

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+        it 'renders JSON with error' do
+          post :create, conference_id: conference.short_title, event_schedule:
+               attributes_for(:event_schedule,
+                              schedule_id: schedule.id,
+                              event_id: nil,
+                              room_id: nil,
+                              start_time: nil)
+
+          expect(JSON.parse(response.body)['status']).to eq("The event couldn't be scheduled. Event can't be blank. Room can't be blank. Start time can't be blank")
```
ChrisBr: Just test the status code.

 Event can't be blank. Room can't be blank. Start time can't be blank etc are validation tests which we already cover in event_schedules_spec.rb

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    describe 'POST #create' do
+      context 'with valid attributes' do
+        it 'saves the event schedule to the database' do
+          expected = expect do
+            post :create, conference_id: conference.short_title, event_schedule:
+                 attributes_for(:event_schedule,
+                                schedule_id: schedule.id,
+                                event_id: create(:event, program: conference.program).id,
```
ChrisBr: Why do you create a new one and don't use the factories you already defined?

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+                                room_id: create(:room, venue: venue).id,
```
ChrisBr: Same here!

spec/controllers/admin/event_schedules_controller_spec.rb
ChrisBr: Ah, you don't have an event and room! My bad, sorry!

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        it 'renders JSON without errors' do
+          patch :update, id: event_schedule.id, conference_id: conference.short_title, event_schedule:
+               attributes_for(:event_schedule,
+                              schedule_id: schedule.id,
+                              event_id: create(:event, program: conference.program).id,
+                              room_id: create(:room, venue: venue).id,
+                              start_time: conference.start_date)
+
+          expect(response).to be_success
+          expect(JSON.parse(response.body)['status']).to eq('ok')
```
ChrisBr: status code is enough

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        end
+      end
+    end
+
+    describe 'POST #update' do
+      context 'with valid attributes' do
+        it 'changes event schedule attributes' do
+          event = create(:event, program: conference.program)
+          room = create(:room, venue: venue)
+          patch :update, id: event_schedule.id, conference_id: conference.short_title, event_schedule:
```
ChrisBr: Can you please move the patch in a before filter and move the expect to their single tests?

You can then create tests like

```it 'will update the room'```
```it 'will update the start_time'```

Wondering if updating the schedule and event is really a use case?

File: spec/controllers/admin/event_schedules_controller_spec.rb [hunk 81]
```diff
@@ -0,0 +1,143 @@
...
+
+        it 'renders JSON with error' do
+          patch :update, id: event_schedule.id, conference_id: conference.short_title, event_schedule:
+               attributes_for(:event_schedule,
+                              schedule_id: schedule.id,
+                              event_id: nil,
+                              room_id: nil,
+                              start_time: nil)
+
+          expect(JSON.parse(response.body)['status']).to eq("The event couldn't be scheduled. Event can't be blank. Room can't be blank. Start time can't be blank")
```
ChrisBr: These are validation tests which we already cover in event_schedule_spec.rb
Just test the status code

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+          delete :destroy, id: event_schedule.id, conference_id: conference.short_title
+        end
+
+        expected.to change { EventSchedule.count }.by(-1)
+      end
+      it 'renders JSON without errors' do
+        delete :destroy, id: event_schedule.id, conference_id: conference.short_title
+
+        expect(response).to be_success
+        expect(JSON.parse(response.body)['status']).to eq('ok')
```
ChrisBr: Status code is enough

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    describe 'DELETE #destroy' do
+      it 'deletes the event schedule' do
+        expected = expect do
+          delete :destroy, id: event_schedule.id, conference_id: conference.short_title
+        end
+
+        expected.to change { EventSchedule.count }.by(-1)
+      end
+      it 'renders JSON without errors' do
+        delete :destroy, id: event_schedule.id, conference_id: conference.short_title
```
ChrisBr: Move the delete to a before filter!

spec/controllers/admin/schedules_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        expect(response).to render_template :index
+      end
+    end
+
+    describe 'POST #create' do
+      it 'saves the schedule to the database' do
+        expected = expect do
+          post :create, conference_id: conference.short_title
+        end
+        expected.to change { Schedule.count }.by 1
```
ChrisBr: Why not moving the post into a ``{}`` block or in a before filter?

spec/controllers/admin/event_schedules_controller_spec.rb
Ana06: Because Rubocop complains:

> Avoid using {...} for multi-line blocks in Rspec change block

There are already other tests doing the same, for example https://github.com/openSUSE/osem/blob/master/spec/controllers/admin/conferences_controller_spec.rb#L81

Some others don't define the expected variable, like this one: https://github.com/openSUSE/osem/blob/master/spec/controllers/admin/sponsorship_levels_controller_spec.rb#L64. But I don't like writing `end.to change`

spec/controllers/admin/schedules_controller_spec.rb
Ana06: If I move it to a before filter how do I check if `Schedule.count` changed?

I can't not move it into a  `{}` block Rubocop complains:

> Avoid using {...} for multi-line blocks in Rspec change block

There are already other tests doing the same, for example https://github.com/openSUSE/osem/blob/master/spec/controllers/admin/conferences_controller_spec.rb#L81

Some others don't define the expected variable, like this one: https://github.com/openSUSE/osem/blob/master/spec/controllers/admin/sponsorship_levels_controller_spec.rb#L64. But I don't like writing `end.to change`

spec/controllers/admin/event_schedules_controller_spec.rb
Ana06: why? :confused:

spec/controllers/admin/event_schedules_controller_spec.rb
Ana06: Not really :expressionless:

spec/controllers/admin/event_schedules_controller_spec.rb
Ana06: why? :confused:

spec/controllers/admin/schedules_controller_spec.rb
ChrisBr: You could move it to a let

``
let(:action) { post :create, conference_id: conference.short_title } 

it 'saves the schedule to the database' do
  action.to change { Schedule.count }.by 1
end
``

spec/controllers/admin/event_schedules_controller_spec.rb
ChrisBr: Because you don't have to call it twice!

spec/controllers/admin/event_schedules_controller_spec.rb
ChrisBr: Then we don't need to test it

spec/controllers/admin/event_schedules_controller_spec.rb
ChrisBr: Because you call it several times!

Ana06:
@ChrisBr 

> Why do you create a new one and don't use the factories you already defined?

`attributes_for(:event_schedule)` is `{}` as I don't give any default value to event_schedule attributes in factory because most of them are relations so I think it is not possible to do it.

Ana06:
@ChrisBr 
> Move the delete to a before filter!
> Because you don't have to call it twice!

In most controller it is called twice, for example: https://github.com/openSUSE/osem/blob/master/spec/controllers/admin/conferences_controller_spec.rb#L81

Do I open a PR to change that?

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        it 'updates the room' do
+          expect(event_schedule.room_id).to eq(room.id)
+        end
+
+        it 'updates the start_time' do
+          expect(event_schedule.start_time).to eq(conference.start_date)
+        end
+
+        it 'renders JSON without errors' do
```
ChrisBr: That's not what the test does!
Something like ``has 200 status code`` is more accurate, isn't it?

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                              event_id: nil,
+                              room_id: nil,
+                              start_time: nil)
+        end
+
+        it 'does not save the event schedule to the database' do
+          expect{ create_action }.to_not change { EventSchedule.count }
+        end
+
+        it 'renders JSON with error' do
```
ChrisBr: That's not what the test does!

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                              event_id: create(:event, program: conference.program).id,
+                              room_id: create(:room, venue: venue).id,
+                              start_time: conference.start_date)
+        end
+
+        it 'saves the event schedule to the database' do
+          expect{ create_action }.to change { EventSchedule.count }.by 1
+        end
+
+        it 'renders JSON without errors' do
```
ChrisBr: That's not what the test does

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                              event_id: nil,
+                              room_id: nil,
+                              start_time: nil)
+        end
+        it 'does not save the event schedule to the database' do
+          expect{ update_action }.to_not change { event_schedule }
+        end
+
+        it 'renders JSON with error' do
+          update_action
```
ChrisBr: That's not what the test does

spec/controllers/admin/event_schedules_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    describe 'DELETE #destroy' do
+      let(:destroy_action) do
+        delete :destroy, id: event_schedule.id, conference_id: conference.short_title
+      end
+
+      it 'deletes the event schedule' do
+        expect{ destroy_action }.to change { EventSchedule.count }.by(-1)
+      end
+
+      it 'renders JSON without errors' do
```
ChrisBr: That's not what the test does

spec/controllers/schedules_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        get :show, conference_id: conference.short_title, format: :xml
+      end
+
+      it 'assigns variables' do
+        expect(assigns(:conference)).to eq conference
+        expect(assigns(:events_xml)).to eq conference.program.selected_event_schedules.map(&:event)
+                                           .group_by{ |event| event.time.to_date }
+      end
+
+      it 'renders successfully' do
```
ChrisBr: That's not what the test does

ChrisBr:
Except some nitpicky test namings which you should improve, it LGTM!
Really good work @Ana06 :clap: :clap: :clap: 

@hennevogel do you also want to have a look at it?

app/assets/javascripts/application.js
```diff
@@ -13,6 +13,7 @@
 //= require jquery
 //= require jquery_ujs
 //= require jquery.mobile.custom.min
+//= require jquery-ui.min
```
hennevogel: We don't just copy dependencies into our code base instead we use gems. Please make use of https://github.com/joliss/jquery-ui-rails or

Ana06:
@ChrisBr @hennevogel Is there anything else to change?

ChrisBr:
From my side it looks good and I assume you already fixed @hennevogel suggestions.
So let's :ship: it!

Great work @Ana06 :clap: :clap: :clap:

Issue #1061, opened 2016-06-27
Provide a way to manually confirm new users


Chaz6:
When some mail hosts block messages at smtp-time (Hotmail, Google, etc), it should be possible for an admin to manually confirm a new user's email address.

shlok007:
I see a closed PR. If someone isn't working already. I would like to work on it :).

shlok007:
As @nishanthvijayan pointed out, @hennevogel  Should we allow the admin to cancel confirmation of a user as well?

Chaz6:
@shlok007 I think so yes, and there should also be an audit trail (i.e. which admin did it).

differentreality:
The audit trail is beyond the scope of this issue at this time, as we are (re) implementing a full changelog. @nishanthvijayan we will merge this change soon, so please make sure we add this in #1019

differentreality:
@shlok007 it would make sense to adapt the view in users#show and users#edit too.

PR #1097, opened 2016-07-10
Ability to confirm users by admin

shlok007:
users

This is how the changes would reflect like.

As this is would be my first contribution to OSEM, It would be great if someone could guide me through this PR about the convections followed and whether the approach taken is correct or not. I'll proceed with the audit trail as specified in #1061 and writing test thereafter :).

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -10,6 +10,17 @@ module Admin
       @users = User.all
     end
 
+    # This action allow admins to manually toggle confirmation state of another user
+    def toggle_confirmation
+      if user_params[:to_confirm] == 'true'
+        @user.confirm
+      elsif user_params[:to_confirm] == 'false'
+        @user.confirmed_at = nil
+        @user.save
+      end
+      head :ok
+    end
+
     def show
       # Variable @show_attributes holds the attributes that are visible for the 'show' action
       # If you want to change the attributes that are shown in the 'show' action of users
@@ -40,7 +51,7 @@ module Admin
 
     def user_params
       params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
-                                   :tshirt, :mobile, :volunteer_experience, :languages, role_ids: [])
+                                   :tshirt, :mobile, :volunteer_experience, :languages, :to_confirm, role_ids: [])
     end
   end
 end
--- a/app/views/admin/conference/_recent_users.html.haml
+++ b/app/views/admin/conference/_recent_users.html.haml
@@ -6,7 +6,7 @@
           %th #
           %th E-Mail
           %th Date registered
-          %th Status
+          %th Confirmed?
       - recent_users.each_with_index do |user, index|
         %tbody
           %tr
@@ -14,10 +14,12 @@
             %td= link_to user.email, admin_user_path(user.id)
             %td= user.created_at.strftime('%m/%d/%Y')
             %td
-              - if user.confirmed?
-                %span.label.label-success Confirmed
-              -else
-                %span.label.label-danger Unconfirmed
+              = check_box_tag user.id, user.id, user.confirmed?, 
+                method: :patch, 
+                url: "/admin/users/#{user.id}/toggle_confirmation?user[to_confirm]=", 
+                class: 'switch-checkbox', 
+                readonly: true,
+                data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
 - else
   %h5.text-warning.text-center
     No sign ups!
--- a/app/views/admin/users/index.html.haml
+++ b/app/views/admin/users/index.html.haml
@@ -10,7 +10,7 @@
     %table.table.table-striped.table-bordered.table-hover.datatable
       %thead
         %th ID
-        %th State
+        %th Confirmed?
         %th Email
         %th Name
         %th Attended Conferences
@@ -23,10 +23,20 @@
             %td
               = user.id
             %td
-              - if user.confirmed?
-                confirmed
+              - if can? :toggle_confirmation, user
+                = check_box_tag user.id, user.id, user.confirmed?, 
+                  method: :patch, 
+                  url: "/admin/users/#{user.id}/toggle_confirmation?user[to_confirm]=", 
+                  class: 'switch-checkbox', 
+                  readonly: false,
+                  data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
               - else
-                unconfirmed
+                = check_box_tag user.id, user.id, user.confirmed?, 
+                  method: :patch, 
+                  url: "/admin/users/#{user.id}/toggle_confirmation?user[to_confirm]=", 
+                  class: 'switch-checkbox', 
+                  readonly: true,
+                  data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' }  
             %td
               = user.email
             %td
--- a/app/views/admin/users/show.html.haml
+++ b/app/views/admin/users/show.html.haml
@@ -14,12 +14,6 @@
           - if can? :edit, @user
             .pull-right
               = link_to 'Edit', edit_admin_user_path(@user), class: 'btn btn-primary'
-          - if @user.confirmed?
-            %span.label.label-success
-              confirmed
-          - else
-            %span.label.label-warning
-              unconfirmed
           %table.table
             - @show_attributes.each do |attr|
               %tr
@@ -32,6 +26,28 @@
                 - elsif attr == 'biography'
                   %td
                     = markdown(@user.biography)
+                - elsif attr == 'email'
+                  %td 
+                    = @user.send(attr)
+                  %tr
+                    %td
+                      %b
+                        Confirmed?
+                    %td
+                      - if can? :toggle_confirmation, @user
+                        = check_box_tag @user.id, @user.id, @user.confirmed?, 
+                          method: :patch, 
+                          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
+                          class: 'switch-checkbox', 
+                          readonly: false,
+                          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
+                      - else
+                        = check_box_tag @user.id, @user.id, @user.confirmed?, 
+                          method: :patch, 
+                          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
+                          class: 'switch-checkbox', 
+                          readonly: true,
+                          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
                 - else
                   %td= @user.send(attr)
         #submissions-content.tab-pane{class: "#{'active' if params[:tab] == 'submissions-content'}"}
--- a/spec/controllers/admin/users_controller_spec.rb
+++ b/spec/controllers/admin/users_controller_spec.rb
@@ -18,6 +18,19 @@ describe Admin::UsersController do
       expect(response).to render_template :index
     end
   end
+  describe 'PATCH #toggle_confirmation' do
+    it 'confirms user' do
+      user_to_confirm = create(:user, email: 'unconfirmed_user@osem.io', confirmed_at: nil)
+      patch :toggle_confirmation, id: user_to_confirm.id, user: { to_confirm: 'true' }
+      user_to_confirm.reload
+      expect(user_to_confirm.confirmed?).to eq true
+    end
+    it 'undo confirmation of user' do
+      patch :toggle_confirmation, id: user.id, user: { to_confirm: 'false' }
+      user.reload
+      expect(user.confirmed?).to eq false
+    end
+  end
   describe 'PATCH #update' do
     context 'valid attributes' do
       before :each do
--- a/app/views/admin/users/_form.html.haml
+++ b/app/views/admin/users/_form.html.haml
@@ -1,5 +1,22 @@
 = semantic_form_for [:admin, @user] do |f|
   = f.inputs 'Basic Information' do
+    .pull-right
+      %b
+        Confirmed?
+      - if can? :toggle_confirmation, @user
+        = check_box_tag @user.id, @user.id, @user.confirmed?, 
+          method: :patch, 
+          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
+          class: 'switch-checkbox', 
+          readonly: false,
+          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
+      - else
+        = check_box_tag @user.id, @user.id, @user.confirmed?, 
+          method: :patch, 
+          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
+          class: 'switch-checkbox', 
+          readonly: true,
+          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
     = f.input :is_admin, hint: 'An admin can create a new conference, manage users and make other users admins.'
     = f.input :name, :as => :string
     = f.input :email
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -19,6 +19,11 @@ Osem::Application.routes.draw do
 
   namespace :admin do
     resources :users
+    resources :users do
+      member do
+        patch :toggle_confirmation
+      end
+    end
     resources :comments, only: [:index]
     resources :conference do
       resource :contact, except: [:index, :new, :create, :show, :destroy]


app/views/admin/users/index.html.haml
```diff
@@ -23,10 +23,9 @@
             %td
               = user.id
             %td
-              - if user.confirmed?
-                confirmed
-              - else
-                unconfirmed
+              - if can? :update, user
+                = check_box_tag user.id, user.id, user.confirmed?, method: :patch, url: "/admin/users/#{user.id}/toggle_confirmation?value=", class: 'switch-checkbox', data: { size: 'small', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
+
```
differentreality: Instead of the parameter **value** you should use a name that reflects what you are actually doing.
And it should be sent inside the user parameters.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,19 @@ def index
       @users = User.all
     end
 
+    def toggle_confirmation
+      user = User.find(params[:id])
```
differentreality: Check how cancancan works. https://github.com/CanCanCommunity/cancancan/wiki/Authorizing-controller-actions
This is already done with **load_and_authorize_resource**

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,19 @@ def index
...
+
+      if params[:value] == "true"
```
differentreality: We use strong parameters, check how we do things in the controllers!

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,19 @@ def index
...
+        user.confirm
+        user.save
+      elsif params[:value] == "false"
+        user.revoke
```
differentreality: This is hard to understand when reading. 
It is also a bit misleading because we are using rolify gem that uses `user.revoke`.  https://github.com/RolifyCommunity/rolify/wiki
I think you should just do whatever you need to do in the controller.

app/views/admin/users/index.html.haml
```diff
@@ -10,7 +10,7 @@
     %table.table.table-striped.table-bordered.table-hover.datatable
       %thead
         %th ID
-        %th State
+        %th Confirmation State
```
differentreality: You don't need to change that.

app/views/admin/users/index.html.haml
shlok007: Okay, and also can you please explain what the three parameters next to the check_box_tag means. I am having a hard time figuring that out.

app/controllers/admin/users_controller.rb
shlok007: Thank you. I'll go through it.

app/views/admin/users/index.html.haml
shlok007: I would suggest to keep it that way. The Bootstrap-Switch-Checkbox would simply toggle between 'Yes' and 'No'. Changing a state of user to 'Yes' or 'No' doesn't convey much.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,17 @@ def index
...
 
+    def toggle_confirmation
+      if params[:user][:confirmation_state] == "true"
+        @user.confirm
+        @user.save
+      elsif params[:user][:confirmation_state] == "false"
+        @user.confirmed_at = nil
+        @user.save
+      end
+      render :nothing => true
```
differentreality: Use new ruby syntax here please.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,17 @@ def index
       @users = User.all
     end
 
+    def toggle_confirmation
+      if params[:user][:confirmation_state] == "true"
```
differentreality: Single quotes here please.

differentreality: Check out how strong parameters work, and let's rewrite this.

shlok007: Okay. I will. :)

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,19 @@ def index
       @users = User.all
     end
 
+    def toggle_confirmation
+      if params[:user][:should_be_confirmed] == 'true'
+        params[:user].delete(:should_be_confirmed)
+        params[:user][:confirmed_at] = Time.now
+        @user.update_attributes(user_params)
+      elsif params[:user][:should_be_confirmed] == 'false'
+        params[:user].delete(:should_be_confirmed)
```
shlok007: It's a bit messy :stuck_out_tongue: . @differentreality Is it possible to send different values other than 'true' and 'false' from the Bootstrap-Switch-Checkbox? I guess it would really help shorten the code.

app/views/admin/users/index.html.haml
differentreality: The text of the switch checkbox can say whatever we want it to say. As long as it fits nicely it could say `Confirmed` and `Unconfirmed`

differentreality:
@shlok007 Please push different commits when you make changes, it is easier to review. Once the code is ready to be merged, you can squash your commits appropriately.

app/controllers/admin/users_controller.rb
differentreality: No you cannot. Did you read about strong parameters? The logic might not apply 100% but I would like that we still whitelist the parameter we want to be using and call user_params instead of calling parameterers directly on Lines 14 and 18.

shlok007: @differentreality `should_be_confirmed` is not a part of user model. Should a parameter not being a part of that model be whitelisted? If yes, I'll permit it in `user_params` too :).

differentreality: Yes, let's rewrite this...
Can we also come up with a better variable name?

app/views/admin/users/index.html.haml
shlok007: toggle
Here is the actual problem, the side which has more number of characters gets squeezed everytime :stuck_out_tongue:. Can this be avoided?

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,18 @@ def index
...
 
+    def toggle_confirmation
+      if user_params[:to_confirm] == 'true'
+        params[:user].delete(:to_confirm)
+        params[:user][:confirmed_at] = Time.now
+      elsif user_params[:to_confirm] == 'false'
+        params[:user].delete(:to_confirm)
+        params[:user][:confirmed_at] = nil
+      end
+      @user.update_attributes(user_params)
```
ChrisBr: What happens if something goes wrong while saving?

shlok007: @ChrisBr The botstrap-switch initiates a request from javascript. `redirect_to` doesn't work in this situation. What should be the best way to handle the this?

differentreality:
@shlok007 we already have something set for that in OSEM. You can do it the same way.
In the meantime, there are some comments that are no more visible about how to proceed with rewriting the controller action in your PR. I would focus on that first. Once the core functionality is there and is properly done, then we can expand it and add the appropriate messages.

app/views/admin/users/index.html.haml
differentreality: Not out of the box. But Confirmation State 'Yes' also doesn't make a lot of sense.  Perhaps the %th should be `Confirmed?` And then Yes/No will make more sense.

app/views/admin/users/index.html.haml
```diff
@@ -23,10 +23,8 @@
             %td
               = user.id
             %td
-              - if user.confirmed?
-                confirmed
-              - else
-                unconfirmed
+              - if can? :update, user
+                = check_box_tag user.id, user.id, user.confirmed?, method: :patch, url: "/admin/users/#{user.id}/toggle_confirmation?user[to_confirm]=", class: 'switch-checkbox', data: { size: 'small', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
```
differentreality: This is one very very long line :)
It would be more readable if you could split it in multiple lines.

app/views/admin/users/index.html.haml
shlok007: Sounds perfect. :+1:

shlok007:
Btw, what's the point of denying confirmation as well? If an admin revokes confirmation of an already confirmed user, they can obviously confirm it again :stuck_out_tongue:. It's not like we could lock them out of their account by not confirming their email address.

differentreality:
We allow the admin to do sth (confirm the user), and we give the admin the option to also undo the action.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,20 @@ def index
       @users = User.all
     end
 
+    def toggle_confirmation
+      #This action allow admins to manually toggle confirmation state of another user 
+      #To change the time of confirmation you can change the value of params[:user][:confirmed_at]
+      if user_params[:to_confirm] == 'true'
+        params[:user].delete(:to_confirm)
+        params[:user][:confirmed_at] = Time.now
```
differentreality: The right way to do this is user.confirm as per Devise's how-to.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,18 @@ def index
...
     end
 
+    def toggle_confirmation
+      #This action allow admins to manually toggle confirmation state of another user 
+      if user_params[:to_confirm] == 'true'
+        @user.confirm
+      elsif user_params[:to_confirm] == 'false'
+        params[:user].delete(:to_confirm)
+        params[:user][:confirmed_at] = nil
+        @user.update_attributes(user_params)
```
differentreality: The way you save the change here can be like you had done it the first time.
Set the attribute you want, save the object.

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,18 @@ def index
...
+      end
+      render nothing: true
```
differentreality: We should change that too.

app/views/admin/users/index.html.haml
```diff
@@ -23,10 +23,11 @@
             %td
               = user.id
             %td
-              - if user.confirmed?
-                confirmed
-              - else
-                unconfirmed
+              - if can? :update, user
+                = check_box_tag user.id, user.id, user.confirmed?, method: :patch, |
```
differentreality: You don't need | at the end of the lines.
Put `method: :patch` on its own line too.

differentreality:
You can run rubocop locally (it is very fast) before pushing your changes to make sure the styling is OK. 
You can use `rubocop -aDc .rubocop.yml` to run rubocop according to our styles file and `-a` option auto-corrects styling, wherever auto-correction is possible.

app/views/admin/users/index.html.haml
differentreality: I have missed this comment, sorry for that. I assume you have already found your answer in http://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-check_box_tag ? :)

app/views/admin/users/index.html.haml
```diff
@@ -23,10 +23,11 @@
             %td
               = user.id
             %td
-              - if user.confirmed?
-                confirmed
-              - else
-                unconfirmed
+              - if can? :update, user
```
differentreality: When that is false, you are not showing anything, and it is impossible to know if the user is confirmed or not!
You can show the switch checkbox disabled. 
Also the action here is not update but `toggle_confirmation`, you should properly authorize. If you are not clear on how to do the authorization, find me on IRC to discuss this more.

app/controllers/admin/users_controller.rb
shlok007: Could replace it with `head :ok`

app/controllers/admin/users_controller.rb
```diff
@@ -40,7 +51,7 @@ def edit; end
 
     def user_params
       params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
-                                   :tshirt, :mobile, :volunteer_experience, :languages, role_ids: [])
+                                   :tshirt, :mobile, :volunteer_experience, :languages, :confirmed_at, :to_confirm, role_ids: [])
```
differentreality: you don't need confirmed_at here any more

app/controllers/admin/users_controller.rb
```diff
@@ -10,6 +10,17 @@ def index
       @users = User.all
     end
 
+    def toggle_confirmation
+      # This action allow admins to manually toggle confirmation state of another user
```
differentreality: This should be on top of the line `def toggle_confirmation`

app/controllers/admin/users_controller.rb
shlok007: I forgot to remove that. Sorry.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -18,6 +18,18 @@
       expect(response).to render_template :index
     end
   end
+  describe 'PATCH #togle_confirmation' do
```
shlok007: Although, it passes, I am not sure if it is the right way to test the action. @differentreality kindly review it.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -18,6 +18,18 @@
...
+    it 'confirms user' do
+      patch :toggle_confirmation, id: user.id, user: { to_confirm: 'true' }
+      confirmed_user = User.find(user.id)
```
differentreality: Why are you fetching that record? We already have the user record.

app/views/admin/users/show.html.haml
```diff
@@ -14,12 +14,22 @@
           - if can? :edit, @user
             .pull-right
               = link_to 'Edit', edit_admin_user_path(@user), class: 'btn btn-primary'
-          - if @user.confirmed?
-            %span.label.label-success
-              confirmed
+          %b
```
differentreality: This can now be part of the table, and the items can then be properly aligned with everything else!

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -18,6 +18,18 @@
       expect(response).to render_template :index
     end
   end
+  describe 'PATCH #togle_confirmation' do
+    it 'confirms user' do
+      patch :toggle_confirmation, id: user.id, user: { to_confirm: 'true' }
```
differentreality: This makes sense only if we make sure that user is not already confirmed. `user` is already confirmed, so we can't really tell if this is working properly.
So before this action, I would create a new user, make sure confirmed_at is nil, and then execute the action.

shlok007:
@differentreality do tell me if any work is pending in this PR.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -18,6 +18,19 @@
       expect(response).to render_template :index
     end
   end
+  describe 'PATCH #togle_confirmation' do
+    it 'confirms user' do
+      unconfirmed_user = create(:user, email: 'unconfirmed_user@osem.io', confirmed_at: nil)
+      patch :toggle_confirmation, id: unconfirmed_user.id, user: { to_confirm: 'true' }
+      confirmed_user = User.find(unconfirmed_user.id)
```
differentreality: You can delete that line, and possibly name the variable `user_to_confirm` or something like that, that makes logical sense

shlok007: The test keeps on failing if I remove that line. 
```ruby
  describe 'PATCH #toggle_confirmation' do
    it 'confirms user' do
      user = create(:user, email: 'unconfirmed_user@osem.io', confirmed_at: nil)
      patch :toggle_confirmation, id: user.id, user: { to_confirm: 'true' }
      expect(user.confirmed?).to eq true
    end
    it 'undo confirmation of user' do
      patch :toggle_confirmation, id: user.id, user: { to_confirm: 'false' }
      expect(user.confirmed?).to eq false
    end
  end
```
This test keeps on failing. I guess it's my unfamiliarity with RSpec that's causing the problem. What am I missing here?

differentreality: Add a `user.reload` before your expect.

shlok007: Thank you!

app/views/admin/conference/_recent_users.html.haml
```diff
@@ -6,18 +6,28 @@
...
             %td= index + 1
             %td= link_to user.email, admin_user_path(user.id)
             %td= user.created_at.strftime('%m/%d/%Y')
             %td
-              - if user.confirmed?
-                %span.label.label-success Confirmed
-              -else
-                %span.label.label-danger Unconfirmed
+              - if can? :toggle_confirmation, @user
+                = check_box_tag user.id, user.id, user.confirmed?,
```
differentreality: I would make this checkbox a disabled one. This page is not where people would usually care to change data. It can be done from admin/users#index and admin/users#show, that sounds good enough to me.

differentreality:
@shlok007 good work! After the last change, please go ahead and squash your commits into 1 for merging.

shlok007:
@differentreality done :+1:

differentreality:
+1

Chaz6:
Thanks!

differentreality:
You are most welcome :)

PR #1101, opened 2016-07-13
Remove cocoon and volunteers tests

differentreality:
While implementing surveys we decided to remove cocoon since it is not being used.

I am putting this into a different PR because we need to remove some other files as well.


differentreality:
This should be merged *AFTER* surveys is implemented, because we are also using nested fields for questions.

hennevogel:
Please push it to the survey PR then :-)

PR #1102, opened 2016-07-14
unused css removed

Ana06:
osem-admin was not included in application.css neither in any other place, so it is not used. 

Some .css from osem-schedule-print do no exit since we merged https://github.com/openSUSE/osem/pull/1030


ChrisBr:
@Ana06 LGTM

PR #1103, opened 2016-07-18
Default value for length in as_json fixed

Ana06:
The default value for length in as_json should be a multiple of EventType::LENGTH_STEP.


ChrisBr:
LGTM :+1:

PR #1104, opened 2016-07-22
Inform that there are no rooms in the schedule

Ana06:
Inform that there are no rooms when creating the schedule with a message instead of creating a room with the title "No rooms". The message also include a link to the place where rooms are created.


ChrisBr:
LGTM :+1:

PR #1106, opened 2016-07-22
Introduce a "Save" button for the admin schedule

Ana06:
The changes were saved after you drop an event, and now it only saves the schedule when you're done with the changes and click the "Save" button.

### DIFF
--- a/app/assets/javascripts/osem-schedule.js
+++ b/app/assets/javascripts/osem-schedule.js
@@ -1,10 +1,13 @@
 var url; // Should be initialize in Schedule.initialize
 var schedule_id; // Should be initialize in Schedule.initialize
 
-function showError(error){
-  // Delete other error messages before showing the new one
+var events_to_save = {};
+var events_to_remove = {};
+
+function showMessage(message, type){
+  // Delete other messages before showing the new one
   $('.unobtrusive-flash-container').empty();
-  UnobtrusiveFlash.showFlashMessage(error, {type: 'error'});
+  UnobtrusiveFlash.showFlashMessage(message, {type: type});
 }
 
 var Schedule = {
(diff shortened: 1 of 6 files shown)


app/controllers/admin/event_schedules_controller.rb
```diff
@@ -31,5 +31,11 @@ def destroy
     def event_schedule_params
       params.require(:event_schedule).permit(:schedule_id, :event_id, :room_id, :start_time)
     end
+
+    def parse_errors(event_schedule)
+      title = event_schedule.event.try(:title).present? ? event_schedule.event.title : 'The event'
```
ChrisBr: Why / How can that happen?
event_schedule validates presence of event and event validates presence of title, so this case shouldn't happen, should it?

app/controllers/admin/event_schedules_controller.rb
```diff
@@ -31,5 +31,11 @@ def destroy
     def event_schedule_params
       params.require(:event_schedule).permit(:schedule_id, :event_id, :room_id, :start_time)
     end
+
+    def parse_errors(event_schedule)
```
ChrisBr: Is this helper function really necessary? I think not ...

ChrisBr:
Also I was able to create two event_schedules for the same event on the same schedule.
Steps to reproduce:

1. Open the schedule in two different browser tabs
2. Schedule on the first tab
3. Schedule on the second tab
4. Save first tab and the second tab
5. Reload the tabs

Result: The event is scheduled on two different times. See attached screenshot.
I see two issues with that: 

1. Something with the JS is broken in this PR
2. We need a validation in the event_schedule.rb model (Please do that in a separate PR!)


schedule

app/assets/javascripts/osem-schedule.js
```diff
@@ -14,66 +17,99 @@ var Schedule = {
...
+  },
+  add: function (previous_parent, new_parent, event) {
+    var params = {
+      previous_parent: previous_parent,
+      new_parent: new_parent,
+      event: event
+    }
+    var event_id = event.attr("event_id");
+    events_to_save[event_id] = params;
+    delete events_to_remove[event_id];
```
ChrisBr: ?
Can we do this somehow different? It's really hard to read/understand!

app/assets/javascripts/osem-schedule.js
```diff
@@ -14,66 +17,99 @@ var Schedule = {
...
-      var error_callback = function(data) {
+      var success_callback_save = function(data) {
+        console.log(data);
+        event.attr("event_schedule_id", data.event_schedule_id);
+        }
+      var error_callback_save = function(data) {
         console.log(data);
-        showError($.parseJSON(data.responseText).errors);
+        errors += $.parseJSON(data.responseText).errors;
+        event.appendTo(events_to_save[key]['previous_parent']);
```
ChrisBr: So in error case it will move the event back to the unscheduled events and shows an error message, right?

Ana06:
@ChrisBr 

> Also I was able to create two event_schedules for the same event on the same schedule.

I know what happened and it is not related to the save feature. When an event is correctly scheduled its `event schedule id `is added to the html as an attribute. So if you have already openned a new tab then the event has no this attribute there and that is why it try to schedule it again. The only way that I think we can avoid that is not using the `event schedule id`, but then we can't use the update and delete method in event_schedule_controller, without id we do not know the route. But I am not sure that we want to do that, at least that will only happens if several organizer are changing the schedule at the same time or if you have several tabs openned. And with the validation we won't have repeated event, only we will get an error when scheduling some of them.

Reloading the page after saving can also be useful to ensure that we have the correct version in case that we had several tabs openned. I can do it in this PR.

app/assets/javascripts/osem-schedule.js
Ana06: To unscheduled events or to its previous time and room if it was already scheduled.

app/assets/javascripts/osem-schedule.js
Ana06: Why is it hard to understand?

I have two arrays, one for the events that must be saved and another one for those that must be removed. So if you add a event to the `events_to_save` array, it has to be deleted from `events_to_remove`, as you don't want to remove it anymore. I don't understand what is not clear.

app/controllers/admin/event_schedules_controller.rb
Ana06: It should, but it is an error, if `.save`, `.update` or `.destroy` fails maybe a validation has failed, so I think we can not assume that event and title are present.

Ana06:
I thinking that I can fix it without changing the routes, adding some checks in `event_schedules_controller.rb`

- Unschedule an event twice works.
- Schedule in another time and/or room and event that has already been scheduled twice keep the last one, so I think it is the expected result.
- Schedule an unscheduled event twice fails. With the validation only the first one will be created. I could try to update it when create fails in the controller, so it would keep the second one (I think that is what is expected).
- Schedule in another time and/or room a scheduled event that has been unscheduled in another tab fails (I have realised that currently the javascript doesn't render an error, but it should because rails get a 404 error, so I have to fix that). I could try to create it when update fails in the controller, so it would keep the second one (I think that is what is expected).

app/controllers/admin/event_schedules_controller.rb
Ana06: Why not? When clicking save I am scheduling several events, so I need to link the error messages to render all together and I use that function to allow that this is done. I add the name of the event (taking into account that validation may have failed), the the errors in brackets and a dot at the end. And I have to do that in `create`, `update` and `destroy`

Look:

image

ChrisBr:
As discussed in IRC: We should introduce batch / bulk :create / :update / :destroy actions. Making a request for each changed event_schedule is ugly and not very high-performance. 

Look for example at this tutorial: http://ericlondon.com/2012/07/18/creating-a-ruby-on-rails-admin-form-to-bulk-update-multiple-models.html

Ana06:
I've introduced the bulk `:create,` `:update` and `:destroy` actions, so every time the save button is clicked only 3 request are done, one per action.

The data saved in `events_to_create`, `events_to_update` and `events_to_remove` hashes use the `event_id` as key. For `events_to_update` I sent a hash which key is the `event_id` and its value is another hash. The key of the second hash is the `event_schedule_id` and the value the params needed to update it. For `events_to_create`  I sent a hash which key is the `event_id` and its value the params needed to create the EventSchedule. Lastly, for `events_to_remove` I sent a hash where the key is the `event_id` and the value the `event_schedule_id` to remove. Basically, if you remove the `event_id` you have what is needed to create, update and delete an `EventSchedule`. So, I don't need to send the `event_id` to the controller but I have decided to send it because for example for `events_to_create` I would like to send an array of objects (with the params), but an array of objects it is sent to the controller like

`{ "0" => {"room_id"=>"25", "start_time"=>"2016-07-14 9:00"}, "1" => {"room_id"=>"3", "start_time"=>"2016-07-14 9:00"}, "2" => {"room_id"=>"7", "start_time"=>"2016-07-14 9:00"}}`

which is a hash where the key is the index and I have to parse it again in the controller, so I directly send the hash with the `event_id`.

Ana06:
I also reload the page after saving the events and I render the error/success message after that.

Ana06:
@ChrisBr @hennevogel @differentreality 
Could you please review this? :birthday:

Ana06:
giphy

differentreality:
@ChrisBr do you wanna finish reviewing this?

differentreality:
Can you please rebase this and take into consideration the changes in https://github.com/openSUSE/osem/pull/1207 ? You have hardcoded some urls in your js that need updating.

differentreality:
Scenario for a little malfunction:
EventA is scheduled at 10:00. Save is fired.
I remove eventA and put in its place eventB (again at 10:00). Save is fired, eventA gets unscheduled, but eventB does not get scheduled.

When save is fired we first fire create (in our case for eventB) and then destroy (in our case for eventA). When we try to schedule eventB, we cannot do that because eventA is still there (since it has not been destroyed yet).

Ana06:
Yes, when I created the PR the replacement feature was not implemented yet, so I didn't take into account.

It'll take me some time to change it so I'll close it and I'll reopen it when it is fixed.

differentreality:
Why do we need to close this? I think only #1168 affects #1082.

Ana06:
But with this the replacement doesn't work, right?

differentreality:
@Ana06 sorry, yes, it's both #1106 and #1168. I left a relevant comment in #1168 about solving it.

PR #1107, opened 2016-07-24
Avoid overlapping in the admin schedule

Ana06:
**NOT FINISHED YET**

Branch copied from #1106, copied from #1104.

Not allow that an event is scheduled at the same time than other event in the admin schedule.

Before:

cargar_el_22_7_2016_a_las_18_04_04


ChrisBr:
as it depends on #1106, we will reopen this one as soon as #1106 is merged!

PR #1108, opened 2016-07-27
Online payment feature - direct stripe integration

rishabhs95:
This PR is in succession to PR #1042.

This makes the feature SAQ-A PCI-compliant which is the easiest and cheapest amongst all available, thereby reducing the PCI burden for our clients.

screen shot 2016-07-27 at 8 36 32 pm

screen shot 2016-07-27 at 8 36 47 pm

screen shot 2016-07-27 at 8 37 47 pm

screen shot 2016-07-27 at 8 38 00 pm


app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    customer = Stripe::Customer.create(
+      :email => params[:stripeEmail],
+      :source  => params[:stripeToken]
+    )
+
+    gateway_response = Stripe::Charge.create(
+      :customer    => customer.id,
+      :amount      => @total_amount_to_pay.cents,
+      :description => 'Rails Stripe customer',
```
lagartoflojo: We should probably improve this message :)

spec/factories/payments.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :payment do
+    user
+    conference
+    last4 4242
+    authorization_code 1234567890
```
lagartoflojo: since `authorization_code` is a string, better make this also a string.

spec/factories/payments.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :payment do
+    user
+    conference
+    last4 4242
```
lagartoflojo: since `last4` is a string, better make this also a string.

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      :currency    => @conference.tickets.first.price_currency
+    )
+
+    payment = Payment.purchase(gateway_response, current_user, @conference)
```
lagartoflojo: Why not handle this the way you did it for active merchant?
You can build a Payment with all the necessary data, and then call `payment.purchase` (no parameters) which will call Stripe and handle all that.

app/models/payment.rb
```diff
@@ -6,8 +6,6 @@ class Payment < ActiveRecord::Base
   attr_accessor :stripeEmail
```
lagartoflojo: why is this camel-cased? looks veeery weird for ruby...

app/models/payment.rb
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  def amount_to_pay
+    Ticket.total_price(conference, user, paid: false).cents
+  end
+
+  def purchase
+    customer = Stripe::Customer.create email: stripe_customer_email,
+                                       source: stripe_customer_token,
+                                       description: user.name
```
lagartoflojo: If the customer orders more than once, will this create the customer many times in Stripe? Would it be better to retrieve the customer if they already exist?

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+    @unpaid_ticket_purchases = current_user.ticket_purchases.unpaid.by_conference(@conference)
+  end
+
+  def create
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+
+    @payment = Payment.new payment_params.merge(stripe_customer_email: params[:stripeEmail],
+                                                stripe_customer_token: params[:stripeToken],
+                                                amount: @total_amount_to_pay.cents,
```
lagartoflojo: Since you are calculating the amount to pay inside the Payment model, why do you need to pass it here?

app/models/payment.rb
rishabhs95: Actually as we make the payment instantly, I think we can call Stripe's `Charge#create` directly instead of creating and saving customer details in our DB. 
I will remove new Customer creation part.

app/views/tickets/index.html.haml
```diff
@@ -35,12 +35,11 @@
...
               %i.fa.fa-shopping-cart
-            = link_to 'Continue without a Ticket!', conference_conference_registration_path(@conference.short_title),
-            class: 'btn btn-danger btn-sm'
+            = link_to 'Cancel registration', conference_conference_registration_path(@conference.short_title), method: :delete, class: 'btn btn-danger btn-sm'
   .row
     .col-md-13
       %p.text-muted.text-center
         %small
-          * Buying a ticket is not mandatory. Checkout will be at the conference registration.
+          * Buying a ticket is mandatory. Your registration will not complete until you buy a ticket.
```
lagartoflojo: In this screen, you can select no tickets and click "continue"...
Do you think we should validate that at least 1 ticket was chosen before continuing?

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,35 @@
...
+    %li
+      %a
+        = humanized_money_with_symbol @total_amount_to_pay
+  %script.stripe-button{ src: "https://checkout.stripe.com/checkout.js",
+        data: { amount: @total_amount_to_pay.cents,
+          currency: @total_amount_to_pay.currency,
+          image: image_url('OSEM_ICON.jpg'),
+          name: ENV['OSEM_NAME'] || 'OSEM',
+          description: "book your tickets",
+          key: Rails.application.secrets.stripe_publishable_key,
```
lagartoflojo: Here we're reading the stripe key from `secrets` instead of `configuration`, so configuring the keys with env variables actually doesn't work.

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,35 @@
...
+          %td
+            = humanized_money_with_symbol ticket.quantity * ticket.price
+
+= form_tag conference_payments_path do
+  %ul.price-tags.pull-right
+    %li
+      %a
+        = humanized_money_with_symbol @total_amount_to_pay
+  %script.stripe-button{ src: "https://checkout.stripe.com/checkout.js",
+        data: { amount: @total_amount_to_pay.cents,
```
lagartoflojo: You can also specify the email here so that users don't have to type it in when the checkout modal opens.

app/views/payments/new.html.haml
```diff
@@ -0,0 +1,18 @@
...
+          %ul
+            - @payment.errors.full_messages.each do |msg|
+              %li= msg
+    .col-xs-8.col-xs-offset-2.well
+      = render partial: 'payment'
+  .row
+    .col-md-13
+      %p.text-muted.text-center
+        %small
+          The payment is totally secure. Your credit card details will be sent directly to our payment processor.
```
lagartoflojo: In American English, "totally" in this context sounds colloquial or even sarcastic.
I would say: 
> All payments are handled securely by our payment processor, [Stripe](https://stripe.com/about).

app/views/payments/_payment.html.haml
rishabhs95: Yes, that's why I am telling users to put their key in `config/secrets.yml` in our documentation

app/views/payments/new.html.haml
rishabhs95: Oh... okay :)

app/views/payments/_payment.html.haml
lagartoflojo: But then why does the `stripe.rb` initializer try to read environment variables and set them in `Rails.configuration`?

lagartoflojo: Using environment variables is, in general, a best practice anyway :)

http://www.clearlytech.com/2014/01/04/12-factor-apps-plain-english/#iclassicon-checkiiii.config---storeconfigintheenvironment

rishabhs95: Okay, I will check the API key like in initialiser.rb then :)

app/views/payments/_payment.html.haml
rishabhs95: Okay, I've added that

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,35 @@
...
+          currency: @total_amount_to_pay.currency,
+          image: image_url('OSEM_ICON.jpg'),
```
lagartoflojo: per the documentation PR https://github.com/openSUSE/osem/pull/1116, we should probably already be loading the image from the env variable, no?

lagartoflojo: Also, the icon you uploaded seems unnecessarily big, considering the small size used in the checkout form.

config/initializers/stripe.rb
```diff
@@ -0,0 +1,6 @@
+Rails.configuration.stripe = {
+  :publishable_key => ENV['STRIPE_PUBLISHABLE_KEY'] || Rails.application.secrets.stripe_publishable_key,
+  :secret_key      => ENV['STRIPE_SECRET_KEY'] || Rails.application.secrets.stripe_secret_key
+}
```
lagartoflojo: In `secrets.yml` you could just do 

```
  stripe_secret_key: <%= ENV["STRIPE_SECRET_KEY"] %>
```

and then you don't need this special code.
Then we can just use `Rails.secrets` directly, forget about `Rails.configuration`, and it will work with both `secrets.yml` and ENV vars.

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def create
+    @payment = Payment.new payment_params.merge(stripe_customer_email: params[:stripeEmail],
+                                                stripe_customer_token: params[:stripeToken],
+                                                user: current_user,
+                                                conference: @conference)
+
+    if @payment.purchase && @payment.save
+      update_purchased_ticket_purchases
+      redirect_to conference_conference_registration_path(@conference.short_title), flash:
+        { success: 'Thanks! You have purchased your tickets successfully.' }
```
differentreality: You should be able to use `success` instead of `flash` directly.

app/views/payments/_payment.html.haml
differentreality: The icon is really cute, but please take into consideration that OSEM is not exclusively used by openSUSE. We should not force openSUSE logo here. The name of the file also seems to say OSEM_icon, but it is really openSUSE icon. We could use OSEM logo instead.

app/models/ability.rb
```diff
@@ -81,6 +81,7 @@ def signed_in(user)
 
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
+    can [:new], Payment, user_id: user.id
```
differentreality: You should probably be authorizing `create` action too.

app/views/payments/_payment.html.haml
rishabhs95: `OSEM_ICON` is the general name for icon used by our application for representing your organisation. As the organisation is openSUSE in the default case, so I've added openSUSE icon in the default case.

app/controllers/payments_controller.rb
rishabhs95: Okay, I've edited that

app/models/ability.rb
rishabhs95: Ah, I will do that... thanks :)

app/views/payments/_payment.html.haml
lagartoflojo: Maybe this line should say: `image: ENV['STRIPE_FORM_ICON']` (or something like that), so organizers can set it or not. If they don't, the checkout form won't show anything (and really, it doesn't HAVE to show anything).

app/views/tickets/index.html.haml
rishabhs95: Okay, I've added a check for that

app/views/payments/_payment.html.haml
rishabhs95: I will remove the icon from it then

app/controllers/ticket_purchases_controller.rb
```diff
@@ -8,9 +8,11 @@ def create
     message = TicketPurchase.purchase(@conference, current_user, params[:tickets][0])
     if message.blank?
       if current_user.ticket_purchases.by_conference(@conference).unpaid.any?
-        redirect_to new_conference_payment_path, notice: 'Please pay here to purchase tickets.'
+        redirect_to new_conference_payment_path,
+                    notice: 'Please pay here to purchase tickets.'
       else
-        redirect_to conference_conference_registration_path(@conference.short_title)
+        redirect_to conference_tickets_path(@conference.short_title),
+                    error: 'Please purchase atleast one ticket to continue.'
```
lagartoflojo: atleast => at least

config/initializers/stripe.rb
```diff
@@ -1,6 +1 @@
-Rails.configuration.stripe = {
-  :publishable_key => ENV['STRIPE_PUBLISHABLE_KEY'] || Rails.application.secrets.stripe_publishable_key,
-  :secret_key      => ENV['STRIPE_SECRET_KEY'] || Rails.application.secrets.stripe_secret_key
-}
-
-Stripe.api_key = Rails.configuration.stripe[:secret_key]
+Stripe.api_key = Rails.application.secret.stripe_secret_key
```
lagartoflojo: This works with env variables if you update secrets.yml as I mentioned in another comment.. will that change be in another commit?

config/secrets.yml.example
```diff
@@ -64,5 +64,5 @@ production:
 
   # Register on stripe and add LIVE keys here
   # https://dashboard.stripe.com/account/apikeys
-  stripe_publishable_key: ''
-  stripe_secret_key: ''
```
lagartoflojo: Please apply this change for the other environments as well :)

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def new
+    @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+    @unpaid_ticket_purchases = current_user.ticket_purchases.unpaid.by_conference(@conference)
+  end
+
+  def create
+    @payment = Payment.new payment_params.merge(stripe_customer_email: params[:stripeEmail],
+                                                stripe_customer_token: params[:stripeToken],
+                                                user: current_user,
+                                                conference: @conference)
```
lagartoflojo: I think if you move the `merge` call to the `payment_params` method, then this line will be cleaner :)
(It will be just `Payment.new payment_params`)

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    let!(:conference) { create(:conference) }
+    let(:ticket_1) { create(:ticket, price: 10, price_currency: 'USD', conference: conference) }
+    let(:payment) { create(:payment, user: user, conference: conference) }
+
+    it ' returns correct unpaid amount' do
+      create(:ticket_purchase, ticket: ticket_1, user: user, quantity: 8)
+      expect(payment.amount_to_pay).to eq(8000)
+    end
+  end
+end
```
lagartoflojo: No tests for the `#purchase` method?

spec/models/payment_spec.rb
```diff
@@ -32,4 +33,61 @@
...
+    end
+
+    context 'if the payment is not successful' do
+      before { StripeMock.prepare_card_error(:invalid_number) }
+
+      let(:payment) { create(:payment, user: user, conference: conference, stripe_customer_token: 'bogus_card_token', stripe_customer_email: user.email) }
+
+      before { payment.purchase }
+
+      context 'when the card is invalid' do
```
lagartoflojo: Can the gem also simulate a TimeOut? Are there any other errors we should consider?

app/models/payment.rb
```diff
@@ -0,0 +1,41 @@
...
+                                             amount: amount_to_pay,
+                                             currency: conference.tickets.first.price_currency
+
+    self.amount = gateway_response[:amount]
+    self.last4 = gateway_response[:source][:last4]
+    self.authorization_code = gateway_response[:id]
+    self.status = 'success'
+    true
+
+  rescue => error
```
lagartoflojo: I recommend against rescuing from all types of errors.
For example, if we have some sort of exception from our own code (I dunno, division by zero or the Stripe gem changes its parameters, whatever), then users will see it as an error in their payment, and WE will never find out that something is wrong with our code.

It's better if we rescue from the exceptions that the Stripe gem can explicitly raise (invalid card? gateway error? timeout? ...?) and if there are any other errors, they will be sent to Errbit for analysis.

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,35 @@
...
+        = humanized_money_with_symbol @total_amount_to_pay
+  %script.stripe-button{ src: "https://checkout.stripe.com/checkout.js",
+        data: { amount: @total_amount_to_pay.cents,
+          email: current_user.email,
+          currency: @total_amount_to_pay.currency,
+          name: ENV['OSEM_NAME'] || 'OSEM',
+          description: "book your tickets",
+          key: Rails.application.secrets.stripe_publishable_key,
+          locale: "auto"}}
+  = link_to 'Edit Purchase', conference_tickets_path(@conference.short_title), class: 'btn btn-primary'
```
lagartoflojo: I'm not too excited about how this page looks. The "Pay with card" button looks nothing like the other buttons in the rest of OSEM, so it looks out of place. Plus we have two blue buttons. And there's the green ribbon which looks a bit out of place.. I know we are not designers 😂 but maybe we can improve it a bit. An idea:

**Before**
screen shot 2016-08-09 at 6 42 46 pm
 

**After**
screen shot 2016-08-09 at 6 42 29 pm

spec/features/ticket_purchases_spec.rb
```diff
@@ -26,26 +26,12 @@
         fill_in "tickets__#{ticket.id}", with: '2'
         expect(current_path).to eq(conference_tickets_path(conference.short_title))
 
-        click_button 'Support'
+        click_button 'Continue'
 
+        expect(current_path).to eq(new_conference_payment_path(conference.short_title))
+        expect(flash).to eq('Please pay here to purchase tickets.')
         purchase = TicketPurchase.where(user_id: participant.id, ticket_id: ticket.id).first
         expect(purchase.quantity).to eq(2)
```
lagartoflojo: Seems like this test is not finished.. the purchase process doesn't end here, does it?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        expect(payment.authorization_code).to eq('test_ch_3')
+      end
+    end
+
+    context 'if the payment is not successful' do
+      before { StripeMock.prepare_card_error(:invalid_number) }
+
+      let(:payment) { create(:payment, user: user, conference: conference, stripe_customer_token: 'bogus_card_token', stripe_customer_email: user.email) }
+
+      before { payment.purchase }
```
lagartoflojo: We could probably merge the two `before` blocks

spec/features/ticket_purchases_spec.rb
rishabhs95: Adding tests

spec/models/payment_spec.rb
rishabhs95: Done.

app/views/payments/_payment.html.haml
rishabhs95: Done.

spec/models/payment_spec.rb
rishabhs95: Considering them

app/models/payment.rb
rishabhs95: Okay

rishabhs95:
We are testing our Payment class... I just added the line `expect{ Stripe::Charge.create }.to raise_error(...)` to show that an exception is raised but it is handled by `rescue`.

app/controllers/payments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def create
+    @payment = Payment.new payment_params
+
+    if @payment.purchase && @payment.save
+      update_purchased_ticket_purchases
+      redirect_to conference_conference_registration_path(@conference.short_title),
+                  notice: 'Thanks! You have purchased your tickets successfully.'
+    else
+      @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
```
ChrisBr: Why don't we show the error in a flash message?
In Payment::purchase we add StripeError to ``errors.add(:base, error.message)`` so we should also propagate it to the users, shouldn't we?

app/controllers/ticket_purchases_controller.rb
```diff
@@ -4,32 +4,22 @@ class TicketPurchasesController < ApplicationController
   authorize_resource :conference_registrations, class: Registration
 
   def create
+    TicketPurchase.by_conference(@conference).unpaid.by_user(current_user).destroy_all
```
ChrisBr: Why not 

``
current_user.ticket_purchases.by_conference(@conference).unpaid.destroy_all
``

app/views/conference_registrations/show.html.haml
```diff
@@ -89,29 +89,27 @@
...
+                = tickets.first.price.symbol
+                = humanized_money tickets.first.price
+            %br
+          - if @tickets.any?
+            = link_to 'Buy more tickets', conference_tickets_path(@conference.short_title), class: "btn btn-default"
+          - else
+            You haven't bought any tickets.
+            = link_to 'Please buy some tickets to support us!', conference_tickets_path(@conference.short_title)
+            %p
+              (Your participation won't be valid without buying a ticket)
```
ChrisBr: Is this really true? I mean for instance we also have free conferences?

app/views/conference_registrations/show.html.haml
```diff
@@ -89,29 +89,27 @@
           %span.fa-stack
             %i.fa.fa-square-o.fa-stack-2x
             %i.fa.fa-ticket.fa-stack-1x
-          Tickets
+          Ticket Purchases
```
ChrisBr: Shouldn't we move the heading also inside the ``tickets.any?``?
If there are no tickets, we don't need this section at all, do we?

lagartoflojo: Well we don't allow registrations with no tickets. We will allow free tickets to allow for this (https://github.com/openSUSE/osem/pull/1118).

ChrisBr: Ok

app/views/conference_registrations/show.html.haml
ChrisBr: So this one is then also obsolete with https://github.com/openSUSE/osem/pull/1118, right?

lagartoflojo: yes :)

rishabhs95: I will just rebase #1118 over this branch, once it gets merged

app/controllers/ticket_purchases_controller.rb
rishabhs95: okay, i have improved that

app/controllers/payments_controller.rb
rishabhs95: I do that in `app/views/payments/new.html.haml#7`

app/views/conference_registrations/show.html.haml
differentreality: I had left a similar comment in the closed PR too.
I still don't like that phrase to exist here. It is not true, not even for this PR, because the registration to the conference is successful, and only after a successful registration we are redirected to the tickets page. I also find it very much against open source conferences, and this is the open source event manage!

lagartoflojo: The idea is that everyone has to get a ticket to come in the conference, even if the conference is free. In the PR #1118, we allow the creation of free tickets.
[Here](https://github.com/openSUSE/osem/pull/1108/files/0bc97925c6addad145afb8ddc830526b7b1e7a03#diff-3160d7d59726e8fa48c13ee0c12dcaf4R10) we check wether the user chose a ticket.

Considering the free-ticket scenario though, I think we should remove the mention of "purchase" or "buy" from the interface, and speak only about "getting" a ticket.

ChrisBr: +1 for @lagartoflojo suggestion!

rishabhs95: Then I think that a message like "Your participation won't be valid without getting a ticket" will be better here.

app/views/payments/_payment.html.haml
```diff
@@ -0,0 +1,32 @@
...
+          %td
+            = ticket.quantity
+          %td
+            = humanized_money_with_symbol ticket.price
+          %td
+            = humanized_money_with_symbol ticket.quantity * ticket.price
+
+= form_tag conference_payments_path do
+  %script.stripe-button{ src: "https://checkout.stripe.com/checkout.js",
+        data: { amount: @total_amount_to_pay.cents,
```
lagartoflojo: These lines are too indented.

spec/support/external_request.rb
```diff
@@ -1,6 +1,6 @@
 # Mock external requests to youtube
 require 'webmock/rspec'
-WebMock.disable_net_connect!(allow_localhost: true)
+WebMock.allow_net_connect!
```
lagartoflojo: We should not allow WebMock to connect to EVERYTHING, only to Stripe.

lagartoflojo:
@differentreality @hennevogel @ChrisBr
Hi guys! We're ready to merge this.. do you have any additional comments before we do? Thanks!

ChrisBr:
@lagartoflojo @rishabhs95 sorry, didn't have time to look at it!

However, congratulations and thanks for your good and continuous work this summer!

PR #1112, opened 2016-07-29
Creation of my_conference repeated in ability_spec

Ana06:
my_conference is created several times in ability_spec but it is not necessary. :smile:


ChrisBr:
LGTM :+1:

PR #1113, opened 2016-07-30
Add cancel/replacement label in All Events page and fixes bug caused by #1082

nishanthvijayan:
-  ~~Access to proposal#show fails if room is not set.(nil)~~
- I've added canceled/replacement labels in All Events like @Ana06  suggested [here](https://github.com/openSUSE/osem/pull/1082#issuecomment-234498947)
  labels_all_events_full

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -503,4 +503,22 @@ module ApplicationHelper
       'deleted'
     end
   end
+
+  def canceled_replacement_event_label(event, event_schedule, *label_classes)
+    if event.state == 'canceled' || event.state == 'withdrawn'
+      content_tag :span, 'CANCELED', class: (['label', 'label-danger'] + label_classes)
+    elsif event.state == 'confirmed' && event_schedule.present? && (!event_schedule.intersecting_events.canceled.empty? || !event_schedule.intersecting_events.withdrawn.empty?)
+      content_tag :span, 'REPLACEMENT', class: (['label', 'label-info'] + label_classes)
+    end
+  end
+
+  def replacement_event_notice(event, event_schedule)
+    if event.state == 'confirmed' && event_schedule.present? && (!event_schedule.intersecting_events.withdrawn.empty? || !event_schedule.intersecting_events.canceled.empty?)
+      replaced_event = (event_schedule.intersecting_events.withdrawn.first || event_schedule.intersecting_events.canceled.first).event
+      content_tag :span do
+        concat content_tag :span, 'Please note that this talk replaces '
+        concat link_to replaced_event.title, conference_program_proposal_path(@conference.short_title, replaced_event.id)
+      end
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -507,13 +507,13 @@ module ApplicationHelper
   def canceled_replacement_event_label(event, event_schedule, *label_classes)
     if event.state == 'canceled' || event.state == 'withdrawn'
       content_tag :span, 'CANCELED', class: (['label', 'label-danger'] + label_classes)
-    elsif event.state == 'confirmed' && event_schedule.present? && (!event_schedule.intersecting_events.canceled.empty? || !event_schedule.intersecting_events.withdrawn.empty?)
+    elsif event_schedule.present? && event_schedule.replacement?
       content_tag :span, 'REPLACEMENT', class: (['label', 'label-info'] + label_classes)
     end
   end
 
-  def replacement_event_notice(event, event_schedule)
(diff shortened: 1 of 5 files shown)


differentreality:
In the schedule the label is before the title, in allevents why do we show it all the way down?

differentreality:
We have also started having a lot of duplication of same code for showing the label and the text for canceled/replacement talks. We should put that in one place, so that if we want to change the text from 'This talk replaces...' to something else, we only have to make the change in 1 place.

nishanthvijayan:
@differentreality  How about near the subtitle?
label_before_subtitle

app/helpers/application_helper.rb
```diff
@@ -366,4 +366,28 @@ def carousel_item_class(number, carousel_number, num_cols, col)
     end
     item_class
   end
+
+  def canceled_replacement_label(event, *label_classes)
```
differentreality: `canceled_replacement_event_label`

differentreality:
I am thinking:

screenshot-7
screenshot-6

nishanthvijayan:
The message here cannot link to the replaced event as it will break the link of the main event box
all_events
replacement_proposal_show
canceled_proposal_show

Ana06:
@nishanthvijayan 

> The message here cannot link to the replaced event as it will break the link of the main event box

Why not? The word `more` currently has a link, it is the same one, but it would work exactly the same if it was different. If you only add the link to the replaced event name I think that there is not any problem and that it may be useful. :smiley:

nishanthvijayan:
@Ana06 
`more` does kinda break the main link (I think).
Check out the demo.
See how, if there's a `more` link, the area below `more` is not linked anymore.
It works correctly if the abstract is small enough and there's no `more` link displayed.In this case it's not really a big problems as the breaking happens to only a small portion.
But in my case since the link to replaced event is at the top it will break the whole link.

Ana06:
@nishanthvijayan you are right, it doesn't work properly. I have fixed it in https://github.com/openSUSE/osem/pull/1132, that way we can keep the `more` link and you can also add a link to the replaced event, which I think is useful. :wink:

nishanthvijayan:
:clap: Wow, that was fast! Thanks! :smile:

differentreality:
So this now depends on #1132

differentreality:
@nishanthvijayan please rebase :)

nishanthvijayan:
@differentreality Done

differentreality:
@nishanthvijayan do you want to set up a few events in the review app to show the different behavior in all cases?

app/helpers/application_helper.rb
```diff
@@ -491,4 +491,22 @@ def user_change_description(version)
       end
     end
   end
+
+  def canceled_replacement_event_label(event, event_schedule, *label_classes)
+    if event.state == 'canceled' || event.state == 'withdrawn'
```
differentreality: That's a bit too random. We don't really want this for any cancelled/withdrawn proposal, just the ones that are scheduled.

differentreality:
We should show the labels in the admin schedule too. Right now a cancelled proposal still shows up in scheduling but the organizers have no way of knowing that it is cancelled, so that they can remove it and add a replacement talk.

differentreality:
There are still things to be done here.

nishanthvijayan:
Before the CANCELD and REPLACEMENT tag was introduced in #1082 
this was the existing situation,

> Currently if a scheduled event is cancelled/withdrawn,
    It remains scheduled.The public schedule page will still display the event without any change
    The event disappears from the admin scheduling interface.

So the definition of `intersecting_events` method was based on this.
Now since canceled events appear in the admin schedule, this definition has to changed.

> Right now a cancelled proposal still shows up in scheduling but the organizers have no way of knowing that it is cancelled, so that they can remove it and add a replacement talk.

With the current system, if a canceled event is unscheduled and then a replacement talk is added in its stead, it will not be detected as a replacement talk.
To do this, the definition of the the `intersecting_events` has to be changed to use PaperTrail::Versions of EventSchedules to see if there exists any versions of EventSchedule (event: 'delete') of the same Schedule that was scheduled in the same time slot and room.
So this can be completed only after #1159 has been merged, as it introduces tracking in EventSchedule and Schedule models.

app/models/event_schedule.rb
```diff
@@ -28,4 +28,8 @@ def end_time
   def intersecting_events
```
differentreality: This returns event_schedules and not events.

differentreality:
Please rebase!

app/helpers/application_helper.rb
ChrisBr: I think this is fine as we only use the function for scheduled events in the view anyway ...

ChrisBr:
@nishanthvijayan can we expect that you rebase anytime soon?

differentreality:
Tracking this in #1253

PR #1114, opened 2016-07-31
Show Event missing info in admin

nishanthvijayan:
(Branch from #1019 - Merge only after it is merged)

Displays missing info tooltip shown in proposal#index in admin events#index as well.

screenshot 2016-07-31 14 33 26
  
  
  
  
Displays who added commercials for an event.

screenshot from 2016-07-31 14 38 40

### DIFF
--- a/app/views/proposal/_form.html.haml
+++ b/app/views/proposal/_form.html.haml
@@ -40,3 +40,6 @@
                         - if can? :destroy, commercial
                           = link_to 'Delete', conference_program_proposal_commercial_path(@conference.short_title, @event.id, commercial.id),
                           :method => :delete, :data => { :confirm => 'Are you sure?' }, class: 'btn btn-danger'
+                    .text-right
+                      = "added by "
+                      = link_to_user(commercial.versions.last.whodunnit)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -383,7 +383,7 @@ module ApplicationHelper
       reverse.sub(',', ' dna ').reverse
   end
 
-  def change_creator_link(user_id)
+  def link_to_user(user_id)
     user = User.find_by(id: user_id)
     if user
       link_to user.name, admin_user_path(id: user_id)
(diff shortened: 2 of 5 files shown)


nishanthvijayan:
@differentreality How about something like this?
missing_info

differentreality:
That's a very good start :D

nishanthvijayan:
What should I improve here?

differentreality:
I would include the speaker name somewhere.

differentreality:
And I guess we can also show in the same table the text present in event.description

nishanthvijayan:
@differentreality Now that table with missing info 's added, the tooltip seems  overkill.What do you think?

differentreality:
I am wondering though, when I have 100 proposals, how can I easily see if my keynote talk is missing anything? Perhaps instead of the tooltip, we could add a similar page in event#show.

nishanthvijayan:
like this?
progress
progress2

differentreality:
I was thinking a tab in events#show (and no more JS). The progress bar (without the pop up) would still be nice to have as a visual.

app/views/proposal/_form.html.haml
```diff
@@ -40,3 +40,7 @@
                         - if can? :destroy, commercial
                           = link_to 'Delete', conference_program_proposal_commercial_path(@conference.short_title, @event.id, commercial.id),
                           :method => :delete, :data => { :confirm => 'Are you sure?' }, class: 'btn btn-danger'
+                    - if commercial.versions.last.whodunnit.to_i != @event.submitter.id
```
differentreality: I think we can show the name of the user that added the commercial even if that user is the submitter. So that our layout is consistent and explicit (always shows the same thing and we don't let users guess what it might mean when there is no name).

nishanthvijayan:
Instead of progress bar, I added a badge at the top indicating how many things need to be completed.

badge

missing_info_separate_tab

differentreality:
As discussed over IRC, we will add a new menu item **Reports** to show all necessary info.
#1155

differentreality:
events#show looks good to me!
events#index missing info should go, as we will put that list in reports

If you want, you can just leave here the events#show with Tasks tab, and some other changes you've made, and Reports can become a new PR.

nishanthvijayan:
Sure.I've removed all the events#index related stuff.I'll open a new PR with Reports soon.

differentreality:
:+1:

PR #1115, opened 2016-07-31
Display correct quantity of tickets in admin views

rishabhs95:
This displays the correct quantity of purchased tickets after integration of online payment feature.
This will be rebased over #1108 after merge.

### DIFF
--- a/app/views/admin/tickets/show.html.haml
+++ b/app/views/admin/tickets/show.html.haml
@@ -26,7 +26,7 @@
             %td
               = buyer.name
               %span.label.label-success
-                = buyer.ticket_purchases.find_by(ticket_id: @ticket.id).quantity
+                = buyer.ticket_purchases.where(ticket_id: @ticket.id).sum('quantity')
             %td
               = buyer.email
             %td
--- a/app/views/admin/tickets/show.html.haml
+++ b/app/views/admin/tickets/show.html.haml
@@ -15,6 +15,7 @@
       %thead
         %th #
         %th Name
+        %th Quantity
         %th E-Mail
         %th Affiliation
         %th Paid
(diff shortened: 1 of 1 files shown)


rishabhs95:
I think these changes are okay @lagartoflojo?

lagartoflojo:
What's the bug that you are solving here? Or is it a refactoring? If it's a bug, I think we also need a test case :) The test case should fail on master, but pass on this branch.

rishabhs95:
Just a small change in admin's ticket purchases view. There are no major model/controller/feature changes that's why all tests are passing at the moment on the master branch.

lagartoflojo:
@rishabhs95 I still don't get it - I see it's a small change, but is it a bug fix or a refactoring?

differentreality:
Maybe I am missing sth with how we are now implementing tickets, but is it possible that we have more than one TicketPurchase for the same user and the same ticket?

rishabhs95:
Yes, it can be possible for a user to have more than one ticket purchase for the same ticket and that's why we are showing the total ticket quantity here. I would say it as a minor bug fix.

lagartoflojo:
If it's a small bug fix, then we need a test case as I said before...

>  If it's a bug, I think we also need a test case :) The test case should fail on master, but pass on this branch.

rishabhs95:
As the change is not related to any controller or any model, rather I am just repairing a `view` change. Currently, we don't have much testing in the views that is why I haven't fixed/changed any test case.

rishabhs95:
Yes, a user can have multiple ticket purchases as they can be paid with different times and needs to be linked with its respective payment. @differentreality

differentreality:
We don't really want to introduce any more view tests btw...

I wonder if it would make more sense to add a new column on that table with the quantity though.

rishabhs95:
Yes, we can add another column for the quantity... the quantity sticker looks nice though :)
What do you think/prefer?

differentreality:
I prefer the column (which is also sortable btw). 
I find the sticker fancy and all, but perhaps for  another page, not the main page for tickets.

rishabhs95:
Okay... Pushing the new changes :)

differentreality:
:+1:

Issue #581, opened 2015-01-15
Loading user page lasts forever


ChrisBr:
Loading the user page lasts very long.
I guess it selects all users from the db, on events.opensuse there are about 1600!
We should load users page by page

differentreality:
What about will-paginate gem?

nishanthvijayan:
Will-paginate only handles pagination.Sorting and seraching will have to be implemented seperately if they are needed

hennevogel:
We already have datatables which can do [server-side processing](https://datatables.net/manual/server-side)

PR #1117, opened 2016-08-01
Implement server-side processing of datatables in users#index

nishanthvijayan:
Solves #581 
-  I've not implemented sorting by Attended Conferences
-  Search checks for matches only in name and email fields


differentreality:
@hennevogel do you have any input about this?

app/controllers/concerns/datatable_serverside_processing.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  private
+
+  def records
+    @records ||= fetch_records
+  end
+
+  def fetch_records
+    sort_direction = params[:sSortDir_0] == 'desc' ? 'desc' : 'asc'
```
ChrisBr: move this to a dedicated function.

```
def sort_direction
...
```

app/controllers/concerns/datatable_serverside_processing.rb
```diff
@@ -0,0 +1,44 @@
...
+    sort_column = @sort_columns[params[:iSortCol_0].to_i]
```
ChrisBr: This also

``` 
def sort_column
...
```

app/controllers/concerns/datatable_serverside_processing.rb
```diff
@@ -0,0 +1,44 @@
...
+    records = @collection.order("#{sort_column} #{sort_direction}")
+
+    if params[:sSearch].present? && @search_columns.present?
```
ChrisBr: Move this also to a dedicated function!

differentreality:
@mdeniz did you have a chance to take a look? Any thoughts?

Ana06:
@differentreality @mdeniz @nishanthvijayan what is the state of this? 🤔

differentreality:
Huge change, but I do not know if there is another way.
We do need the feature one way or the other.

@mdeniz had mentioned at some point he can have a look, but review is still pending.

I don't know if perhaps @bear454 is also up for this.

bear454:
@nishanthvijayan: I pulled down this branch, and tried it against master, it doesn't work for me. I spent a little time diving into making changes, and I just don't see an easy route forward without rearchitecting.

IMO the best way forward would be to implement https://github.com/jbox-web/ajax-datatables-rails instead of reinventing the wheel.

hennevogel:
I agree with @bear454, let's close this. Sorry @nishanthvijayan

PR #1118, opened 2016-08-01
ability to create and purchase free tickets

rishabhs95:
This provides the ability for organiser to create free tickets and attendees to purchase them.
This will be rebased over #1108 after merge.

### DIFF
--- a/app/controllers/ticket_purchases_controller.rb
+++ b/app/controllers/ticket_purchases_controller.rb
@@ -11,8 +11,13 @@ class TicketPurchasesController < ApplicationController
         redirect_to new_conference_payment_path,
                     notice: 'Please pay here to get tickets.'
       else
-        redirect_to conference_tickets_path(@conference.short_title),
-                    error: 'Please get at least one ticket to continue.'
+        if current_user.ticket_purchases.by_conference(@conference).paid.any?
+          redirect_to conference_conference_registration_path(@conference.short_title),
+                      notice: 'You have free tickets for the conference.'
+        else
+          redirect_to conference_tickets_path(@conference.short_title),
+                      error: 'Please get at least one ticket to continue.'
+        end
       end
     else
       redirect_to conference_conference_registration_path(@conference.short_title),
--- a/app/controllers/ticket_purchases_controller.rb
+++ b/app/controllers/ticket_purchases_controller.rb
@@ -10,14 +10,12 @@ class TicketPurchasesController < ApplicationController
       if current_user.ticket_purchases.by_conference(@conference).unpaid.any?
         redirect_to new_conference_payment_path,
                     notice: 'Please pay here to get tickets.'
+      elsif current_user.ticket_purchases.by_conference(@conference).paid.any?
+        redirect_to conference_conference_registration_path(@conference.short_title),
+                    notice: 'You have free tickets for the conference.'
       else
-        if current_user.ticket_purchases.by_conference(@conference).paid.any?
-          redirect_to conference_conference_registration_path(@conference.short_title),
-                      notice: 'You have free tickets for the conference.'
-        else
-          redirect_to conference_tickets_path(@conference.short_title),
-                      error: 'Please get at least one ticket to continue.'
-        end
+        redirect_to conference_tickets_path(@conference.short_title),
+                    error: 'Please get at least one ticket to continue.'
       end
     else
       redirect_to conference_conference_registration_path(@conference.short_title),
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -23,7 +23,7 @@ class TicketPurchase < ActiveRecord::Base
     ActiveRecord::Base.transaction do
       conference.tickets.each do |ticket|
         quantity = purchases[ticket.id.to_s].to_i
-        # if the user bought the ticket, just update the quantity
+        # if the user bought the ticket and is still unpaid, just update the quantity
         if ticket.bought?(user) && ticket.unpaid?(user)
           purchase = update_quantity(conference, quantity, ticket, user)
         else
@@ -39,10 +39,20 @@ class TicketPurchase < ActiveRecord::Base
   end
 
   def self.purchase_ticket(conference, quantity, ticket, user)
-    purchase = new(ticket_id: ticket.id,
-                   conference_id: conference.id,
-                   user_id: user.id,
-                   quantity: quantity) if quantity > 0
+    if quantity > 0
+      if ticket.price_cents.zero?
+        purchase = new(ticket_id: ticket.id,
+                       conference_id: conference.id,
+                       user_id: user.id,
+                       quantity: quantity,
+                       paid: true)
+      else
+        purchase = new(ticket_id: ticket.id,
+                       conference_id: conference.id,
+                       user_id: user.id,
+                       quantity: quantity)
+      end
+    end
     purchase
   end
 
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -40,18 +40,11 @@ class TicketPurchase < ActiveRecord::Base
 
   def self.purchase_ticket(conference, quantity, ticket, user)
     if quantity > 0
-      if ticket.price_cents.zero?
-        purchase = new(ticket_id: ticket.id,
-                       conference_id: conference.id,
-                       user_id: user.id,
-                       quantity: quantity,
-                       paid: true)
-      else
-        purchase = new(ticket_id: ticket.id,
-                       conference_id: conference.id,
-                       user_id: user.id,
-                       quantity: quantity)
-      end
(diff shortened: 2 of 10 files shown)


app/controllers/ticket_purchases_controller.rb
```diff
@@ -11,8 +11,13 @@ def create
         redirect_to new_conference_payment_path,
                     notice: 'Please pay here to get tickets.'
       else
-        redirect_to conference_tickets_path(@conference.short_title),
-                    error: 'Please get at least one ticket to continue.'
+        if current_user.ticket_purchases.by_conference(@conference).paid.any?
+          redirect_to conference_conference_registration_path(@conference.short_title),
+                      notice: 'You have free tickets for the conference.'
```
lagartoflojo: This notice will be displayed on the "registration confirmation" page, right?
Maybe the message could be something like "You have successfully registered" or something similar?

rishabhs95: This notice will only be displayed if the user chooses ONLY "free tickets" out of all available tickets.
I am adding this message as they may be getting those "free tickets" for the second time and something like "You have successfully registered" can be confusing in that case

hennevogel:
What's the workflow of this one? I register -> buy a ticket -> "buy" a free ticket?

rishabhs95:
The workflow goes like this:
User registers -> gets a choice to "get" some tickets out of available tickets(free/paid) -> User has to get atleast one of them.

differentreality:
So what if the free ticket is optional?

rishabhs95:
With this PR we plan to provide flexibility for organisers. It totally depends upon the organiser if they want to create an option for `free-tickets` or not. 

1. If organisers create `free-ticket`, attendees can get that ticket and mark their presence for the conference.
2. If organisers doesn't create any `free-ticket`, attendees will have to pay for the tickets they want to get and will be able to pay for them online.

We will be able to cover both the scenarios with this PR.

differentreality:
What if the organizer wants to create an **_optional_** free ticket?

Eg. I want to create a free ticket for a social event. It should **not** be mandatory.

rishabhs95:
We were tending towards making the tickets mandatory for the attendees as it was most suitable option for events like conferences, etc.
What do you think about this @lagartoflojo?

lagartoflojo:
@differentreality We talked about this with @rishabhs95 and @kalabiyau, and we came to the conclusion that making tickets mandatory for attendees is a good idea for a few reasons:

- Simpler workflow, less options, less code. By making every attendee get a ticket, we don't have to handle different workflows in the code, nor do we have to give more options to the organizers. In general, more code means more bugs, and more options means more documentation and confusion.
- Simpler counting of attendees. Right now, registrations can include more than one attendee, since one person can register, but can get many tickets. Therefore, the registration count is not really the full count of attendees. If everyone gets a ticket, everyone is accounted for, and organizers just need one "source of truth": the ticket count.
- Better expectations for attendees. I would say that to attend the grand majority of events (be it conferences, concerts, or anything else), people have to get a ticket. That's a normal expectation. Getting a ticket is a "standard" prerequisite to attend an event. But if we have this duality between registration / ticket, that sometimes you need a ticket, sometimes you don't, this becomes quite confusing. Let's keep it simple.
- Finally, making tickets mandatory is not really that much of a hassle. All it requires is clicking on the "up arrow" at least once on the ticket selection page. 

Since many conferences that use OSEM are free to attend, we also thought about creating a free ticket by default, which organizers can leave as a sane default or delete/modify if necessary.

What do you think? Do you think making tickets optional is necessary?

differentreality:
I could swear we had discussed this before.
How about I give some use cases and you confirm how they will work with the new implementation?

* openSUSE conference is free to attend. You come, you enjoy. No tickets, no hassle, right?
* openSUSE conference is free to attend but you can support it with personal donations if you want to. For that purpose we have a supporter ticket, which you can buy, if you want to. What you propose in this case is that I create 1 ticket with money, and 1 free ticket, and that all attendees are obliged to get the free ticket if they want to register?
* opensSUSE conference is free to attend. We have a party, for which we have booked a place, where only openSUSE ppl are allowed to enter. So people, need to  print out a free ticket, to gain access to the party. We create a free ticket, but it makes no sense for everyone to "buy" it, unless of course they are actually coming, because that would mess up our numbers and our preparation for the party (think number of free drinks we offer based on the estimated number of attendees).

lagartoflojo:
> I could swear we had discussed this before.
Hmm maybe I wasn't involved in that discussion, I don't recall..

> openSUSE conference is free to attend. You come, you enjoy. No tickets, no hassle, right?

Well, openSUSE organizers *do* encourage people to register, so that they have a better count of who's going. So when people register, they'll just have to choose the free ticket.

> openSUSE conference is free to attend but you can support it with personal donations if you want to. For that purpose we have a supporter ticket, which you can buy, if you want to. What you propose in this case is that I create 1 ticket with money, and 1 free ticket, and that all attendees are obliged to get the free ticket if they want to register?

Correct. When attendees who don't want to donate get to the ticket selection page, they just select the free ticket.

> opensSUSE conference is free to attend. We have a party, for which we have booked a place, where only openSUSE ppl are allowed to enter. So people, need to  print out a free ticket, to gain access to the party. We create a free ticket, but it makes no sense for everyone to "buy" it, unless of course they are actually coming, because that would mess up our numbers and our preparation for the party (think number of free drinks we offer based on the estimated number of attendees).

When you say "openSUSE people", you mean the conference attendees? They don't really have to print out anything... Actually, currently, tickets are a "virtual" thing, there's no QR, nothing special. Check-in at the venue is just as it is today, by looking up the attendee's name in the list.

But by adding an additional "Party" ticket (free or not), then people who want to go to the party get it, and organizers get a better idea of how many people want to attend.

lagartoflojo:
@rishabhs95 As per conversation on IRC with the OSEM team, could you please create a PR so that when an organizer creates a new conference, OSEM creates a free ticket for that conference by default?

rishabhs95:
Okay, I will do that @lagartoflojo

rishabhs95:
I have added those changes in this PR itself.

rishabhs95:
I have updated the PR with latest improvements

rishabhs95:
Please check if this PR is good to merge. also, do tell if it requires any improvements

rishabhs95:
can you check this PR too @differentreality? :)

differentreality:
This is already assigned to @lagartoflojo and he knows the workflow you agreed upon much better than me, so I think that he should be the one to review this.

rishabhs95:
cool... @lagartoflojo... I think we were almost good with this PR... what do you think?

ChrisBr:
@lagartoflojo @rishabhs95 what's missing here?

rishabhs95:
sorry for being late @ChrisBr, was busy with some work these days... I just have to go through some PR reviews

ChrisBr:
@rishabhs95 cool :)

rishabhs95:
I've made some changes with respect to the code reviews and all the tests are passing now, please do have a look @lagartoflojo

spec/models/conference_spec.rb
```diff
@@ -1585,11 +1585,11 @@
   end
 
   describe 'after_create' do
-    let!(:conference) { Conference.new(title: 'ABC',short_title: 'XYZ',start_date: Date.today, end_date: Date.today + 10) }
-    let!(:free_ticket) { conference.tickets.first }
+    before { @conference = Conference.new }
```
lagartoflojo: I actually like the previous version of this test better.
In the original version, we create a conference and test that, once created, it has at least one free ticket.

The new version checks the *implementation*. Meaning: when we create a conference, make sure that a certain method is called.
But what if we break that method and it DOESN'T create a free ticket anymore? This test will still pass, but the conference will not have a free ticket anymore.

app/models/ticket_purchase.rb
```diff
@@ -39,10 +39,16 @@ def self.purchase(conference, user, purchases)
...
-    purchase = new(ticket_id: ticket.id,
-                   conference_id: conference.id,
-                   user_id: user.id,
-                   quantity: quantity) if quantity > 0
+    if quantity > 0
+      purchase = new(ticket_id: ticket.id,
+                     conference_id: conference.id,
+                     user_id: user.id,
+                     quantity: quantity,
+                     paid: false)
```
lagartoflojo: why not just say `paid: ticket.price_cents.zero?)` ?
Then you don't need that second `if ticket.price_cents.zero?....`

spec/models/conference_spec.rb
rishabhs95: Yeah, and the thing is that I haven't included a free ticket in factory, so even after running all the sub models, the free ticket should come from the `after_create` callback, not from the factory.

rishabhs95:
The new tests should look better to me and are working as expected. Please do have a look @lagartoflojo

lagartoflojo:
Thanks @rishabhs95!! Merging this now :)

PR #1123, opened 2016-08-05
style: if ! changed by unless

Ana06:
- if ! changed by unless as it is more readable
- !object.blank? is equivalent to object.present?, also more readable


ChrisBr:
LGTM :+1:

PR #1124, opened 2016-08-06
Allow organizers to filter changes by conference

nishanthvijayan:
Branch from #1019
Adds a dropdown using which organizers can view changes in a specific conference
- The dropdown lists all conferences in which the current user is an organizer (all for admin)

per_conference_filter

### DIFF
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,26 +1,23 @@
 module Admin
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
       authorize! :index, PaperTrail::Version.new(item_type: 'User')
       @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
       @conference_id = params[:conference_id].to_i unless params[:conference_id].nil?
-      if @conference_id.nil?
-        @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", @conf_ids_for_organizer])
-      elsif !Conference.exists?(id: @conference_id)
+
+      unless @conference_id.nil? || Conference.exists?(id: @conference_id)
         redirect_to admin_revision_history_path, error: "Conference with ID #{@conference_id} does not exist!"
         return
-      else
-        authorize! :index, PaperTrail::Version.new(conference_id: @conference_id)
-        @versions = PaperTrail::Version.where(conference_id:  @conference_id)
       end
+
+      return unless @conference_id.present?
+      authorize! :index, PaperTrail::Version.new(conference_id: @conference_id)
+      @versions = @versions.where(conference_id:  @conference_id)
     end
 
     def revert_attribute
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_attribute, @version
-
       if params[:attribute] && @version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.keys.include?(params[:attribute])
         if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
           flash[:error] = 'The item is already in the state that you are trying to revert it back to'
@@ -42,9 +39,6 @@ module Admin
     end
 
     def revert_object
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_object, @version
-
       if @version.event != 'create'
         if @version.reify.save
           flash[:notice] = 'The selected change was successfully reverted'
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,15 +1,14 @@
 module Admin
   class VersionsController < Admin::BaseController
+    load_resource :conference, find_by: :short_title
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @conference_id = params[:conference_id].to_i unless params[:conference_id].nil?
+      @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role(:organizer, current_user).pluck(:short_title)
 
-      return unless @conference_id.present?
-      authorize! :index, PaperTrail::Version.new(conference_id: @conference_id)
-      @versions = @versions.where(conference_id:  @conference_id)
+      return unless @conference.present?
+      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
+      @versions = @versions.where(conference_id:  @conference.id)
     end
 
     def revert_attribute
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -8,9 +8,12 @@ module Admin
       @conference_id = params[:conference_id].to_i unless params[:conference_id].nil?
       if @conference_id.nil?
         @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", @conf_ids_for_organizer])
+      elsif !Conference.exists?(id: @conference_id)
+        redirect_to admin_revision_history_path, error: "Conference with ID #{@conference_id} does not exist!"
+        return
       else
-        @versions = PaperTrail::Version.where(['conference_id IN (?)', (@conf_ids_for_organizer & [@conference_id])]).
-                                                                    where.not(item_type: 'User')
+        authorize! :index, PaperTrail::Version.new(conference_id: @conference_id)
+        @versions = PaperTrail::Version.where(conference_id:  @conference_id)
       end
     end
 
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -4,7 +4,7 @@ module Admin
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role(:organizer, current_user).pluck(:short_title)
+      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
 
       return unless @conference.present?
       authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -7,11 +7,6 @@ module Admin
       @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
       @conference_id = params[:conference_id].to_i unless params[:conference_id].nil?
 
-      unless @conference_id.nil? || Conference.exists?(id: @conference_id)
-        redirect_to admin_revision_history_path, error: "Conference with ID #{@conference_id} does not exist!"
-        return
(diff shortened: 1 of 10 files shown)


differentreality:
I would also add a conference specific route, and add a link for the changelog in the left conference menu.

app/controllers/admin/versions_controller.rb
```diff
@@ -4,8 +4,14 @@ class VersionsController < Admin::BaseController
 
     def index
       authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
+      @conference_id = params[:conference_id].to_i unless params[:conference_id].nil?
```
differentreality: Can't you just load and authorize conference resource?

app/controllers/admin/versions_controller.rb
```diff
@@ -4,8 +4,14 @@ class VersionsController < Admin::BaseController
 
     def index
       authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
```
differentreality: `Conference.accessible_by(current_ability)` should work

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,18 @@
 module Admin
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_resource :conference, find_by: :short_title
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
       authorize! :index, PaperTrail::Version.new(item_type: 'User')
```
differentreality: Why do we need this line if we have L4 ?

differentreality:
Why don't you add abilities for the other roles as well, like you started doing in https://github.com/openSUSE/osem/pull/1128/files ?

spec/controllers/admin/versions_controller_spec.rb
```diff
@@ -97,5 +97,20 @@
         expect(flash[:error]).to match('Revert failed. Attribute missing or invalid')
       end
     end
+
+    describe 'GET #index' do
+      it 'raises error if conference_id is invalid' do
+        sign_in admin
+        get :index, conference_id: 88
```
differentreality: conference_id should be its short_title

differentreality:
I bumped on to this:

https://github.com/openSUSE/osem/blob/master/app/views/admin/versions/_object_desc_and_link.html.haml#L157 here you need to check if the version.conference_id and the conference actually exist, remember there are version records that exist prior to the changelog, and even though we run a rake task to include the conference_id attribute, that might not always be successful (bad things happen and things that shouldn't get deleted).

app/controllers/admin/versions_controller.rb
differentreality: Do we need this like that?

nishanthvijayan:
@differentreality I didn't understand.What other abilities needs to be added?

differentreality:
You only added abilities for the role organizer. What about the others? 
With cancancan we can easily show whatever each user can see. More over you plan on allowing users with cfp role to revert changes (from the event history tab), so that needs to be authorized too. I have added some comments in the gdoc about some things we need to add for the changelog. Let's talk more about it on IRC.

nishanthvijayan:
@differentreality 
  - I've opened up revision_history pages to users with organizer, info_desk and cfp roles.
  - I've added ability rules such that they view and revert everything they `:manage` 
  - volunteers_coordinators do not manage any models that are being tracked, so they don't have access to the pages.

app/models/ability.rb
```diff
@@ -210,8 +210,10 @@ def signed_in_with_cfp_role(user)
...
     end
 
-    can [:index, :revert_object, :revert_attribute], PaperTrail::Version, item_type: 'Event', conference_id: conf_ids_for_cfp
-    can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version,
+        item_type: %w(Event EventType Track DifficultyLevel EmailSettings Room Cfp Program Comment), conference_id: conf_ids_for_cfp
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version,
+        ["item_type = 'Commercial' AND conference_id IN (?) AND (object LIKE '%Event%' OR object_changes LIKE '%Event%')", conf_ids_for_cfp] do |version|
       version.item_type == 'Commercial' && conf_ids_for_cfp.include?(version.conference_id) &&
       (version.object.to_s.include?('Event') || version.object_changes.to_s.include?('Event'))
```
differentreality: why do we also keep these lines after you added L216?

nishanthvijayan: The conditions inside the block is for checking authentication.The sql is there so that accessible_by(current_ability) can fetch the accessible records.
https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities-with-Blocks#fetching-records

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -136,7 +136,7 @@
       = link_to(admin_conference_roles_path(@conference.short_title)) do
         %span.fa.fa-group
         Roles
-  - if can? :index, PaperTrail::Version.new(item_type: 'User')
+  - if can?(:index, PaperTrail::Version.new(conference_id: @conference.id, item_type: 'Event')) || can?(:index, PaperTrail::Version.new(conference_id: @conference.id, item_type: 'Registration'))
```
differentreality: Can we just have -` if can? :index, PaperTrail::Version.new` ?

nishanthvijayan: No.
conference_id: @conference.id part is so that the link appears only if in conferences that the user has access to.Without this part, the Revision Histroy(conference sidebar) link will appear for all conferences if user has organizer/cfp/info_desk role in atleast 1 conference.

Again `can?(:index, PaperTrail::Version.new(conference_id: @conference.id)` is not enough
as this will make the link appear only for organizers.

Organizers have :index permission for all versions that for which the user is an organizer and 'User'. Cfp users have :index permission only in Event EventType Track DifficultyLevel EmailSettings Room Cfp Program Comment versions of confs that they have cfp_role
and Info Desk users have access to Registration type versions
So `can?(:index, PaperTrail::Version.new(conference_id: @conference.id, item_type: 'Event'))` will authorize organizers and cfp team users and `can?(:index, PaperTrail::Version.new(conference_id: @conference.id, item_type: 'Registration')` will authorize info_desk users

differentreality:
https://github.com/nishanthvijayan/osem/blob/441d2873d2d206d363f50894823a6c95d1f12401/app/views/admin/versions/_object_desc_and_link.html.haml#L153

raises 

> undefined method `unserialize_attributes_for_paper_trail!' for #<Class:0x007f88622a13d0>

nishanthvijayan:
Fixed it.It was caused by versions missing conference_id.

differentreality:
How is it fixed? The line isn't touched.

nishanthvijayan:
I pushed some temp commits with begin rescue blocks to figure out where the error was happening in the demo.The demo was failing  fro versions 12, 22, 32, 52, 63 and 72.They were failing because they didn't have conference_id and the errors were because in places like this https://github.com/nishanthvijayan/osem/blob/441d2873d2d206d363f50894823a6c95d1f12401/app/views/admin/versions/_object_desc_and_link.html.haml#L139 the Conference.find() would fail if there's no conference_id.



> 
  https://github.com/nishanthvijayan/osem/blob/441d2873d2d206d363f50894823a6c95d1f12401/app/views/admin/versions/_object_desc_and_link.html.haml#L153
  raises
      undefined method `unserialize_attributes_for_paper_trail!' for #Class:0x007f88622a13d0

Did this happen in your local machine?

differentreality:
Yeap, and still happening.

nishanthvijayan:
Could you please use byebug and tell me which model's version is failing.?
If you had tested #1159 also locally, this error might be because of that.It introduces tracking for Schedule and EventSchedule and if there are those versions in your db, that might be causing this.

differentreality:
Solved by removing versions from EventSchedule (introduced in another, yet unmerged, PR)

nishanthvijayan:
Anything left to be done here?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -153,6 +158,9 @@
...
 - unless %w(Conference Subscription Registration User).include?(version.item_type)
-  = "in conference"
-  = link_to Conference.find(version.conference_id).short_title,
-          edit_admin_conference_path(id: Conference.find(version.conference_id).short_title)
+  - if conference
+    = "in conference"
+    = link_to conference_short_title,
+            edit_admin_conference_path(id: conference_short_title)
+  - else
+    in deleted conference
```
differentreality: We still have the ID the conference had though right?
We should show that...
`in deleted conference with ID 123`

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,15 +1,19 @@
+- conference = Conference.find_by(id: version.conference_id)
+- conference_short_title = conference.try(:short_title) || 'deleted_conference'
+- conference_title = conference.try(:title) || 'Deleted Conference'
+
 - if version.item_type == 'UsersRole'
   - users_role = get_version_object(version)
   = 'role'
-  = link_to users_role.role.name, admin_conference_role_path(conference_id: Conference.find(version.conference_id).short_title, id: users_role.role.name)
+  = link_to users_role.role.name, admin_conference_role_path(conference_id: conference_short_title, id: users_role.role.name)
```
differentreality: Hmmm I didn't properly catch this before.

Here it is possible that we will create a link that does not exist. That does not sound right :/
If there is no conference record there is no point in linking anywhere, because we can't visit that link anyway.

differentreality:
@nishanthvijayan can you rebasee this please?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -25,149 +31,152 @@
...
 - when 'Schedule'
   = link_if_alive version, "Schedule #{version.item_id}",
-            admin_conference_schedule_path(conference_id: Conference.find(version.conference_id).short_title, id: version.item_id)
+            admin_conference_schedule_path(conference_id: conference_short_title, id: version.item_id)
 
 - when 'Conference'
   = 'conference'
-  = link_to Conference.find(version.conference_id).title,
-          edit_admin_conference_path(id: Conference.find(version.conference_id).short_title)
+  = link_to Conference.find(version.item_id).title,
```
differentreality: Let's change that in a similar way as in https://github.com/openSUSE/osem/pull/1124/files#diff-0f35c627650a51bc88770fd668fe2e0eR182 ?

differentreality:
@nishanthvijayan ping :)

differentreality:
@nishanthvijayan can you help me out with this?
https://github.com/openSUSE/osem/blob/master/app/views/admin/versions/_object_desc_and_link.html.haml#L48
When is `object.try(:event_id)` going to return a result?

app/views/admin/versions/_object_desc_and_link.html.haml
differentreality: This no longer needs any attention since our definition for @versions in the controller depends on the version.conference_id to correspond to an existing conference (for which the user is organizer).
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/versions_controller.rb#L8

differentreality:
Closing this in favor of #1530

PR #1125, opened 2016-08-07
Schedule tags improved

Ana06:
As @differentreality suggested I have created this new PR to improve the tags introduced in https://github.com/openSUSE/osem/pull/1082 by @nishanthvijayan 

The whole `CANCELED` tag is displayed even in small devices. The `REPLACEMENT` one is too big for small devices, so it is shorted, but it is better than before I think. Some screenshots:

**Before:**

image

image

**Now:**

image

image

image

image


ChrisBr:
LGTM

PR #1126, opened 2016-08-08
Do not use unless with else

Ana06:
Solve Rubocop offense introduced in https://github.com/openSUSE/osem/pull/1123


differentreality:
:+1:

PR #1128, opened 2016-08-09
Improve event history page

nishanthvijayan:
( Branch from #1019 )
The History page in event#show has been changed to match the look of the main changelog page introduced in #1019 .

event_history_content


nishanthvijayan:
@differentreality Since cfp_team users [can manage Events](https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L175), I think we should allow cfp_team users to  revert Event changes.What do you think?

differentreality:
Yes that sounds good!

differentreality:
We should be probably also showing here the commercials of the event.

nishanthvijayan:
@differentreality  It now displays changes commercials as well.

app/views/admin/events/show.html.haml
```diff
@@ -9,29 +9,36 @@
...
+
+                    %small.text-muted
+                      = distance_of_time_in_words(Time.now,version.created_at) + ' ago'
+                      %br
+                      = "(#{version.created_at.strftime('%B %-d, %Y %H:%M')})"
+
                   %br
+                  = render partial: 'shared/object_changes', locals: { version: version }
+                %td
+                  = render partial: 'shared/changelog_actions', locals: { version: version }
```
differentreality: Empty line at the end of file needed

nishanthvijayan: Done

app/models/ability.rb
```diff
@@ -210,6 +210,10 @@ def signed_in_with_cfp_role(user)
       role.resource_type == 'Conference' && role.name == 'cfp' &&
       (Conference.with_role(:cfp, user).pluck(:id).include? role.resource_id)
     end
+
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
+      %w(Event Commercial).include?(version.item_type) && (conf_ids_for_cfp.include? version.conference_id)
```
differentreality: That would allow users with cfp role to also edit Commercials of conference. You should provide a quick fix for this here, and let's discuss a bit about improving the authorization of versions in general. I opened #1144

app/models/ability.rb
```diff
@@ -211,8 +211,10 @@ def signed_in_with_cfp_role(user)
       (Conference.with_role(:cfp, user).pluck(:id).include? role.resource_id)
     end
 
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version, item_type: 'Event', conference_id: conf_ids_for_cfp
     can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
-      %w(Event Commercial).include?(version.item_type) && (conf_ids_for_cfp.include? version.conference_id)
+      version.item_type == 'Commercial' && conf_ids_for_cfp.include?(version.conference_id) &&
+      (version.object.to_s.include?('Event') || version.object_changes.to_s.include?('Event'))
```
nishanthvijayan: Querying for versions of event commercials is tricky.
For all PaperTrail::Version objects,
 -  object field has a representation of the object before the changes in that version.
    So object is nil for 'create' type versions
 -  object_changes field has representation of the changes in the object from its previous state.So 'create' type versions' object_changes field will have its commercialable_type in it.
Hence the two checks here.

app/controllers/admin/events_controller.rb
```diff
@@ -86,7 +86,9 @@ def show
       @comment_count = @event.comment_threads.count
       @ratings = @event.votes.includes(:user)
       @difficulty_levels = @program.difficulty_levels
-      @versions = @event.versions | PaperTrail::Version.where(item_type: 'Commercial', item_id: @event.commercials.pluck(:id))
+      @versions = @event.versions |
+       PaperTrail::Version.where(item_type: 'Commercial').where_object(commercialable_id: @event.id, commercialable_type: 'Event') |
+       PaperTrail::Version.where(item_type: 'Commercial').where_object_changes(commercialable_id: @event.id, commercialable_type: 'Event')
```
nishanthvijayan: Same here!

differentreality:
:+1:

PR #1129, opened 2016-08-09
Ichain redirection

### LINKED ISSUE #1054: Register with Ichain: redirect to login page

arun1595:
Fixes #1054

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -52,7 +52,7 @@ class Ability
 
       can [:new, :create], Registration do |registration|
         conference = registration.conference
-        conference.registration_open? && registration.new_record? && !conference.registration_limit_exceeded?
+        can_register?(conference)
       end
 
       can :show, Event do |event|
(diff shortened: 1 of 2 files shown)


app/models/ability.rb
```diff
@@ -45,14 +45,20 @@ def not_signed_in
     # can view Commercials of confirmed Events
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can [:show, :create], User
-    unless ENV['OSEM_ICHAIN_ENABLED'] == 'true'
+    if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
+      can [:new, :create], Registration do |registration|
```
differentreality: That's not accurate. This is authorization rules for a not_signed_in user. 
If ICHAIN is enabled and the user is **not** logged in, then the user **cannot** access new and create actions for Registration.

arun1595: I am missing something here. In the controller method, we're redirecting the user to ichain login page when he tries to create a new registration. So, the user can only  create a new registration after he's signed in right?

Or am I wrong?

differentreality: That's right

differentreality:
Closing due to inactivity. There is also another PR about this now.

PR #1132, opened 2016-08-10
Make the whole event panel clickable in all events

Ana06:
Make the whole event panel clickable in all events in the public schedule. As @nishanthvijayan pointed out in https://github.com/openSUSE/osem/pull/1113 it didn't work properly when having links inside the panel.


differentreality:
Nice work :D

It would be awesome if we could also open it in a new tab (ctrl + click does not currently work, and it used to).

Ana06:
@differentreality 

> It would be awesome if we could also open it in a new tab (ctrl + click does not currently work, and it used to).

Done :wink:

differentreality:
:+1:

PR #1134, opened 2016-08-11
Make links inside event-panel work with ctrl+click

Ana06:
Function introduce to be used in links inside event-panel to make ctrl + click work. It can be used in https://github.com/openSUSE/osem/pull/1113

### DIFF
--- a/app/assets/javascripts/osem-schedule.js
+++ b/app/assets/javascripts/osem-schedule.js
@@ -116,3 +116,28 @@ $(document).ready( function() {
       }
   });
 });
+
+function eventClicked(e, element){
+  var url = $(element).data('url');
+  if(e.ctrlKey)
+    window.open(url,'_blank');
+  else
+    window.location = url;
+}
+
+/* Links inside event-panel (to make ctrl + click work for these links):
+ = link_to text, '#', onClick: 'insideLinkClicked();', 'data-url' => url
+*/
+function insideLinkClicked(event){
+  // stops the click from propagating
+  if (!event) // for IE
+    var event = window.event;
+  event.cancelBubble = true;
+  if (event.stopPropagation) event.stopPropagation();
+
+  var url = $(event.target).data('url');
+  if(event.ctrlKey)
+    window.open(url,'_blank');
+  else
+    window.location = url;
+}
--- a/app/views/schedules/_event.html.haml
+++ b/app/views/schedules/_event.html.haml
@@ -1,4 +1,4 @@
-.panel.panel-default.event-panel{ id: "link-#{event.id}" }
+.panel.panel-default.event-panel{ onClick: 'eventClicked(event, this);', "data-url" => "#{url_for(conference_program_proposal_path(@conference.short_title, event.id))}" }
   .panel-body
     - if speaker = event.speakers.first
       = image_tag speaker.gravatar_url, :class => "img-circle pull-right all-speaker-pic", |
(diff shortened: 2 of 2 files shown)


Ana06:
It is not necessary that `insideLinkClicked()` function is inline javascript. I put it there because it will be easy to find it when adding links inside `event-panel` but I can move it if you think it is not a good idea.

ChrisBr:
Please rebase

app/views/schedules/_event.html.haml
```diff
@@ -33,6 +33,22 @@
 
 :javascript
 
+  /* Links inside event-panel (to make ctrl + click work for these links):
+   = link_to text, '#', onClick: 'insideLinkClicked();', 'data-url' => url
+  */
+  function insideLinkClicked(e){
```
ChrisBr: Where is this function called / bind?

Ana06: It is not called anywhere because there is not any link inside event-panel apart from the `more` link, which is the same link than the event-panel one, so it is not necessary to use it. But @nishanthvijayan is going to add a link inside event-panel in https://github.com/openSUSE/osem/pull/1113.

Then if we merge this he can use the function, or if you prefer to merge https://github.com/openSUSE/osem/pull/1113 first, I'll addapt this PR afterwards to use this function in the new link.

differentreality:
I checked this PR together with #1113 and Ctrl+click on the link inside the event-panel still doesn't work. But click on the link (and right click -> open in new tab) does. 

@ChrisBr have you reviewed this?

Ana06:
@differentreality it works. I have just tried it and if the link looks like that it works:

`= link_to 'text', '#', onClick: 'insideLinkClicked();', 'data-url' => url`

Ana06:
I have moved the function to the javascript file and change the name of the function to avoid merge conflicts with https://github.com/openSUSE/osem/pull/1142

app/views/schedules/_event.html.haml
```diff
@@ -14,6 +14,7 @@
     %p
       = truncate(event.abstract, :length => 400)
       = link_to 'more', conference_program_proposal_path(@conference.short_title, event.id) if event.abstract.length > 400
+      = link_to 'text2', '#', onClick: 'insideLinkClicked();', 'data-url' => 'bla'
```
differentreality: Test line that persisted?

Ana06: :confounded: I'll remove it!

differentreality: Please do :)

differentreality:
Ping

app/views/schedules/_event.html.haml
ChrisBr: ping

app/assets/javascripts/osem-schedule.js
```diff
@@ -116,3 +116,27 @@ $(document).ready( function() {
...
+  if(e.ctrlKey)
+    window.open(url,'_blank');
+  else
+    window.location = url;
+}
+
+/* Links inside event-panel (to make ctrl + click work for these links):
+ = link_to text, '#', onClick: 'insideLinkClicked();', 'data-url' => url
+*/
+function insideLinkClicked(e){
```
ChrisBr: ```s/e/event/a``` please

ChrisBr:
Otherwise LGTM @Ana06

app/assets/javascripts/osem-schedule.js
differentreality: The variable name. 
Just e is not very self explanatory when reading the code

PR #987, opened 2016-05-04
Blind voting

differentreality:

PR #1136, opened 2016-08-11
voting period is open when voting dates are not set

differentreality:
Fixes bug from #987

### DIFF
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -72,16 +72,16 @@ class Program < ActiveRecord::Base
   def show_voting?
     return true unless blind_voting
 
-    Date.today > voting_end_date
+    Time.current > voting_end_date
   end
 
   ##
   # Checks if we are still in voting period
   # ====Returns
-  # * +true+ -> If the voting period is not over yet
+  # * +true+ -> If the voting period is not over yet (or if the voting dates are not set)
   # * +false+ -> If the voting period is over
   def voting_period?
-    return false unless voting_start_date && voting_end_date
+    return true unless voting_start_date && voting_end_date
 
     (voting_start_date.to_datetime..voting_end_date.to_datetime).cover? Time.current
   end
--- a/spec/models/program_spec.rb
+++ b/spec/models/program_spec.rb
@@ -90,8 +90,8 @@ describe Program do
   end
 
   describe 'voting_period?' do
-    it 'retuns false when voting dates are not set' do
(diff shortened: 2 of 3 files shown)


spec/models/program_spec.rb
```diff
@@ -90,7 +90,7 @@
 
   describe 'voting_period?' do
     it 'retuns false when voting dates are not set' do
-      expect(program.voting_period?).to eq false
+      expect(program.voting_period?).to eq true
```
ChrisBr: Huh?

differentreality: Until we did not have voting dates, the voting period was always on (technically even after the conference). I want to maintain that behavior for past conferences.

ChrisBr: Ok, but the test is then not doing what it does say ;)

differentreality: Yeah that was unfortunate :)

differentreality:
Can someone review this?

differentreality:
@ChrisBr @hennevogel can you review this? (Or anyone else?)

spec/models/program_spec.rb
```diff
@@ -90,8 +90,8 @@
   end
 
   describe 'voting_period?' do
-    it 'retuns false when voting dates are not set' do
-      expect(program.voting_period?).to eq false
+    it 'retuns true when voting dates are not set' do
```
ChrisBr: Why is that necessary now?

differentreality: Compatibility with pre-existing conferences that do not have voting dates, and essentially no change in workflow for ppl already used to that logic. That practically means voting is still possible after the conference, which might not make much sense, but that's what happens now in our existing conferences.

ChrisBr: Ah ok!

app/models/program.rb
```diff
@@ -72,16 +72,16 @@ def selected_event_schedules
...
 
   ##
   # Checks if we are still in voting period
   # ====Returns
-  # * +true+ -> If the voting period is not over yet
+  # * +true+ -> If the voting period is not over yet (or if the voting dates are not set)
   # * +false+ -> If the voting period is over
   def voting_period?
-    return false unless voting_start_date && voting_end_date
+    return true unless voting_start_date && voting_end_date
```
ChrisBr: Only thing I'm now concerned is that the function name does clearly not cover what it is doing :(

PR #1137, opened 2016-08-11
Update CONTRIBUTING.md to setup ichain in test mode using dotenv

arun1595:


ChrisBr:
Ah, thanks @arun1595! LGTM

PR #1142, opened 2016-08-15
User schedule

Ana06:
-   Introduce an association between User and Event
-   Make it possible to select events in both the schedule and events
-   Show the individual User schedule 

image

image

There is one problem. If there are less than 7 rooms, as the speaker picture get bigger to use all the space, in devices with less than 450px weight there is not enough space for the star:

image

For less rooms or bigger devices it works fine:
image

**Solutions**
1. Keep the speaker picture size always the same. I don't like that as the speaker picture is really important and it should be as bigger as possible.
2. Put the star at the beginning of the title. The problem with that is that the star would be a little bit smaller and that it takes a little bit of space of the title. On the other hand, it is more consistent with the all events (as there the star is before the title) and I think it looks nice. It would look like this:

image

I prefer the second one. Reducing the number of lines for the title I think it is not an alternative, as the title is more important than the speaker picture and the star.

### DIFF
--- a/app/assets/javascripts/osem-schedule.js
+++ b/app/assets/javascripts/osem-schedule.js
@@ -1,3 +1,5 @@
+// ADMIN SCHEDULE
+
 var url; // Should be initialize in Schedule.initialize
 var schedule_id; // Should be initialize in Schedule.initialize
 
@@ -116,6 +118,34 @@ $(document).ready( function() {
       }
   });
 });
+
+
+// PUBLIC SCHEDULE
+
+function starClicked(e){
+  // stops the click from propagating
+  if (!e) var e = window.event;
+  e.preventDefault();
+  e.cancelBubble = true;
+  if (e.stopPropagation) e.stopPropagation();
+
+  var callback = function(data) {
+    $(e.target).toggleClass('fa-star fa-star-o');
+  }
+
+  var params = { favourite_user_id: $(e.target).data('user') };
+  if($(e.target).hasClass('fa-star-o')){
+    params['add'] = true;
+  }
+
+  $.ajax({
+    url: $(e.target).data('url'),
+    type: 'PATCH',
+    data: params,
+    success: callback,
+    dataType : 'json'
+  });
+}
 
 function eventClicked(e, element){
   var url = $(element).data('url');
--- a/app/assets/javascripts/osem-schedule.js
+++ b/app/assets/javascripts/osem-schedule.js
@@ -134,9 +134,6 @@ function starClicked(e){
   }
 
   var params = { favourite_user_id: $(e.target).data('user') };
-  if($(e.target).hasClass('fa-star-o')){
-    params['add'] = true;
-  }
 
   $.ajax({
     url: $(e.target).data('url'),
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -1,10 +1,11 @@
 class ProposalsController < ApplicationController
+  skip_authorization_check
   before_action :authenticate_user!, except: [:show, :new, :create]
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true
   load_and_authorize_resource :event, parent: false, through: :program
   # We authorize manually in these actions
-  skip_authorize_resource :event, only: [:confirm, :restart, :withdraw]
+  skip_authorize_resource :event, only: [:confirm, :restart, :toogle_favorite, :withdraw]
 
   def index
     @event = @program.events.new
@@ -63,29 +64,26 @@ class ProposalsController < ApplicationController
   end
 
   def update
-    respond_to do |format|
-      format.html do
-        @url = conference_program_proposal_path(@conference.short_title, params[:id])
+    @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
-        if @event.update(event_params)
-          redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
-                      notice: 'Proposal was successfully updated.'
-        else
-          flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
-          render action: 'edit'
-        end
-      end
-      format.json do
-        user = User.find(params[:favourite_user_id])
-        users = @event.favourite_users
-        if users.include? user
-          @event.favourite_users.delete(user)
-        else
-          @event.favourite_users << User.find(params[:favourite_user_id])
-        end
-        render json: {}
-      end
+    if @event.update(event_params)
+      redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
+                  notice: 'Proposal was successfully updated.'
+    else
+      flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
+      render action: 'edit'
     end
+  end
+
+  def toogle_favorite
+    user = User.find(params[:favourite_user_id])
+    users = @event.favourite_users
+    if users.include? user
+      @event.favourite_users.delete(user)
+    else
+      @event.favourite_users << User.find(params[:favourite_user_id])
+    end
+    render json: {}
   end
 
   def withdraw
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -63,14 +63,28 @@ class ProposalsController < ApplicationController
   end
 
   def update
-    @url = conference_program_proposal_path(@conference.short_title, params[:id])
+    respond_to do |format|
+      format.html do
+        @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
-    if @event.update(event_params)
-      redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
-                  notice: 'Proposal was successfully updated.'
-    else
-      flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
-      render action: 'edit'
+        if @event.update(event_params)
+          redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
+                      notice: 'Proposal was successfully updated.'
+        else
+          flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
+          render action: 'edit'
+        end
+      end
+      format.json do
+        user = User.find(params[:favourite_user_id])
+        users = @event.favourite_users
+        if users.include? user
+          @event.favourite_users.delete(user)
+        else
+          @event.favourite_users << User.find(params[:favourite_user_id])
+        end
+        render json: {}
+      end
     end
   end
 
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -1,6 +1,7 @@
 class SchedulesController < ApplicationController
   protect_from_forgery with: :null_session
   before_action :respond_to_options
+  before_action :favourites
   load_and_authorize_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
 
@@ -29,19 +30,26 @@ class SchedulesController < ApplicationController
     @dates = @conference.start_date..@conference.end_date
 
     @events_schedules = @program.selected_event_schedules
+    @events_schedules = @events_schedules.select{ |e| e.event.favourite_users.exists?(current_user.id) } if @events_schedules && current_user && @favourites
     @events_schedules = [] unless @events_schedules
+    @favourites = params[:favourites] == 'true'
 
     @unscheduled_events = if @program.selected_schedule
                             @program.events.confirmed - @program.selected_schedule.events
                           else
                             @program.events.confirmed
                           end
+    @unscheduled_events = @unscheduled_events.select{ |e| e.favourite_users.exists?(current_user.id) } if current_user && @favourites
 
     day = @conference.current_conference_day
     @tag = day.strftime('%Y-%m-%d') if day
   end
 
   private
+
+  def favourites
+    @favourites = params[:favourites] == 'true'
+  end
 
   def respond_to_options
     respond_to do |format|
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -22,8 +22,11 @@ class Event < ActiveRecord::Base
   belongs_to :difficulty_level
   belongs_to :program
 
+  has_and_belongs_to_many :favourite_users, class_name: 'User'
+
   accepts_nested_attributes_for :event_users, allow_destroy: true
   accepts_nested_attributes_for :users
+  accepts_nested_attributes_for :favourite_users
 
   before_create :generate_guid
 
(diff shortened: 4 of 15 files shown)


db/migrate/20160811143427_create_users_events.rb
```diff
@@ -0,0 +1,8 @@
+class CreateUsersEvents < ActiveRecord::Migration
```
ChrisBr: The table is called ``:events_users`` and so should be the migration

db/schema.rb
```diff
@@ -98,6 +98,8 @@
     t.text     "description"
     t.integer  "registration_limit", default: 0
     t.string   "picture"
+    t.integer  "start_hour"
```
ChrisBr: Where does this come from?

db/migrate/20160811143427_create_users_events.rb
```diff
@@ -0,0 +1,8 @@
+class CreateEventsUsers < ActiveRecord::Migration
```
differentreality: The name of the file needs to match the name of this class.

differentreality:
In https://osem-demo-pr-1142.herokuapp.com/conference/osemdemo/schedule?favourites=false
I can no longer click on the event and open the proposal#show page. Is that on purpose?

Ana06:
@differentreality No, that shouldn't happen, I'll look at it. :wink:

Ana06:
The last commit solves what @differentreality pointed out. :wink:

app/views/schedules/_event.html.haml
```diff
@@ -1,9 +1,15 @@
-.panel.panel-default.event-panel{ id: "link-#{event.id}" }
+.panel.panel-default.event-panel{ onClick: 'eventClicked(event, this);', "data-url" => "#{url_for(conference_program_proposal_path(@conference.short_title, event.id))}" }
   .panel-body
     - if speaker = event.speakers.first
       = image_tag speaker.gravatar_url, :class => "img-circle pull-right all-speaker-pic", |
                                         :alt => speaker.name, |
                                         :title => speaker.name |
+    - if  current_user && !@favourites
```
differentreality: I think we should show the star in favorites too, so that it is easy for the attendee to unstar a talk.

differentreality:
I see this error in Console:
`events?favourites=true:284 Uncaught TypeError: Cannot read property 'scrollIntoView' of null(…)`

Referring to line:
`          document.getElementById("").scrollIntoView();`

of code:
```

<script>
        $('.program-selector').on('click', function(e) {
          $('.li-dropdown-program').removeClass('active');
        });
        
        // Go to current date and time
        $(document).ready(function(){
          document.getElementById("").scrollIntoView();
        });
</script>
```

app/views/schedules/_event.html.haml
Ana06: ok!

Ana06:
> I see this error in Console:
> events?favourites=true:284 Uncaught TypeError: Cannot read property 'scrollIntoView' of null(…)

That is because it is not a day of the conference and it doesn't have to go to the current date and time. So it is not important that it fails, but I will add an `if` or something so the error doesn't appear in that case.

differentreality:
Ah OK - yes, it would be best not to show an error in console all together! Thanks :)

differentreality:
After those minor changes, would you also like to rebase + squash your commits to prepare this for merging?

ChrisBr:
@Ana06 please rebase! And should we first merge #1106 before continuing with this one?

Ana06:
@ChrisBr #1106 and this are independent

differentreality:
Do you know what would be nice?
In schedule?favourites=true if we would show only the rows/rooms where the user has favored an event. That way the user does not see empty rows, or has to scroll down to see the favorite event. What do you think? If it is too much to implement now in this PR, we can open an issue for it instead.

differentreality:
Also this needs to be rebased :)

Ana06:
@differentreality 

> In schedule?favourites=true if we would show only the rows/rooms where the user has favored an event. That way the user does not see empty rows, or has to scroll down to see the favorite event. What do you think? If it is too much to implement now in this PR, we can open an issue for it instead.

I am not sure if it would be more confusing that helpful. Because then when clicking favourites the whole screen would change. But we could first merge this and then discussing improvements.

differentreality:
@Ana06 please rebase. @ChrisBr any more comments on this?

app/models/event.rb
```diff
@@ -22,8 +22,11 @@ class Event < ActiveRecord::Base
   belongs_to :difficulty_level
   belongs_to :program
 
+  has_and_belongs_to_many :public_users, class_name: 'User'
```
ChrisBr: Why ``public_users``? The name doesn't really make sense for me?

app/models/user.rb
```diff
@@ -49,6 +49,9 @@ class User < ActiveRecord::Base
   has_many :votes, dependent: :destroy
   has_many :voted_events, through: :votes, source: :events
   has_many :subscriptions, dependent: :destroy
+
+  has_and_belongs_to_many :public_events, class_name: 'Event'
```
ChrisBr: Why public_events? The name doesn't really make sense for me?
Aren't they ``favorite_events``?

app/controllers/proposals_controller.rb
```diff
@@ -63,14 +63,33 @@ def create
...
   def update
-    @url = conference_program_proposal_path(@conference.short_title, params[:id])
-
-    if @event.update(event_params)
-      redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
-                  notice: 'Proposal was successfully updated.'
-    else
-      flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
-      render action: 'edit'
+    respond_to do |format|
```
ChrisBr: I would prefer to have a dedicated action for this and not merging it into update. As it update implies that the owner of the proposal updates his proposal. And I even think that this will cause problems with the permissions as you do ``load_and_authorize :event`` in the before filters. Which means only users with permissions to update the proposal/event can star it, right?

app/assets/javascripts/osem-schedule.js
```diff
@@ -117,6 +119,40 @@ $(document).ready( function() {
...
+
+function starClicked(e){
+  // stops the click from propagating
+  if (!e) var e = window.event;
+  e.preventDefault();
+  e.cancelBubble = true;
+  if (e.stopPropagation) e.stopPropagation();
+
+  var callback = function(data) {
+    if($(e.target).hasClass('fa-star-o')){
```
ChrisBr:
```
$(e.target).toggleClass( "fa-star-o" )
$(e.target).toggleClass( "fa-star" )
```

and you can drop line 133-141

app/assets/javascripts/osem-schedule.js
```diff
@@ -117,6 +119,40 @@ $(document).ready( function() {
...
+      $(e.target).addClass('fa-star');
+    }
+    else{
+    $(e.target).removeClass('fa-star');
+      $(e.target).addClass('fa-star-o');
+    }
+  }
+
+  var params = { public_user_id: $(e.target).data('user') };
+  if($(e.target).hasClass('fa-star-o')){
```
ChrisBr: Why not just also handle this in the action? 
If it is stared then drop the star, if not add the star, e.g. ``toggle_favorite`` or something similar!

app/controllers/schedules_controller.rb
```diff
@@ -30,20 +31,27 @@ def events
     @dates = @conference.start_date..@conference.end_date
 
     @events_schedules = @program.selected_event_schedules
+    @events_schedules = @events_schedules.select{ |e| e.event.public_users.exists?(current_user.id) } if @events_schedules && current_user && @favourites
     @events_schedules = [] unless @events_schedules
+    @favourites = params[:favourites] == 'true'
```
ChrisBr: Not necessary, right?

app/controllers/schedules_controller.rb
```diff
@@ -30,20 +31,27 @@ def events
     @dates = @conference.start_date..@conference.end_date
 
     @events_schedules = @program.selected_event_schedules
+    @events_schedules = @events_schedules.select{ |e| e.event.public_users.exists?(current_user.id) } if @events_schedules && current_user && @favourites
```
ChrisBr: And I would also prefer to have a dedicated view for this. Or is it too much intersection?

Ana06:
@ChrisBr 

> Why public_users? The name doesn't really make sense for me?

What about `favourite_users`? :thinking:

app/assets/javascripts/osem-schedule.js
Ana06: but that is to know if I have to add or delete the user in the controller. But I could also do that without this param, just deleting the user if already exists or adding it otherwise.

app/controllers/schedules_controller.rb
Ana06: why not?

app/controllers/schedules_controller.rb
Ana06: It is exactly the same view, but rendering less events. I don't see the point in splitting it, because we will duplicate code.

differentreality:
I am very eager to get this in :-)

Ana06:
@ChrisBr I did the requested changes, please review it! :wink:

app/assets/javascripts/osem-schedule.js
```diff
@@ -117,6 +119,31 @@ $(document).ready( function() {
...
+    $(e.target).toggleClass('fa-star fa-star-o');
+  }
+
+  var params = { favourite_user_id: $(e.target).data('user') };
+
+  $.ajax({
+    url: $(e.target).data('url'),
+    type: 'PATCH',
+    data: params,
+    success: callback,
```
ChrisBr: error?

app/assets/javascripts/osem-schedule.js
```diff
@@ -1,3 +1,5 @@
+// ADMIN SCHEDULE
```
ChrisBr: Wondering if we should move this in a dedicated file / subdirectory then?

app/controllers/proposals_controller.rb
```diff
@@ -1,10 +1,11 @@
 class ProposalsController < ApplicationController
+  skip_authorization_check
```
ChrisBr: Why?

app/controllers/proposals_controller.rb
```diff
@@ -74,6 +75,17 @@ def update
     end
   end
 
+  def toogle_favorite
+    user = User.find(params[:favourite_user_id])
+    users = @event.favourite_users
+    if users.include? user
+      @event.favourite_users.delete(user)
+    else
+      @event.favourite_users << User.find(params[:favourite_user_id])
```
ChrisBr: Why finding the user twice and not using the user from line 79?

app/controllers/proposals_controller.rb
```diff
@@ -74,6 +75,17 @@ def update
     end
   end
 
+  def toogle_favorite
+    user = User.find(params[:favourite_user_id])
+    users = @event.favourite_users
+    if users.include? user
```
ChrisBr: Parenthesis?

app/controllers/proposals_controller.rb
```diff
@@ -74,6 +75,17 @@ def update
...
+      @event.favourite_users.delete(user)
+    else
+      @event.favourite_users << User.find(params[:favourite_user_id])
+    end
+    render json: {}
```
ChrisBr: What happens if something goes wrong?

app/views/schedules/_event.html.haml
```diff
@@ -4,11 +4,15 @@
       = image_tag speaker.gravatar_url, :class => "img-circle pull-right all-speaker-pic", |
                                         :alt => speaker.name, |
                                         :title => speaker.name |
-
     %p
       = canceled_replacement_event_label(event, event_schedule)
       = replacement_event_notice(event_schedule)
-
+    - if current_user
+      = link_to('#', onClick: 'starClicked();') do
```
ChrisBr: Why not using a jQuery clickhandler and binding it to an id instead of inline js?

app/views/schedules/_schedule_item.html.haml
```diff
@@ -1,12 +1,18 @@
 %td.event{ width: "#{ width * span }%" , |
            colspan: span, |
            role: "button" }
-  %a.unstyled-link{href: url_for(conference_program_proposal_path(@conference.short_title, event.id))}
+  %div{ onClick: 'eventClicked(event, this);', "data-url" => "#{url_for(conference_program_proposal_path(@conference.short_title, event.id))}" }
```
ChrisBr: Same here

app/views/schedules/_schedule_item.html.haml
```diff
@@ -1,12 +1,18 @@
...
     %div{ class: "elipsis break-words event-title", |
           style: "-webkit-line-clamp: #{ event_lines(@rooms) }; height: #{ event_height(@rooms) }px;"} |
 
       = canceled_replacement_event_label(event, event_schedule, 'schedule-label')
 
+      - if current_user
+        = link_to('#', onClick: 'starClicked();') do
```
ChrisBr: Same here

app/controllers/proposals_controller.rb
differentreality: We should not have that, because we do want other actions to get authorized. You need to remove this line and add an ability for the new action `:toggle_favorite`

Once you have that in ability, you can either let load_and_authorize_resource work its magic, or you manually authorize your action.

app/controllers/proposals_controller.rb
differentreality: `if users.include?(user)` ? Because that's not needed, do we prefer it? I don't.

Ana06: According to Ruby style guide it seems that parenthesis are preferred: https://github.com/bbatsov/ruby-style-guide#method-invocation-parens

app/controllers/proposals_controller.rb
Ana06: what could go wrong?

`delete` and `<<` can not fail

we could check that the record was updated or deleted and then render an error.

app/assets/javascripts/osem-schedule.js
Ana06: there is no error currently, if this fails the start is not updated.

differentreality:
@Ana06 @ChrisBr do you think we can finalize this in the next couple of weeks?

differentreality:
@Ana06 please rebase

app/controllers/proposals_controller.rb
differentreality: @Ana06 I think we should remove this, isn't  line 8 enough?

differentreality:
@Ana06 would you like to rebase this and finally get it merged? :D

Ana06:
@differentreality sorry, I do not find time for everything. 🙈 I'll try to do it in the following days 💪

differentreality:
How about I rebase this myself and open a new PR? @Ana06

hennevogel:
@differentreality please do so if you can find the time

PR #1143, opened 2016-08-15
Make the schedule times changeable

Ana06:


Ana06:
What happens if an event is already scheduled at 9 and then the conference start hour is changed to 10? I suppose I should delete all EventSchedules that are not in the hours range when updating them.

differentreality:
Isn't the schedule itself that defines when a conference starts?

If there is sth scheduled at 9am, then the conference starts at 9am.

Ana06:
@differentreality No. Currently the public schedules goes from 9:00 to 20:00. We could do what you say in the public schedule. However we need to set start and end hours for the admin schedule, as currently you can not schedule an event at 7:00.

image

That the public schedule itself defines when a conference starts has some things I don't like:
- The hours range in the public schedule will change when an event is scheduled earlier, so maybe it is a little bit confusing for users.
- It is more complicated to implement and it makes almost no difference.

differentreality:
OK so we will enable the organizers to schedule events at any (random) time they want (eg. earlier than 9am). Alright, that would be interesting :)

differentreality:
I am wondering if it makes sense to add times for the Schedule model itself instead of putting more info on Conference.

As far as deleting the earlier talks is concerned, while it makes some sense, what an organizer would probably want to do is move everything 1 hour later instead.

Ana06:
But then we could have schedules with different times, do we want that?

differentreality:
Why not? I mean since we are allowing organizers to make suggestions about how the schedule should look like, that can include options about the starting time of the keynotes. In the end, the selected_schedule will reflect what is decided. 
What do others think?

.rubocop.yml
```diff
@@ -95,7 +95,7 @@ Metrics/BlockNesting:
   Max: 4
 
 Metrics/ClassLength:
-  Max: 575
+  Max: 585
```
differentreality: Let's disable the cop in conference model all together, it seems to be a regular issue.
https://github.com/openSUSE/osem/pull/1118/files#diff-a8447a840df703618503bdc301e044e9R1
https://github.com/openSUSE/osem/pull/1118/files#diff-a8447a840df703618503bdc301e044e9R1067

differentreality:
Can you please rebase?

differentreality:
This seems to be working :-) Do we want to provision that the hours will not change after a schedule has been selected?

Ana06:
@differentreality 

> Do we want to provision that the hours will not change after a schedule has been selected?

But you could want to make a change after a schedule has been selected right?

differentreality:
Yeah, let's leave this at that. Looks good to be merged to me. @ChrisBr ?

PR #1148, opened 2016-08-15
Validation in EventSchedule

Ana06:
As @ChrisBr pointed out in https://github.com/openSUSE/osem/pull/1106#issuecomment-239422358 we need this validation to avoid scheduling the same event twice for the same schedule.

### DIFF
--- a/db/migrate/20160815140302_add_index_to_event_schedule.rb
+++ b/db/migrate/20160815140302_add_index_to_event_schedule.rb
@@ -1,0 +1,5 @@
+class AddIndexToEventSchedule < ActiveRecord::Migration
+  def change
+    add_index :event_schedules, [:event_id, :schedule_id], unique: true
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20160815140302_add_index_to_event_schedule.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToEventSchedule < ActiveRecord::Migration
+  def change
+    add_index :event_schedules, [:event_id, :schedule_id], unique: true
```
ChrisBr: Why do we need the index?

ChrisBr: Rails guide advises so to prevent errors on database level:

http://guides.rubyonrails.org/active_record_validations.html#uniqueness

ChrisBr:
LGTM

PR #1149, opened 2016-08-15
Bug fix nil parameter ApplicationHelper.markdown

### LINKED ISSUE #1145: Provision for nil value in markdown(text)
https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb#L263 
This line breaks when text is nil

charliequinn:
Fix #1145 with specs


ChrisBr:
@charliequinn thanks for your contribution! LGTM :+1:

PR #1150, opened 2016-08-16
Added specs for ApplicationHelper.show_time

charliequinn:
refactor of code to use divmod and handle decimal number parameter

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -165,17 +165,16 @@ module ApplicationHelper
   end
 
   def show_time(length)
-    h = length / 60
-    min = length - h * 60
+    return '0 h 0 min' if length.blank?
 
-    if h != 0
-      if min != 0
-      "#{h} h #{min} min"
-      else
-        "#{h} h"
-      end
+    h, min = length.divmod(60)
+
+    if h == 0
+      "#{min.round} min"
+    elsif min == 0
+      "#{h} h"
     else
-      "#{min} min"
+      "#{h} h #{min.round} min"
     end
   end
 
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -15,6 +15,28 @@ describe ApplicationHelper, type: :helper do
     end
   end
 
+  describe 'show_time' do
+    it 'when length > 60' do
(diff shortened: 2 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -165,17 +165,14 @@ def active_nav_li(link)
   end
 
   def show_time(length)
-    h = length / 60
-    min = length - h * 60
+    h, min = length.divmod(60)
```
differentreality: Shall we also ensure that length ain't nil/blank? 
`return unless length.present?`

ChrisBr: +1

``
nil.divmod(60)
``

will throw an undefined method  exception!

(I know that it did it before, however, I think it would be nice to fix it).

ChrisBr:
Otherwise it LGTM :+1: 

:clap: :clap: :clap: Thanks @charliequinn

app/helpers/application_helper.rb
ChrisBr: Maybe we should discuss what return value we expect if length is nil?

nil, "", "0 h 0m" ?

@differentreality

differentreality: That's a very good question. I was initially thinking we could just throw a nil there, but in the unlikely event that things miserably break, if we show 0h 0m it would make it visually easier to find the failure.
Right now we use this function for the event_type.length in the dropdown of proposal#new

So 0h 0m ?

charliequinn: I've added '0 h 0 min' return value for when length is nil

app/helpers/application_helper.rb
```diff
@@ -165,17 +165,16 @@ def active_nav_li(link)
   end
 
   def show_time(length)
-    h = length / 60
-    min = length - h * 60
+    return '0 h 0 min' if length.nil?
```
ChrisBr: What about ``blank?``

http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F

charliequinn: updated to handle blank

ChrisBr:
Awesome @charliequinn 
Thanks :+1:

ChrisBr:
(I will merge as soon CI reports green ...)

PR #1151, opened 2016-08-16
added minimum length to abstract in event model

redasus:
#1092


differentreality:
That won't work. Please take a closer look at the code. Minimum abstract length is based on the type of the proposal, is not always the same.

PR #1159, opened 2016-08-17
Misc improvements to Revision History

nishanthvijayan:
-  Fixes few minor inconsistent linking problems
- When displaying which attributes changed, use names of associated objects instead of just ID to make is more human readable.
  Eg: 
  Previously: 
  prev
  Now:
  now
- Tracks model Schedule and EventSchedule

schedule

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -413,12 +413,6 @@ module ApplicationHelper
       when 'new' then 'resubmitted'
       end
 
-    when version.changeset['start_time'] && version.changeset['start_time'][0].nil?
-      'scheduled'
-
-    when version.changeset['start_time'] && version.changeset['start_time'][1].nil?
-      'unscheduled'
-
     else
       "updated #{updated_attributes(version)} of"
     end
@@ -491,6 +485,14 @@ module ApplicationHelper
       else
         link_to_user(version.whodunnit) + " updated #{updated_attributes(version)} of"
       end
+    end
+  end
+
+  def event_schedule_change_description(version)
+    case version.event
+    when 'create' then 'scheduled'
+    when 'update' then 'rescheduled'
+    when 'destroy' then 'unscheduled'
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -397,6 +397,7 @@ module ApplicationHelper
   # Returns object as it was before version's change(unless its a create event's version)
   # Else Returns object as it was after version's change
   def get_version_object(version)
+    return nil unless version
     version.item || version.reify || version.next.reify
   end
 
@@ -501,5 +502,11 @@ module ApplicationHelper
     else
       'deleted'
     end
+  end
+
+  def object_last_description(model_name, id)
+    object_version = PaperTrail::Version.where(item_type: model_name, item_id: id).last
+    object = (object_version.reify || object_version.item) if object_version
+    object.try(:name) || object.try(:title)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -397,6 +397,7 @@ module ApplicationHelper
   # Returns object in its current state if its alive
   # Otherwise Returns object state just before deletion
   def current_or_last_object_state(model_name, id)
+    return nil unless id.present? && model_name.present?
     begin
       object = model_name.constantize.find_by(id: id)
     rescue NameError
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -392,13 +392,22 @@ module ApplicationHelper
     end
   end
 
-  # Recieves a PaperTrail::Version object
+  # Recieves a model_name and id
+  # Returns nil if model_name is invalid
   # Returns object in its current state if its alive
-  # Returns object as it was before version's change(unless its a create event's version)
-  # Else Returns object as it was after version's change
-  def get_version_object(version)
-    return nil unless version
-    version.item || version.reify || version.next.reify
+  # Otherwise Returns object state just before deletion
+  def current_or_last_object_state(model_name, id)
+    begin
+      object = model_name.constantize.find_by(id: id)
+    rescue NameError
+      return nil
+    end
+
+    if object.nil?
+      object_last_version = PaperTrail::Version.where(item_type: model_name, item_id: id).last
+      object = object_last_version.reify if object_last_version
+    end
+    object
   end
 
   def event_change_description(version)
@@ -423,18 +432,17 @@ module ApplicationHelper
   end
 
   def subscription_change_description(version)
-    user = get_version_object(version).user
+    user = current_or_last_object_state(version.item_type, version.item_id).user
     user_name = user.name unless user.id.to_s == version.whodunnit
     version.event == 'create' ? "subscribed #{user_name} to" : "unsubscribed #{user_name} from"
   end
 
   def registration_change_description(version)
     if version.item_type == 'Registration'
-      user = get_version_object(version).user
-    else
-      registration_id = get_version_object(version).registration_id
-      registration_last_version = PaperTrail::Version.where(item_type: 'Registration', item_id: registration_id).last
-      user = get_version_object(registration_last_version).user
+      user = current_or_last_object_state(version.item_type, version.item_id).user
+    elsif version.item_type == 'EventsRegistration'
+      registration_id = current_or_last_object_state(version.item_type, version.item_id).registration_id
+      user = current_or_last_object_state('Registration', registration_id).user
     end
 
     if user.id.to_s == version.whodunnit
@@ -453,7 +461,7 @@ module ApplicationHelper
   end
 
   def comment_change_description(version)
-    user = get_version_object(version).user
+    user = current_or_last_object_state(version.item_type, version.item_id).user
     if version.event == 'create'
       version.previous.nil? ? 'commented on' : "re-added #{user.name}'s comment on"
     else
@@ -462,7 +470,7 @@ module ApplicationHelper
   end
 
   def vote_change_description(version)
-    user = get_version_object(version).user
+    user = current_or_last_object_state(version.item_type, version.item_id).user
     if version.event == 'create'
       version.previous.nil? ? 'voted on' : "re-added #{user.name}'s vote on"
     elsif version.event == 'update'
@@ -506,9 +514,7 @@ module ApplicationHelper
     end
   end
 
-  def object_last_description(model_name, id)
-    object_version = PaperTrail::Version.where(item_type: model_name, item_id: id).last
-    object = (object_version.reify || object_version.item) if object_version
-    object.try(:name) || object.try(:title)
+  def link_if_alive(version, link_text, link_url)
+    version.item ? link_to(link_text, link_url) : link_text
   end
 end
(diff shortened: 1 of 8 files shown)


app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,162 @@
...
     = link_to 'commercial',
             admin_conference_commercials_path(conference_id: Conference.find(version.conference_id).short_title)
 
-- elsif %w(EventsRegistration Comment Vote).include?(version.item_type)
+- when 'EventsRegistration', 'Comment', 'Vote', 'Event'
   = 'event'
-  - event_id = get_version_object(version).try(:event_id) || get_version_object(version).try(:commentable_id)
+  - object = current_or_last_object_state(version)
+  - event_id = object.try(:id) || object.try(:event_id) || object.try(:commentable_id)
   = link_to Event.find(event_id).title,
```
differentreality: We should make sure the event exists before trying to show its title.

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,162 @@
...
+          admin_conference_splashpage_path(conference_id: Conference.find(version.conference_id).short_title)
+
+- when 'EmailSettings'
+  = link_if_alive version, 'email settings',
+          admin_conference_emails_path(conference_id: Conference.find(version.conference_id).short_title)
+
+- when 'User'
+  - if version.event == 'update'
+    = 'user'
+    = link_to User.find(version.item_id).name, admin_user_path(id: version.item_id)
```
differentreality: We need to make sure the user exists before trying to display the name.

app/views/shared/_object_changes.html.haml
```diff
@@ -12,17 +12,37 @@
...
-            %td= values[1].blank? ?  '-' : values[1]
+            - if attribute.include?('_id')
+              - model_name = attribute.chomp('_id').camelize
+              %td= attribute.chomp('_id')
+              - if version.event != 'create'
+                %td
+                  = object_last_description(model_name, values[0])
+                  = "(ID: #{values[0].blank? ?  '-' : values[0]})"
+              %td
+                = object_last_description(model_name, values[1])
```
differentreality: We update track_id of Event. There is a version record about it (item_type being 'Event').
We want to show the name of the new track, why do we need to have a record in table versions for that to happen?

app/views/shared/_object_changes.html.haml
```diff
@@ -12,17 +12,37 @@
...
+                = "(ID: #{values[1].blank? ?  '-' : values[1]})"
```
differentreality: If `blank?` is true, this will produce sth resembling a smiley face which is more confusing than helpful, in my opinion. Just the dash we had before was pretty self-explanatory, if you ask me.

app/views/admin/versions/_object_desc_and_link.html.haml
nishanthvijayan: Is it necessary? I assumed it's okay since there is no interface to delete an event.

differentreality: Yes it is, because someone might have done it through the console. openSUSE has such cases.

app/views/admin/versions/_object_desc_and_link.html.haml
nishanthvijayan: I've changed this, now it first check if the user exists, if not it will get the name from its last version.

app/views/admin/versions/_object_desc_and_link.html.haml
nishanthvijayan: I've changed this, now it first check if the event exists, if not it will get the name from its last version.

app/views/shared/_object_changes.html.haml
nishanthvijayan: Same here.I've changed this to check if the associated object (track in this case) exists first and checks in versions only otherwise.

app/views/shared/_object_changes.html.haml
nishanthvijayan: Done

differentreality:
Sorry perhaps I didn't make the comment clear. I think we should show the ID, when one exists. We should just not show the whole text `(ID: -)` when an ID does not really exist, and just show `-` like we did before.

nishanthvijayan:
Okay.I've fixed it
Example:
associated_id_change

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,161 @@
...
             admin_conference_commercials_path(conference_id: Conference.find(version.conference_id).short_title)
 
-- elsif %w(EventsRegistration Comment Vote).include?(version.item_type)
+- when 'EventsRegistration', 'Comment', 'Vote', 'Event'
   = 'event'
-  - event_id = get_version_object(version).try(:event_id) || get_version_object(version).try(:commentable_id)
-  = link_to Event.find(event_id).title,
+  - object = current_or_last_object_state(version.item_type, version.item_id)
+  - event_id = object.try(:id) || object.try(:event_id) || object.try(:commentable_id)
+  = link_to current_or_last_object_state('Event', event_id).title,
```
differentreality: What if there is no event record?

app/helpers/application_helper.rb
```diff
@@ -428,18 +433,17 @@ def users_role_change_description(version)
...
+    user = current_or_last_object_state(version.item_type, version.item_id).user
     user_name = user.name unless user.id.to_s == version.whodunnit
     version.event == 'create' ? "subscribed #{user_name} to" : "unsubscribed #{user_name} from"
   end
 
   def registration_change_description(version)
     if version.item_type == 'Registration'
-      user = get_version_object(version).user
+      user = current_or_last_object_state(version.item_type, version.item_id).user
     else
```
differentreality: It would be useful to have a comment on this line indicating which is the else scenario. I think right now it is only the EventsRegistration.

app/views/shared/_object_changes.html.haml
```diff
@@ -12,17 +12,40 @@
...
+      - if version.event != 'destroy'
         - version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.each do |attribute, values|
           %tr
-            %td= attribute
-            - if version.event != 'create'
-              %td= values[0].blank? ?  '-' : values[0]
-            %td= values[1].blank? ?  '-' : values[1]
+            - if attribute.include?('_id')
+              - model_name = attribute.chomp('_id').camelize
+              %td= attribute.chomp('_id')
```
differentreality: I'd rather we don't chomp it and show the actual attribute.

app/views/admin/versions/_object_desc_and_link.html.haml
nishanthvijayan: That seems unlikely considering how Event was already being tracked

differentreality: Well we either provision for that here, or we need a task deleting versions of unused Events. I'd rather we provision for that here for the time being, and we can also create a relevant task later (another PR).

nishanthvijayan: I didn't understand your comment.

To make it clear, the reason why its assumed here that there will be event records is because the Event model was already being tracked, even before #1019.They would be fetched properly by current_or_last_object_state even if the rake task for adding conference_id has not been run as `current_or_last_object_state` uses model_name and id

nishanthvijayan:
Anything left to be done here?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,163 @@
...
 
-  - case commercial_last_version.commercialable_type
+  - case commercial.commercialable_type
   - when 'Event'
-    =  link_to 'commercial',
-            edit_admin_conference_program_event_path(conference_id: Conference.find(version.conference_id).short_title,
-                id: commercialable_last_version.id, anchor: 'commercials-content')
-    = "in event #{commercialable_last_version.title}"
+    commercial in event
+    =  link_to (commercialable.try(:title) || 'deleted event'),
```
differentreality: It doesn't make sense to link to an event that does not exist and is going to raise an error anyway :/

If there is an event, we can link to it, just like you do.
If there is no event, we should show just the text (and possible the ID of the event: `Deleted event with ID 123`)

app/views/shared/_object_changes.html.haml
```diff
@@ -12,17 +12,40 @@
...
     %tbody
-      - if version.event!= 'destroy'
+      - if version.event != 'destroy'
         - version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.each do |attribute, values|
           %tr
             %td= attribute
-            - if version.event != 'create'
-              %td= values[0].blank? ?  '-' : values[0]
-            %td= values[1].blank? ?  '-' : values[1]
+
```
differentreality: Since we have another change to make anyway, let's add a comment for the following if. So that people can understand all the magic happening below. Sth like:

> If the attribute is an associated model, show the value of the record it corresponds to, not just the ID. 
> Eg. when the version is of an Event, if the attribute is event_type_id it shows
> Workshop (ID: 116) instead of just 116

differentreality:
I've made a couple of comments, other than that this is good to go. Please also rebase + squash commits after the changes!

differentreality:
@nishanthvijayan can you complete this?

differentreality:
I will merge this and send a new PR for the last 2 comments.

differentreality:
:+1: that's awesome @nishanthvijayan, it's great to see this being finalized :-)
#1113 depended on this one, I guess now we can proceed with it as well?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,171 @@
...
-    = link_to 'contact details',
-            edit_admin_conference_contact_path(conference_id: Conference.find(version.conference_id).short_title)
+  - target = current_or_last_object_state(version.item_type, version.item_id)
+  = link_if_alive version, target.to_s, admin_conference_targets_path(conference_id: Conference.find(version.conference_id).short_title)
+
+- when 'EventSchedule'
+  - event_schedule = current_or_last_object_state(version.item_type, version.item_id)
+  event
+  = link_to (current_or_last_object_state('Event', event_schedule.event_id).try(:title) || 'deleted event'),
+           admin_conference_program_event_path(conference_id: Conference.find(version.conference_id).short_title, id: event_schedule.event_id)
```
differentreality: We still seem to have some wrong links being created here for Event but I really want to keep going with the rest of the PRs, so I am putting this on a note here to fix in a following PR.

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,156 +1,171 @@
...
+  - users_role = current_or_last_object_state(version.item_type, version.item_id)
+  - user = current_or_last_object_state('User', users_role.user_id)
   = 'role'
   = link_to users_role.role.name, admin_conference_role_path(conference_id: Conference.find(version.conference_id).short_title, id: users_role.role.name)
   = version.event == 'create' ? 'to' : 'from'
   = 'user'
-  = link_to users_role.user.name, admin_user_path(id: users_role.user.id)
 
-- elsif version.item_type == 'Subscription' || version.item_type == 'Registration'
+  = link_to (user.try(:name) || 'deleted user'), admin_user_path(id: users_role.user.id)
```
differentreality: make this link_to_user(user_id) too?

PR #1160, opened 2016-08-18
Show all events instead of the schedule

Ana06:
Show all events instead od the schedule if there are less than 2 rooms. The schedule looks ugly with only one room:

image

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -7,7 +7,7 @@ class SchedulesController < ApplicationController
   def show
     @rooms = @conference.venue.rooms if @conference.venue
     schedules = @program.selected_event_schedules
-    unless schedules
+    unless schedules && @rooms && @rooms.length > 1
       redirect_to events_conference_schedule_path(@conference.short_title)
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/schedules_controller.rb
```diff
@@ -29,6 +29,9 @@ def show
   def events
     @dates = @conference.start_date..@conference.end_date
 
+    rooms = @conference.venue.try(:rooms)
+    @several_rooms = rooms && rooms.length > 1
```
ChrisBr: The temp variable does not add any value!
```
@several_rooms = @conference.venue.try(:rooms) && @conference.venue.try(:rooms) > 1
```

ChrisBr:
Otherwise LGTM

differentreality:
I don't agree with this change.
If people click on schedule, they expect to see a schedule (as long as it exists, eg. when we are in the middle of the conference).
Moreover we are changing the layout (hiding the schedule tab) and that's just confusing for people.

I think we should let people see the schedule as it is. If you think there is something we can do to beautify it, cool, let's do that. Otherwise people will get to see something uglier. They can always manually change to the list of events if they want to.

Ana06:
I have been thinking about it and @differentreality is right, it would be confusing. I'll close this PR.

PR #1161, opened 2016-08-18
Delete EventSchedules when a Room is deleted

Ana06:
We should remove all the EventSchedules that belongs_to a Room when it is deleted. Otherwise both the public and admin schedule would fail.


ChrisBr:
LGTM :+1:

ChrisBr:
@Ana06 please fix the failing test!

PR #1162, opened 2016-08-18
Make tests pass on MySQL

lagartoflojo:
Tests were not working when you ran them on MySQL.

- Some failures were due to a bad configuration of DatabaseCleaner. Plus, there was an instance of DatabaseCleaner being called directly from a test 😱 
- Other failures were due to different handling of dates between SQLite and MySQL.
- Other (most) failures were due to hard-coding id's on the tests, which I fixed by asking the DB for the proper id.
- Finally, maybe the most "controversial" change, and related to the previous point, was in the ConferenceSerializer spec. There, we were comparing the `to_json` output of the serializer to an expected JSON. That expected JSON also contained hard-coded ids, which made the test fail. I removed that expected JSON and introduced the [json-schema gem](https://github.com/ruby-json-schema/json-schema), which allows us to test the *structure* of the JSON, rather than the specific contents. Personally I think the test looks much nicer now. If you guys like it, I can also modify the other Serializer specs to use it :)

### DIFF
--- a/spec/features/campaign_spec.rb
+++ b/spec/features/campaign_spec.rb
@@ -17,8 +17,7 @@ feature Campaign do
 
       click_button 'Create Campaign'
 
-      expect(flash).
-	  to eq("Campaign creation failed. Name can't be blank and Utm campaign can't be blank")
+      expect(flash).to eq("Campaign creation failed. Name can't be blank and Utm campaign can't be blank")
 
       fill_in 'campaign_name', with: 'Test Campaign'
       fill_in 'campaign_utm_campaign', with: 'campaign'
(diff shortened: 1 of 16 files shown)


spec/features/campaign_spec.rb
```diff
@@ -30,13 +29,12 @@
...
       # Validations
-      expect(flash).
-          to eq('Campaign successfully created.')
+      expect(flash).to eq('Campaign successfully created.')
 
-      expect(find('#name_1').text).to eq('Test Campaign')
-      expect(find('#visits_1').text).to eq('0')
-      expect(find('#registrations_1').text).to eq('0')
-      expect(find('#submissions_1').text).to eq('0')
+      expect(page).to have_selector('.campaign-name', text: 'Test Campaign')
```
differentreality: Can't we do this using the ID instead of introducing a class we don't need anywhere else except for the test?

lagartoflojo: Yeah, I didn't like that very much either.
I pushed a commit so that it now uses the CSS "starts with" matcher. How about it?

hennevogel:
Awesome :hankey: @lagartoflojo!

lagartoflojo:
I changed the Travis config, so that the tests run on MariaDB there. I introduced a `database.yml.travis` file with the necessary database config.

In general, I would prefer that we run the tests in an environment as close as production as possible, so I think this is a good change. The only issue is that, if devs are running the tests on SQLite locally, they might pass there, but fail on Travis.

What do you think?

lagartoflojo:
Another question: Should I port all serializer tests to use the `json-schema` gem? I think it makes sense to test all serializers in the same way.

hennevogel:
:+1: to mysql in travis. porting to `json-schema` in another PR please :-)

Issue #1140, opened 2016-08-11
Unreachable code in ConferenceRegistrationsController#new


charliequinn:
Whilst writing specs in pull request #1138 I noticed that there are a couple of instances of unreachable code in ConferenceRegistrationsController new method.

**First instance** (check if current user and is registered) - https://github.com/openSUSE/osem/blob/master/app/controllers/conference_registrations_controller.rb#L22-L24

cannot be reached due to line 12 @conference.user_registered?(current_user)

**Second instance** (if OSEM_ICHAIN_ENABLED is true and user is guest) - https://github.com/openSUSE/osem/blob/master/app/controllers/conference_registrations_controller.rb#L16-L18

cannot be reached because of this guy https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L48 meaning that guest users won't have authorization to register if OSEM_ICHAIN_ENABLED = 'true' 

However, this causes a rather ambiguous error message to be displayed to the user ("Sorry, you can not register for #{conference.title}. Registration limit exceeded or the registration is not open. Maybe you need to sign in?"). 

I believe it makes sense to remove this from the ability model (provided it doesn't affect anything else) and keep it on the controller so that we can give the user a much more precise reason ("You must be signed in") as to why the user cannot register.

What do you think? Would greatly appreciate any feedback on the above.

ChrisBr:
I would confirm first instance, @differentreality can you also please have a look!

ChrisBr:
Second instance: I would prefer to have it in the ability.rb, however, you're right that the error message is misleading :( 
@differentreality what do you think?

differentreality:
The issue was also discussed in https://github.com/openSUSE/osem/issues/1054 but there hasn't been a proper solution.
I can confirm both instances mentioned. They definitely should be fixed.

ability.rb needs to represent all authorization rules we have. One of them is that a user which is not signed_in cannot access the new action of Registration. 
It is the controller that needs a bit of refactoring, removing what's not necessary, and putting things in the right order so that it makes sense to visitors.

charliequinn:
Yeah in hindsight I agree that keeping the authorization rules together (as opposed to dotted about controllers) is best practice.

I've written up a proposed set of 8 expected outcomes based on the 3 input parameters we currently have:

1. User signed in
1. OSEM_ICHAIN_ENABLED
1. registration period open && registration.new_record? && registration limit not exceeded

[...]

PR #1165, opened 2016-08-19
Update ConferenceRegistrationsController#new

charliequinn:
ref #1140 
- authenticate on all controller actions unless (OSEM_ICHAIN_ENABLED != true and
  action is 'new' or 'create')
- before_filter rename to before_action
- addition of before_action :user_already_registered on :new action (redirects
  to edit if user already registered)
- increased specs

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -1,27 +1,13 @@
 class ConferenceRegistrationsController < ApplicationController
-  before_filter :authenticate_user!, except: [:new, :create]
+  before_action :authenticate_user!, unless: :allow_guests?
   load_resource :conference, find_by: :short_title
+  before_action :user_already_registered, only: :new
   authorize_resource :conference_registrations, class: Registration, except: [:new, :create]
   before_action :set_registration, only: [:edit, :update, :destroy, :show]
 
   def new
     @registration = Registration.new(conference_id: @conference.id)
     authorize! :new, @registration, message: "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
-
-    # Redirect to registration edit when user is already registered
-    if @conference.user_registered?(current_user)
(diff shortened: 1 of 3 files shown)


app/controllers/conference_registrations_controller.rb
```diff
@@ -1,28 +1,14 @@
 class ConferenceRegistrationsController < ApplicationController
-  before_filter :authenticate_user!, except: [:new, :create]
+  before_action :authenticate_user!, unless: :allow_guests?
   load_resource :conference, find_by: :short_title
+  before_action :user_already_registered, only: :new
```
differentreality: Since this is only used in new action, I would just leave it in the action, instead of creating a new function.

app/controllers/conference_registrations_controller.rb
```diff
@@ -96,6 +82,18 @@ def destroy
 
   protected
 
+  def allow_guests?
+    ENV['OSEM_ICHAIN_ENABLED'] != 'true' && params[:action].in?(['new', 'create'])
```
differentreality: So for destroy action we will not authenticate the user?

app/models/ability.rb
```diff
@@ -45,16 +45,17 @@ def not_signed_in
     # can view Commercials of confirmed Events
     can :show, Commercial, commercialable_type: 'Event', commercialable_id: Event.where(state: 'confirmed').pluck(:id)
     can [:show, :create], User
+
+    can [:new, :create], Registration do |registration|
```
differentreality: That implies that a not_signed_in user can create a Registration even if ichain is enabled. That's not a valid ability.

differentreality:
@charliequinn Hey, do you need more help proceeding with this PR?

differentreality:
@charliequinn hey, what's the progress of this PR?

differentreality:
I will close this due to inactivity.

PR #1166, opened 2016-08-20
[#1092] ensure empty abstract

anandjose-arackal:
i suggest this fix
keep server side custom validation for the abstract. Remove the code for client side validation

### DIFF
--- a/app/views/proposal/new.html.haml
+++ b/app/views/proposal/new.html.haml
@@ -43,7 +43,7 @@
               :javascript
                 $("##{@program.event_types.first.id}-help").collapse('show');
 
-              = f.input :abstract, required: true, input_html: { rows: 5, data: { provide: 'markdown-editable' } },
+              = f.input :abstract, input_html: { rows: 5, data: { provide: 'markdown-editable' } },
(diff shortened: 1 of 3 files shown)


differentreality:
Please rebase

app/views/proposal/new.html.haml
```diff
@@ -43,7 +43,7 @@
               :javascript
                 $("##{@program.event_types.first.id}-help").collapse('show');
 
-              = f.input :abstract, required: true, input_html: { rows: 5, data: { provide: 'markdown-editable' } },
+              = f.input :abstract, input_html: { rows: 5, data: { provide: 'markdown-editable' } },
```
ChrisBr: Why do we want to remove client side validation @anandjose-arackal

ChrisBr:
and yes, please rebase @anandjose-arackal

app/views/proposal/new.html.haml
differentreality: @ChrisBr do you think we should keep it with a condition? (If min_abstract_length == 0 not required, otherwise required)

differentreality:
I will close this due to inactivity. 
@ChrisBr I am still interested in knowing if you think we should keep the client side validation.

PR #1167, opened 2016-08-22
Fix bug in the schedule

Ana06:
The events which start time was the one of the last cell of the table were not rendered.

image


ChrisBr:
LGTM

PR #1168, opened 2016-08-23
Overlapping

Ana06:
**Branch copied from #1106, so maybe I have to change this if #1106 changes.**

Not allow that an event is scheduled at the same time than other event in the admin schedule, as previously:

image

I've also added a validation to avoid that this happens when scheduling in different tabs. I've added tests for this validation with four cases that are representative.

### DIFF
--- a/app/assets/javascripts/osem-schedule.js
+++ b/app/assets/javascripts/osem-schedule.js
@@ -1,10 +1,13 @@
 var url; // Should be initialize in Schedule.initialize
 var schedule_id; // Should be initialize in Schedule.initialize
 
-function showError(error){
-  // Delete other error messages before showing the new one
+var events_to_save = {};
+var events_to_remove = {};
+
+function showMessage(message, type){
+  // Delete other messages before showing the new one
   $('.unobtrusive-flash-container').empty();
-  UnobtrusiveFlash.showFlashMessage(error, {type: 'error'});
+  UnobtrusiveFlash.showFlashMessage(message, {type: type});
 }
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/osem-schedule.js
```diff
@@ -86,8 +133,18 @@ $(document).ready( function() {
     snap: '.schedule-room-slot',
     revertDuration: 200,
     revert: function (event, ui) {
-        console.log(event.attr);
-        return !event;
+      console.log(event.attr);
```
differentreality: undefined
This raises an error. Doesn't affect the PR's functionality, which seems to be working OK.

```
osem-schedule.js?body=1:136 function ( name, value ) {
		return access( this, jQuery.attr, name, value, arguments.length > 1 );
	}

```

Ana06: I can remove that because it is the only log that is still here and I think it is not needed.

differentreality: Sounds good

Ana06:
As I closed #1106 and this depends on it I will close this too.

differentreality:
Would it make sense to show on the schedule-grid only events that are confirmed? That would make schedule ignore if the events are scheduled, so that events that are scheduled and cancelled/withdrawn will be left out of the schedule grid. 
That would allow us to schedule another event in their slot. 

I am thinking something along those lines:
http://paste.opensuse.org/47699581

PR #1175, opened 2016-08-27
using sprockets-rails instead of quiet-assets

shlok007:
Gem `quiet_assets` is deprecated. This PR uses `sprocket-rails` instead of quiet assets to mute the log messages.


ChrisBr:
LGTM

Issue #1081, opened 2016-07-03
Add more info on user profile


differentreality:
Something that came up during Hackweek. 
Eg. user's twitter handle.

altsalt:
I would at a minimum like to see these included:
Short Bio
Personal Site
Twitter

shlok007:
@differentreality I would like to work on this.

shlok007:
What info do you think could be asked ?

Something like:
- [ ] **Twitter handle**
- [ ] **Personal Site**
- [ ] **LinkedIn** and **github**?
- [ ] **Gender**
- [ ] **Designation**
- [ ] **Company**/**Organizations** 
- [ ] **Short Bio**? ( We already have a biography field )
- [ ] If not the address we could atleast ask the **country** they belong to
- [ ] **Technology they work on**

hennevogel:
Please do not collect too much data from people. This is enough:

* Link to Personal Page/Blog
* Social Media (Diaspora, GNU Social, Twitter, Google+, LinkedIn) 
* Social Code (Gitlab, [Gna!](http://gna.org/), [Savannah](http://savannah.gnu.org/), Github)

namangupta01:
i would like to work on this issue?

differentreality:
@namangupta01 this is a new feature, you should post how you plan on implementing it prior to starting working on it.

PR #1180, opened 2016-09-01
User profile info

AnithaPal:
Add more info on user profile #1081:  added social media and code information to the user model. It can be updated through edit_user_path
[User Profile](http://i.imgur.com/CTokp7E.png)


app/views/users/edit.html.haml
```diff
@@ -14,6 +14,18 @@
         = link_to 'Change your avatar here', 'https://gravatar.com'
         = f.input :affiliation, as: :string,
           hint: 'This could be a company, a user group, or nothing at all.'
+        = f.input :website_url, label: 'Personal Website'
+        = f.inputs :name => 'Social Media' do
+          = f.input :linkedin
```
ChrisBr: We use html5 so you could use input types:
http://www.w3schools.com/html/html_form_input_types.asp

app/views/users/show.html.haml
```diff
@@ -9,6 +9,28 @@
             = @user.nickname
       %p
         = markdown(@user.biography)
+      %h3 Social Media and Social Code
+      %ul.list-unstyled
+        %li
+          = @user.website_url
```
ChrisBr: Please make this links! Think about what the user should provide as data and how you want to display!
E.g. for twitter I would expect that I enter ``@bruckmayer`` and that it displays ``https://twitter.com/bruckmayer``. So you have to think about each of these services...
Just displaying it in a list does not add much value ...

ChrisBr:
Could you please squash your commits together into one single commit?

ChrisBr:
Think about what the user should provide as data and how you want to display!
E.g. for twitter I would expect that I enter @bruckmayer and that it displays https://twitter.com/bruckmayer. So you have to think about each of these services...
Just displaying it in a list does not add much value ...

AnithaPal:
Since services sometimes change how they construct username urls I thought I could leave it up to the osem user to input the correct url.

ChrisBr:
One test is failing:
``/spec/views/conference/show.html.haml_spec.rb:11``

ChrisBr:
@AnithaPal you have a lot commits in this PR which doesn't belong to it, e.g.:

a371547536e406160d42cf6f8e711dcc193c6a02
b066781ed9a627483269efdc9b2e366d6c07450e
5a6a8a3c81dac7b7ec13f9b1c545fe81e49ade75
fbb295042525505e4ec0188e81c6b95985919c64

and all of the merge commits! The only commit which represents this PR is 
ebb118a8835da382c356435977b663b17afe5c13

This may sound picky, however, using small and atomar commits and proficiency in git is really essential! Can one of your mentors teach you about advanced git technologies? If not, please join the our IRC channel!

app/views/users/edit.html.haml
```diff
@@ -14,6 +14,18 @@
         = link_to 'Change your avatar here', 'https://gravatar.com'
         = f.input :affiliation, as: :string,
           hint: 'This could be a company, a user group, or nothing at all.'
+        = f.input :website_url, type: :url, label: 'Personal Website'
+        = f.inputs :name => 'Social Media' do
+          = f.input :linkedin, type: :url
```
ChrisBr: I would like to see some hints for the different types. E.g. for the twitter handle I don't know if I have to enter ``@bruckmayer`` or ``https://twitter.com/bruckmayer``.

spec/factories/users.rb
```diff
@@ -20,6 +20,17 @@
     sequence(:username) { |n| "username#{n}" }
     password 'changeme'
     password_confirmation 'changeme'
+    googleplus { Faker::Internet.url('http://plus.google.com/testosemurl') }
```
ChrisBr: There is a best practice for factories: Just Enough Data
https://semaphoreci.com/community/tutorials/working-effectively-with-data-factories-using-factorygirl

That means, you should only put data in your factory which is necessary to create a valid user. As you don't validate the presence of these fields you should drop it.

And furthermore, according to the [faker gem docs ](https://github.com/stympy/faker)this will produce urls like:
```http://http://linkedin.com/testosemuser/francis```

File: spec/features/user_spec.rb [hunk 5]
```diff
@@ -2,11 +2,46 @@
 
 feature User do
 
+  let!(:user) {create(:user)}
+
+  describe 'update user profile' do
```
ChrisBr: What about model tests for the validation?

app/views/users/show.html.haml
```diff
@@ -11,14 +11,69 @@
...
-                %strong
-                  at
-                = event.program.conference.title
+      %h3 Social Media and Social Code
+      %table.table
+        %tr
+          %td
+            %b Website
+          %td
+            =  @user.website_url.present? ? @user.website_url : '-'
```
ChrisBr: These are still no links!

ChrisBr:
What about using font awesome icons on the user show page?
http://fontawesome.io/icons/

For most fields they should have icons (github, google+, gitlab, linkedin, twitter, etc)

differentreality:
I think the user page has become way too long, what about splitting the form into 2 columns?

screenshot from 2017-01-11 18-33-56

ChrisBr:
@differentreality I wouldn't do it because it breaks with the reading flow...
Maybe a second tab with "Contact Information" / "Social Media"?

differentreality:
Perhaps. The user, however, will be saving data that are not visible (ie data on the other tab will be saved as well). 

I've also considered a drop down that lets you add the social info you actually want and does not show everything all together.

BelieveC:
@differentreality, Can I work on this? I think we can implement Tab in form as @ChrisBr suggested. Similar to facebook profile info update, we would only save data fields which are there in that tab instead of saving a complete form. This would ensure User will update only desired fields which are visible to him in particular tab.

ChrisBr:
@BelieveC go ahead!

BelieveC:
Thank you @ChrisBr.

PR #1187, opened 2016-09-08
Remove unnecessary code from conference_controller

Ana06:
Remove unnecessary code from conference_controller#update.

`@conference = Conference.find_by(short_title: params[:id])` is not needed as it is loaded in https://github.com/Ana06/osem/blob/2fc19f1b6c67e56409de6333d0f0661ec0427add/app/controllers/admin/conference_controller.rb#L3

### DIFF
--- a/app/controllers/admin/conference_controller.rb
+++ b/app/controllers/admin/conference_controller.rb
@@ -79,7 +79,6 @@ module Admin
     end
 
     def update
-      @conference = Conference.find_by(short_title: params[:id])
       short_title = @conference.short_title
       @conference.assign_attributes(conference_params)
       send_mail_on_conf_update = @conference.notify_on_dates_changed?


ChrisBr:
@Ana06 I really like these changes, however you still have failing tests! Have you tried your changes? Because I assume load_and_authorize is not working properly as @conference.short_title is ...

app/controllers/admin/conference_controller.rb
```diff
@@ -89,7 +87,7 @@ def update
         redirect_to edit_admin_conference_path(id: @conference.short_title),
                     notice: 'Conference was successfully updated.'
       else
-        redirect_to edit_admin_conference_path(id: short_title),
+        redirect_to edit_admin_conference_path(id: @conference.short_title),
```
differentreality: If the user tries to update the conference and the short_title is left blank, this line is gonna crash.

differentreality:
Due to the short_title existence validation, I think it makes sense to leave this code as it is. @Ana06 @ChrisBr  what do you think?

app/controllers/admin/conference_controller.rb
Ana06: you are right  :confounded:

Ana06:
@differentreality It should be possible to remove this line: https://github.com/openSUSE/osem/pull/1187/files#diff-2b80456463c1fbc204231eef0ac434c9L82 as it is loaded here: https://github.com/Ana06/osem/blob/2fc19f1b6c67e56409de6333d0f0661ec0427add/app/controllers/admin/conference_controller.rb#L3

Although the **`short_title` variable is necessary**. I'll left that as it was.

differentreality:
Yeap that's right! :+1:

PR #1189, opened 2016-09-12
Support JSONP callbacks for the API/v1 endpoints

kormoc:
SeaGL currently uses OSEM as a backend, but still has a static website
serving as our public interface. The API endpoints have a great amount
of data, but without the jsonp callback paramater, embedding the data
is a bit difficult. This change just adds the jsonp callback param
to all the API endpoints. This should be a NOOP for anyone using direct
json calls, but will allow cross domain requests that require jsonp.

### DIFF
--- a/app/controllers/api/v1/conferences_controller.rb
+++ b/app/controllers/api/v1/conferences_controller.rb
@@ -5,11 +5,11 @@ module Api
       respond_to :json
 
       def index
-        render json: @conferences, serializer: ConferencesArraySerializer
+        render json: @conferences, serializer: ConferencesArraySerializer, callback: params['callback']
       end
 
       def show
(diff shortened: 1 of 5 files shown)


differentreality:
Is this supposed to produce anything different than before?
https://osem-demo-pr-1189.herokuapp.com/api/v1/conferences

kormoc:
My apologies. In my cleanup of my patch, I removed some required lines for the functionality.

Once this updated PR deploys, https://osem-demo-pr-1189.herokuapp.com/api/v1/conferences will still display the same data, but https://osem-demo-pr-1189.herokuapp.com/api/v1/conferences?callback=foobar will display a jsonp resource, which is needed for embedded cross domain requests.

differentreality:
:+1: thanks @kormoc

Issue #1096, opened 2016-07-10
Navbar does not collapse when clicked outside the navbar


arun1595:
After clicking on the `.navbar-toggle` button, the navbar expands showing the links. When a user clicks oustide the navbar, the button's background changes but the navbar doesn't collapse.

[Watch here](https://vid.me/VLqZ)

itaditya:
I would like to take on this bug , is anyone on this one already . Please tell so that otherwise I can contribute ;)

bear454:
See https://github.com/openSUSE/osem/pull/1957#issuecomment-373197255

PR #1199, opened 2016-09-19
To enable navbar collapse on clicking outside the navbar

itaditya:
To enable navbar collapse on clicking outside the navbar .
To make it mergeable I have to make some more commits . Please guide me if there is a problem with that . ;)

### DIFF
--- a/app/controllers/api/v1/conferences_controller.rb
+++ b/app/controllers/api/v1/conferences_controller.rb
@@ -4,12 +4,15 @@ module Api
       load_resource find_by: :short_title
       respond_to :json
 
+      # Disable forgery protection for any json requests. This is required for jsonp support
+      skip_before_action :verify_authenticity_token
+
       def index
-        render json: @conferences, serializer: ConferencesArraySerializer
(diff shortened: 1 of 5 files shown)


differentreality:
Hey :-) 

Could you write in more detail which problem you are trying to solve?

itaditya:
Thanks for replying : ) . In mobile screen , when the navbar hamburger(menu) btn is clicked , it expands , but it only caollapse when the user clicks on that button only. So I have made this improvement in which it also collapse when the user clicks anywher on the screen . Thus I think it imroves UX . (Maybe you have seen this behaviour for modals in which on clicking outside the modal it dissapears .)

differentreality:
@itaditya thank you for working on this! Can you please split the 2 issues in different PRs? You also have some changes that are irrelevant to what you are trying to fix.

itaditya:
Ok I'll do that
;)

itaditya:
Hello, I found that this issue was a duplicate of #1096 , and so I have sent a PR on that properly , please review it .

Issue #1197, opened 2016-09-18
Admin conferences program events#show page- always shows a red X indicating commercial is missing


richiethomas:
This is caused by a mis-named "event.progress_status['commercial']" parameter being passed to the loop.  It should be "event.progress_status['commercials']".

I've already pushed a fix; just capturing the issue here for visibility.

differentreality:
Good that you opened an issue about it, thanks, good catch! It would be awesome if you could open a separate PR fixing just this issue. Let us know here, or on IRC, if you have trouble doing that.

richiethomas:
Sure thing.  BTW I noticed that the events/show.html.haml page does not have test coverage, which is why this issue wasn't fixed.  I am writing a view spec for that page now, if that's OK.

richiethomas:
Also, I noticed that the view spec for events/index.html.haml only checks for the presence of the column header labels, not for specific values for each event.  Are those values being checked in a feature spec somewhere, and if so, should I follow the same pattern with the events/show view spec?

differentreality:
Actually, we don't really want to introduce any more view specs.

differentreality:
You can open a new Pull Request only with this change at any time - you don't have to wait for your other PR to be merged first. Just make sure the new PR is a branch coming from the master, so that it only contains relevant changes.

richiethomas:
Not sure I understand, do you mean push to a local branch with just the relevant commit onto my forked repo and do the PR from that branch?  I'll give that a shot now.

PR #1200, opened 2016-09-19
Fix admin conferences program events#show page, now only shows a red X when commercial is missing

richiethomas:
Issue #1197


differentreality:
:+1:  
Thanks @richiethomas!

Issue #1158, opened 2016-08-17
Event ratings


hennevogel:
People should be able to rate Events after they have occurred. Maybe even for different qualities like

* Speaker Quality
* Competence
* Slides

shlok007:
There isn't an `event#show` page yet. The details of an event can be found in `proposals#show`. Can the rating be implemented on the same page? 
Is something like this okay?
![rate_osem](https://cloud.githubusercontent.com/assets/14155445/18751774/3f7788c2-80ae-11e6-94bb-8dbc8604e673.png)

ChrisBr:
@shlok007 event#show == proposal#show! I will have a look at your PR!

shlok007:
Thank you!

PR #1100, opened 2016-07-13
Surveys

differentreality:
Add Surveys
* the organizer can generate survey questions and possible replies manually
* surveys have a specific target, ie they are shown either during registration or after the conference
* substitute questions

This feature resulted from the work of multiple ppl during HackWeek 14

PR #1204, opened 2016-09-22
Implement Event Ratings

shlok007:
Refers to issue #1158 .

Rfc: https://docs.google.com/document/d/1sWImuP14RFeTWrs3nqdRZt8xfD0-xr0hBj10B2_KCQQ/edit?usp=sharing

To-Do:
- [x] Introduce VotableFields and admin views to manage votable fields.
- [x] Enable votes for events ( proposals ) by users if the event has ended.
- [x] Add tests for votable fields introduced.
- [x] Introduce new voting on admin side as well.
- [x] Drop the whole current voting system on admin side.
- [x] Rake task to migrate current voting of past conferences to new 'overall' in VotableField.


shlok007:
The code generated is by `ratyrate` gem . Can somebody please help with the failing test here?
hakiri

app/models/event.rb
```diff
@@ -84,6 +85,10 @@ def registration_possible?
     registrations.count < max_attendees
   end
 
+  def ended?
+    Time.now.to_i > event_schedules.find_by(schedule_id: program.selected_schedule_id).end_time.to_i
```
shlok007: I am not sure if this is the right way to compute if an event has ended .

Ana06: I think you should use the timezone of the conference this event belongs to. And you can compare times directly without converting them to string.

``` Ruby
timezone = program.conference.timezone
Time.find_zone(timezone).now > event_schedules.find_by(schedule: program.selected_schedule).end_time
```

differentreality:
@shlok007 did you review other gems for this as well?

differentreality:
@shlok007 ?

shlok007:
Sorry for the late response. Yes, i stumbled upon [letsrate](https://github.com/muratguzel/letsrate) gem as well. There is always a possibility to write our own rating system to rate proposals.

differentreality:
Nah we don't have to do that, if there is something well maintained with updates that suits our needs. But I would like to know what other gems you evaluated and what the one you chose offers us. So can you provide some more details about your choice of gem? What advantages/disadvantages it has in comparison with other alternatives? Things like that which will help us make a conscious decision. 

We will use voting extensively, so we should choose the gem wisely and then use it for all the places we have/will have voting.

shlok007:
I stumbled upon [`letsrate`](https://github.com/muratguzel/letsrate) and its fork [`ratyrate`](https://github.com/wazery/ratyrate). Seems like ratyrate is still maintained after letsrate. Also,  ratyrate has a lot of options and well documented guide that we could use. There is also [`jquery-raty-rails`](https://github.com/bmc/jquery-raty-rails) which I avoided as people seem to be having issues with rails 4 https://github.com/bmc/jquery-raty-rails/issues/7 .

This [poster](https://camo.githubusercontent.com/d5c0f4ac5be31adbb930b78c5e561232e7a91f12/68747470733a2f2f646c2e64726f70626f7875736572636f6e74656e742e636f6d2f752f37313630353038302f526174795261746525323046656174757265732e706e67 ) lists the additional features that ratyrate has over letsrate.

At the end of the day, all the rating gems wraps the [JQuery Raty](https://github.com/wbotelhos/raty) library so if none works for us we have an option to manually add and use the library ourself.

shlok007:
Do let me know what should be appropriate here. I'll proceed with this PR and try fixing the Hakiri test failure if we decide to use `ratyrate` gem . :slightly_smiling_face:

differentreality:
Please rebase!

app/assets/javascripts/jquery.raty.js
```diff
@@ -0,0 +1,760 @@
+/*!
```
differentreality: Why is this file here?

app/models/event.rb
```diff
@@ -1,5 +1,6 @@
 class Event < ActiveRecord::Base
   include ActiveRecord::Transitions
+  ratyrate_rateable 'speaker_quality', 'competence', 'slides'
```
differentreality: Can we make this custom? Meaning an organizer chooses the qualities to rate on.

differentreality:
What about making this available on the admin side as well? Let's first decide on how to implement it, in a way that does not disrupt too much conferences using the current way of voting (and only want 1 point of voting, overall voting - not voting on multiple qualities).

shlok007:
This is what I propose:

- The votable fields can be decided by organizers in a separate view `admin/votes#index` and `admin/votes#new` .

Screenshot of `admin/votes#index` :
vote
 
- Rake task to migrate the current votes from the admin side into a new field `overall rating` .

- The new fields added could have an option to be global ( like questions ) for all conferences and hence could be enabled or disabled for each conference .

- VotableFields would have a has_and_belongs_to_many relationship with conference .

app/assets/javascripts/ratyrate.js.erb
```diff
@@ -0,0 +1,62 @@
+$.fn.raty.defaults.half = false;
```
differentreality: Why do we need this file if we have the gem?

app/models/rate.rb
```diff
@@ -0,0 +1,6 @@
+class Rate < ActiveRecord::Base
+  belongs_to :rater, class_name: 'User'
+  belongs_to :rateable, polymorphic: true
+
+  # attr_accessible :rate, :dimension
```
differentreality: Why do we need this comment here?

differentreality:
Do we truly need to introduce all those new models?

app/assets/javascripts/ratyrate.js.erb
shlok007: The gem seems to have included it by itself. I am wondering the same. Although, I'll give it a check if the gem still works the same when I get rid of it.

app/models/rate.rb
shlok007: I don't think we do. This is automatically generated code. I'll look up more for comments/things like these that can be safely removed. Although, I shouldn't change much except comments as I won't know reason things were introduced.

app/models/votable_field.rb
```diff
@@ -0,0 +1,18 @@
+class VotableField < ActiveRecord::Base
+  belongs_to :conference
+  validates :title, :votable_type, presence: true
+
+  # ratyrate does not allow criterias to have spaces in them
+  validate :no_spaces_in_title
```
shlok007: Another solution could be to replace spaces with another character like underscore ( _ ) and convert them back to spaces while rendering them in proposals#show . But this could be a problem if the user actually defines a field that has a underscore.

differentreality: We ask for proper names.

app/models/votable_field.rb
```diff
@@ -0,0 +1,18 @@
...
+  validate :correct_votable_type
+
+  private
+
+  def no_spaces_in_title
+    errors.add(:title, 'should not have spaces') unless title.match(/\s/).nil?
+  end
+
+  def correct_votable_type
+    errors.add(:votable_type, "should be one of the following: 'Event'") unless ['Event'].include? votable_type
```
differentreality: available votable_types should be in a variable

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Osem::Application.routes.draw do
 
+  post '/rate' => 'rater#create', :as => 'rate'
```
differentreality: as: 'rate'

app/models/votable_field.rb
shlok007: By proper names you mean names without spaces, the way it is currently done, right? :slightly_smiling_face:

differentreality: You commented on the code that ratyrate does not allow votable_fields to have spaces, so that's what's proper. No spaces.

shlok007: Would it be better if I use regexp instead of this to check for no spaces ?

app/models/votable_field.rb
shlok007: agreed .. :sweat_smile:

differentreality:
So I guess the remaining things to take into account here are:
* Enable the gem for the admin voting on events
* Show results of current implemented voting in admin

app/assets/javascripts/ratyrate.js.erb
shlok007: It doesn't seem to work without it.

shlok007:
It seems like ratyrate does not allow using same names ( `dimension` as they call it ) to vote for the same model. Therefore, it won't be possible to have the same field name for admin and non admin side.

Ticket in ratyrate to address the same: https://github.com/wazery/ratyrate/issues/143 .
Official Documentation for ratyrate: https://www.sitepoint.com/ratyrate-add-rating-rails-app/

shlok007:
The two drawbacks of using this gem are: 

- Title for the votable fields cannot have spaces so fields like `speaker's familiarity with subject` which cannot be represented by a single word might cause a problem

    **Possible Solution**
    Replace spaces with `_` or some other character in         https://github.com/openSUSE/osem/pull/1204/files#diff-6ecd61439b970b0e5ae49be738861446R4 ( with         this approach `speaker competance` and `speaker_competance` will be treated the same, we might have to take care of that as well )

- We cannot have same field name to rate a model in admin and non admin side. ( eg: if the attendees have an option to rate on 'overall' then the admins can't have an 'overall' field, if they do the ratings of both will be combined together )

    **Possible Solution**
    Prepend `admin_` to the votable fields that are created for the admin side voting by the admin.

In both the cases we have an option to just go with the default behaviour although, it might reduce the usability.

Gemfile
```diff
@@ -23,6 +23,9 @@ gem 'mysql2'
 # for observing records
 gem 'rails-observers'
 
+# for rating
+gem 'ratyrate', github: 'wazery/ratyrate', branch: 'master'
```
shlok007: Something I am really **unsure** about. The released version of ratyrate contains a bug in the [rating_for_user](https://github.com/wazery/ratyrate#other-helpers) method that shows a particular user's rating. This method is required for blind voting. However, the master is quite stable and does not contain the bug. This could be a temporary change until we get a response regarding https://github.com/wazery/ratyrate/issues/125 ( issue asking for a new version release )

shlok007: Does using the gem from the **master** branch instead of the released version something we can agree upon?

shlok007: Using the gem from master:
ratyrate_gem

Using the released version:
rayrate

The function to show current_user's vote renders only a single star in the released version which has been fixed in master but a new version hasn't been released yet.

delblues: What's this master branch version you guys are talking about? I'm trying the master, but keep comes the alpha, is there another file to download? 

I'm having a lot of issues with this gem for helper `rating_for_user`. I can't record at `overrall_averages` and `average_cache` tables, as long record the` rater_id` at `rate table. `
The method on this file is also comment:
https://github.com/wazery/ratyrate/blob/master/lib/ratyrate/model.rb

tks

shlok007: Using the master branch is as simple as using this line above in the Gemfile. It would be great if we can discuss this more on the issue itself rather than here. :slightly_smiling_face:

shlok007:
Except tests, the work on this PR should be complete. Can it be please temporarily deployed to heroku at this stage for everyone else to review ?

differentreality:
@shlok007 it would be useful if you added the necessary records in the review app.

differentreality:
Please rebase, and feel free to squash those commits to make rebasing easier (you can leave a couple of commits, if you think that's useful).

shlok007:
Documenting the new rating feature: https://docs.google.com/document/d/1sWImuP14RFeTWrs3nqdRZt8xfD0-xr0hBj10B2_KCQQ/edit?usp=sharing

app/assets/stylesheets/osem-rating.css.scss
```diff
@@ -1,19 +1,3 @@
...
-.myrating.glow { background-image: image-url("star-glow.png"); }
-.othersrating.bright { background-image: image-url("star-bright.png"); }
-.avgrating.bright { background-image: image-url("star-bright.png"); }
-.avgrating {
-  background: image-url("star.png") 0 0;
-  margin-right: -2px;
-  width: 24px;
-  height: 24px;
-  display: inline-block;
+img.raty-cancel {
```
shlok007: The cancel button doesn't cancel the rating but instead clicking on cancel changes the rating to '0' stars. This property is to prevent the cancel button from being displayed.

app/controllers/admin/votable_fields_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+      else
+        redirect_to admin_conference_votable_fields_path(@conference.short_title),
+                    error: 'Votable field could not be destroyed.' \
+                    "#{@votable_field.errors.full_messages.join('. ')}."
+      end
+    end
+
+    private
+
+    def remove_rates
```
shlok007: To delete the existing rating for a votable field so when a new field of same title is created, it does not have any previous rating present already.

differentreality:
I am thinking that if we manually have to introduce the custom votable fields we want, then we might as well not use the gem, and just extend the capabilities of the existing voting options we have, and refactor where needed.

app/views/admin/votable_fields/_form.html.haml
```diff
@@ -0,0 +1,16 @@
...
+        - if @votable_field.new_record?
+          New
+        Votable Field
+.row
+  .col-md-8
+    = semantic_form_for(@votable_field, :url => (@votable_field.new_record? ? admin_conference_votable_fields_path : admin_conference_votable_field_path(@conference.short_title, @votable_field))) do |f|
+      = f.input :title
+      = f.input :votable_type, collection: options_for_select(['Event'])
+      = f.input :stars, label: 'Add the total number of stars for the field'
+      = f.input :for_admin, label: 'Add this field for voting in admin side only'
```
differentreality: How about we have 2 boolean options, one for admin namespace, one for attendees? That way we can have votable fields available only for admin, or only for attendees, or for both.

app/views/admin/programs/_form.html.haml
```diff
@@ -7,11 +7,11 @@
...
       = f.input :schedule_interval, hint: "It is the minimal time interval of your schedule. The value should be 5, 6, 10, 12, 15, 20, 30 or 60. Warning! Some events could be unscheduled when changing this value."
-      = f.input :blind_voting, hint: 'Enable this feature if you do not want to show voting results and voters prior to user submitting a vote. For the feature to work you need to set the voting dates below as well'
-      = f.input :voting_start_date, as: :string, input_html: { id: 'datetimepicker-voting_start_date', readonly: true, value: (f.object.voting_start_date.to_formatted_s(:db_without_seconds) unless f.object.voting_start_date.nil?) }
-      = f.input :voting_end_date, as: :string, input_html: { id: 'datetimepicker-voting_start_date', readonly: true, value: (f.object.voting_end_date.to_formatted_s(:db_without_seconds) unless f.object.voting_end_date.nil?) }
+      = f.input :rating_enabled, label: 'Enable voting', hint: 'To enable voting you need to set voting dates as well'
```
differentreality: Let's add a new section to separate the voting options from the rest.

app/models/vote.rb
```diff
@@ -1,14 +1,2 @@
 class Vote < ActiveRecord::Base
```
shlok007: There is a rake task to migrate old votes into new. The Vote model class is required for that task. If this class is removed we might have to write raw SQL inside the task to access the old votes data table.

app/models/conference.rb
```diff
@@ -22,7 +22,7 @@ class Conference < ActiveRecord::Base
   has_many :supporters, through: :ticket_purchases, source: :user
   has_many :tickets, dependent: :destroy
   has_many :resources, dependent: :destroy
-
+  has_many :votable_fields, dependent: :destroy
```
shlok007: Then probably we should move VotableField to be a program option instead of conference in all places including views and routes of the app as well. I will be doing that as well. :)

app/models/vote.rb
differentreality: There are other ways to do that. Eg. https://github.com/openSUSE/osem/blob/master/db/migrate/20140610165551_migrate_data_person_to_user.rb#L2

bear454:
I would like to close this PR: 

* The discussion looks like we could better solve the problem with enhancements to #1100 .
* There have been no changes in over half a year, and it's now very conflicted with master.

differentreality:
What I had in mind was to keep the stars rating, but maybe we should indeed look into integrating something specifically for this reason into surveys. 

Indeed it's been a while since we worked on this. Closing.

PR #1210, opened 2016-10-05
Fix event type description display on proposals#new and proposals#edit

### LINKED ISSUE #1208: proposal#new and proposal#edit do not show description of event_type on load

jmks:
Fix #1208  

On load the correct event type description is displayed by osem.js:166 when the dropdown value is set. Re-showing the text would be harmless, but it triggers a change event listener [here](https://github.com/openSUSE/osem/blob/7cc2609662dd6e1c581c74a36c87edd57e6b82ab/app/assets/javascripts/osem.js#L34-L39)

Just removing these lines seems to display the description correctly for different event types and renders from errors, on both `proposals#new` and `proposals#edit`.

### DIFF
--- a/app/views/proposals/_proposal_form.html.haml
+++ b/app/views/proposals/_proposal_form.html.haml
@@ -23,9 +23,6 @@
       %span{ class: 'help-block select-help-text event_event_type_id collapse', id: "#{event_type.id}-help" }
         = event_type.description
 
-    :javascript
-      $("##{@conference.program.event_types.first.id}-help").collapse('show');
-
     = f.input :difficulty_level, as: :select, collection: @conference.program.difficulty_levels, input_html: { class: 'select-help-toggle' },
       include_blank: '(Please select)' if @conference.program.difficulty_levels.any?
 
--- a/app/views/proposals/new.html.haml
+++ b/app/views/proposals/new.html.haml
@@ -40,9 +40,6 @@
                 %span{ class: 'help-block event_event_type_id collapse', id: "#{event_type.id}-help" }
                   = event_type.description
 
-              :javascript
-                $("##{@program.event_types.first.id}-help").collapse('show');
-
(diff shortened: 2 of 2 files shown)


differentreality:
That seems to be working :-) Thank you @jmks for working on this!

A suggestion: when you want to refer to files and lines in files, please include a link. If you see a file in github you can click on the line number and then the line is included in the url and you can use that url in your comments to properly link to the file/line you want.

differentreality:
:+1:

differentreality:
@jmks there is actually one more thing you need to check. The files you touched are shared with the admin view of editing the proposal, and that's not showing up properly. 
We should fix that as well!

jmks:
@differentreality I'm looking at routes like `admin/conferences/:conference_id/program/events/:id/edit` and the help text shows up correctly.

 The admin view [file that's rendered](https://github.com/openSUSE/osem/blob/7cc2609662dd6e1c581c74a36c87edd57e6b82ab/app/views/admin/events/edit.html.haml#L1) just [renders one of the files](https://github.com/openSUSE/osem/blob/7cc2609662dd6e1c581c74a36c87edd57e6b82ab/app/views/proposals/_form.html.haml#L9) I changed.

Should I be looking somewhere else?

differentreality:
Try editing a proposal with event type which is *not* the first one in the list.

jmks:
Am I looking in the right spot?

osem admin edit events highlights

Maybe a browser difference, and I should explicitly set the selected item.

differentreality:
@jmks is that what you see when the page is loaded without touching anything? Without selecting the event type from the drop down?

jmks:
@differentreality Sorry for the delay. Yes, that is what I see immediately after the page loads (Chrome 53.0.2785.116 (64-bit)).

differentreality:
OK this seems to be working :)
:+1: good work @jmks

PR #1215, opened 2016-10-18
fixed #1096

### LINKED ISSUE #1096: Navbar does not collapse when clicked outside the navbar

itaditya:
In mobile screen , when the navbar hamburger(menu) btn is clicked , it expands , but it only collapse when the user clicks on that button only. So I have made this improvement in which it also collapse when the user clicks anywhere on the screen . Thus I think it improves UX . (Maybe you have seen this behaviour for modals in which on clicking outside the modal it disappears .)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -54,4 +54,8 @@ $(document).ready(function() {
     $('body').smoothScroll({
         delegateSelector: 'a.smoothscroll'
     });
+    // Navbar collapse when clicked elsewhere
+    $('.navbar-toggle').blur(function() {
+        $(this).click();
+    });
 });


differentreality:
This solution doesn't seem to allow for sub-menus to appear.
Eg. I cannot click on the sign_in dropdown that should make the form appear

screenshot from 2016-11-28 14-08-05

differentreality:
@itaditya thanks for working on this. Unfortunately this does not seem to be the best approach (check my previous comment). I will close this PR so that you can open a new one with a revised solution, OK?

itaditya:
Ok , no problem

Issue #1222, opened 2016-11-20
Cross-Site Request Forgery Warning in app/controllers/api/base_controller.rb


rishabhs95:
[Security issue from Hakiri](https://hakiri.io/projects/d1eb39a7696349/stacks/70eaa9009f1f2b/builds/21ffb2eea97058/warnings/5170ff4a2b6fbb): protect_from_forgery should be configured with 'with: :exception' in app/controllers/api/base_controller.rb

PR #1223, opened 2016-11-20
hakiri: fix security issue

rishabhs95:


app/controllers/api/base_controller.rb
```diff
@@ -1,5 +1,5 @@
 module Api
   class BaseController < ActionController::Base
-    protect_from_forgery
+    protect_from_forgery(with: :exception)
```
differentreality: Shall we lose the parenthesis?

rishabhs95: Done 👍

differentreality:
:+1: Thanks for the quick fix @rishabhs95 !

differentreality:
Fixes https://github.com/openSUSE/osem/issues/1222

PR #1228, opened 2016-12-15
Fix multiple choice questions

Ana06:

PR #1229, opened 2016-12-16
Fixes for PostreSQL support

commandprompt:
Hi, I've made a bunch of fixes in order to make OSEM run on PosgreSQL, I would appreciate if someone can take a look at this.
Here is a brief overview of changes I've done:

1. fix database migrations to use the correct _boolean_ type of default values for boolean columns (0 -> false) commit c7cc8f8f34090baa1641013af8c3c6aa168052e9
2. contitionally use the add_monetize column instead of add_money for the ticket price DB migration; commit 06d8f6d3685e60042fa06458626e5e69c87d66aa
3. add missing columns to the ActiveRecord DB querie so Postgres does not complain about the missing columns in GROUP By statements; commit 02cdf05b5d77d7878681029e5c9ba491e27f0908
4. add .count(:all) ActiveRecord call to force inclusion of the "count_all" column in query result set in queries which referred to this column (this happens automaticall for other DBs, but not for Postgres); commit 02cdf05b5d77d7878681029e5c9ba491e27f0908
5. add UUID monkey patch to solve this ActiveUUID issue https://github.com/jashmenn/activeuuid/issues/74; commit a799ea7ca9cbf53eb6058bf636cf289694d16572
6. add PG database driver gem into the project's Gemfile; commit 78824559136b2c8424a44cae1a9d084ada8b212c
7. add updated schema.rb to the project source tree; commit 8903d88ef9cce38c50443d20008fad76ec65968b
8. fix the column types for ahoy_events.visit_id to use INTEGER instead of BINARY(16), so it matches with the FK type on the visits table (this has nothing to do with Postgres, just a bug I've spotted in the process); commit 5e7d5d60d838e28b4e731568ff43349338849a46

Thanks

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -131,7 +131,7 @@ class Conference < ActiveRecord::Base
     result = []
 
     if program && program.cfp && program.events
-      submissions = program.events.group(:week).count
+      submissions = program.events.select(:week).group(:week).order(:week).count
       start_week = program.cfp.start_week
       weeks = program.cfp.weeks
       result = calculate_items_per_week(start_week, weeks, submissions)
@@ -179,7 +179,7 @@ class Conference < ActiveRecord::Base
         registration_period.start_date &&
         registration_period.end_date
 
-      reg = registrations.group(:week).count
+      reg = registrations.group(:week).order(:week).count
       start_week = get_registration_start_week
       weeks = registration_weeks
       result = calculate_items_per_week(start_week, weeks, reg)
@@ -337,8 +337,8 @@ class Conference < ActiveRecord::Base
   # ====Returns
   # * +hash+ -> user: submissions
   def self.get_top_submitter(limit = 5)
-    submitter = EventUser.where('event_role = ?', 'submitter').limit(limit).group(:user_id)
-    counter = submitter.order('count_all desc').count
+    submitter = EventUser.select(:user_id).where('event_role = ?', 'submitter').limit(limit).group(:user_id)
+    counter = submitter.order('count_all desc').count(:all)
     calculate_user_submission_hash(submitter, counter)
   end
 
@@ -348,10 +348,10 @@ class Conference < ActiveRecord::Base
   # ====Returns
   # * +hash+ -> user: submissions
   def get_top_submitter(limit = 5)
-    submitter = EventUser.joins(:event)
+    submitter = EventUser.joins(:event).select(:user_id)
         .where('event_role = ? and program_id = ?', 'submitter', Conference.find(id).program.id)
         .limit(limit).group(:user_id)
-    counter = submitter.order('count_all desc').count
+    counter = submitter.order('count_all desc').count(:all)
     Conference.calculate_user_submission_hash(submitter, counter)
   end
 
@@ -1100,7 +1100,8 @@ class Conference < ActiveRecord::Base
   def self.calculate_user_submission_hash(submitters, counter)
     result = ActiveSupport::OrderedHash.new
     counter.each do |key, value|
-      submitter = submitters.where(user_id: key).first
+      # make PG happy by including the user_id in ORDER
+      submitter = submitters.where(user_id: key).order(:user_id).first
       if submitter
         result[submitter.user] = value
       end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,17 +13,20 @@
 
 ActiveRecord::Schema.define(version: 20170531094819) do
 
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
+
   create_table "ahoy_events", force: :cascade do |t|
-    t.uuid     "visit_id",   limit: 16
+    t.integer  "visit_id"
     t.integer  "user_id"
     t.string   "name"
     t.text     "properties"
     t.datetime "time"
   end
 
-  add_index "ahoy_events", ["time"], name: "index_ahoy_events_on_time"
-  add_index "ahoy_events", ["user_id"], name: "index_ahoy_events_on_user_id"
-  add_index "ahoy_events", ["visit_id"], name: "index_ahoy_events_on_visit_id"
+  add_index "ahoy_events", ["time"], name: "index_ahoy_events_on_time", using: :btree
+  add_index "ahoy_events", ["user_id"], name: "index_ahoy_events_on_user_id", using: :btree
+  add_index "ahoy_events", ["visit_id"], name: "index_ahoy_events_on_visit_id", using: :btree
 
   create_table "answers", force: :cascade do |t|
     t.string   "title"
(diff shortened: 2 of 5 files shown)


commandprompt:
Hi, @differentreality @ChrisBr I saw you've merged a couple of other pull requests today(e.g #1228 ), it seems that our work is also marked as broken by the CI system. Can you please take a look at this PR too?
Thanks.

differentreality:
Hey :-) We hope to work more intensively on the open PRs in the coming days! It might take us a bit longer to review larger PRs, so please bear with us! Regarding Travis, it seems it is complaining about some styling thingies and also some other things which I am not sure your PR affects. We will look more closely. 
Thank you very much for making the effort to make the necessary changes for postgresql and taking the time to send a PR for it!

ChrisBr:
@commandprompt We would be happy to get your changes upstream :)
Did you test if your changes still work with MySQL / SQLite?

commandprompt:
Hey @ChrisBr, yes, I've ran tests using all three database drivers(MySQL, SQLite and Postgres). 
Some tests are failing on all database drivers on the master branch, but there are no difference in RSpec results for different DBs, although it would not hurt if someone else can take a look at the merged version with a fresh pair of eyes :) .
Most of the Activerecord queries are DB agnostic and no raw SQL was used, there are some Postgres-specific adjustments though, but those are conditional (i.e. 06d8f6d and a799ea7)

commandprompt:
@ChrisBr
Regarding the UUID conditional monkey patch here is the reference https://github.com/jashmenn/activeuuid/issues/74 (it was mentioned in the commit comments).

@hennevogel 
As for the migration for the **add_monetize** vs **add_money**, for SQLite/MySQL it is behaving in the same way as it was before Postgres changes; the reason I did that in the original migration is because it will not run correctly with Postgres Activerecord adapter (here is a reference https://github.com/RubyMoney/money-rails/issues/396).

I know that altering already existing migrations is not a good practice in general, but in this situation the only alternative I see is to monkey-patch the SchemaStatements module in the money-rails  gem, which seems to be hacky and even less elegant to me, please correct me if I am wrong.

I must admit that I also altered some other migrations, changing the default values for boolean-type fields to the correct types (0 -> false, see c7cc8f8) and this will apply to all DB drivers at the moment, but I can wrap it with the same conditional for Postgres if you would consider this approach acceptable.
Please let me know whay you think.
Thanks.

differentreality:
If you create new migrations **before** enabling pg gem and migrate them, schema.rb will change. With a proper schema.rb a new installation should successfully run `rake db:schema:load`. Was that not the case with you? Or were you perhaps running `rake db:migrate` instead?

As far as add_monetize is concerned, doesn't it add the same exact columns (so that schema:load should still work)?

Can you help me reproduce the problem that requires us to add the initializer for the uuid patch?

eug3nix:
I've used the rake db:migrate to initialize the database, that's why the migrations also were adjusted.
for the db:schema:load there is a couple of not yet fixed issues in the schema.rb file like this:

create_table "visits", force: :cascade do |t|
    t.uuid     "visitor_id",       **limit: 16**

which are still causing exceptions in PostgreSQL, since the UUID field size should not be explicitly defined (and the UUID should be 16 bytes in its binary representation anyways...)

As for the money please refer to the issue comments 396 for the money-rails project or its readme file, here is a short excerpt from it:
**Notice for Rails >= 4.2 and PG adapter
Due to the addition of the money column type for PostgreSQL in Rails 4.2, you will need to use add_monetize instead of add_money column. If you're adding the column within a create_table block, use t.monetize, and use remove_monetize to remove the column.**
This applies to the db:migrate scenario of course. The schema.rb file with the interger definition for the ticket price should work fine on all databases.

differentreality:
Based on the discussion we had on IRC, it seems we can remove all the changes to the existing migrations, add new migrations that show those changes (they do not affect current installations), and remove the initializer as well.

db/schema.rb
```diff
@@ -13,17 +13,20 @@
 
 ActiveRecord::Schema.define(version: 20160815140302) do
 
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
+
   create_table "ahoy_events", force: :cascade do |t|
-    t.uuid     "visit_id",   limit: 16
+    t.integer  "visit_id"
```
differentreality: We need to introduce a migration that makes this change.

config/initializers/postgres_uuid_patch.rb
```diff
@@ -0,0 +1,19 @@
+module PostgresqlUuidQuotingPatch
```
differentreality: Do we still need this?

app/models/conference.rb
```diff
@@ -280,8 +280,8 @@ def get_status
   # ====Returns
   # * +hash+ -> user: submissions
   def self.get_top_submitter(limit = 5)
-    submitter = EventUser.where('event_role = ?', 'submitter').limit(limit).group(:user_id)
-    counter = submitter.order('count_all desc').count
+    submitter = EventUser.select(:user_id).where('event_role = ?', 'submitter').limit(limit).group(:user_id)
```
differentreality: @ChrisBr I think you had initially written this, right? Could you please review these changes? On mysql they seem to return different results, but I am not sure if that affects the end result we are getting. I am also a bit confused about the result counter gives us.

differentreality:
@eug3nix do we still need the initializer?

eug3nix:
No, I am going to push the new version of this PR tomorrow

db/schema.rb
differentreality: @eug3nix not sure if you plan on doing this or if you missed the comment.

eug3nix: I am still trying to figure out the best solution to that, here are some concerns I have:
1 - the type of visit_id column depends on the Ahoy store being used, so I am not sure if it is a good idea to just override the column type in the migration
2 - with the migration scenario another problem may arise in case if we are updating already existing records - how can we convert UUID to and Integer types without ruining the relations for such records?
3 - I am not exactly sure how the database ends in an invalid state with the different column types for the visit_id vs visits.id

If you know how the Ahoy works in detail, please advise.
Of course, I can just add the migration to change the column type but I don't want to risk breaking anything :)

app/models/conference.rb
differentreality: @ChrisBr do you wanna take a closer look at this?

db/schema.rb
differentreality: @eug3nix the PR description mentions that this is **not** related to the postgres implementation. If not, let's remove unrelated changes that keep this PR from being merged. I think everything else is set :)

db/schema.rb
```diff
@@ -13,17 +13,20 @@
 
 ActiveRecord::Schema.define(version: 20160815140302) do
 
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
differentreality: Is it OK to have this line in our schema.rb even when postgres is **not** being used?
And will this remain after migrating with a different db?

differentreality: This line does not seem to create any problems when using mysql

db/schema.rb
```diff
@@ -13,17 +13,20 @@
...
     t.integer  "user_id"
     t.string   "name"
     t.text     "properties"
     t.datetime "time"
   end
 
-  add_index "ahoy_events", ["time"], name: "index_ahoy_events_on_time"
-  add_index "ahoy_events", ["user_id"], name: "index_ahoy_events_on_user_id"
-  add_index "ahoy_events", ["visit_id"], name: "index_ahoy_events_on_visit_id"
+  add_index "ahoy_events", ["time"], name: "index_ahoy_events_on_time", using: :btree
```
differentreality: @eug3nix following up on our IRC conversation, these changes to the index might also be introduced with mysql, so they don't seem to be an issue either.

db/schema.rb
differentreality: Yes, indeed we need this.  At first glance, records seem to be creating for ahoy_events and visits

differentreality:
@lagartoflojo

differentreality:
@commandprompt any chance you can rebase this so that we can merge it soon? Thanks! :)

differentreality:
There is a space missing, which makes our linters complain, I am going to merge this anyway, as I think it is more important to finally add this.

@eug3nix thanks a lot for this addition! We are looking forward to receiving more! :+1:

PR #1231, opened 2016-12-20
Open graph

ChrisBr:


app/views/conferences/show.html.haml
```diff
@@ -1,3 +1,12 @@
+- content_for :head do
+  %meta{ property: "og:title", content: @conference.title }
+  %meta{ property: "og:site_name", content: (ENV['OSEM_NAME'] || 'OSEM') }
+  %meta{ property: "og:description", content: @conference.description }
+  %meta{ property: "og:url", content: conference_path(@conference.short_title) }
+  - @conference.picture?
```
differentreality: We need an if here

ChrisBr: of course we do ...

ChrisBr:
@differentreality please review.
You can test here:
https://developers.facebook.com/tools/debug/sharing
(I already created a test event, note that it has to be accepted, otherwise the link redirects of course to the conference index page)

differentreality:
:+1:

PR #1233, opened 2016-12-21
Avoid a Javascript error when @tag is null

Ana06:
As it was pointed out in https://github.com/openSUSE/osem/pull/1142 if `@tag` is `null` there is an error in the Console when trying to take the user to the current day and time.


ChrisBr:
LGTM

PR #1235, opened 2016-12-29
caching comment counts

### LINKED ISSUE #1230: Making comment counter more efficient
By introducing a counter cache.
See my comment from #1186 

But we can improve your changes with some advanced Rails techniques :) With your changes, every time I access events#index we make several COUNT calls. Instead hitting the database all the time we could instead cache it. Rails already implements this :smile:

From the rails documentation:

belongs_to :author, dependent: :destroy, counter_cache: true

http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to

Of course you would need a migration to add the column for storing the hash. However, I think it's worth the effort. These adaptions don't make the code more complicated and improve the performance A LOT!

shlok007:
Closes #1230

### DIFF
--- a/db/migrate/20161229080315_add_comments_count_to_events.rb
+++ b/db/migrate/20161229080315_add_comments_count_to_events.rb
@@ -1,0 +1,10 @@
+class AddCommentsCountToEvents < ActiveRecord::Migration
+  def change
+    add_column :events, :comments_count, :integer, default: 0, null: false
+
+    Event.find_each do |event|
+      comments_count = event.comment_threads.count
+      event.update_attribute(:comments_count, comments_count) unless comments_count.zero?
+    end
+  end
+end
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -98,6 +98,29 @@ describe Event do
     end
   end
 
+  describe '#comments_count' do
+    context 'has a valid counter cache' do
+      before do
+        create(:comment, commentable: event)
+      end
+
+      it 'successfully increments comments_count' do
+        expected = expect do
+          create(:comment, commentable: event)
+        end
+        expected.to change { event.comments_count }.by(1)
+      end
+
+      it 'successfully decrements comments_count' do
+        expected = expect do
+          event.comment_threads.last.destroy
+          event.reload
+        end
+        expected.to change { event.comments_count }.by(-1)
+      end
+    end
+  end
+
   describe 'scope ' do
     context 'confirmed' do
       it 'returns only confirmed events' do
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -8,7 +8,7 @@ class Comment < ActiveRecord::Base
   # want user to vote on the quality of comments.
   #acts_as_votable
 
-  belongs_to :commentable, polymorphic: true
+  belongs_to :commentable, counter_cache: true, polymorphic: true
 
   # NOTE: Comments belong to a user
   belongs_to :user
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -184,4 +184,4 @@
                 %ul.dropdown-menu{ role: 'menu' }
                   = render 'change_state_dropdown', event: event
             %td.text-center
-              = link_to "#{event.comment_threads.count}", admin_conference_program_event_path(@conference.short_title, event), anchor: 'comments-div'
+              = link_to "#{event.comments_count}", admin_conference_program_event_path(@conference.short_title, event), anchor: 'comments-div'
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,20 +13,17 @@
 
 ActiveRecord::Schema.define(version: 20170629232817) do
 
-  # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
-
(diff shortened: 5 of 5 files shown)


ChrisBr:
Does this also work for polymorphic?

shlok007:
It was somewhat an issue before but was fixed in the release of rails 5. Seems like it hasn't been backported to 4.x .
Related: https://github.com/rails/rails/issues/16407

Sorry, maybe i'll try to figure out a workaround .

differentreality:
@ChrisBr what shall we do with this?

shlok007:
It seems to me that we haven't encountered the above issue and it works fine for us.

db/migrate/20161229080315_add_comments_count_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddCommentsCountToEvents < ActiveRecord::Migration
+  def change
+    add_column :events, :comments_count, :integer, default: 0
```
lagartoflojo: I'd add `null: false`

spec/models/event_spec.rb
```diff
@@ -98,6 +98,23 @@
     end
   end
 
+  describe '#comments_count' do
+    context 'has a valid counter cache' do
+      before :each do
```
lagartoflojo: No `:each` needed: `before do`

spec/models/event_spec.rb
```diff
@@ -98,6 +98,23 @@
...
+        create(:comment, commentable_type: 'Event', commentable_id: event.id)
```
lagartoflojo: `create :comment, commentable: event`

spec/models/event_spec.rb
```diff
@@ -98,6 +98,23 @@
...
+      end
+
+      it 'successfully increments comments_count' do
+        expect(event.reload.comments_count).to eq 1
```
lagartoflojo: I think the test should be more like:
`expect { create :comment, commentable: event }.to change { event.comments_count }.by(1)`

spec/models/event_spec.rb
```diff
@@ -98,6 +98,23 @@
...
+      end
+
+      it 'successfully decrements comments_count' do
+        event.comment_threads.destroy_all
+        expect(event.reload.comments_count).to eq 0
```
lagartoflojo: Same comment as above, using `....to change...`

db/migrate/20161229080315_add_comments_count_to_events.rb
shlok007: yes sure!

db/migrate/20161229080315_add_comments_count_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddCommentsCountToEvents < ActiveRecord::Migration
+  def change
+    add_column :events, :comments_count, :integer, default: 0, null: false
```
lagartoflojo: So this PR looks good. The only thing we'd be missing which I just noticed is that we need to populate the `comments_count` field of existing events.

db/migrate/20161229080315_add_comments_count_to_events.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  class TempEvent < ActiveRecord::Base
+    self.table_name = 'events'
+  end
+
+  def change
+    add_column :events, :comments_count, :integer, default: 0, null: false
+
+    TempEvent.all.each do |event|
```
lagartoflojo: `TempEvent.all.each` will load ALL events into memory, which is potentially bad.
It's better to use [`TempEvent.find_each`](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/Batches.html#method-i-find_each), which will load the events in batches.

db/migrate/20161229080315_add_comments_count_to_events.rb
```diff
@@ -0,0 +1,18 @@
+class AddCommentsCountToEvents < ActiveRecord::Migration
+  class TempComment < ActiveRecord::Base
+    self.table_name = 'comments'
+  end
+
+  class TempEvent < ActiveRecord::Base
```
lagartoflojo: Why use `TempEvent` instead of just `Event`?

db/migrate/20161229080315_add_comments_count_to_events.rb
```diff
@@ -0,0 +1,18 @@
...
+    self.table_name = 'events'
+  end
+
+  def change
+    add_column :events, :comments_count, :integer, default: 0, null: false
+
+    TempEvent.all.each do |event|
+      comments_count = Comment.where(commentable_id: event.id, commentable_type: 'Event').count
```
lagartoflojo: If you used just `Event` you could do `event.comments` instead of that entire `where`-clause.

lagartoflojo: Or at least you can add `has_many :comments.....` to `TempEvent`

db/migrate/20161229080315_add_comments_count_to_events.rb
shlok007: thank you for this! :slightly_smiling_face:

PR #1239, opened 2017-01-06
Fixed revision count and dropped observers

### LINKED ISSUE #1236: Broken revision count
I don't understand the intended behavior of  [revision_observer.rb](https://github.com/openSUSE/osem/blob/master/app/models/revision_observer.rb) .

It seems like it should increment `revision` column on every change of `room` , `track` , `event` and `conference` itself. If that's the case then then the count increments for the changes in `conference` model only and needs to be fixed.

shlok007:
Closes #1236 .

This PR fixes the `revision` count in conference and also switches to use concerns instead of observers to calculate revision count .


shlok007:
Since, random travis failures were fixed in #1246 . Can the travis build for this PR be restarted please? :slightly_smiling_face:

differentreality:
What do we want to do about this?

shlok007:
Are the changes okay? should I squash and rebase?

differentreality:
@lagartoflojo was it you that would take a look at this?

lagartoflojo:
@differentreality will do!

app/models/concerns/revision_count.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module RevisionCount
+  extend ActiveSupport::Concern
+
+  included do
+    after_update :increment_revision
+  end
+
+  def increment_revision
+    conference = if is_a?(Room)
```
lagartoflojo: I reaaaaally don't like checking the type of the class to see what method to call.
Prefer [duck typing](https://robots.thoughtbot.com/back-to-basics-polymorphism-and-ruby#duck-typing)!

We can just set a "requirement" for the `RevisionCount` module: if a class includes this module, it has to implement the `conference` method.
So:
Room could have `has_one :conference, through: :venue`
Conference: `def conference; self; end`
Event: `has_one :conference, through: :program`
That way, the `increment_revision` method is reduced to:
```ruby
def increment_revision
  conference.increment!(:revision)
end
```

app/models/concerns/revision_count.rb
```diff
@@ -0,0 +1,21 @@
...
+                   venue.conference
+                 else
+                   is_a?(Conference) ? self : program.conference
+                 end
+    conference.reload.increment(:revision)
+    conference.update_column(:revision, conference.revision)
+  rescue
```
lagartoflojo: Why do we need to rescue?

db/migrate/20170108053041_add_default_to_revision_in_conference.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultToRevisionInConference < ActiveRecord::Migration
+  def change
+    change_column :conferences, :revision, :integer, default: 0
```
lagartoflojo: I'd add: `null: false`

app/models/concerns/revision_count.rb
```diff
@@ -0,0 +1,21 @@
+require 'active_support/concern'
```
shlok007: I seems to work fine without it.

lagartoflojo: Let's remove it then :)

app/models/concerns/revision_count.rb
lagartoflojo: Ok, so the problem is not with duck typing, which works fine!

The problem is with `#increment!`. Let's first consider that `#increment_revision` gets called on `after_update`. With that in mind, the problem is that `#increment!` ALSO triggers callbacks, so it calls `after_update` and thus `#increment_revision` AGAIN. And again. And again. Which is why you get "stack level too deep".

So to update the `revision`, we need to do it using a method that will NOT trigger callbacks.

db/schema.rb
```diff
@@ -13,20 +13,17 @@
 
 ActiveRecord::Schema.define(version: 20170603095900) do
 
-  # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
-
   create_table "ahoy_events", force: :cascade do |t|
-    t.integer  "visit_id"
+    t.uuid     "visit_id",   limit: 16
```
lagartoflojo: Hmm why are we seeing all these changes in schema.rb?

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,32 @@
       expect{ conference.save }.to change{ EventSchedule.count }.from(2).to(1)
     end
   end
+
+  describe 'revision count' do
```
lagartoflojo: `describe '#revision'`

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,32 @@
...
+
+    let(:track) { create(:track, program: subject.program) }
+    let(:event) { create(:event, program: subject.program, track: track) }
+    let(:venue) { create(:venue, conference: subject) }
+    let(:room)  { create(:room, venue: venue) }
+
+    context 'should increment' do
```
lagartoflojo: Contexts are used when you have different cases, such as:

```ruby
context 'when user is logged in' do #....
context 'when user is logged out' do #....
```

As such, it does not make much sense to use one single `context`, so in this case we can just remove it.

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,32 @@
...
+
+      it 'for change in conference' do
+        subject.title = 'changed'
+        expect{ subject.save }.to change { subject.revision }.by(1)
+      end
+      it 'for change in event' do
```
lagartoflojo: Please put a blank line between `it` blocks :)
(I wonder why we don't have a RuboCop rule for this...)

app/models/concerns/revision_count.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module RevisionCount
+  extend ActiveSupport::Concern
+
+  included do
+    after_update :increment_revision
+  end
+
+  def increment_revision
+    conference.increment!(:revision)
```
lagartoflojo: Ok, so the problem is not with duck typing, which works fine!

The problem is with `#increment!`. Let's first consider that `#increment_revision` gets called on `after_update`. With that in mind, the problem is that `#increment!` ALSO triggers callbacks, so it calls `after_update` and thus `#increment_revision` AGAIN. And again. And again. Which is why you get "stack level too deep".

So to update the `revision`, we need to do it using a method that will NOT trigger callbacks.

db/schema.rb
shlok007: seems like something messed up while rebasing the PR. I'll correct it.

app/models/concerns/revision_count.rb
shlok007: Oh! The way it was before did not trigger the callbacks and that is why we didn't encounter the error at that time.
`update_attribute` or `update_column` will not trigger callbacks so we can maybe leave it the way it was before:
```ruby
conference.reload.increment(:revision)
conference.update_column(:revision, conference.revision)
```

lagartoflojo:
I just realized that we have to remove this line:
https://github.com/openSUSE/osem/blob/master/config/application.rb#L28

shlok007:
@lagartoflojo Do you mean this line which has been removed here: https://github.com/openSUSE/osem/pull/1239/files#diff-b1fe55db50c712fef0673345e5b9c0d9L28

lagartoflojo:
Woop! I missed that 😳

PR #1244, opened 2017-01-10
Adds table header in admin/events#index

AEtherC0r3:


differentreality:
That was supposed to be there, not sure what happened to it! Good catch :+1: 
PS. for next time, a screenshot for such changes would be appreciated.

PR #1245, opened 2017-01-10
Find correct event id in versions

### LINKED ISSUE #1226: Version of vote shows vote ID instead of event ID
In versions#index versions referring to votes try to link to an event based on the version id.

AEtherC0r3:
Fixes #1226

Screenshots:
Before Before
After After


differentreality:
Thanks for the fix! :+1:

PR #1246, opened 2017-01-12
Improvements in the DatabaseCleaner

Ana06:
Move all what is related to the `DatabaseCleaner` to the `database_cleaner.rb` file. Load the seeds only when they have been deleted and not before every test.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -55,13 +55,6 @@ RSpec.configure do |config|
   #     --seed 1234
   config.order = 'random'
 
-  config.around(:each) do |example|
-    DatabaseCleaner.cleaning do
-      Rails.application.load_seed
-      example.run
-    end
-  end
-
   # poltergeist as a underlying mech for Capybara
   Capybara.javascript_driver = :poltergeist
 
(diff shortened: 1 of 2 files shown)


ChrisBr:
LGTM

hennevogel:
truncation != transaction

Ana06:
I know it, but it happens something strange, after truncating the database and loading the seeds and before the first test is executed some data is introduced in the database. I have no idea of where this data comes from.

So this strange data keeps there until a truncation is executed (after a feature). If a test query the information in the tables where strange data is introduced (for example User and Conference) before they are truncated it fails.

I added some puts here that illustrate what happens:

https://travis-ci.org/Ana06/osem/builds/191338450

This only happens in Travis, locally it works.

So the second commit (to truncate the database before each test) is just a quick solution until we find out why is this happening. Because otherwise merging the opened PR seems no to be much safe.

Ana06:
@hennevogel  I know it, but it happens something strange, after truncating the database and loading the seeds and before the first test is executed some data is introduced in the database. I have no idea of where this data comes from.

So this strange data keeps there until a truncation is executed (after a feature). If a test query the information in the tables where strange data is introduced (for example User and Conference) before they are truncated it fails.

I added some puts here that illustrate what happens:

https://travis-ci.org/Ana06/osem/builds/191338450

This only happens in Travis, locally it works.

So the second commit (to truncate the database before each test) is just a quick solution until we find out why is this happening. Because otherwise merging the opened PR seems no to be much safe.

Issue #1220, opened 2016-11-16
Feature request: More granular time slots


kormoc:
Currently the time slots are limited to 15 minute chunks. We would prefer 5 minute chunks, as generally our talks are ~50 minutes and then there's a 10 minute break to setup the room for the next talk. Currently we're forced to set the talks as 45 minutes or 1 hour.

differentreality:
This is done, however I think this should be an option conference-wise not app-wise. @Ana06 @eug3nix  this is not currently happening, right? Meaning we cannot have 5-minute slots for schedule of conf A, and 15-minute slots for conf B?

eug3nix:
@differentreality yes, that's correct, the schedule grid density is currently configured in the .dotenv files, but this can be adjusted of course

differentreality:
This will also be extremely useful and flexible for when we implement the hosted version of OSEM.

kormoc:
Thank you very much @Ana06, @differentreality, @eug3nix, and @JewelSam ! This will make things a lot easier for us to manage for our next conference 🎉

PR #1249, opened 2017-01-17
implements configurable schedule granularity

eug3nix:
this pull request may be related to the https://github.com/openSUSE/osem/issues/1220 
description of changes made:
schedule time slot size now can be adjusted using the OSEM_SCHEDULE_CELL_SIZE configuration directive in dotenv configuration file, internally this is done by defining the EventType::LENGTH_STEP dynamically, some sanity checks for the configuration var values are done in the schedule_parameters.rb initialilzer

For the schedule timeslot sizes less that 15 minutes, the application layout will be switched to the compact mode in which the admin schedule editor UI will use smaller cell sizes.

Thanks,
Eugene/CommandPrompt Inc.

### DIFF
--- a/app/assets/stylesheets/osem-schedule.css.scss
+++ b/app/assets/stylesheets/osem-schedule.css.scss
@@ -16,7 +16,7 @@
 }
 
 .schedule-room-slot{
-  padding: 2px 5px;
+  padding: 2px 3px;
   border: 1px solid #848484;
   height: 58px;
   line-height: 20px;
@@ -29,6 +29,12 @@
   position:relative;
   z-index:1;
   cursor:move;
+}
+
+.schedule-room-slot .schedule-event.compact {
+  margin-top: -23px;
+  margin-left: 20%;
+  width: 80%;
 }
 
 .schedule-event-text{
@@ -167,7 +173,7 @@ td.room{
 
 th.date{
   font-size: 12px;
-  padding: 8px 0px 8px 1% !important;
+  padding: 4px 0px 4px 2px !important;
 }
 
 .speaker-pic{
--- a/app/views/admin/schedules/_day_tab.html.haml
+++ b/app/views/admin/schedules/_day_tab.html.haml
@@ -1,3 +1,7 @@
+- compact_grid = EventType::LENGTH_STEP < 15
+- cells_per_hour = 60 / EventType::LENGTH_STEP
+/ use smaller cell heights for more compact grids
+- cell_height = compact_grid ? 32 : 58
 - date_event_schedules = @event_schedules.select{ |e| e.start_time.to_date.eql? date }
 .row
   - @rooms.each do |room|
@@ -5,14 +9,15 @@
       .room-name
         - room_date_event_schedules = date_event_schedules.select{ |e| e.room == room }
         = room.name
-      - (9*4..18*4).each do |slot|
-        - hour = slot / 4
-        - minutes = (15 * (slot % 4) == 0) ? '00' : 15 * (slot % 4)
+      - (9*cells_per_hour..18*cells_per_hour).each do |slot|
+        - hour = slot / cells_per_hour
+        - minutes = (EventType::LENGTH_STEP * (slot % cells_per_hour)).to_s.rjust(2, '0')
         - time = "#{hour}:#{minutes}"
         .schedule-room-slot{ id: "schedule-room-#{room.guid}-#{hour}-#{minutes}", |
                              room_id: room.id, |
                              hour: time, |
-                             date: date}
+                             date: date, |
+                             style: "height: #{cell_height}px"}
           .div
             = time
           - event_schedules = room_date_event_schedules.select{ |e| (e.start_time.hour.to_s + e.start_time.strftime(':%M')).eql? time }
--- a/app/views/admin/schedules/_event.html.haml
+++ b/app/views/admin/schedules/_event.html.haml
@@ -1,6 +1,9 @@
 - cells_length = event.event_type.length / EventType::LENGTH_STEP
 / this height fits the room cells
-- height = (cells_length * 58) - 23
+- compact_grid = EventType::LENGTH_STEP < 15
+- single_cell_height =  compact_grid ? 32 : 58
+- height = (cells_length * single_cell_height)
+- height -= 23 unless compact_grid
 / subtracting the padding before calculate the number of lines
 - lines = (height - 7) / 23
 - color = event.track.try(:color).present? ? event.track.try(:color) : 'FFFFFF'
@@ -8,7 +11,8 @@
                  id: "event-#{event.id}", |
                  event_id: event.id, |
                  length: cells_length, |
-                 event_schedule_id: event_schedule_id }
+                 event_schedule_id: event_schedule_id, |
+                 class: ('compact' if compact_grid) }
   .schedule-event-text{ style: "-webkit-line-clamp: #{lines}; height: #{lines * 23}px;"}
     %span.schedule-event-delete-button{ onclick: "Schedule.remove(\'event-#{event.id}\');" } X
     = event.title
--- a/app/views/schedules/_carousel.html.haml
+++ b/app/views/schedules/_carousel.html.haml
@@ -1,7 +1,7 @@
-- intervals = number_columns * 60 / EventType::LENGTH_STEP + 1
+- intervals = hrs_per_slide * 60 / EventType::LENGTH_STEP + 1
 - width = 85 / intervals
-- carousel_number = (@conf_period / number_columns.to_f).ceil
-.carousel.slide{ id: "carousel-#{ date }-#{ number_columns }", |
+- carousel_number = (@conf_period / hrs_per_slide.to_f).ceil
+.carousel.slide{ id: "carousel-#{ date }-#{ hrs_per_slide }", |
                      "data-ride" => "carousel", |
                      "data-wrap" => "false", |
                      "data-interval" => "false" }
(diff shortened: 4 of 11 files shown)


Ana06:
The problem of allowing less than 15 minutes in the schedule is that it looks horrible for small devices and it even break the whole schedule.

**10 minutes interval for 300x600 screen**

image

**5 minutes interval for 300x600 screen**

image


That was the reason why we originally decided to keep 15 at the minimum value. So the question is do we want to allow intervals smaller than 15 minutes? and if so how?

One 'easy' thing we could do is to allow to download the schedule in pdf and for the devices we know it is not possible to render the schedule properly with the given interval just only allow to download to pdf. Mobiles are getting bigger and bigger so... 

And then we can try to improve how the schedule is rendered so it is displayed properly for small devices. I come up with the following idea:

image

- Not rendering all the hours in small devices
- Removing the speaker picture in small devices

These two things are not difficult to implement and it improve it a lot. At least it won't be able to break the schedule.

But it would still be horrible in some cases, like 5 minutes events, so then I would adjust the number of hours rendered in every section of the carousel taking into account the device size and the interval. So for example in small devices:

**Currently**
for 5 minutes: 65 minutes
for 15 minutes: 75 minutes
etc
**Desired**
for 5 minutes: 35 minutes
for 15 minutes: 75 minutes
etc

but that could be not that easy to implement.


@ChrisBr @hennevogel @differentreality what do you thing?

app/views/admin/schedules/_event.html.haml
```diff
@@ -1,14 +1,18 @@
 - cells_length = event.event_type.length / EventType::LENGTH_STEP
 / this height fits the room cells
-- height = (cells_length * 58) - 23
+- compact_grid = EventType::LENGTH_STEP < 15
```
Ana06: I wonder if we really need to have two cases or we could just generalise. The problem of having bigger cells is that the admin schedule page gets to long and hard to manage. On the other hand if they are too small you can see only the few first characters of the event title and that makes really difficult to know what event you are scheduling. That was the reason what we keep it as big as it is know.

So maybe making it always compact and showing the event title when you put the mouse over the event could be a solution for both problems. :thinking: 

@hennevogel @ChrisBr @differentreality ideas?

app/views/admin/schedules/_day_tab.html.haml
```diff
@@ -1,18 +1,24 @@
+- cell_duration = EventType::LENGTH_STEP
```
eug3nix: I did that just for the code readability sake (as for some other vars)

app/views/admin/schedules/_day_tab.html.haml
```diff
@@ -1,18 +1,24 @@
...
+- compact_grid = EventType::LENGTH_STEP < 15
+- cells_per_hour = 60 / cell_duration
```
eug3nix: same here, it seems to be more readable to me, but I may adjust it to your preference

app/views/admin/schedules/_day_tab.html.haml
```diff
@@ -1,18 +1,24 @@
...
+      - (SCHEDULE_START_HOUR * cells_per_hour..SCHEDULE_END_HOUR * cells_per_hour).each do |slot|
+        - hour = slot / cells_per_hour
+        - minutes = (cell_duration * (slot % cells_per_hour)).to_s.rjust(2, '0')
         - time = "#{hour}:#{minutes}"
         .schedule-room-slot{ id: "schedule-room-#{room.guid}-#{hour}-#{minutes}", |
                              room_id: room.id, |
                              hour: time, |
-                             date: date}
+                             date: date, |
+                             style: "height: #{cell_height}px"}
```
eug3nix: this also boils down to code readability in the price of one extra var
it is debatable, I would like to know what do others think on this code-style concern

app/views/schedules/_carousel.html.haml
```diff
@@ -1,3 +1,4 @@
+/ number_columns is actually hrs per page/slide
```
eug3nix: will do it

app/views/admin/schedules/_event.html.haml
eug3nix: My intention here is to keep the original, less dense grids unless the end user really wants to go to more granular schedules.
Regarding the mouseover - this does not solve the issue with touch platforms.
Also for the too-narrow columns with unreadable text - this only applies to the events 15min or less, so it works ok for the longer events

app/views/admin/schedules/_day_tab.html.haml
Ana06: I would say `EventType::LENGTH_STEP` is already readable enough. And if not it would be better to rename it than having several variables that are the same thing.

app/views/admin/schedules/_day_tab.html.haml
Ana06: But then you need to go to the definition of all the variables to understand what `cell_height` is, while with that it would be just in one place. But yes, what others thing? @differentreality @ChrisBr ^

differentreality: I think that Line21 looks OK like this because it already has a lot of information. I would however merge L2 and L5, as variable `compact_grid` seems rather redundant, we can have `cell_height` as an one-liner if.

That said, I think the bigger issue for this PR at the moment is not grooming it, but addressing the issue with the screen size. So I would focus on that. If that's solved, we can merge the PR, after we merge it, we can refine the code fast and easily.

Ana06:
@eug3nix can you please add a screenshot of your last commit?  :slightly_smiling_face:

eug3nix:
@Ana06 sure, here you go
the idea is simple, the HH and MM timestamp components are wrapped in span tags, so these will wrap if there is not enough space to fit both of them horizontally 
screenshot-localhost 3000-2017-01-26-13-26-45

app/views/admin/schedules/_event.html.haml
Ana06: ok, then I am fine with this :wink:

app/assets/stylesheets/osem-schedule.css.scss
```diff
@@ -173,7 +173,7 @@ td.room{
 
 th.date{
   font-size: 12px;
-  padding: 8px 0px 8px 1% !important;
+  padding: 4px 0px 4px 2px !important;
```
Ana06: what about changing this only for small devices?

dotenv.example
```diff
@@ -57,3 +57,8 @@ OSEM_SMTP_DOMAIN=""
 
 # Enable the usage of the devise ichain plugin
 OSEM_ICHAIN_ENABLED=false
+
+# Schedule grid parameters, cell size in minutes
+OSEM_SCHEDULE_CELL_SIZE=10
+OSEM_SCHEDULE_START_HOUR = 9
```
Ana06: This variable shouldn't be in that PR right?

dotenv.example
```diff
@@ -57,3 +57,8 @@ OSEM_SMTP_DOMAIN=""
...
+OSEM_SCHEDULE_END_HOUR = 19
```
Ana06: This variable shouldn't be in that PR right?

config/initializers/schedule_parameters.rb
```diff
@@ -0,0 +1,6 @@
+#sanitize the OSEM_SCHEUDLE_CELL_SIZE to be used for EventType::LENGTH_STEP
```
Ana06: do we really need that file?

app/views/admin/schedules/_day_tab.html.haml
```diff
@@ -1,18 +1,23 @@
+- compact_grid = EventType::LENGTH_STEP < 15
```
Ana06: I would remove `compact_grid` and add this code directly in `cell_height`

Ana06:
@eug3nix 

> the idea is simple, the HH and MM timestamp components are wrapped in span tags, so these will wrap if there is not enough space to fit both of them horizontally

Great idea! :tada:  Not sure if for 5 minutes interval it would work in small devices (300x600) properly, but for 10 minutes it works perfectly. So I would propose that we merge this and then we can consider if it is worthwhile to continue improving this.

dotenv.example
eug3nix: yep, that one was actually remove, but somehow sneaked in back to the PR branch while I am worked on it. will get that fixed

app/assets/stylesheets/osem-schedule.css.scss
eug3nix: actually, it makes sense for the full-sized grids too, especially for the denser grids.
And in general, it seems to be more obvious to the end-user where the timestamp belongs in the cell, since the time in timestamp is where the cell starts (e.g its left border), when the timestamp is aligned to the center it looks more confusing IMO, please see our schedule with the left-aligned timestamps, we have exactly this style applied http://www.pgconf.us/conferences/2017/schedule

config/initializers/schedule_parameters.rb
eug3nix: I think we do, since it makes the system a bit more foolproof, we should not trust the user supplied data.

Ana06: But could it be in the EventType model instead of in this new file?

We are modifying there the `SCHEDULE_CELL_SIZE` variable anyway: https://github.com/openSUSE/osem/pull/1249/files#diff-490f3d30636bbfbfaf4b4489e64bd430R20

eug3nix: ok, here is my position on that and why I did that:
1 - there should be some sanity checks since we don't know, if the one who will be setting up the software has spent some time reading the docs etc. there IS a room for human error. Otherwise it may be done in the model itself
2 - all config stuff initialization should be done in the initializers, that's why they are for, this is a common way of doing it. And I am not sure if it is a responsibility of the model to do such sanity checks

Ana06:
Can u remove 29e1aa1194521a6962420b7c9a8611d9f045f3fc? Apart from that LGTM, @differentreality can u take a look to see if we can merge it?

differentreality:
@Ana06 there is some scheduling magic that you know of much better than I do. So if you think that's what we need, then it's all good :)

I would, though, appreciate some comments about what an organizer is now supposed to do for scheduling. We have an initializer and we also have an ENV variable. 
If the organizer does not touch anything, default will be intervals of 15 minutes? 
And organizer can choose something else in the ENV variable? 
We should let people know how to use this.

eug3nix:
@differentreality yes, the organizer can set the schedule slot time in the environment, or, in our case thru the .env files. the OSEM_SCHEDULE_CELL_SIZE, it will default to the 15 minutes period in two cases:
1 - the value for OSEM_SCHEDULE_CELL_SIZE is not set at all (unlikely, since we have this string in .env sample file)
2 - the value for OSEM_SCHEDULE_CELL_SIZE is invalid (not an integer or it is zero or it is not divisor of 60))

I've added the comments in dotenv.example file for this new conf directive, it is also mentioned in app.json file.

eug3nix:
@Ana06 I've removed the merge commit from history by rebasing the OSEM's master branch to my pull request branch locally, then force pushed. So it seems that we are good to go :)

Ana06:
@eug3nix I think you should also add the variables here https://github.com/openSUSE/osem/blob/master/INSTALL.md#configure

PR #1251, opened 2017-01-18
Fix bug in the schedule

Ana06:
As @eug3nix pointed out there is a bug in the schedule: events that should be rendered in more than one carousel section are only rendered in the first one.


differentreality:
Are there before/after screenshots for this?
How can we reproduce?

Ana06:
Well, I could add a screenshot but it would be an empty schedule, because the problem is that the event is not shown :joy: 

**To reproduce:** Schedule an event at 9:45 which lasts one hour. Then look at the schedule with small screen size (300x600 for example) and you will see that the event apears from 9:45 to 10:15 in first part of the carousel but not from 10:00 to 10:45 in the second part of the carousel.

differentreality:
:+1:

PR #1253, opened 2017-01-20
Rebases #1113

differentreality:
Rebases #1113

### DIFF
--- a/app/models/event_schedule.rb
+++ b/app/models/event_schedule.rb
@@ -31,6 +31,10 @@ class EventSchedule < ActiveRecord::Base
     room.event_schedules.where(start_time: start_time, schedule: schedule).where.not(id: id)
   end
 
+  def replacement?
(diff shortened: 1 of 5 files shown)


app/models/event_schedule.rb
```diff
@@ -31,6 +31,10 @@ def intersecting_events
     room.event_schedules.where(start_time: start_time, schedule: schedule).where.not(id: id)
   end
 
+  def replacement?
+    event.state == 'confirmed' && (!intersecting_events.canceled.empty? || !intersecting_events.withdrawn.empty?)
```
Ana06: This should be the same as:
`event.state == 'confirmed' && intersecting_events.present?`

right? because we shouldn't have several confirmed events scheduled at the same time.

differentreality: @Ana06 yes, that's right.
I like how it is written though, because it makes it clear that we are talking about events that are cancelled or withdrawn, and I think it will protect the result we want to achieve, in case any magic happens in the DB. What do you think?

Ana06: ok! :wink:

PR #1254, opened 2017-01-21
Updated devise to 4.2.0

shlok007:
Devise changelog: https://github.com/plataformatec/devise/blob/master/CHANGELOG.md


differentreality:
I think we should get this in. What do others think?

differentreality:
:+1: Thanks @shlok007 :tada:

PR #1255, opened 2017-01-23
Display paid and unpaid tickets in admin/tickets#show

### LINKED ISSUE #1190: admin/tickets#show
The paid column shows true even if not all of the tickets are paid (some paid, some not). That's misleading and a problem when it comes to the financial side of the tickets.

If a user purchased 2 tickets in 2 different occasions, and only 1 ticket is paid, an organizer should know how many tickets are paid, and how many are not.

shlok007:
Closes #1190 .

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -19,8 +19,10 @@ class Ticket < ActiveRecord::Base
     buyers.include?(user)
   end
 
-  def paid?(user)
-    ticket_purchases.paid.by_user(user).present?
+  def tickets_paid(user)
+    paid_tickets    = quantity_bought_by(user, paid: true)
+    unpaid_tickets  = quantity_bought_by(user, paid: false)
+    "#{paid_tickets}/#{paid_tickets+unpaid_tickets}"
   end
 
   def quantity_bought_by(user, paid: false)
--- a/app/views/admin/tickets/show.html.haml
+++ b/app/views/admin/tickets/show.html.haml
@@ -33,5 +33,5 @@
             %td
               = buyer.affiliation
             %td
-              = @ticket.paid?(buyer)
+              = @ticket.tickets_paid(buyer)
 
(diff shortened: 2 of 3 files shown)


app/views/admin/tickets/show.html.haml
```diff
@@ -19,6 +19,7 @@
         %th E-Mail
         %th Affiliation
         %th Paid
+        %th Unpaid
```
Ana06: I think that we don't need an unpaid column. I would render in the paid one how many tickets out of the total are paid.

I mean, if a user bought 3 tickets and paid 2 render: `2/3`

app/views/admin/tickets/show.html.haml
```diff
@@ -33,5 +34,7 @@
             %td
               = buyer.affiliation
             %td
-              = @ticket.paid?(buyer)
```
Ana06: the `paid? `method is only used here, so you can remove it form the ticket model. You will also have to modify the tests.

app/models/ticket.rb
```diff
@@ -23,6 +23,14 @@ def paid?(user)
     ticket_purchases.paid.by_user(user).present?
   end
 
+  def no_of_paid(user)
```
Ana06: I would change this method to return an string with how many tickets out of the total are paid. I mean "2/3" if a user bought 3 tickets and paid 2.

And I am not sure if the name of the method is clear, what about `tickets_paid(user)`?

app/views/admin/tickets/show.html.haml
shlok007: Thank you. I'll update it. 🙂

Ana06:
LGTM :wink: 

Can you please squash the two commits into one: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one ?

PR #1257, opened 2017-01-23
implements adding the proposal by admin users

eug3nix:
implements manual addition of events by admin users
adds some RSPEC tests for the new functionality
the new functionality is available in the Events section of Conference admin (the Add Event button near the pdf/xls/csv export)

the form allows specifying the submitter and speaker for the event being created
the newly created events are confirmed from the start, so there is no need to manually approve those

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -305,12 +305,12 @@ class Event < ActiveRecord::Base
     if validate_owners
       errors.add(:speaker_id, "can't be blank!") unless self.speaker_id.present?
       if self.speaker_id.present?
-        errors.add(:speaker_id, "user should exist!") unless User.where(id: self.speaker_id).take
+        errors.add(:speaker_id, 'user should exist!') unless User.where(id: self.speaker_id).take
       end
 
       errors.add(:submitter_id, "can't be blank!") unless self.submitter_id.present?
       if self.submitter_id.present?
-        errors.add(:submitter_id, "user should exist!") unless User.where(id: self.submitter_id).take
+        errors.add(:submitter_id, 'user should exist!') unless User.where(id: self.submitter_id).take
       end
     end
   end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,4 +1,8 @@
 class Event < ActiveRecord::Base
+  attr_accessor :submitter_id
+  attr_accessor :speaker_id
+  attr_accessor :validate_owners
+
   include ActiveRecord::Transitions
   has_paper_trail on: [:create, :update], ignore: [:updated_at, :guid, :week], meta: { conference_id: :conference_id }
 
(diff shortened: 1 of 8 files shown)


app/views/admin/events/_form.html.haml
```diff
@@ -0,0 +1,77 @@
+.row
+  .col-md-12
+    .page-header
+      %h1
+        New Event
+.row
+  .col-md-12
+    = semantic_form_for(@event, url: @url) do |f|
+      = f.inputs name: 'Event Information' do
```
differentreality: Couldn't we use the existing proposals/proposal_form with minor adjustments and add speaker and submitter at the bottom in the admin/events/_form ?

app/views/admin/events/index.html.haml
```diff
@@ -4,7 +4,10 @@
       %h1
         Events
         = "(#{@events.length})" if @events.any?
-        .btn-group.pull-right
+
+        .pull-right
+          - if can? :create, Event
+            =link_to 'Add Event', new_admin_conference_program_event_path(@conference.short_title), :class => 'button btn btn-default btn-info'
```
differentreality: btn-success for uniformity purposes with the rest of the app.

app/models/event.rb
```diff
@@ -294,4 +300,18 @@ def before_end_of_conference
   def conference_id
     program.conference_id
   end
+
+  def submitter_and_speaker_present
+    if validate_owners
+      errors.add(:speaker_id, "can't be blank!") unless self.speaker_id.present?
```
differentreality: `self` is redundant

app/views/admin/events/_form.html.haml
eug3nix: I'll think on how to make it more DRY,
another thing - should be have submitter/speaker fields at the very bottom, these are not 2nd grade fields in this context, that's why I put them at the beginning of the form

app/views/admin/events/index.html.haml
eug3nix: I would like to discuss it more, my position is:
In comparison to the xls/pdf/csv list export this is the primary action for this page, so it should be visually separated from other actions. Here is some reference http://patternry.com/p=primary-secondary-actions/
If all buttons in green is the standard for the whole project UI - I can make it the regular btn-success
But in general the UI with the clearly separated actions is more usable per many studies. Shouldnt we try to improve the app in this aspect?

app/models/event.rb
eug3nix: oh, yep, my bad, will remove that

app/views/admin/events/index.html.haml
differentreality: I got your point, it is just that we don't have an info button anywhere else. Yes, we should decide on a specific course of action for multiple buttons (perhaps not in this particular PR but in a new issue). Unfortunately, no, we don't have UI guidelines. In terms of guidelines (not UI), we got rubocop for styling and we use new ruby syntax for the code.

app/views/admin/events/_form.html.haml
differentreality: Yes, they could be on the beginning, absolutely. It's probably even gonna look better if we add the `= semantic_form_for` in admin/events/_form and then add the speaker/submitter fields and then show the rest of the form from `= render partial: 'proposals/proposal_form', locals: { f: f }`

differentreality:
Will all of this PR be included in #1278 ?

eug3nix:
@differentreality not exactly (from the code standpoint), since the #1278 also has the manual event management functionality but is implemented more clean way, so this PR should be treated as obsolete.

PR #1258, opened 2017-01-24
implement user creation by admin

eug3nix:
the new Add User functionality can be accessed using the *Add User* button on admin/users URI
the manually created users are in confirmed state and thus does not need any extra confirmation from the user.

best regards,
Eugene/CommandPrompt Inc.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -6,6 +6,17 @@ module Admin
       @user = User.new
     end
 
+    def create
+      @user = User.new(user_params)
+      @user.skip_confirmation!
+      if @user.save
+        redirect_to admin_users_path, notice: 'User successfully created.'
+      else
+        flash.now[:error] = "Creating User failed: #{@user.errors.full_messages.join('. ')}."
+        render :new
+      end
+    end
+
     def index
       @users = User.all
     end
@@ -50,8 +61,9 @@ module Admin
     private
 
     def user_params
-      params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
-                                   :tshirt, :mobile, :volunteer_experience, :languages, :to_confirm, role_ids: [])
+      params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin,
+                                   :username, :login, :is_disabled, :tshirt, :mobile, :volunteer_experience,
+                                   :languages, :to_confirm, :password, role_ids: [])
     end
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -119,6 +119,9 @@ class Ability
     # for admins
     can :manage, :all if user.is_admin
 
+    # even admin cannot create new users with ICHAIN enabled
+    cannot [:new, :create], User if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
+
     cannot :revert_object, PaperTrail::Version do |version|
       (version.event == 'create' && %w(Conference User Event).include?(version.item_type))
     end
--- a/app/views/admin/users/_form.html.haml
+++ b/app/views/admin/users/_form.html.haml
@@ -1,25 +1,28 @@
 = semantic_form_for [:admin, @user] do |f|
   = f.inputs 'Basic Information' do
-    .pull-right
-      %b
-        Confirmed?
-      - if can? :toggle_confirmation, @user
-        = check_box_tag @user.id, @user.id, @user.confirmed?, 
-          method: :patch, 
-          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
-          class: 'switch-checkbox', 
-          readonly: false,
-          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
-      - else
-        = check_box_tag @user.id, @user.id, @user.confirmed?, 
-          method: :patch, 
-          url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=", 
-          class: 'switch-checkbox', 
-          readonly: true,
-          data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' } 
+    - unless @user.new_record?
+      .pull-right
+        %b
+          Confirmed?
+        - if can? :toggle_confirmation, @user
+          = check_box_tag @user.id, @user.id, @user.confirmed?,
+            method: :patch,
+            url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=",
+            class: 'switch-checkbox',
+            readonly: false,
+            data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
+        - else
+          = check_box_tag @user.id, @user.id, @user.confirmed?,
+            method: :patch,
+            url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=",
+            class: 'switch-checkbox',
+            readonly: true,
+            data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
     = f.input :is_admin, hint: 'An admin can create a new conference, manage users and make other users admins.'
     = f.input :name, as: :string
+    = f.input :username, :as => :string if @user.new_record?
     = f.input :email
+    = f.input :password if @user.new_record?
     = f.input :affiliation, as: :string
     = f.input :biography, input_html: { rows: 10, data: { provide: 'markdown-editable' } },
       hint: markdown_hint
--- a/app/views/admin/users/index.html.haml
+++ b/app/views/admin/users/index.html.haml
@@ -1,10 +1,13 @@
 .row
   .col-md-12
     .page-header
-      %h2
+      %h1
         Users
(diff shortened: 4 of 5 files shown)


hennevogel:
Hey @eug3nix thank you for your contribution! 💐

eug3nix:
Hey, @hennevogel do you mean the ICHAIN auth setting of OSEM? 
Can I use the ENV['OSEM_ICHAIN_ENABLED'] == 'true' to conditionally disable the user management ?
Thanks.

hennevogel:
yes you can :-)

eug3nix:
here you go, @hennevogel I've adjusted the code per your recomendations and added some tests for such cases

app/controllers/admin/users_controller.rb
```diff
@@ -50,8 +66,14 @@ def edit; end
     private
 
     def user_params
-      params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
+      # no password attribute in ICHAIN mode
+      if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
+        params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
```
hennevogel: No need to do this all again. You can [merge](http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-merge) `:user` into the params when `ENV['OSEM_ICHAIN_ENABLED']` is set.

app/controllers/admin/users_controller.rb
```diff
@@ -50,8 +66,11 @@ def edit; end
     private
 
     def user_params
-      params.require(:user).permit(:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
-                                   :tshirt, :mobile, :volunteer_experience, :languages, :to_confirm, role_ids: [])
+      permit_base = [:email, :name, :email_public, :biography, :nickname, :affiliation, :is_admin, :username, :login, :is_disabled,
+                     :tshirt, :mobile, :volunteer_experience, :languages, :to_confirm, role_ids: []]
+      # no password for ICHAIN-enabled mode
+      permit_base << :password unless ENV['OSEM_ICHAIN_ENABLED'] == 'true'
```
ChrisBr: As this is the admin controller I would just add the ``:password`` to the permit without the conditional.

app/views/admin/users/index.html.haml
```diff
@@ -1,10 +1,14 @@
 .row
   .col-md-12
     .page-header
-      %h2
+      %h1
         Users
         - if @users
           = "(#{@users.length})"
+
+        - if can? :create, User and ENV['OSEM_ICHAIN_ENABLED'] != 'true'
```
ChrisBr: ``
if can? :create, User && ENV['OSEM_ICHAIN_ENABLED'] != 'true'
``

app/views/admin/users/index.html.haml
```diff
@@ -1,10 +1,14 @@
...
+          .pull-right
+            =link_to 'Add User', new_admin_user_path, :class => 'button btn btn-default btn-info'
```
ChrisBr:
```
= link_to 'Add User', new_admin_user_path, class: 'button btn btn-default btn-info' 
```

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -50,4 +50,66 @@
...
+        expect(flash[:error]).to match("Creating User` failed: #{user.errors.full_messages.join('. ')}.")
+      end
+
+      it 'does not create new user' do
+        expect do
+          post :create, user: attributes_for(:user)
+        end.not_to change{ Event.count }
+      end
+    end
+    context 'OSEM_ICHAIN_ENABLED is true' do
```
ChrisBr: Line break missing

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -50,4 +50,66 @@
...
+      before do
+        stub_const('ENV', ENV.to_hash.merge('OSEM_ICHAIN_ENABLED' => 'true'))
+        # allow_any_instance_of(User).to receive(:save).and_return(false)
```
ChrisBr: Please remove the commented code

app/views/admin/users/index.html.haml
eug3nix: @ChrisBr we had a discussion on that with @differentreality some time ago, please see this PR https://github.com/openSUSE/osem/pull/1257
long story short - it is nice to have a visually separated primary and secondary actions from the UX perspective, I used the same approach in the https://github.com/openSUSE/osem/pull/1278
If you guys strongly against that, I would not argue and change the btn to the default green, but I think it is a good step toward more user-friendlt UI

ChrisBr:
LGTM! Restarted travis, hopefully it will resolve the timeout failure ...

app/views/admin/users/index.html.haml
```diff
@@ -1,10 +1,14 @@
 .row
   .col-md-12
     .page-header
-      %h2
+      %h1
         Users
         - if @users
           = "(#{@users.length})"
+
+        - if can? :create, User && ENV['OSEM_ICHAIN_ENABLED'] != 'true'
```
differentreality: We should add the rule in abilities

cannot [:new, :create], User if ENV['OSEM_ICHAIN_ENABLED'] == 'true'

differentreality:
@eug3nix can you please rebase?

app/views/admin/users/index.html.haml
eug3nix: @differentreality commits squashed, I've also adjusted the code to get rid of the flash vs flash.now problem that we discussed recently (ref #1429)

app/views/admin/users/_form.html.haml
```diff
@@ -1,26 +1,29 @@
...
+        - else
+          = check_box_tag @user.id, @user.id, @user.confirmed?,
+            method: :patch,
+            url: "/admin/users/#{@user.id}/toggle_confirmation?user[to_confirm]=",
+            class: 'switch-checkbox',
+            readonly: true,
+            data: { size: 'small', on_color: 'success', off_color: 'warning', on_text: 'Yes', off_text: 'No' }
     = f.input :is_admin, hint: 'An admin can create a new conference, manage users and make other users admins.'
-    = f.input :name, as: :string
+    = f.input :name, :as => :string
```
differentreality: We should stick to the new ruby syntax

app/controllers/admin/users_controller.rb
```diff
@@ -6,6 +6,22 @@ def new
...
 
+    def create
+      # there is no usual user creation process when ICHAIN auth is enabled
+      if ENV['OSEM_ICHAIN_ENABLED'] != 'true'
+        @user = User.new(user_params)
+        @user.skip_confirmation!
+        if @user.save
+          redirect_to admin_users_path, notice: 'User successfully created.'
+        else
+          flash.now[:error] = "Creating User` failed: #{@user.errors.full_messages.join('. ')}."
```
differentreality: There is a ` after User, not sure what it is doing there

differentreality:
Apart from a few instances of old ruby syntax this looks good, @hennevogel @ChrisBr do you also want to take a look?

app/controllers/admin/users_controller.rb
```diff
@@ -6,6 +6,22 @@ def new
...
+        @user = User.new(user_params)
+        @user.skip_confirmation!
+        if @user.save
+          redirect_to admin_users_path, notice: 'User successfully created.'
+        else
+          flash.now[:error] = "Creating User failed: #{@user.errors.full_messages.join('. ')}."
+          render :new
+        end
+      else
+        redirect_to admin_users_path, error: 'Cannot create user with ICHAIN auth enabled.'
```
differentreality: Cancancan takes care of this with https://github.com/commandprompt/osem-pg/blob/f9a816eed44625683746192c43acd8fc53f92999/app/controllers/admin/users_controller.rb#L3
So this if statement should not be necessary here, as we should only be able to access the action `create`, if the abilities allow that.

If the authorization fails, cancancan raises an error https://github.com/openSUSE/osem/blob/master/app/controllers/application_controller.rb#L46

app/views/admin/users/index.html.haml
differentreality: Please make the addition in the models/ability.rb then you don't need `&& ENV['OSEM_ICHAIN_ENABLED'] != 'true'` here or the if statement in the controller admin/users#create.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -50,4 +50,66 @@
       end
     end
   end
+  describe 'GET #new' do
+    it 'populates an array of users' do
```
differentreality: not a very accurate description

differentreality:
@eug3nix please rebase

app/models/ability.rb
```diff
@@ -119,6 +119,9 @@ def signed_in_with_roles(user)
     # for admins
     can :manage, :all if user.is_admin
 
+    # even admin cannot create new users with ICHAIN enabled
+    cannot [:new, :create], user if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
```
differentreality: This needs to be the `User` instead of `user`, as in the User model.

differentreality:
@eug3nix can you please rebase?

differentreality:
Thanks @eug3nix :smile: 
Let's get it in :+1: :tada:

PR #1261, opened 2017-02-01
Ability to track resources

### LINKED ISSUE #1224: Track resources
As part of event planning, it would be great to track resources that are required for the event.  Catering, Staff Support, Equipment like audio video, computer & network access, badges, security, important points of contacts, tables, chairs among other needed furnishing etc

shlok007:
Closes #1224 .

- Resources#index
Quantity used can directly be changed from index.

resource_index

- Resource#show
resource_show

- Resource#new

new_resource


app/views/admin/resources/index.html.haml
```diff
@@ -0,0 +1,41 @@
...
+      %h1 Resources
+      %p.text-muted
+        Manage resources in conference
+- if @conference.resources.any?
+  .row
+    .col-md-12
+      %table.table.table-hover.datatable
+        %thead
+          %th Name
+          %th Total Quantity
```
Ana06: And I would put the Left one before the used one

db/migrate/20170129075432_create_resources_table.rb
```diff
@@ -0,0 +1,11 @@
+class CreateResourcesTable < ActiveRecord::Migration
+  def change
+    create_table :resources do |t|
+      t.string  :name
+      t.string  :description
```
Ana06: I would say description should be a text and not a string as it could be longer.

app/models/resource.rb
```diff
@@ -0,0 +1,7 @@
+class Resource < ActiveRecord::Base
```
Ana06: I would add a validation to check than `used` is less than `quantity`

config/routes.rb
```diff
@@ -120,6 +121,7 @@
     # TODO: change conference_registrations to singular resource
     resource :conference_registration, path: 'register'
     resources :tickets, only: [:index]
+    resources :resources, only: [:index]
```
Ana06: Why do we want this route? :confused:

app/views/admin/resources/_form.html.haml
```diff
@@ -0,0 +1,16 @@
...
+  .col-md-12
+    .page-header
+      %h1
+        - if @resource.new_record?
+          New
+        = @resource.name
+        Resource
+.row
+  .col-md-8
+    = semantic_form_for(@resource, :url => (@resource.new_record? ? admin_conference_resources_path : admin_conference_resource_path(@conference.short_title, @resource))) do |f|
```
Ana06: used should also be in this form as you could also want to edit it while being in the edit page. :slightly_smiling_face:

app/views/admin/resources/index.html.haml
```diff
@@ -0,0 +1,41 @@
...
+          %th Used
+          %th Left( Used/Total )
```
shlok007: oops! :stuck_out_tongue:

config/routes.rb
shlok007: Sorry. We don't. :+1:

db/migrate/20170129075432_create_resources_table.rb
shlok007: Sure. I'll make the required changes :slightly_smiling_face:

shlok007:
@Ana06 Thank you for the review :yellow_heart: 

I have made the requested changes.

differentreality:
@Ana06 yes, personally I do like the functionality. However changing the quantity of resources should be simpler/easier. If you are in a hectic registration desk and doing registrations and giving out t-shirts, you don't have time to hit edit, change the quantity in a field, hit save. So perhaps there should be a show page with a big button or something. If we want this functionality in the app, we should give some love to the UI, so that it is not just a list of things, but has a functionality that gives some added value to the organizers.

app/views/admin/resources/index.html.haml
```diff
@@ -0,0 +1,38 @@
...
+          - @conference.resources.each do |resource|
+            %tr
+              =semantic_form_for(resource, url: admin_conference_resource_path(@conference.short_title, resource.id), html: { method: :put }) do |f|
+                %td
+                  = link_to(admin_conference_resource_path(@conference.short_title, resource.id)) do
+                    = resource.name
+                %td
+                  = resource_quantity_left(resource)
+                %td.col-sm-1.col-md-1
+                  = f.input :used, label: false
```
Ana06: use `number_field` instead of `input` as this is always a number

Ana06:
@differentreality 

> @Ana06 yes, personally I do like the functionality. However changing the quantity of resources should be simpler/easier. If you are in a hectic registration desk and doing registrations and giving out t-shirts, you don't have time to hit edit, change the quantity in a field, hit save. So perhaps there should be a show page with a big button or something. If we want this functionality in the app, we should give some love to the UI, so that it is not just a list of things, but has a functionality that gives some added value to the organizers.

Right, do we want to this in this PR or merge this and then improve it? Because it is already a quite big PR.

differentreality:
@Ana06 up to you! (I haven't reviewed the PR, just looked at the screenshots)

app/views/admin/resources/index.html.haml
shlok007: I used `input` keeping in mind that we have used `input` throughout the app even if the input required is a number. The input field does not accept alphabets or characters as input here because of [`formtastic-bootstrap`](https://github.com/mjbellantoni/formtastic-bootstrap) I guess. Changing it to `number_field` would remove the styling as well.

app/helpers/application_helper.rb
```diff
@@ -536,4 +536,8 @@ def replacement_event_notice(event_schedule)
       end
     end
   end
+
+  def resource_quantity_left(resource)
```
Ana06: what about changing the name to `quantity_left_of(resource)`

app/controllers/admin/resources_controller.rb
```diff
@@ -0,0 +1,52 @@
+module Admin
+  class ResourcesController < ApplicationController
+    load_and_authorize_resource :conference, find_by: :short_title
+    load_and_authorize_resource :resource, only: [:show, :edit, :update, :destroy]
```
Ana06: who should be able to create, modify, detroy and see resources? Currently only admins would be allowed right? I would say that organizers should be able to manage all this too. And users signed in with info desk role and volunteers coordinator role should be able to see and maybe update them too.

differentreality: Yeah abilities should be added. Organizer, reg_desk too for sure.

app/helpers/application_helper.rb
shlok007: Sure :)

app/controllers/admin/resources_controller.rb
shlok007: Yes, I have changed the permissions now to exactly as said . :slightly_smiling_face:

app/models/ability.rb
```diff
@@ -244,6 +247,7 @@ def signed_in_with_volunteers_coordinator_role(user)
     # ids of all the conferences for which the user has the 'volunteers_coordinator' role
     conf_ids_for_volunteers_coordinator = Conference.with_role(:volunteers_coordinator, user).pluck(:id)
 
+    can [:index, :show, :update], Resource, conference_id: conf_ids_for_coordinator
```
differentreality: And that's why you should first write some tests for the new abilities you are introducing :)

shlok007: Lesson learned :sweat_smile:

app/models/ability.rb
```diff
@@ -187,6 +188,7 @@ def signed_in_with_cfp_role(user)
     # ids of all the conferences for which the user has the 'cfp' role
     conf_ids_for_cfp = Conference.with_role(:cfp, user).pluck(:id)
 
+    can [:index, :show, :update], Resource, conference_id: conf_ids_for_cfp
```
Ana06: `:index` and `:show` can be both replaced by `:read`: https://github.com/ryanb/cancan/wiki/Action-Aliases

There are some more places were this appear, please change it there too 🙂

app/models/resource.rb
```diff
@@ -0,0 +1,10 @@
...
+  belongs_to :conference
+  validate :used_less_than_quantity
+
+  private
+
+  def used_less_than_quantity
+    errors.add(:used, 'cannot be less than total quantity') unless used <= quantity
```
Ana06: The error message is wrong, it should be _can not be *higher* than total quantity_ :wink:

app/models/ability.rb
differentreality: I would prefer to be explicit, ie use the actual actions instead of aliases. Aliases might change or we don't really know what they correspond to immediately.

shlok007:
Do let me know when its ready to merge and i'll squash the commits. :slightly_smiling_face:

Ana06:
@shlok007 LGTM :tada: please squash your commits :wink:

app/views/admin/resources/show.html.haml
```diff
@@ -0,0 +1,26 @@
...
+      %tr
+        %td
+          %b Total Quantity
+        %td
+          = @resource.quantity
+      %tr
+        %td
+          %b Quantity left
+        %td
+          = resource_quantity_left(@resource)
```
differentreality: Where is this defined?

differentreality:
Index is not really the place to have an input box. This should be addressed some other way, perhaps with buttons Add/Remove to increase/reduce quantity.

app/views/admin/resources/show.html.haml
Ana06: in the helper

Ana06:
@differentreality 
> Index is not really the place to have an input box. This should be addressed some other way, perhaps with buttons Add/Remove to increase/reduce quantity.

Agree, but this PR is already big and how to make this usable would requiere a little bit of discussion, so I would merge this first and then discussing ways to improve it. Would you by fine with this?

app/models/ability.rb
Ana06: Well it is more readable. And `update` is also an alias, as it really means `update` + `edit`. And all this is done on purpose in the gem.

Anyway, it is not that important, I'm fine with this as it is. It is already like that in other parts of the ability file.

app/views/admin/resources/show.html.haml
shlok007: the helper method should be `quantity_left_of` . I probably messed up this file while rebasing. :sweat:

shlok007:
Updated squashed and double checked for any other unwanted changes :slightly_smiling_face:

app/models/ability.rb
differentreality: @Ana06 as far as I am concerned, read is confusing, and currently our ability file does not use it anywhere (that was done on purpose, there have been comments about it in the past - which I cannot dig up though).

We also don't use update as an alias, meaning that, when we imply edit action, we write it as well. 

CanCanCan at some point had aliased different actions to `read`, and there were comments/issues about that being changed. If that happens again, I am afraid we might not catch it in time. 

I would hugely prefer that we are explicit with our abilities, rather than implicit (ie using aliases), and I strongly believe it is very important. Abilities define our authorization, if we miss something there it would mean that it gives improper access to people in our app.

differentreality:
I agree that perhaps it makes sense for both you and @shlok007 that we merge this with basic usage, and decide later how to beautify it.
But still input box in index is **not** the standard practice and we should not have it there. For the time being I would suggest we can remove the input box (and the save button) all together.  We merge it, and we address its usability in another PR.

Ana06:
@differentreality 

> But still input box in index is not the standard practice and we should not have it there. For the time being I would suggest we can remove the input box (and the save button) all together. We merge it, and we address its usability in another PR.

Ok, then @shlok007  please change this  🙂

shlok007:
> But still input box in index is not the standard practice and we should not have it there. For the time being I would suggest we can remove the input box (and the save button) all together. We merge it, and we address its usability in another PR.

Sounds good. Input field is now removed:slightly_smiling_face:

differentreality:
Nice work :)
It would be useful if you could change the screenshot of index in the first comment.

app/models/ability.rb
Ana06: What I meant was that `update` include `edit` even if you don't add it explicitly, as CanCan has an alias for it. But I am perfectly fine with leaving it as it is now. :slightly_smiling_face:

PR #1265, opened 2017-02-01
[WIP] Adds domain support to OSEM

captn3m0:
- This is a very basic support. All it affects is the listings
  on the homepage. Every conference can set a domain, and as
  long as the request host matches this domain, the conference
  will be shown on the list
- If a conference does not have a domain set (which will be the
  case for older conferences), such conferences are shown
  always
- Conference creation and edits now have an optional domain field
- Also adds corresponding migrations (+index)

The use case is for the osem instance powering hackbeach. We want to re-use the same instance for hillhacks, and maintain the user accounts. This will let both of the services use the same rails instance+db, while showing a different set of conferences at `osem.{hillhacks|hackbeach}.in`.

### DIFF
--- a/app/controllers/admin/conferences_controller.rb
+++ b/app/controllers/admin/conferences_controller.rb
@@ -176,7 +176,7 @@ module Admin
     private
 
     def conference_params
-      params.require(:conference).permit(:title, :short_title, :description, :timezone,
+      params.require(:conference).permit(:title, :short_title, :description, :timezone, :domain,
                                          :start_date, :end_date, :rooms_attributes, :tracks_attributes,
                                          :tickets_attributes, :event_types_attributes,
                                          :picture, :picture_cache, :questions_attributes,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,7 +37,8 @@ class ApplicationController < ActionController::Base
   end
 
   def get_conferences
-    @conferences =Conference.all
+    domain = request.host
+    @conferences = Conference.get_conferences_to_list(domain, 0)
   end
 
   def current_ability
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -5,7 +5,8 @@ class ConferencesController < ApplicationController
   load_resource :program, through: :conference, singleton: true, except: :index
 
   def index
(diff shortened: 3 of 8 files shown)


captn3m0:
One issue is with the OSEM_NAME variable, which remains constant across all conferences. We are considering patching it via a `X-Osem-Name` header if present in the request.

differentreality:
@captn3m0 I don't understand why you needs this. 

Old conferences are not shown, unless you click 'Other Conferences'

captn3m0:
We want to display different conferences on the following links:

https://osem.hackbeach.in/
https://osem.hillhacks.in/

Both of these share the same osem instance. This lets us have muti-tenancy to some degree. When you visit the hillhacks site, you get the hillhacks conferences only (even after clicking "other conferences").

differentreality:
@captn3m0 I see what you are saying, but I am still not convinced we should mess up the code so much for this reason, specially since old conferences do not show up by default.

Do others think this is useful?

captn3m0:
As it stands, I agree, this is a very special case.

Would this be more useful, if it enabled full-fledged multi-tenancy? Create organizations that can manage their own conferences, with conferences only for your org being shown.

differentreality:
@captn3m0 Can you elaborate on your idea? I am not sure I understand. Also are you talking about what attendees see in the home page or admin side? We don't need organizations that manage conferences, because we have organizers (users with role organizer) that manage conferences.

captn3m0:
What I meant by "organization" was basically "conference-collective". Community might be a better word.

This would let multiple communities share the same osem instance (with the same users) while maintaining their own set of conferences.

So I might setup a osem instance that can cater to multiple communities, and each of those can have their own organizers to manage their conferences.

differentreality:
Yeah we plan to implement this actually as a GSoC project, @lagartoflojo knows more about it, but for now I think we can close this pr.

lagartoflojo:
Hi @captn3m0, thanks for your contribution.
As @differentreality said, we plan on creating a [hosted version of OSEM as a GSoC project](https://github.com/openSUSE/mentoring/issues/72), which would solve the "conference collective" use case that you described. Part of that project includes custom domains for each conference.

Since it will be a GSoC project, we'll have to wait until GSoC starts to actually work on this, and then we will very much appreciate your feedback!

differentreality:
Closing this as it will be part of GSoC project this year.

PR #1266, opened 2017-02-01
Remove the unreachable code and modify the error message in ConferenceRegistrationsController#new

### LINKED ISSUE #1140: Unreachable code in ConferenceRegistrationsController#new

AEtherC0r3:
When iChain is enabled if the user is not signed in the message "You are not authorized to access this page. Maybe you need to sign in?" will be displayed, while if the user is signed in and something else goes wrong, the message "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open." will be shown.

Fixes #1140

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -6,22 +6,21 @@ class ConferenceRegistrationsController < ApplicationController
 
   def new
     @registration = Registration.new(conference_id: @conference.id)
-    authorize! :new, @registration, message: "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
 
     # Redirect to registration edit when user is already registered
     if @conference.user_registered?(current_user)
+      # Authorization needs to happen in every action before the return statement
+      # We authorize the #edit action, since we redirect to it
+      authorize! :edit, current_user.registrations.find_by(conference_id: @conference.id)
       redirect_to edit_conference_conference_registration_path(@conference.short_title)
-      return
-    # ichain does not allow us to create users during registration
-    elsif (ENV['OSEM_ICHAIN_ENABLED'] == 'true') && !current_user
-      redirect_to root_path, alert: 'You need to sign in or sign up before continuing.'
       return
     end
 
-    # avoid openid sign_in to redirect to register/new when the sign_in user had already a registration
-    if current_user && @conference.user_registered?(current_user)
-      redirect_to edit_conference_conference_registration_path(@conference.short_title)
+    if !@conference.registration_open? || @conference.registration_limit_exceeded?
+      message = "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
+      @ignore_not_signed_in_user = true
     end
+    authorize! :new, @registration, message: message
 
     # @user variable needs to be set so that _sign_up_form_embedded works properly
     @user = @registration.build_user
--- a/spec/controllers/conference_registration_controller_spec.rb
+++ b/spec/controllers/conference_registration_controller_spec.rb
@@ -1,51 +1,167 @@
 require 'spec_helper'
 
 describe ConferenceRegistrationsController, type: :controller do
-  let(:conference) { create(:conference) }
+  let(:conference) { create(:conference, title: 'My Conference', short_title: 'myconf') }
   let(:user) { create(:user) }
+  let(:not_registered_user) { create(:user) }
+  let(:registered_user) { create(:user) }
+  let!(:registration) { create(:registration, conference: conference, user: registered_user, created_at: 1.day.ago) }
+
+  shared_examples 'access #new action' do |user, ichain, path, message|
+    before :each do
+      sign_in send(user) if user
+      stub_const('ENV', ENV.to_hash.merge('OSEM_ICHAIN_ENABLED' => ichain))
+      get :new, conference_id: conference.short_title
+    end
+
+    it 'redirects' do
+      expect(response).to redirect_to path
+    end
+
+    it 'shows flash alert' do
+      expect(flash[:alert]).to eq message
+    end
+  end
 
   context 'user is signed in' do
-    before { sign_in(user) }
+    before :each do
+      sign_in user
+    end
 
     describe 'GET #new' do
       context 'registration period open' do
-        before do
-          @registration_period = create(:registration_period, conference: conference)
+        before :each do
+          create(:registration_period, conference: conference, start_date: 3.days.ago, end_date: 1.day.from_now)
         end
 
-        context 'user registered' do
-          before do
-            @registration = create(:registration, conference: conference, user: user)
-            get :new, conference_id: conference.short_title
+        context 'registration limit not exceeded' do
+          before :each do
+            conference.registration_limit = 0
+            conference.save!
           end
 
-          it 'redirects to edit conference registration' do
-            expect(response).to redirect_to edit_conference_conference_registration_path(conference.short_title)
+          context 'OSEM_ICHAIN_ENABLED is true' do
+            before :each do
+              stub_const('ENV', ENV.to_hash.merge('OSEM_ICHAIN_ENABLED' => 'true'))
+            end
+
+            context 'user registered' do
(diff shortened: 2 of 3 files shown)


AEtherC0r3:
The changes correspond to the following truth table

OSEM_ICHAIN_ENABLED | user signed in | period open && registration limit not exceeded && registration new record | result
---------------------------------- | ------------------ | ------------------------------------------------------------------------- | ------
true | true | true | Continue to /conference/:id/register/new
true | true | false | Show message "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
true | false | true |  Show message "You are not authorized to access this page. Maybe you need to sign in?"
true | false | false | Show message "You are not authorized to access this page. Maybe you need to sign in?"
false | true | true | Continue to /conference/:id/register/new
false | true | false | Show message "Sorry, you can not register for #{conference.title}. Registration limit exceeded or the registration is not open."
false | false | true | Continue to /conference/:id/register/new
false | false | false | Show message "Sorry, you can not register for #{conference.title}. Registration limit exceeded or the registration is not open. Maybe you need to sign in?"

differentreality:
I added numbering to the different cases, so that we can easily reference them.
No 2 and 4, should probably show the same message, message of No 2 is what is actually going on.  Asking the user to sign in only to tell the user registration is closed, is useless and kinda misleading. Just like No 6 and 8.

AEtherC0r3:
The truth table for the updated code is as follows

No | OSEM_ICHAIN_ENABLED | user signed in | period open && registration limit not exceeded && registration new record | result
--- | ---------------------------------- | ------------------ | ------------------------------------------------------------------------- | ------
1 | X | X | false | Show message "Sorry, you can not register for #{@conference.title}. Registration  limit exceeded or the registration is not open."
2 | true | false | true |  Show message "You are not authorized to access this page. Maybe you need to sign in?"
3 | true | true | true | Continue to /conference/:id/register/new
4 | false | X | true | Continue to /conference/:id/register/new

app/controllers/conference_registrations_controller.rb
```diff
@@ -6,22 +6,19 @@ class ConferenceRegistrationsController < ApplicationController
 
   def new
     @registration = Registration.new(conference_id: @conference.id)
-    authorize! :new, @registration, message: "Sorry, you can not register for #{@conference.title}. Registration limit exceeded or the registration is not open."
 
     # Redirect to registration edit when user is already registered
     if @conference.user_registered?(current_user)
+      authorize! :edit, current_user.registrations.find_by(conference_id: @conference.id)
```
differentreality: Why do we need this? Please add a comment.

app/controllers/conference_registrations_controller.rb
```diff
@@ -6,22 +6,19 @@ class ConferenceRegistrationsController < ApplicationController
...
-    # ichain does not allow us to create users during registration
-    elsif (ENV['OSEM_ICHAIN_ENABLED'] == 'true') && !current_user
-      redirect_to root_path, alert: 'You need to sign in or sign up before continuing.'
-      return
     end
 
-    # avoid openid sign_in to redirect to register/new when the sign_in user had already a registration
-    if current_user && @conference.user_registered?(current_user)
-      redirect_to edit_conference_conference_registration_path(@conference.short_title)
+    unless @conference.registration_open? && !@conference.registration_limit_exceeded?
```
differentreality: Could this be a `can? :new, @registration` condition?

AEtherC0r3: There are cases where `can? :new, ,@registration` will return false while the required result is true, for example if OSEM_ICHAIN_ENABLED is true, the user is not signed in, the registrations are open and the registration limit hasn't been exceeded. Therefore it isn't a suitable replacement.

AEtherC0r3: An equivalent expression using can? that can replace the current one is the following: `(can? :new, @registration) || (can? :new, @registration) && ENV['OSEM_ICHAIN_ENABLED'] == 'true' && current_user == nil`
Also, code readability can be improved by replacing the unless statement with `if !@conference.registration_open? || @conference.registration_limit_exceeded?`, if that's an issue.

differentreality:
Can we create a registered_user and a not_registered_user with let() in the beginning, so that we don't  repeat the creation of the registration every time we test what happens with the registered_user?

You could also check passing a variable user in your shared_examples and based on that sign in the appropriate user in the shared examples - if the user is nil, that means there is no signed_in user.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -1,51 +1,179 @@
...
+          before :each do
+            conference.registration_limit = 0
+            conference.save!
           end
 
-          it 'redirects to edit conference registration' do
-            expect(response).to redirect_to edit_conference_conference_registration_path(conference.short_title)
+          context 'OSEM_ICHAIN_ENABLED is true' do
+            before :each do
+              stub_const('ENV', ENV.to_hash.merge('OSEM_ICHAIN_ENABLED' => 'false'))
```
differentreality: in the context text it says ichain is true

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -1,51 +1,179 @@
...
 
 describe ConferenceRegistrationsController, type: :controller do
-  let(:conference) { create(:conference) }
+  let(:conference) { create(:conference, title: 'My Conference', short_title: 'myconf') }
   let(:user) { create(:user) }
+  let(:not_registered_user) { create(:user) }
+  let(:registered_user) { create(:user) }
+  let!(:registration) { create(:registration, conference: conference, user: registered_user, created_at: 1.day.ago) }
+
+  shared_examples 'access action without ability' do |user, ichain, path, message|
```
differentreality: You call the shared example access action *without* ability, but you are using it for cases when the user has the ability to access the action. I would suggest we rename this. Perhaps even just `#new` ? What do you think?

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -1,51 +1,179 @@
...
-            get :new, conference_id: conference.short_title
+        context 'registration limit exceeded' do
+          before :each do
+            conference.registration_limit = 1
+            conference.save!
           end
 
-          it 'user variable exists' do
-            expect(assigns(:user)).not_to be_nil
+          context 'OSEM_ICHAIN_ENABLED is true' do
```
differentreality: This context does not really offer any value here, specially since there is no before block anymore. You can just include the info that ichain is true in the 2 contexts below (for register and not registered user).

spec/controllers/conference_registration_controller_spec.rb
AEtherC0r3: Yup, the 'without ability' part is misleading. I will change it to 'access #new action'

spec/controllers/conference_registration_controller_spec.rb
AEtherC0r3: Sorry about that

differentreality:
:+1: please rebase and squash your commits to merge this! Good work!

differentreality:
:tada:

PR #1268, opened 2017-02-05
Added feature tests for Roles

shlok007:


spec/features/roles_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    scenario 'assign info desk role to user' do
+      click_link('Users', href: admin_conference_role_path(conference.short_title, 'info_desk'))
+
+      fill_in 'user_email', with: other_user.email
+
+      click_button 'Add'
+      other_user.reload
+
+      expect(other_user.has_role?(:info_desk, conference)).to be_truthy
```
differentreality: In our test suite we mainly use `eq true`

differentreality:
You are using shared_examples, but the way you have structured the code, you are not really sharing any tests. The naming is also a bit misleading, as you are not testing the management of the roles, you are only testing the addition of role. 
What you should be doing is having 1 test/scenario in the shared_examples for the action you want to test (eg. add role), and call the shared examples for every different role.

differentreality:
This looks really nice :) Good job!
How about you add some more contexts for the other roles as well before we merge it?

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let!(:organizer) { create(:user, role_ids: [organizer_role.id]) }
+
+  shared_examples 'successfully' do |role_name|
+    let!(:role) { Role.find_by(name: role_name, resource: conference) }
+    before(:each) do
+      sign_in organizer
+      visit admin_conference_roles_path(conference.short_title)
+    end
+
+    scenario 'add role' do
```
differentreality: add**s** role reads better (same for edit role -> edits role)

shlok007:
> What you should be doing is having 1 test/scenario in the shared_examples for the action you want to test (eg. add role), and call the shared examples for every different Editing of roles is only possible by organizers hence the scenario of `edits role` would fail for other users. If we have to include test for 'adds role' for all roles, we would have to put it in a separate `shared_example`. Do let me know if there is any cleaner approach to tackle this.

differentreality:
It's equally important to test that role X cannot perform an action. So, we should not merely test successfully performed actions. 
How about you do a write-up of all the cases we have here? Then we can structure the tests.
There is always the possibility of including the expected result `true/false` in your parameters, eg. https://github.com/openSUSE/osem/blob/master/spec/models/program_spec.rb#L108

shlok007:
Actions that have to be tested are `adds role`, `removes role` and `edits role`. Am I missing something?

> There is always the possibility of including the expected result true/false in your parameters, eg. https://github.com/openSUSE/osem/blob/master/spec/models/program_spec.rb#L108

Thank you but I am having trouble using that while adding all scenario in a single `shared_example` because for roles who cannot perform an action the respective elements ( like `bootstrap-switch` ) wouldn't be present on the page which would return ElementNotFound exception . :sweat_smile: .

I have tried testing all the roles using a single `shared_example`. Please do review the changes.

differentreality:
@shlok007 why isn't there a second commit for the most recent changes that you pushed?

shlok007:
Sorry. I thought it would be better to squash the commits and start anew since almost all lines were changed.

differentreality:
@shlok007 the thing is that like this, I was previously reviewing something completely different, and now I have no idea what that was.

The way you were running tests before is a totally different approach than before though, so we won't apply the same structure (ie shared examples).

shlok007:
> @shlok007 the thing is that like this, I was previously reviewing something completely different, and now I have no idea what that was.

I apologize. I have undid the rebase now.

Also, I have split the test into multiple shared_examples so now we are also testing if role X cannot perform an action.

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    let!(:role) { Role.find_by(name: role_name, resource: conference) }
+    let!(:by_role) { Role.find_by(name: by_role_name, resource: conference) }
+    let!(:user_to_sign_in) { create(:user, role_ids: [by_role.id]) }
+
+    before :each do
+      sign_in user_to_sign_in
+      visit admin_conference_roles_path(conference.short_title)
+    end
+
+    scenario 'edits role', feature: true, js: true do
```
differentreality: This should be more specific and descriptive, for example something like
"edits role #{role_name}, when user has role #{by_role_name}"

differentreality:
If a test fails we see the following:

> Role organizer for volunteers_coordinator role behaves like edit role successfully edits role

We can probably rephrase that better.

shlok007:
Test passes for edit role:
`Role organizer behaves like should successfully edit role volunteers_coordinator`

Test passes for add and remove users:
`Role organizer behaves like should successfully add role volunteers_coordinator`
`Role organizer behaves like should successfully edit role volunteers_coordinator`

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    scenario "edit role #{role_name}", feature: true, js: true do
+      click_link('Users', href: admin_conference_role_path(conference.short_title, role_name))
+
+      expect(first('td').has_css?('.bootstrap-switch-container')).to eq false
+    end
+  end
+
+  shared_examples 'should not successfully edit' do |role_name, by_role_name|
```
differentreality: does not, instead of should not

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      click_link('Users', href: admin_conference_role_path(conference.short_title, role_name))
+
+      fill_in 'user_email', with: user_with_no_role.email
+      click_button 'Add'
+      user_with_no_role.reload
+
+      expect(user_with_no_role.has_role?(role.name, conference)).to eq true
+    end
+
+    scenario "edit role #{role_name}", feature: true, js: true do
```
differentreality: This is not edit role, this is remove user from a role (or remove role, from a user, as you have phrased it accordingly in `add role`)

spec/features/roles_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      visit admin_conference_roles_path(conference.short_title)
+    end
+
+    scenario "add role #{role_name}", feature: true, js: true do
+      click_link('Users', href: admin_conference_role_path(conference.short_title, role_name))
+
+      expect(page.has_field?('user_email')).to eq false
+    end
+
+    scenario "edit role #{role_name}", feature: true, js: true do
```
differentreality: This is for when the user cannot remove a role from a user, not edit the role

differentreality:
@shlok007 I have made a few wording comments, please fix those so that I can review it all, other than that I think it looks good!

differentreality:
Please rephrase all occurrences, removing should, so that tests read like
successfully edits
does not successfully edit

Please note the 3rd person applied in affirmative statement as well.

differentreality:
Looks good :) Please rebase and squash!

differentreality:
:+1:

PR #1270, opened 2017-02-06
Fixed schedule creation when there is no room.

### LINKED ISSUE #1209: Fix schedule creation when there are no room
If there are not rooms you can create an schedule and then you see that:

screenshot_20160926_135117

We should render something like that in the schedules page and not allow to create an schedule if there are no rooms, as this is quite confusing.

Also, the link to `Create rooms` doesn't work if there is not a venue. So, we should change it by `Create a venue with rooms` with a link to where the venue is created if there is not a venue.

siddhantbajaj:
Closes #1209

### DIFF
--- a/app/views/admin/schedules/index.html.haml
+++ b/app/views/admin/schedules/index.html.haml
@@ -25,8 +25,21 @@
                 method: :get, class: 'btn btn-primary'
                 = link_to 'Delete', admin_conference_schedule_path(@conference.short_title, schedule.id),
                 method: :delete, class: 'btn btn-danger', data: { confirm: "Do you really want to delete Schedule #{schedule.id}?" }
-
 .row
-  .col-md-12.text-right
-    = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
-    method: :post, class: 'btn btn-primary'
+  .col-md-12
+    - if @venue.try(:rooms).present?
+      .text-right
+        = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
+        method: :post, class: 'btn btn-primary'
+    - elsif @venue
+      .h3
+        No Rooms!
+        %small
+          = link_to 'Create rooms', admin_conference_venue_rooms_path
+          before creating the schedule.
+    - else
+      .h3
+        No Venue!
+        %small
+          = link_to 'Create a venue with rooms', new_admin_conference_venue_path
+          before creating the schedule.


app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,20 @@
...
+  .row
+    .col-md-12.text-right
+      = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
+      method: :post, class: 'btn btn-primary'
+- else
+  .h3
+    No Rooms!
+    %small
+      = link_to 'Create rooms', admin_conference_venue_rooms_path
+      before creating the schedule.
```
differentreality: Please add an empty line at the end of the file. That's gonna make the red indication disappear. Your editor might have an option to do this automatically.

differentreality:
Please provide screenshots for your changes.

app/controllers/admin/schedules_controller.rb
```diff
@@ -9,7 +9,9 @@ class SchedulesController < Admin::BaseController
     load_resource :selected_schedule, through: :program, singleton: true
     load_resource :venue, through: :conference, singleton: true
 
-    def index; end
+    def index 
+      @rooms = @conference.venue.rooms if @conference.venue
```
differentreality: Variable @rooms does not need to hold the records of all the rooms. You are not using the rooms anywhere, you are merely checking if rooms exist.

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,21 @@
...
+    No Venue!
+    %small
+      = link_to 'Create a venue with rooms', new_admin_conference_venue_path
+      before creating the schedule.
+- elsif @rooms.present?
+  .row
+    .col-md-12.text-right
+      = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
+      method: :post, class: 'btn btn-primary'
+- else
```
differentreality: The logic of the if statement is not very straight forward here. The existence of @rooms implies there is a venue, however when reading the code, it is not explicit what the else means (ie. no venue and no rooms). How could we re-write the if-statement in a way that is easier to read and understand later?

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,21 @@
...
+  .h3
+    No Rooms!
+    %small
+      = link_to 'Create rooms', admin_conference_venue_rooms_path
+      before creating the schedule.
+
```
differentreality: This is still here. If you are having trouble with it, find me on IRC to figure it out.

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,21 @@
...
-    = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
-    method: :post, class: 'btn btn-primary'
+- if @rooms.nil?
+  .h3
+    No Venue!
+    %small
+      = link_to 'Create a venue with rooms', new_admin_conference_venue_path
+      before creating the schedule.
+- elsif @rooms.present?
+  .row
```
differentreality: .row and .col-md-12 should exist in all cases.

app/controllers/admin/schedules_controller.rb
siddhantbajaj: I think we need it in the layout file to check whether rooms are present or not.

app/views/admin/schedules/index.html.haml
siddhantbajaj: Yeah I know looks a bit messy, but yeah will come up with a better logic.

app/views/admin/schedules/index.html.haml
siddhantbajaj: Ahhhh looks like I accidently typed space in the new line as well.

app/views/admin/schedules/index.html.haml
siddhantbajaj: okay, will do these changes.

app/controllers/admin/schedules_controller.rb
differentreality: In the index file we need to check if the are any rooms (true or false), we don't need to have the whole collection of rooms available.

siddhantbajaj:
CHANGES:
screenshot from 2017-02-08 04 56 43

screenshot from 2017-02-08 04 58 01

screenshot from 2017-02-08 04 59 24

app/controllers/admin/schedules_controller.rb
```diff
@@ -9,7 +9,10 @@ class SchedulesController < Admin::BaseController
     load_resource :selected_schedule, through: :program, singleton: true
     load_resource :venue, through: :conference, singleton: true
```
differentreality: This line defines @venue variable, you can read more about it at https://github.com/CanCanCommunity/cancancan/wiki/Authorizing-controller-actions#choosing-actions

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,24 @@
                 method: :get, class: 'btn btn-primary'
                 = link_to 'Delete', admin_conference_schedule_path(@conference.short_title, schedule.id),
                 method: :delete, class: 'btn btn-danger', data: { confirm: "Do you really want to delete Schedule #{schedule.id}?" }
-
-.row
-  .col-md-12.text-right
-    = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
-    method: :post, class: 'btn btn-primary'
+- if @venue_present && @rooms_present
```
differentreality: You can simply do `if @venue && @venue.rooms.any?`

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,24 @@
...
-    = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
-    method: :post, class: 'btn btn-primary'
+- if @venue && @venue.rooms.any?
+  .row
+    .col-md-12.text-right
+      = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
+      method: :post, class: 'btn btn-primary'
+- elsif @venue
+  .row
+    .col-md-12
```
differentreality: But if `.row` and `.col-md-12` are always needed, we can just define them once outside the if statements, no need to repeat them.

differentreality:
This looks ok :+1: 
After making changes for my last comment, please squash your commits into one, so that we can merge this!

:tada: Thank you for contributing :tada:

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,24 @@
                 method: :get, class: 'btn btn-primary'
                 = link_to 'Delete', admin_conference_schedule_path(@conference.short_title, schedule.id),
                 method: :delete, class: 'btn btn-danger', data: { confirm: "Do you really want to delete Schedule #{schedule.id}?" }
-
-.row
-  .col-md-12.text-right
-    = link_to 'Add Schedule', admin_conference_schedules_path(@conference.short_title),
-    method: :post, class: 'btn btn-primary'
+- if @venue && @venue.rooms.any?
```
Ana06: This can also be written as `if @venue.try(:rooms).present?`

differentreality: Do you think we should prefer writing this like that?
Because it is less explicit and the same amount of code.

Ana06: I like it more, as it only focus on if there are rooms or not, but I am also fine with it as it is now.

differentreality: @Ana06 sorry my bad, you meant writing just `if @venue.try(:rooms).present?`, right? 
I was thinking `if @venue && @venue.try(:rooms).present?`
Yes, that's indeed a nicer approach

app/views/admin/schedules/index.html.haml
```diff
@@ -25,8 +25,21 @@
                 method: :get, class: 'btn btn-primary'
                 = link_to 'Delete', admin_conference_schedule_path(@conference.short_title, schedule.id),
                 method: :delete, class: 'btn btn-danger', data: { confirm: "Do you really want to delete Schedule #{schedule.id}?" }
-
 .row
-  .col-md-12.text-right
+  .col-md-12
+- if @venue && @venue.rooms.any?
```
differentreality: Your indentation is wrong here, the code needs be inside .col-md-12

siddhantbajaj: Really sorry, was in a hurry to get my first issue done. Would take proper care of indentation from now on.

differentreality:
:+1:

Issue #1234, opened 2016-12-29
Show Concurrent Events on Event#show view


ChrisBr:
It would be interesting to see Concurrent Events while on the event#show page.

abit2:
Hey i'm new here 
Can you please elaborate on this

PR #1275, opened 2017-02-06
Display concurrent events

### LINKED ISSUE #1205: function intersecting_events no longer returns a collect of events
https://github.com/openSUSE/osem/blame/master/app/models/event_schedule.rb#L28

Returns event_schedules instead of events

shlok007:
Closes #1205 and #1234 .

[ updated screenshot after b64291b ]
updated

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -575,4 +575,21 @@ module ApplicationHelper
   def quantity_left_of(resource)
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
+    return nil unless event.scheduled? && event.program.selected_event_schedules
+    event_schedule = event.program.selected_event_schedules.find_by(event: event)
+    other_event_schedules = event.program.selected_event_schedules.reject { |other_event_schedule| other_event_schedule == event_schedule }
+    concurrent_events = []
+
+    event_time_range = (event_schedule.start_time.strftime '%Y-%m-%d %H:%M')...(event_schedule.end_time.strftime '%Y-%m-%d %H:%M')
+    other_event_schedules.each do |other_event_schedule|
+      next unless other_event_schedule.event.confirmed?
+      other_event_time_range = (other_event_schedule.start_time.strftime '%Y-%m-%d %H:%M')...(other_event_schedule.end_time.strftime '%Y-%m-%d %H:%M')
+      if (event_time_range.to_a & other_event_time_range.to_a).present?
+        concurrent_events << other_event_schedule.event
+      end
+    end
+    concurrent_events
+  end
 end
--- a/app/views/proposals/show.html.haml
+++ b/app/views/proposals/show.html.haml
@@ -97,3 +97,20 @@
               %dt Requires Registration:
               %dd
                 = link_to "Yes (#{registered_text(@event)})", new_conference_conference_registration_path(@conference.short_title), class: 'btn btn-xs btn-danger', disabled: !@event.registration_possible?
+          - if concurrent_events(@event).present?
+            .col-md-12
+              %hr
+              %h4 Happening at the same time:
+              %ol
+                - concurrent_events(@event).each do |event|
+                  %li
+                    = link_to conference_program_proposal_path(@conference.short_title, event.id) do
+                      = event.title
+                  %dl
+                    %dt Start Time:
+                    %dd
+                      = event.program.selected_event_schedules.find_by(event: event).start_time.strftime("%Y %B %e %H:%M")
+                    %br
+                      %dt Room:
+                      %dd
+                        = event.room.name
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -92,4 +92,42 @@ describe ApplicationHelper, type: :helper do
       end
     end
   end
+
+  describe '#concurrent_events' do
+    before :each do
+      @other_event = create(:event, program: conference.program, state: 'confirmed')
+      schedule = create(:schedule, program: conference.program)
+      conference.program.update_attributes!(selected_schedule: schedule)
+      @event_schedule = create(:event_schedule, event: event, start_time: conference.start_date, room: create(:room), schedule: schedule)
+      @other_event_schedule = create(:event_schedule, event: @other_event, start_time: conference.start_date, room: create(:room), schedule: schedule)
+    end
+
+    describe 'does return correct concurrent events' do
+      it 'when events starts at the same time' do
+        expect(concurrent_events(event).include?(@other_event)).to eq true
+      end
+
+      it 'when event is in between the other event' do
+        @event_schedule.update_attributes!(start_time: @other_event_schedule.start_time + 10.minutes)
+        expect(concurrent_events(event).include?(@other_event)).to eq true
+      end
+    end
+
+    describe 'does not return as concurrent event ' do
+      it 'when event is not scheduled' do
+        @event_schedule.destroy
+        expect(concurrent_events(event).present?).to eq false
+      end
+
+      it 'when one event starts and other ends at the same time' do
+        @event_schedule.update_attributes!(start_time: @other_event_schedule.end_time)
+        expect(concurrent_events(event).present?).to eq false
+      end
+
+      it 'when conference program does not have a selected schedule' do
+        conference.program.update_attributes!(selected_schedule_id: nil)
+        expect(concurrent_events(event).present?).to eq false
+      end
+    end
+  end
 end


differentreality:
Intersecting events is when an event is scheduled, then cancelled/withdrawn, then another event is scheduled on its place.

Concurrent event is when event1 is happening in roomA and event2 is happening in roomB at the same time.

This differentiation is not clear in the PR.

shlok007:
> Intersecting events is when an event is scheduled, then cancelled/withdrawn, then another event is scheduled on its place.
Concurrent event is when event1 is happening in roomA and event2 is happening in roomB at the same time.

Thank you. I have updated the PR accordingly now.

app/helpers/application_helper.rb
```diff
@@ -529,11 +529,29 @@ def canceled_replacement_event_label(event, event_schedule, *label_classes)
...
         concat content_tag :span, 'Please note that this talk replaces '
         concat link_to replaced_event.title, conference_program_proposal_path(@conference.short_title, replaced_event.id)
       end
     end
   end
+
+  def concurrent_events(event)
+    return nil unless event.event_schedules
+    schedule = Schedule.find(event.program.selected_schedule_id)
+    event_schedule = EventSchedule.find_by(schedule: schedule.id, event: event)
```
differentreality: Can't this be `schedule.event_schedules.find_by(event: event)`?

app/helpers/application_helper.rb
```diff
@@ -529,11 +529,29 @@ def canceled_replacement_event_label(event, event_schedule, *label_classes)
...
+    other_event_schedules = schedule.event_schedules
+    concurrent_events = []
+
+    other_event_schedules.each do |other_event_schedule|
+      unless other_event_schedule.event.unconfirmed?
```
differentreality: It's better if you do this with an `if` statement, instead of `unless`

app/models/event_schedule.rb
```diff
@@ -27,12 +27,12 @@ def end_time
   ##
   # Returns events that are scheduled in the same room and start_time as event
   #
-  def intersecting_events
+  def intersecting_events_schedule
     room.event_schedules.where(start_time: start_time, schedule: schedule).where.not(id: id)
```
differentreality: intersecting_event_schedules

app/models/event_schedule.rb
```diff
@@ -27,12 +27,12 @@ def end_time
   ##
   # Returns events that are scheduled in the same room and start_time as event
```
differentreality: This should be updated as well

app/views/proposals/show.html.haml
```diff
@@ -97,3 +97,10 @@
               %dt Requires Registration:
               %dd
                 = link_to "Yes (#{registered_text(@event)})", new_conference_conference_registration_path(@conference.short_title), class: 'btn btn-xs btn-danger', disabled: !@event.registration_possible?
+          - if concurrent_events(@event).present?
+            .col-md-12#concurrent-events
+              %h4
+                = "Other Events on \n #{@event_schedule.start_time.strftime("%Y %B %e")} at #{@event_schedule.start_time.strftime("%H:%M")}:"
+              - concurrent_events(@event).each do |event|
+                %h4
+                  = link_to event.title, conference_program_proposal_path(@conference.short_title, event.id)
```
differentreality: This might look better if we show it with %li, as some event titles might use more than one line and then it gets confusing

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,22 @@ def replacement_event_notice(event_schedule)
...
+    return nil unless event.event_schedules
+    schedule = Schedule.find(event.program.selected_schedule_id)
+    event_schedule = schedule.event_schedules.find_by(event: event)
+    other_event_schedules = schedule.event_schedules
+    concurrent_events = []
+
+    other_event_schedules.each do |other_event_schedule|
+      if other_event_schedule.event.confirmed?
+        if other_event_schedule.start_time == event_schedule.start_time &&
+           other_event_schedule.room       != event_schedule.room
```
differentreality: This big space shouldn't be here

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,22 @@ def replacement_event_notice(event_schedule)
   def quantity_left_of(resource)
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
+    return nil unless event.event_schedules
+    schedule = Schedule.find(event.program.selected_schedule_id)
```
differentreality: What if there is no selected_schedule yet?

differentreality:
The renaming of the function is fine, though should probably have been a separate PR.

As far as the concurrent events are concerned, we should probably show the room along with the title as well. 
I would suggest that we don't limit concurrency on same start_time, eg. eventA happens 9:00 - 10:00, eventB happens 9:30 - 10:00, eventC happens 8:45 - 9:15, they are all still pretty concurrent.

app/views/proposals/show.html.haml
```diff
@@ -97,3 +97,11 @@
               %dt Requires Registration:
               %dd
                 = link_to "Yes (#{registered_text(@event)})", new_conference_conference_registration_path(@conference.short_title), class: 'btn btn-xs btn-danger', disabled: !@event.registration_possible?
+          - if concurrent_events(@event).present?
+            .col-md-12#concurrent-events
+              %h4
+                = "Other Events on \n #{@event_schedule.start_time.strftime("%Y %B %e")} at #{@event_schedule.start_time.strftime("%H:%M")}:"
```
differentreality: Events is a term we only use in the admin namespace of OSEM, its meaning is not very straightforward to attendees.
This could simply be
**Happening at the same time**

As we are mentioning the date and time a few lines above.

shlok007:
> The renaming of the function is fine, though should probably have been a separate PR.

At first, It seemed to me that intersecting events and concurrent events are one and the same(the reason for clubbing it in a single PR), but after your clarification it turns out they are far from similar. I could probably move that commit to a separate PR if you wish.

>As far as the concurrent events are concerned, we should probably show the room and start time along with the title.
I would suggest that we don't limit concurrency on same start_time, eg. eventA happens 9:00 - 10:00, eventB happens 9:30 - 10:00, eventC happens 8:45 - 9:15, they are all still pretty concurrent.

I have just added a couple more commits to address the same. :slightly_smiling_face:

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,23 @@ def replacement_event_notice(event_schedule)
   def quantity_left_of(resource)
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
+    return nil unless event.program.selected_event_schedules
+    event_schedule = event.program.selected_event_schedules.find_by(event: event)
+    other_event_schedules = event.program.selected_event_schedules
```
differentreality: Why don't you already reject the event_schedule from this collection?

shlok007: Thank you!

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,22 @@ def replacement_event_notice(event_schedule)
   def quantity_left_of(resource)
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
+    return nil unless event.program.selected_event_schedules
+    event_schedule = event.program.selected_event_schedules.find_by(event: event)
```
differentreality: What if the event is not scheduled yet, so this is nil? L550 will raise an error.

shlok007: :disappointed_relieved:

differentreality:
@Ana06 @hennevogel any input for how this looks on proposals#show?

app/helpers/application_helper.rb
differentreality: @shlok007 please run a local instance of OSEM and check the changes you are making.

differentreality:
@ChrisBr anything else you had in mind about this feature?

app/views/proposals/show.html.haml
```diff
@@ -97,3 +97,23 @@
...
+                  %dl
+                    .col-md-12
+                      %dt Start Time:
+                      %dd
+                        = event.program.selected_event_schedules.find_by(event: event).start_time.strftime("%Y %B %e %H:%M")
+                    .col-md-12
+                      %dt Room:
+                      %dd
+                        = event.room.name
+                  %br
```
shlok007: I could provide sufficient padding to `%dl` and we could remove the `%br` .

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,22 @@ def replacement_event_notice(event_schedule)
...
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
+    return nil unless event.scheduled? && event.program.selected_event_schedules
+    event_schedule = event.program.selected_event_schedules.find_by(event: event)
+    other_event_schedules = event.program.selected_event_schedules.reject { |other_event_schedule| other_event_schedule == event_schedule }
+    concurrent_events = []
+
+    event_time_range = (event_schedule.start_time.strftime '%Y-%m-%d %H:%M')...(event_schedule.end_time.strftime '%Y-%m-%d %H:%M')
```
shlok007: I am using `.strftime` to remove the seconds part of the time. Suppose for an one hour event the number of elements in the array would be 60 each corresponding to each minute in event_time_range and other_event_time_range . Using `.to_datetime`  the number of elements would be 3600 in both arrays which will increase the number of comparisons made in line 554 . :sweat_smile:

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,22 @@ def replacement_event_notice(event_schedule)
...
+    other_event_schedules.each do |other_event_schedule|
+      if other_event_schedule.event.confirmed?
```
shlok007: thank you

app/views/proposals/show.html.haml
differentreality: I don't think we really need any of the CSS you are adding (or the `.nav.nav-pills` for that matter). Why don't you try a simpler approach? Perhaps a horizontal line can separate the proposal info from the `Happening at the same time` info. I would also use numbering for showing the concurrent events.

shlok007: I'll remove the CSS :slightly_smiling_face:

app/helpers/application_helper.rb
```diff
@@ -540,4 +540,21 @@ def replacement_event_notice(event_schedule)
   def quantity_left_of(resource)
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
+
+  def concurrent_events(event)
```
differentreality: This definitely needs to be tested before merging the PR! :)

differentreality:
@shlok007 Please add relevant tests for your new method so that we can merge this!

shlok007:
On it! :)

spec/helpers/application_helper_spec.rb
```diff
@@ -92,4 +92,16 @@
       end
     end
   end
+
+  describe 'concurrent_events' do
```
differentreality: #concurrent_events

differentreality:
@shlok007 the test for the new function needs to be way more extensive than that! It needs to include possible points of failure and edge cases.

spec/helpers/application_helper_spec.rb
```diff
@@ -92,4 +92,40 @@
...
+
+      it 'when event is in between the other event' do
+        @event_schedule.update_attributes(start_time: rand(@other_event_schedule.start_time...@other_event_schedule.end_time))
+        expect(concurrent_events(event).include?(@other_event)).to eq true
+      end
+    end
+
+    describe 'does not return as concurrent event ' do
+      it 'when event is not scheduled' do
+        @event_schedule.destroy
```
shlok007: Sorry, doesn't https://github.com/openSUSE/osem/blob/master/spec/factories/event_schedule.rb#L17 create a selected schedule for the conference while creating `@event_schedule` in L99?

differentreality: Absolutely it does, it just ain't obvious in this test. And if the factory were to change, this test would still pass but it would not be testing what it is supposed to. In my opinion we should make our assumptions about the test explicit.

spec/helpers/application_helper_spec.rb
```diff
@@ -96,8 +96,10 @@
   describe '#concurrent_events' do
     before :each do
       @other_event = create(:event, program: conference.program, state: 'confirmed')
-      @event_schedule = create(:event_schedule, event: event, start_time: conference.start_date, room: create(:room))
-      @other_event_schedule = create(:event_schedule, event: @other_event, start_time: conference.start_date, room: create(:room))
+      schedule = create(:schedule, program: conference.program)
+      conference.program.update_attributes(selected_schedule: schedule)
```
differentreality: `update_attributes!`  let's do it like this in all occurrences.

differentreality:
:+1:  Nice @shlok007, please rebase and squash commits after travis succeeds!

PR #1276, opened 2017-02-06
Fix Proposal update page's requirement field. 

BelieveC:
Requirement field collapses only when requirement data is empty else it shows requirement data.
Expected Behavior Fixed:
1) Collapse when requirement data is Empty
2) Doesn't collapse  if requirement data is present.
require

### DIFF
--- a/app/views/proposals/_proposal_form.html.haml
+++ b/app/views/proposals/_proposal_form.html.haml
@@ -55,9 +55,9 @@
     %p.text-right
       = link_to '#description', 'data-toggle' => 'collapse' do
         Do you require something special?
-    .collapse#description
+  
+    #description{class:"collapse#{'in' if @event.description.present?}"}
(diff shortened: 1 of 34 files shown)


app/views/proposals/_proposal_form.html.haml
```diff
@@ -55,7 +55,11 @@
     %p.text-right
       = link_to '#description', 'data-toggle' => 'collapse' do
         Do you require something special?
-    .collapse#description
+  -if @event.description.blank?
```
Ana06: I would try to avoid repeating the input line, something like:

``` css
#description{ class: "collapse#{ ' in' if @event.description.present? }" }
= f.input :description, input_html: { rows: 5 }, label: 'Requirements', placeholder: 'Eg. Whiteboard, printer, or something like that.'
```

BelieveC: Thank you @Ana06 ! I would make described changes.

PR #1278, opened 2017-02-07
Implements support for multiple speakers per event and adding events manually by conf admins

commandprompt:
Hi.
This is a draft implementation of the support for multiple speakers per event, this pull request also includes the support for adding events manually through the Admin->Conference->Events screen
The manual event management was reimplemented in a bit different way than it was in the PR #1257
Features:
Allows submitter to add extra speakers to the conference using the My Submissions->Edit screen
Shows multiple speakers' userpics in the public schedule event blocks and event descriptions
Shows multiple speakers with their userpics and biographies on the Event public info screen
Admin users can now add Events manually on the Conference Event listing screen (using the Add Event button)
Event Info page in the Conference Admin section also displays multiple speakers and their biographies (if the speaker has one)

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -5,7 +5,7 @@ module Admin
     load_and_authorize_resource :event, through: :program
     load_and_authorize_resource :events_registration, only: :toggle_attendance
 
-    before_action :get_event, except: [:index, :create]
+    before_action :get_event, except: [:index, :create, :new]
 
     # FIXME: The timezome should only be applied on output, otherwise
     # you get lost in timezone conversions...
@@ -62,6 +62,7 @@ module Admin
       @comments = @event.root_comments
       @comment_count = @event.comment_threads.count
       @user = @event.submitter
+      @users = User.all.order(:name)
       @url = admin_conference_program_event_path(@conference.short_title, @event)
       @languages = @program.languages_list
     end
@@ -79,6 +80,8 @@ module Admin
     end
 
     def update
+      @users = User.all.order(:name)
+      @languages = @program.languages_list
       if @event.update_attributes(event_params)
 
         if request.xhr?
@@ -94,7 +97,26 @@ module Admin
       end
     end
 
-    def create; end
+    def create
+      @url = admin_conference_program_events_path(@conference.short_title, @event)
+      @users = User.all.order(:name)
+      @languages = @program.languages_list
+      @event.submitter = current_user
+
+      if @event.save
+        ahoy.track 'Event submission', title: 'New submission'
+        redirect_to admin_conference_program_events_path(@conference.short_title), notice: 'Event was successfully submitted.'
+      else
+        flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
+        render action: 'new'
+      end
+    end
+
+    def new
+      @url = admin_conference_program_events_path(@conference.short_title, @event)
+      @languages = @program.languages_list
+      @users = User.all.order(:name)
+    end
 
     def accept
       send_mail = @event.program.conference.email_settings.send_on_accepted
@@ -161,7 +183,8 @@ module Admin
                                     # Set only in admin/events controller
                                     :track_id, :state, :language, :is_highlight, :max_attendees,
                                     # Not used anymore?
-                                    :proposal_additional_speakers, :user, :users_attributes)
+                                    :proposal_additional_speakers, :user, :users_attributes,
+                                    speaker_ids: [])
     end
 
     def comment_params
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -8,7 +8,13 @@ class Event < ActiveRecord::Base
 
   has_many :event_users, dependent: :destroy
   has_many :users, through: :event_users
-  has_many :speakers, through: :event_users, source: :user
+
+  has_many :speaker_event_users, -> { where(event_role: 'speaker') }, class_name: 'EventUser'
+  has_many :speakers, through: :speaker_event_users, source: :user
+
+  has_one :submitter_event_user, -> { where(event_role: 'submitter') }, class_name: 'EventUser'
+  has_one  :submitter, through: :submitter_event_user, source: :user
+
   has_many :votes, dependent: :destroy
   has_many :voters, through: :votes, source: :user
   has_many :commercials, as: :commercialable, dependent: :destroy
@@ -23,6 +29,7 @@ class Event < ActiveRecord::Base
   belongs_to :program
 
   accepts_nested_attributes_for :event_users, allow_destroy: true
+  accepts_nested_attributes_for :speakers, allow_destroy: true
   accepts_nested_attributes_for :users
 
   before_create :generate_guid
@@ -33,6 +40,7 @@ class Event < ActiveRecord::Base
   validates :abstract, presence: true
   validates :event_type, presence: true
   validates :program, presence: true
+  validates :speakers, presence: true
   validates :max_attendees, numericality: { only_integer: true, greater_than_or_equal_to: 1, allow_nil: true }
 
   validate :max_attendees_no_more_than_room_size
@@ -113,20 +121,16 @@ class Event < ActiveRecord::Base
     @total_rating > 0 ? number_with_precision(@total_rating / @total.to_f, precision: 2, strip_insignificant_zeros: true) : 0
   end
 
-  def submitter
-    result = event_users.where(event_role: 'submitter').first
-    if result.nil?
-      user = nil
-      # Perhaps the event_users haven't been saved, if this is a new proposal
-      event_users.each do |u|
-        if u.event_role == 'submitter'
-          user = u.user
-        end
-      end
-      user
-    else
-      result.user
+  # get event speakers with the event sumbmitter at the first position
+  # if the submitter is also a speaker for this event
+  def speakers_ordered
+    speakers_list = speakers.to_a
+
+    if speakers_list.reject! { |speaker| speaker == submitter }
+      speakers_list.unshift(submitter)
     end
+
+    speakers_list
   end
 
   def transition_possible?(transition)
--- a/app/views/admin/events/index.html.haml
+++ b/app/views/admin/events/index.html.haml
@@ -4,7 +4,9 @@
       %h1
         Events
         = "(#{@events.length})" if @events.any?
-        .btn-group.pull-right
+        .pull-right
+          - if can? :create, Event
+            =link_to 'Add Event', new_admin_conference_program_event_path(@conference.short_title), class: 'button btn btn-default btn-info'
           - if can? :read, Event
             .btn-group
               %button.btn.btn-default.dropdown-toggle{ 'data-toggle' => 'dropdown', type: 'button', class: 'btn btn-success' }
@@ -54,8 +56,8 @@
           %th
             %b Submitter
           %th
-            %b Speaker
-          - if @program.languages.present?
+            %b Speakers
+          -if @program.languages.present?
             %th
               %b Language
           %th
@@ -99,10 +101,9 @@
                 %i Hidden
             %td
               - if @program.show_voting?
-                - if speaker = event.speakers.first
-                  = link_to speaker.name, admin_user_path(speaker)
-                - else
-                  Unknown speaker
+                - event.speakers_ordered.each do |speaker|
+                  .speaker
+                    = link_to speaker.name, admin_user_path(speaker)
               - else
                 %i Hidden
 
--- a/app/views/proposals/_proposal_form.html.haml
+++ b/app/views/proposals/_proposal_form.html.haml
@@ -4,6 +4,10 @@
 
     = f.input :subtitle, as: :string
 
+    = f.input :speakers, as: :select,
+      collection: options_for_select(@users.map {|user| ["#{user.name} (#{user.email})", user.id]}, @event.speakers.map(&:id)),
+      include_blank: false, label: 'Speakers', input_html: { class: 'select-help-toggle', multiple: 'true' }
+
     - if @program.tracks.any?
       = f.input :track_id, as: :select,
                 collection: @program.tracks.map {|track| ["#{track.name}", track.id] },
@@ -60,4 +64,15 @@
 
 
   %p.text-right
-    = f.submit 'Update Proposal', class: 'btn btn-success'
+    - if @event.new_record?
+      = f.submit 'Create Proposal', class: 'btn btn-success'
+    - else
+      = f.submit 'Update Proposal', class: 'btn btn-success'
+
+:javascript
+  $(document).ready(function() {
+    $('#event_speaker_ids').selectize({
+      plugins: ['remove_button'],
+      maxItems: 5
+    } )
+  });
--- a/app/views/proposals/show.html.haml
+++ b/app/views/proposals/show.html.haml
@@ -3,8 +3,9 @@
   %meta{ property: "og:url", content: conference_program_proposal_url(@conference.short_title, @event) }
   %meta{ property: "og:description", content: @event.abstract }
   %meta{ property: "og:site_name", content: (ENV['OSEM_NAME'] || 'OSEM') }
-  %meta{ property: "og:image", content: @speaker.gravatar_url }
-  %meta{ property: "og:image:secure_url", content: @speaker.gravatar_url }
+  - if @speakers_ordered.any?
+    %meta{ property: "og:image", content: @speakers_ordered.first.gravatar_url }
+    %meta{ property: "og:image:secure_url", content: @speakers_ordered.first.gravatar_url }
 
 .container
   .row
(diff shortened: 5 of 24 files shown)


commandprompt:
This is a draft implementation which is quite likely to be improved and extended, so we are open to discussion, please share your thoughts.
Here is our TODO list for the moment:

- Implement some kind of event speaker invitation support, so the new speaker could either accept or reject being the speaker for the event.
- Make the limit for the number of speakers per event configurable (currently hard-coded to 5)
- Write the tests for the new functionality once the implementation is finalized

best regards
Eugene/CommandPrompt Inc.

differentreality:
I am wondering if inviting speakers makes much sense. Do you really see a use for this? I mean, if you want to have multiple speakers, and the speakers agree to give that talk, the least they can do is create an account to OSEM app. And then you can simply add them in the proposal.

differentreality:
Where is the Add Event button in admin? 
Some screenshots of all those changes would be nice to have in the PR btw.

differentreality:
https://github.com/openSUSE/osem/blob/master/app/views/proposals/show.html.haml#L6 needs some attention too

differentreality:
I am happy to review this, but what do others think about the overall functionality it adds?

differentreality:
Currently someone can submit a talk and assign a different speaker. The speaker does not see the talk in the proposals#index. I wonder if we want that, or rather prefer speakers see those talks as well, in order to be able to then add their slides, for example.

eug3nix:
1 - the speaker invitation workflow may make sense to prevent the potential abuse of this feature, I mean the case when one wants to add the _already existing_ user to the speaker list, this is just an idea
2 - the Add Event button is on the top right of the Admin -> Conference -> Events screen 
addbtn
3 - I can adjust the list of events on proposals#index to show events where one either a speaker or submitter, but in this case the name of the navbar menu item should be reconsidered (My Submissions vs My Events?)

Thanks for your attention and comments, I am looking forward to making adjustments based on input from you and other team members, we just need to find some common ground on the requirements for this feature

differentreality:
@eug3nix the light blue button 'Add Event' is **not** part of this PR.

differentreality:
The Gemfile.lock that was committed in this PR still contains sprockets, which you have removed in Gemfile.

differentreality:
https://github.com/openSUSE/osem/blob/master/app/views/proposals/show.html.haml#L6 this line still raises an error and I cannot access `http://localhost:3001/admin/conferences/tconf0/program/events/new` due to get_event being called in the controller. Is it possible that you are not running this branch from master but from another branch (you made similar changes in #1257)

eug3nix:
@differentreality all of the mentioned issues were fixed

app/controllers/admin/events_controller.rb
```diff
@@ -92,7 +95,27 @@ def update
       end
     end
 
-    def create; end
+    def create
+      @url = admin_conference_program_events_path(@conference.short_title, @event)
+      @users = User.all.order(:name)
+      @languages = @program.languages_list
+      @event.submitter = current_user
+      @event.state = :confirmed
```
differentreality: I find this a bit arbitrary. I think we should allow for the evaluation process to be in place even with events created by organizers. An event might be created by an organizer long before the conference, and it makes sense that it is evaluated by the whole CfP committee before being accepted. Also I think the workflow should be that the organizer adds the real speaker of the event, and it is then the responsibility of the speaker to do the final confirmation that marks the event as confirmed. 
I realize organizer created events might be things like 'Lunch' as well, however in case it is for keynotes and such stuff it sounds safer to me if we keep the basic workflows in place. 
Others' thoughts?

eug3nix: @differentreality it is up to the team whether to keep this event.state override, the code exported in this pull request is based on out internal needs at CommandPrompt, so I just decided to keep it for now, I can remove that if you wish.
Besides that, is there anything else you wish to have changed?

differentreality: I think we should allow speakers to confirm their event manually, as confirmation of an event means a final acknowledgement that the event is accepted and the speaker will be coming. If needed, an organizer can always manually confirm events from events#index.

differentreality:
I still get an error here
https://github.com/openSUSE/osem/blob/master/app/views/proposals/show.html.haml#L6

app/controllers/admin/events_controller.rb
```diff
@@ -92,7 +95,27 @@ def update
...
+
+      if @event.save
+        ahoy.track 'Event submission', title: 'New submission'
+        redirect_to admin_conference_program_events_path(@conference.short_title), notice: 'Event was successfully submitted.'
+      else
+        flash[:error] = "Could not submit proposal: #{@event.errors.full_messages.join(', ')}"
+        render action: 'new'
```
differentreality: In this case we loose the speakers that were added, and they are no longer shown in the form. Any ideas for easily fixing this?

lagartoflojo:
How does the `selectize` lib work with potentially thousands of entries? I'm kind of worried about `@users = User.all.order(:name)`, rendering ALL the users in the db in the HTML, delivering that HTML to the clients, and the performance of `selectize` with thousands of entries.
Have you tried this scenario?

differentreality:
That will not work with ICHAIN, because when ichain is used, the users remain unconfirmed (of course that might be something we want to change).

app/views/proposals/_proposal_form.html.haml
```diff
@@ -4,6 +4,10 @@
 
     = f.input :subtitle, as: :string
 
+    = f.input :speakers, as: :select,
+      collection: options_for_select(@users.map {|user| ["#{user.name}", user.id]}, @event.speakers.pluck(:id)),
```
differentreality: Perhaps it makes sense to also show the email of the user next to the name (helps differentiate between duplicate accounts with the same name)

lagartoflojo:
@eug3nix I tried the codepen in a couple of other machines (using Firefox) and it also seems to work pretty well. So I think we can work with the current solution for now.
Still, `Users.all` looks bad, we don't need to instantiate all the users when we just need their name, email and id. I would still go for `@users = User.order(:name).pluck(:id, :name, :email)` (taking @differentreality's of showing the email as well).
Now this brings me to another point, which is that the `@users = ...` line is repeated a lot. You could move the entire `f.input :speakers...` input to a helper, and then have that helper load the users. That way, you don't need to load `@users` all over the place.

app/views/admin/events/index.html.haml
```diff
@@ -99,10 +101,12 @@
                 %i Hidden
             %td
               - if @program.show_voting?
-                - if speaker = event.speakers.first
-                  = link_to speaker.name, admin_user_path(speaker)
-                - else
-                  Unknown speaker
+                - event.speakers_ordered.each do |speaker|
```
lagartoflojo: The name of this column should be changed to "Speakers" or "Speaker(s)"

app/views/proposals/show.html.haml
```diff
@@ -1,10 +1,11 @@
+- speakers_ordered = @event.speakers_ordered
```
lagartoflojo: How about loading this variable from the controller?

app/views/proposals/show.html.haml
```diff
@@ -25,21 +26,24 @@
...
-                = @speaker.affiliation
-          -if @speaker.biography?
-            = markdown(@speaker.biography)
+      %h3
+        Presented by:
+      - speakers_ordered.each do |speaker|
+        .speakerinfo
+          .row
+            .col-md-4
+              = image_tag speaker.gravatar_url(:size => 120), :class => "img-responsive img-rounded"
```
lagartoflojo: old style ruby hash 
`:class => "img-responsive img-rounded"`
`class: "img-responsive img-rounded"`

app/views/proposals/show.html.haml
```diff
@@ -25,21 +26,24 @@
...
+            .col-md-8
+              %h4
+                = link_to speaker.name, user_path(speaker.id)
+                = "(#{speaker.email})"
+              - if speaker.affiliation?
+                .text-muted
+                  = speaker.affiliation
+          .row.speakerbio
+            .col-md-12
+              -if speaker.biography?
```
lagartoflojo: I think this `if` should go above `.row.speakerbio`, no?

app/models/event.rb
```diff
@@ -113,20 +121,32 @@ def average_rating
...
+  #     user
+  #   else
+  #     result.user
+  #   end
+  # end
+
+  # get event speakers with the event sumbmitter at the first position
+  # if the submitter is also a speaker for this event
+  def speakers_ordered
+    @speakers = speakers.to_a
```
lagartoflojo: Is an instance variable really needed here? (`@speakers`)

app/models/event.rb
```diff
@@ -113,20 +121,32 @@ def average_rating
...
+    @speakers << submitter unless @speakers.any?
+
+    if @speakers.reject! { |speaker| speaker == submitter }
+      @speakers.insert(0, submitter)
```
lagartoflojo: Line 144 adds the submitter as a speaker if there's no speakers.
Then line 146 deletes the submitter if he's one of the speakers, and then line 147 adds him again. 😕 

I'm wondering: should an event be valid if it has no speakers? Then we wouldn't need line 144.

How about:

```ruby
def speakers_ordered
  speakers_list = speakers.to_a
  submitter_index = speakers_list.find_index(submitter)
  speakers_list.unshift(speakers_list.delete_at(submitter_index)) if submitter_index && submitter_index > 0
  speakers_list
end
```

app/views/admin/events/index.html.haml
```diff
@@ -99,10 +101,12 @@
...
+                  .speaker
+                    - if speaker
+                      = link_to speaker.name, admin_user_path(speaker)
+                    - else
```
eug3nix: this was a leftover from the previous version of the code, I've removed these extra conditions completely

app/views/proposals/show.html.haml
eug3nix: the whole template uses the old syntax, I just followed that style, is there any point of having just one line using the new hash syntax? I don't think so, IMO all templates should be refactored to use new syntax, but this is out of scope of this PR

app/views/proposals/show.html.haml
eug3nix: yep, that's right, all of the adjustments I've made per your comments will be pushed to this PR soon :)

app/views/proposals/show.html.haml
differentreality: I agree we should refactor our views and while this would not be for me no 1 priority in such feature, we do try to use the new syntax as well as rubocop styling (eg. in this case single quotes) in any new code we submit.

eug3nix: Oh, thanks for the clarification I'll make the adjustments then to keep it happy :)

app/models/event.rb
eug3nix: Insertion of the submitter if there are no speakers was just a guard code which is not really needed anymore, since the should ALWAYS have at least one speaker, so this line was remove, and per your feedback regarding the instance vars I'ev ended up with this
```ruby
def speakers_ordered
  speakers_list = speakers.to_a

  if speakers_list.reject! { |speaker| speaker == submitter }
    speakers_list.insert(0, submitter)
  end

  speakers_list
end
```

lagartoflojo: K cool.
I still don't find this line `if speakers_list.reject! { |speaker| speaker == submitter }` very clear, it's not easy to read. (When does `reject!` return a falsey value?)
Also, `Array#unshift(...)` does the same as `Array#insert(0,...)`

Maybe this is more readable?

```ruby
def speakers_ordered
  speakers_list = speakers.to_a
  submitter_index = speakers_list.find_index(submitter)
  if submitter_index && submitter_index > 0
    speakers_list.delete_at(submitter_index)
    speakers_list.unshift(submitter) 
  end
  speakers_list
end
```

Although now I'm questioning whether the submitter should be the first in the list at all? Then we can just forgo this method completely.

eug3nix: While working on this I had a concept of "main speaker" for the events, it seems logical that the one who submitted an event and who also is a speaker should be shown first in the speakers list on the event public pages.

As for reject! straight from the Ruby docs

> Equivalent to #delete_if, deleting elements from self for which the block evaluates to true, but returns nil if no changes were made.

I agree that Array.unshift is more elegant and will update the code per your input.
Besides that, can we settle on something like this:
```ruby
def speakers_ordered
    speakers_list = speakers.to_a

    if speakers_list.reject! { |speaker| speaker == submitter }
      speakers_list.unshift(submitter)
    end

    speakers_list
end
```

app/controllers/admin/events_controller.rb
eug3nix: I am aware of this issues, but don't have simple and elegant solution for this ATM.
@lagartoflojo maybe you have some thoughts on the elegant the solution for this?
The thing is that form loses its nested attributes on re-render, since the event.speakers is not yet persisted, the preselected values are lost for the speakers dropdown on the form.

app/models/event.rb
lagartoflojo: > the one who submitted an event and who also is a speaker should be shown first

On the other hand, why should one speaker be "more important" than another? :)
Personally, I would keep things simple and order speakers alphabetically, but I won't insist.

app/controllers/admin/events_controller.rb
lagartoflojo: I'll test it out tonight and let you know :)

eug3nix: Nevermind, I found a solution,
it happened because the **@event.speakers.pluck(:id)**  will not return the ids for not-yet-persisted relations.
replacing this code with **@event.speakers.map(&:id) solved** the issue.

app/views/proposals/_proposal_form.html.haml
eug3nix: Agreed. I've adjusted the code per your feedback

differentreality:
Any other feedback for this feature?

lagartoflojo:
@differentreality yes, my comment here https://github.com/openSUSE/osem/pull/1278#issuecomment-283924951 still stands.

differentreality:
I would also prefer `@users = User.order(:name).pluck(:id, :name, :email)` instead of `@users = User.all`

For the sake of readability, I would **not** like to move the = f.input :speakers anywhere else, personally I am OK with loading the @users variable wherever needed. @lagartoflojo do you feel it would make sense to load it in the application controller instead? (perhaps as @users_all)

lagartoflojo:
> do you feel it would make sense to load it in the application controller instead?

I'd prefer to leave it as it is instead of making it available to the entire application via ApplicationController.

But I think that for readability, it's actually even nicer to move this input to a helper.
The HAML file would read like:

```haml
-# ...
= f.input :title, as: :string, required: true
= f.input :subtitle, as: :string
= speakers_selector_input(f)
-# ...
```

Instead of:

```haml
-# ...
= f.input :title, as: :string, required: true
= f.input :subtitle, as: :string
= f.input :speakers, as: :select,
  collection: options_for_select(@users.map {|user| ["#{user.name} (#{user.email})", user.id]}, @event.speakers.map(&:id)),
  include_blank: false, label: 'Speakers', input_html: { class: 'select-help-toggle', multiple: 'true' }
-# ...
```


And the helper would load the user data as well as render the drop down.

app/views/proposals/_proposal_form.html.haml
```diff
@@ -60,4 +64,15 @@
...
+    - else
+      = f.submit 'Update Proposal', class: 'btn btn-success'
+
+:javascript
+  $(document).ready(function() {
+    $('#event_speaker_ids').selectize({
+      plugins: ['remove_button'],
+      maxItems: 5
+    } )
+  });
```
differentreality: New line at the end of file

app/views/proposals/show.html.haml
```diff
@@ -25,21 +25,24 @@
...
-                = @speaker.affiliation
-          -if @speaker.biography?
-            = markdown(@speaker.biography)
+      %h3
+        Presented by:
+      - @speakers_ordered.each do |speaker|
+        .speakerinfo
+          .row
+            .col-md-4
+              = image_tag speaker.gravatar_url(:size => 120), class: "img-responsive img-rounded"
```
differentreality: size: 120 and single quotes for string

app/views/admin/events/index.html.haml
```diff
@@ -4,7 +4,9 @@
       %h1
         Events
         = "(#{@events.length})" if @events.any?
-        .btn-group.pull-right
+        .pull-right
+          - if can? :create, Event
+            =link_to 'Add Event', new_admin_conference_program_event_path(@conference.short_title), :class => 'button btn btn-default btn-info'
```
differentreality: new ruby style

differentreality:
@lagartoflojo unless fixed with rebasing, I will make it a junior issue to move the select. 

It's been a while, and there are no more comments or feedback, so I would like to get this merged.

differentreality:
@eug3nix please rebase and squash commits :)

I have also added a review app for this PR, https://osem-demo-pr-1278.herokuapp.com/
We would need to add examples of proposals with multiple speakers to see how the feature works.

app/controllers/admin/events_controller.rb
```diff
@@ -5,7 +5,7 @@ class EventsController < Admin::BaseController
     load_and_authorize_resource :event, through: :program
     load_and_authorize_resource :events_registration, only: :toggle_attendance
 
-    before_action :get_event, except: [:index, :create]
+    before_action :get_event, except: [:index, :create, :reports, :new]
```
differentreality: We no longer have an action `reports`

differentreality:
Let's get this in! :tada: 
Thanks @eug3nix :+1:

PR #1281, opened 2017-02-08
Better "short title" UI

sheamunion:
better_short_title_ui

### DIFF
--- a/app/views/admin/conferences/new.html.haml
+++ b/app/views/admin/conferences/new.html.haml
@@ -5,7 +5,7 @@
         = f.input :title, hint: "The name of your conference as it shall appear throughout the site. Example: 'OpenSUSE Conference 2013'",
                           input_html: { required: 'required' }
         = f.input :short_title, hint: "A short and unique handle for your conference, using only lower-case letters, numbers and underscores. This will be used to identify your conference in URLs etc. Example: 'froscon2011'",
-                                input_html: { required: 'required', pattern: '[a-z0-9-]+', title: 'Only numbers, dashes, and/or lower cased letters.' }, :prepend => 'http://myconf.com/'
+                                input_html: { required: 'required', pattern: '[a-z0-9_]+', title: 'Only lower-case letters, numbers and underscores.' }, :prepend => conferences_url + '/'
       = f.inputs 'Scheduling' do
         = f.input :timezone, as: :time_zone, default: Time.zone.name, hint: 'Please select in what time zone your conference will take place.'
         = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', required: 'required' }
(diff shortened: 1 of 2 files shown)


app/views/admin/conferences/new.html.haml
```diff
@@ -5,7 +5,7 @@
         = f.input :title, hint: "The name of your conference as it shall appear throughout the site. Example: 'OpenSUSE Conference 2013'",
                           input_html: { required: 'required' }
         = f.input :short_title, hint: "A short and unique handle for your conference, using only lower-case letters, numbers and underscores. This will be used to identify your conference in URLs etc. Example: 'froscon2011'",
-                                input_html: { required: 'required' }
+                                input_html: { required: 'required', pattern: '[a-z0-9-]+', title: 'Only numbers, dashes, and/or lower cased letters.' }, :prepend => 'http://myconf.com/'
```
lagartoflojo: Instead of hard-coding `http://myconf.com/`, could we put `conferences_url` instead, so that it uses the site's actual URL?

lagartoflojo: Also, the message should be "Only lower-case letters, numbers and underscores" (so that it's consistent with the help message, and it should allow underscores `_` rather than dashes `-`.

sheamunion: Absolutely! One point of clarification: I noted that the format validation of `:short_title` in the Conference model includes dashes `-`. I removed that from the regex to align with the the text "Only lower-case letters, numbers and underscores."

lagartoflojo:
Thanks a lot, @sheamunion!!

differentreality:
Our conference factory might produce short_titles with dash (-)
https://github.com/openSUSE/osem/blob/master/spec/factories/conferences.rb#L6
```
irb(main):046:0> SecureRandom.urlsafe_base64(4)
=> "p-GiFg"
```
Which is why some we have some random weird travis failures on all factories today.

lagartoflojo:
@differentreality Good catch! So we in theory DO allow dashes in the short_title, even though our UI did not say so.
@sheamunion would you be able to re-add the dash to the allowed characters? Sorry =/

sheamunion:
@lagartoflojo No problem! I'll update the validation. Want me to update the input field to allow dashes as well, or, do we still want to disallow dashes publicly?

lagartoflojo:
Great, thanks! I think if we allow them in the model, there's no reason not to allow them in the UI as well.. So yeah, I think adding to both places makes sense :-)

sheamunion:
@lagartoflojo It looks the conference factory will also produce capital letters so I removed all my changes to the model validation and updated the UI to include lower and upper-case letters, dashes, and underscores.

PR #1283, opened 2017-02-09
show vote changes

### LINKED ISSUE #1227: Show changes for votes in admin/events#show (history tab)
History of event does not currently show changes for votes. Versions for votes show in the overall changelog, they should also be made available in the history of the event.

siddhantbajaj:
fixed #1227 

screenshot from 2017-02-09 06 46 35

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -51,7 +51,9 @@ module Admin
       @difficulty_levels = @program.difficulty_levels
       @versions = @event.versions |
        PaperTrail::Version.where(item_type: 'Commercial').where_object(commercialable_id: @event.id, commercialable_type: 'Event') |
-       PaperTrail::Version.where(item_type: 'Commercial').where_object_changes(commercialable_id: @event.id, commercialable_type: 'Event')
+       PaperTrail::Version.where(item_type: 'Commercial').where_object_changes(commercialable_id: @event.id, commercialable_type: 'Event') |
+       PaperTrail::Version.where(item_type: 'Vote').where('object_changes LIKE ?', "%\nevent_id:\n- \n- #{@event.id}\n%") |
+       PaperTrail::Version.where(item_type: 'Vote').where('object LIKE ?', "%\nevent_id: #{@event.id}\n%")
     end
 
     def edit
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -215,6 +215,7 @@ class Ability
     end
 
     can [:index, :revert_object, :revert_attribute], PaperTrail::Version, item_type: 'Event', conference_id: conf_ids_for_cfp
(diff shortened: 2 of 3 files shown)


differentreality:
We should include the vote versions in the @versions variable of the controller instead.

app/controllers/admin/events_controller.rb
```diff
@@ -52,6 +52,7 @@ def show
       @versions = @event.versions |
        PaperTrail::Version.where(item_type: 'Commercial').where_object(commercialable_id: @event.id, commercialable_type: 'Event') |
        PaperTrail::Version.where(item_type: 'Commercial').where_object_changes(commercialable_id: @event.id, commercialable_type: 'Event')
+      add_vote_versions_to_versions
```
differentreality: Why don't you add votes the way we add commercials in the above lines?

differentreality:
Ping

app/controllers/admin/events_controller.rb
siddhantbajaj: Sorry for late response but I tried that approach it did not work out. Since it creates a like query on object_changes attribute some vote versions of a particular event were shown in other events too.

differentreality: Why was that happening and how can I reproduce it?

siddhantbajaj:
okay so suppose there are 2 events
Event 1{ id:1}
Event 2{ id:2}
A user have rated Event 1 lets say,
Vote 1{ id:2, event_id:1}
Now the object_change attribute of this vote version will be something like this "event_id:1 , id:2 "(NOTE it contains "event_id" and "2" though event_id is not 2)
Now the query =>PaperTrail::Version.where(item_type: 'Vote').where_object_changes(event_id: 2)
will search for "event_id" and "2" in object_change attribute.
So this vote version will be shown in Event 2 history as well.

differentreality:
True, it's a known issue in paper trail as well. 
In that case, why aren't we manually forming the where statement instead of using `where_object` and `where_object_changes`?
I am also wondering if the commercial has the same issue, worth exploring, perhaps you can open a new issue about it?

siddhantbajaj:
> In that case, why aren't we manually forming the where statement instead of using where_object and where_object_changes?

yes I was thinking the same, will do the changes.

> I am also wondering if the commercial has the same issue, worth exploring, perhaps you can open a new issue about it?

yep will look into it and create an issue if needed.

differentreality:
What about reverting those changes? Except for organizer, other roles are able to conditionally revert things related to Event and Commercial, but not Vote.

differentreality:
Ping @siddhantbajaj

siddhantbajaj:
@differentreality Sorry for the late response, might have missed the notification mail.

differentreality:
Looks good :+1: 
Please rebase and squash your commits for merging!

siddhantbajaj:
done. :+1:

Issue #1259, opened 2017-01-30
Change homepage to next conference#show


davidsf:
We use osem as annual conference organizer, currently osem show a list of next conferences as homepage but should be great if it could go to the next conference#show if there are only one conference configured.

We have done a hack to do this in http://www.opensouthcode.org but if there are any interest I could try to implement it.

lagartoflojo:
I very much like this idea. @Ana06 @differentreality @ChrisBr thoughts?

differentreality:
I'd go for it too.

lagartoflojo:
@davidsf would you like to implement this?

ChrisBr:
@lagartoflojo http://www.pgconf.us/conferences/2017 did the same! So yes, I also think it makes sense

Ana06:
I like it too

divyanshumehta:
Can I try it out @Ana06 , please. Doesn't it need only the routes file to be changed?

ChrisBr:
@divyanshumehta go ahead, you don't need to ask if you want to work on an issue!

Rupal-IIITD:
@davidsf 
I didn't quite understand as to what you mean by "go to the next conference#show if there are only one conference configured"?
Can I request you to elaborate a bit more?

davidsf:
@Rupal-IIITD 
We use osem for an annual conference, for us is better to go to the conference#show of the next conference, instead of listing the conferences.

davidsf:
What is the state of this? I see some implementation by @divyanshumehta. It is completed ?

differentreality:
No, unfortunately not @davidsf but it is in my plans to have this done soon.

bear454:
@davidsf please consider closing this PR, and supporting the more flexible change in #1994 .

differentreality:
It seems we should let people choose what they want:
1. always redirect to conferences#show (if there is an upcoming conference), this will make it impossible to access conferences#index, and thus see a list of past conferences.
2. no change - you get to see conferences#index (which btw has buttons for different options for each conference listed there!)
3. we have been also discussing redirecting to conferences#show only if there is only 1 conference in the OSEM installation - I see this being helpful and relevant only for one-time-only conferences (unlikely they will install OSEM just to host 1 single conference and nothing more, ever)

https://github.com/openSUSE/osem/pull/1867

bear454:
All of these scenarios are handled by #1994.

PR #1289, opened 2017-02-11
Issue #1259

divyanshumehta:
Solves the Issue #1259 . Counts currently ongoing conference and redirects to the conference#show view if there is only one live conference

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -7,7 +7,8 @@ class ConferencesController < ApplicationController
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
     if @current.count == 1
-      redirect_to "/conferences/#{@current.first.id}"
+      # redirect_to "/conferences/#{@current.first.short_title}"
+      redirect_to conference_path(@current.first.short_title)
     end
     @antiquated = @conferences - @current
   end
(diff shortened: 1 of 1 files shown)


differentreality:
@divyanshumehta can you please add in the comments a description of what your PR is doing and which issue it is solving?

app/controllers/conferences_controller.rb
```diff
@@ -6,6 +6,9 @@ class ConferencesController < ApplicationController
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+    if @current.count == 1
+      redirect_to "/conferences/#{@current.first.id}"
```
shlok007: Did you test the changes in your local machine first? This won't work as routing for conferences is by `short_title` and not `id` .
Also, I suggest using named routes instead. :slightly_smiling_face:

divyanshumehta: Yes @shlok007 it does not work, as routing is based on short_title, also when I the URL `/conferences/:id` with the short_title in `params[:id]`, it redirects me back to homepage saying to login and even after login in it only works for the admin, any user or non-user cannot acess the conference#show show authorization needs to be relaxed in this case.

nikhilgupta1211: @divyanshumehta  the conference should have a splash page and public access for anyone to visit the link

PR #1291, opened 2017-02-12
Redirect to conference#show if there is only one current conference

divyanshumehta:
#1259 
If only one conference is live then it redirects the homepage to conference#show view.

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -2,7 +2,7 @@ class ConferencesController < ApplicationController
   protect_from_forgery with: :null_session
   before_action :respond_to_options
   load_and_authorize_resource find_by: :short_title
-  load_resource :program, through: :conference, singleton: true, except: :index
+  load_resource :program, through: :conference, singleton: true, except: [:index, :current]
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
@@ -10,6 +10,11 @@ class ConferencesController < ApplicationController
   end
 
   def show; end
+
+  def current
+    current = Conference.first
+    redirect_to conference_path(current.short_title)
+  end
 
   private
 
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -13,6 +13,7 @@ class ConferencesController < ApplicationController
 
   def current
     current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current).first
+    current = Conference.where('start_date = ?', Date.current + 1).first if current.blank?
     redirect_to conference_path(current.short_title)
   end
 
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -12,7 +12,7 @@ class ConferencesController < ApplicationController
   def show; end
 
   def current
-    current = Conference.first
+    current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current).first
     redirect_to conference_path(current.short_title)
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,4 @@
+require "root_route_constraint"
 Osem::Application.routes.draw do
 
   if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
@@ -147,6 +148,8 @@ Osem::Application.routes.draw do
   end
 
   get '/admin' => redirect('/admin/conferences')
-
+  constraints RootRouteConstraint.new do
+    get '/' => 'conferences#current'
+  end
   root to: 'conferences#index', via: [:get, :options]
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,4 +1,4 @@
-require "root_route_constraint"
+require 'root_route_constraint'
 Osem::Application.routes.draw do
 
   if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
--- a/lib/root_route_constraint.rb
+++ b/lib/root_route_constraint.rb
@@ -1,6 +1,7 @@
 class RootRouteConstraint
   def initialize
-    @current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current).reorder(start_date: :asc)
+    @current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current)
+    @current = Conference.where('start_date = ?', Date.current + 1) if @current.blank?
   end
 
   ##
@@ -10,10 +11,6 @@ class RootRouteConstraint
   # * +true+ -> only one conferene is live AND the conference has public splashpage
   # * +false+ -> no or more than one conferences are live or the only live conferece has no public splashpage
   def matches?(*)
-    if @current.present? && @current.first.splashpage.present?
-      @current.count == 1 && @current.first.splashpage.public?
-    else
-      false
-    end
+    @current.present? && @current.first.splashpage.present? && @current.count == 1 && @current.first.splashpage.public?
   end
 end
--- a/lib/root_route_constraint.rb
+++ b/lib/root_route_constraint.rb
@@ -1,0 +1,10 @@
+class RootRouteConstraint
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+  end
+
+  def matches?(*)
+    return unless @current.present? && @current.first.splashpage.present?
+    @current.count == 1 && @current.first.splashpage.public?
+  end
+end
--- a/lib/root_route_constraint.rb
+++ b/lib/root_route_constraint.rb
@@ -1,10 +1,19 @@
 class RootRouteConstraint
   def initialize
-    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+    @current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current).reorder(start_date: :asc)
   end
 
+  ##
+  # Checks if only one conference is live and has a public splashpage
+  # If any conference is live it checks how many live conferences are there
+  # ====Returns
+  # * +true+ -> only one conferene is live AND the conference has public splashpage
+  # * +false+ -> no or more than one conferences are live or the only live conferece has no public splashpage
   def matches?(*)
-    return unless @current.present? && @current.first.splashpage.present?
-    @current.count == 1 && @current.first.splashpage.public?
+    if @current.present? && @current.first.splashpage.present?
+      @current.count == 1 && @current.first.splashpage.public?
+    else
+      false
+    end
   end
 end
--- a/spec/controllers/conferences_controller_spec.rb
+++ b/spec/controllers/conferences_controller_spec.rb
@@ -26,6 +26,13 @@ describe ConferencesController do
     end
   end
 
+  describe 'GET #current' do
+    it 'redirects to first conference splashpage' do
+      get :current
+      expect(response).to redirect_to(conference_path(Conference.first.short_title))
+    end
+  end
+
   describe 'OPTIONS #index' do
     it 'Response code is 200' do
       process :index, 'OPTIONS'
--- a/spec/lib/root_route_constraint_spec.rb
+++ b/spec/lib/root_route_constraint_spec.rb
@@ -1,0 +1,39 @@
+require 'spec_helper'
+
+describe RootRouteConstraint do
+  describe '#matches?' do
+
+    it 'returns false, if no conference is live' do
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq false
+    end
+
+    it 'returns false, when one conference is live but no splashpage' do
+      create(:conference)
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq false
+    end
+
+    it 'returns false, when one conference is live but no public splashpage' do
+      conference = create(:full_conference)
+      conference.splashpage.public = false
+      conference.splashpage.save
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq false
+    end
+
+    it 'returns true, when one conference is live and has public splashpage' do
+      create(:full_conference)
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq true
+    end
+
+    it 'returns false, if more than one conference is live' do
+      create(:full_conference)
+      create(:full_conference)
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq false
+    end
+
+  end
+end
--- a/spec/lib/root_route_constraint_spec.rb
+++ b/spec/lib/root_route_constraint_spec.rb
@@ -34,6 +34,5 @@ describe RootRouteConstraint do
       constraint = RootRouteConstraint.new
       expect(constraint.matches?).to eq false
     end
-
   end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -29,7 +29,7 @@ class Ability
 
   # Abilities for not signed in users (guests)
   def not_signed_in
-    can [:index], Conference
+    can [:index, :current], Conference
     can [:show], Conference do |conference|
       conference.splashpage && conference.splashpage.public == true
     end


differentreality:
The problem with this approach is that we can never really access conferences#index, thus we can never browse the old conferences.

nikhilgupta1211:
There is one more problem in this i.e if the conference available does not have a public access then the root path will not work as conference#show will be unavailable except for the admin.

nikhilgupta1211:
There is one more problem in this i.e if the conference available does not have a public access then the root path will not work as conference#show will be unavailable except for the admin.

divyanshumehta:
Yes. but isn't that what the issue demands. In order to access conference#index we can make a params[:ref] which will be assigned to a link `Conferences Index` on the navbar and clicking it will redirect to conference#index .

nikhilgupta1211:
i think we should only redirect if `@current.first.splashpage.public` is true .

differentreality:
There are 2 things to take into consideration:
* The redirection can take place only if the user has access to the page you are redirecting to, ie conferences#show
* We need to still be able to access conferences#index, if we want to

divyanshumehta:
Now clicking on the logo in the navbar will redirect to conference#index and the also checks for splashpage with public access.

app/views/layouts/_navigation.html.haml
```diff
@@ -7,7 +7,7 @@
         %span.icon-bar
         %span.icon-bar
         %span.icon-bar
-      = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path, class: 'navbar-brand', title: 'Open Source Event Manager'
+      = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path(:ref => 'nav'), class: 'navbar-brand', title: 'Open Source Event Manager'
```
differentreality: This now appends `/?ref=nav` to our link, which is not that nice!

differentreality:
So basically now, I still cannot navigate to the conferences#index

In my opinion, what we should be changing is the behavior of the root_path, not the behavior of conferences#index

divyanshumehta:
OK. We can set up the root to conference#show in case of a single current conference using routing constraints, and conferences#index elsewise .In order to access the `conferences#index` we can set up a new route to ``` '/index' => 'conferences#index' ``` so we can directly acces both the views, beacause at a time either conferences#index or conferences#show can be the root, so will be need to setup an independent route to conferences#index

What you say @differentreality

differentreality:
I think we should only change the root path route.

divyanshumehta:
Hmm so, i have got another idea the root path routes to a method in conferences which then tanfers to either show or index based on the condition ie If true redirect to conferences#show else conferences#index where both would be having independent routes

divyanshumehta:
Check this implementation. The root is set confernces#show if condition is satisfied and you can still access the confereces#index via the URL `/conferences`

differentreality:
@divyanshumehta let's write up in details what we need to do in each case.

1. If there is only 1 upcoming conference
  - root path should redirect to conference#show `/ => /conferences/myconf`
  - conference#index should be accessible in `/conferences`
2. else (there are multiple upcoming conferences), things remain as they are

divyanshumehta:
In the latest commit I solved all the points mentioned above, and made changes only to the routes file as you said. Just check them once. It look good to me now.

differentreality:
@divyanshumehta I have checked your changes. You are not *redirecting* to conferences#show.
Rendering the content of conferences#show and redirect to that path are two different things. Do you understand what I mean?

differentreality:
@divyanshumehta I thought your reaction to my comment meant you understood what I was saying, but I see on IRC that you still have questions. So what's your question exactly? 
In your PR you render the conferences#show template, but you don't redirect there, meaning the link in the browser is not /conferences/myconf. Do you understand what I mean?

divyanshumehta:
Oh yes I see what you say is right. The URL is "/" whereas the page is the conferences#show. That is because the I set the root to conferences#show. So I need to send the user the URL rather than bringing the page to the home URL

differentreality:
Yes, that is what I would suggest.

divyanshumehta:
Please check this one I think this is what we need.

divyanshumehta:
@differentreality how can we improve this PR to so that is is good enough to be merged?

config/routes.rb
```diff
@@ -1,3 +1,13 @@
+class RootConstraint
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+  end
+ 
+  def matches?(request)
+    @current.count==1 && !@current.first.splashpage.blank? && @current.first.splashpage.public?
```
ChrisBr: ``
@current.count == 1 && @current.first.splashpage.present? && @current.first.splashpage.public?
``

config/routes.rb
```diff
@@ -1,3 +1,13 @@
+class RootConstraint
```
ChrisBr: RootRouteConstraint

config/routes.rb
```diff
@@ -1,3 +1,13 @@
...
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
```
ChrisBr: Thinking about if it makes sense to just pass the conference as parameter

```
def initialize(conference)
  @current = conference
end
```
and then instantiate like this

```
conference = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
get '/' => redirect("/conferences/#{conference.first.short_title}"), constraints: RootConstraint.new(conference)
```

config/routes.rb
```diff
@@ -147,6 +157,7 @@
   end
 
   get '/admin' => redirect('/admin/conferences')
-
+  conf = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc).first
+  get '/' => redirect("/conferences/#{conf.short_title}"), constraints: RootConstraint.new
```
ChrisBr: Why do you create the root different here?
Does 
```
    root to: redirect("/conferences/#{conf.short_title}"), constraints: RootConstraint.new
```
not work? (or something similar...)

config/routes.rb
```diff
@@ -147,6 +157,7 @@
   end
 
   get '/admin' => redirect('/admin/conferences')
-
+  conf = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc).first
```
ChrisBr:
```
conference = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc).first
```

ChrisBr:
@differentreality I would go for the route constraint!

ChrisBr:
@divyanshumehta also a test would be nice!

ChrisBr:
@divyanshumehta is the OSEM logo on the top left navbar a link to conferences#index?

divyanshumehta:
No the OSEM logo is link to root. Earlier I used it to go to root with a reference, but now it is just going to root.

config/routes.rb
divyanshumehta: I tried it that way but it throws and error saying that a named route is being used for two different URLs. So I did it this way and it worked.

divyanshumehta:
@ChrisBr what type of test should be added.

config/routes.rb
divyanshumehta: This creates as this conference would be passed to the initialize method of the RootRouteConstraint and there `@current.count` becomes undefined. so I am replacing it with 
`conference = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)`

divyanshumehta:
Can anybody please tell why the builds are failing?

siddhantbajaj:
I am not sure about this but I think since you are using Conference model in routes.rb it assumes that the table exist in the db but since the migration haven't run yet so the table is not yet created.We can fix this locally but commenting that line and the running our migrations but i don't know how to deal with this on travis.

nikhilgupta1211:
i think a better approach will be 
defining the class `RootRouteConstraint` in  `/app/constraints/root_route_constraint.rb`
image

for routes.rb 
image

and the action in ` conferences_controller.rb`
image
It should also fix the build failing.

And it also covers the points mentioned by @ChrisBr

differentreality:
@divyanshumehta does the test suite run successfully on your machine?

divyanshumehta:
@nikhilgupta1211 Putting the `redirect_to` in the `conferences_controller.rb` creates a too many redirects error. Also `app/constraints/` is not in the load path so I am putting the `root_constraints.rb` in `lib/`. I read that it is always better to make new user classes in `lib/` .

nikhilgupta1211:
@divyanshumehta did you allowed the current action in ability.rb for CanCan?If not please do
change load_resource in `conferences_controller.rb `to
` load_resource :program, through: :conference, singleton: true, except: [:index, :current]`

and in `ability.rb`
```

def not_signed_in
    can [:index,:current], Conference
            ....

```

divyanshumehta:
@nikhilgupta1211 It works but it fails the test.

@differentreality test suite runs on my local.

nikhilgupta1211:
@divyanshumehta one of the failure is due to the offenses by Rubocop(so please remove them) and the other tests they are checking a return to `root_path` but as we are redirecting to one of the conferences they are failing to pass the tests we have to redirect when no user is signed in or we have to change the tests accordingly . 
Before implementing this please wait for @differentreality `s feedback

divyanshumehta:
The rubocop tests are now passing only the redirect ones remain.
@differentreality please look into it. Need your help.

app/controllers/conferences_controller.rb
```diff
@@ -11,6 +11,11 @@ def index
 
   def show; end
 
+  def current
+    current = Conference.first
+    redirect_to "/conferences/#{current.short_title}"
```
ChrisBr: Don't! Please use the rails url helper http://guides.rubyonrails.org/routing.html

lib/root_route_constraint.rb
```diff
@@ -0,0 +1,9 @@
+class RootRouteConstraint
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+  end
+
+  def matches?(*)
+    @current.count == 1 && @current.first.splashpage.public?
```
ChrisBr: You need to check for ``@current`` and also for ``splashpage``. Your tests fail because ``splashpage`` is nil!

divyanshumehta: So basically I need to do this 
```
 def matches?(*)
  	unless @current.blank? && @current.first.splashpage.blank?
  		if @current.present? && @current.first.splashpage.present?
  			@current.count == 1 && @current.first.splashpage.public?
  		end
  	end
  end
```

divyanshumehta:
@ChrisBr what tests do we need could you please elaborate a little.

app/controllers/conferences_controller.rb
divyanshumehta: This is working fine 
```
redirect_to conference_path(current.short_title)
```

ChrisBr:
@divyanshumehta e.g. a controller test of the current action and also a test for RootRouteConstraint.
Please also rebase to fix the hakiri failure!

divyanshumehta:
I wrote the tests, but I think I am making some mistake, can anyone please help.

spec/controllers/conferences_controller_spec.rb
```diff
@@ -26,6 +26,13 @@
     end
   end
 
+  describe 'GET #current' do
+    it 'redirects to first conference splashpage' do
+      current = Conference.first
+      expect(response).to redirect_to(conference_path(current.short_title))
```
differentreality: You are not invoking an action in the controller. The thing this test is testing, when is it supposed to happen?

spec/controllers/conferences_controller_spec.rb
```diff
@@ -26,6 +26,13 @@
     end
   end
 
+  describe 'GET #current' do
+    it 'redirects to first conference splashpage' do
+      current = Conference.first
```
differentreality: You are not supposed to set this variable in the test.

spec/lib/root_route_constraint_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+
+describe RootRouteConstraint do
+  describe '#matches' do
+    it 'return true if only one conference is live' do
+      current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
```
differentreality: You are not supposed to set this in the test, you are supposed to create a test environment with all the necessary resources you need in order to check if what you are expecting to happen, is actually happening.
In this case you are testing
1. if a conference exists, and the conference has the attributes we are interested in, then it matches
2. if a conference does not exist, then it does not match
3. if a conference exists but it does not have the necessary attributes, then it does not match

spec/controllers/conferences_controller_spec.rb
divyanshumehta: this test will only test that the action redirects to the splashpage for the first conferences. That is the only thing this `conference#current`  is supposed to do.

This is my first ever time trying to write tests, so I could use a bit of help please.

spec/controllers/conferences_controller_spec.rb
divyanshumehta: ok should i directly write
`expect(response).to redirect_to(conference_path(Conference.first.short_title))`

spec/controllers/conferences_controller_spec.rb
divyanshumehta: So I should add a line `get :current` just after the it 'redirects to first conference splashpage' do`

differentreality:
@divyanshumehta you have failing tests. Please fix them.

divyanshumehta:
int the `spec/lib/root_route_constraint_spec.rb` will the following tests do good
```

    it 'does match when only one live conference and has public splashpage' do
      if Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc).first.splashpage.public?
        RootRouteConstraint.new.matches?.should be_true
      else
        RootRouteConstraint.new.matches?.should be_false
      end
    end

    it 'does not match when live conference does not exist' do
      if Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc).blank?
        RootRouteConstraint.new.matches?.should be_false
      end
    end
```

differentreality:
No! You are not going to use if statements, you are going to create the situation you want, eg. create a conference,enable a splashpage, etc. 

And please add a commit with the code  you need reviewed. If it doesn't work out, you can always revert that commit.

differentreality:
@divyanshumehta ping

divyanshumehta:
@differentreality I am not able to find a way to do the `root_route_constraint_spec.rb` . I dont understand how to implement the test. Only implementation of the test is failing rest features are good.

divyanshumehta:
The test passes on my local but fails here on travis. Here is my test result  https://pastebin.com/GVRDcPS1

spec/factories/conferences.rb
```diff
@@ -54,5 +54,41 @@
         conference.save!
       end
     end
+
+    factory :semi_conference do
```
divyanshumehta: This factory created a full_conference but with a non public splashpage. As a situation may occur, where only one conference is love but has no public splashpage. So to test this situation i made a semi _conference .

divyanshumehta: It creates a conference but the splashpage is not public

app/controllers/conferences_controller.rb
```diff
@@ -11,6 +11,11 @@ def index
 
   def show; end
 
+  def current
+    current = Conference.first
```
divyanshumehta: Yes this needs to be changed.

lib/root_route_constraint.rb
```diff
@@ -0,0 +1,13 @@
+class RootRouteConstraint
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+  end
+
+  def matches?(*)
+    if @current.present? && @current.first.splashpage.present?
+      @current.count == 1 && @current.first.splashpage.public?
```
divyanshumehta: yes if `@current.count == 1 && @current.first.splashpage.public?` is true then it return true else false, but first there need to check if `@current` and `@current.first.splashpage` are not nil.

spec/factories/conferences.rb
divyanshumehta: i think it can also be called a private_conference

differentreality: all those names are confusing, and you are duplicating code in the factory, why don't you just use the existing full_conference factory and simply make the splashpage **not** be public in your test?

lib/root_route_constraint.rb
differentreality: so why aren't these in one line?

lib/root_route_constraint.rb
```diff
@@ -0,0 +1,13 @@
+class RootRouteConstraint
+  def initialize
+    @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
+  end
+
+  def matches?(*)
```
differentreality: Did you check the format of the comments as I showed you?

lib/root_route_constraint.rb
divyanshumehta: there needs to be a check for the presence of any conferences or the presence of splashpage or else `@current.first.splashpage.public?` give Nil:Class error. So if statement is used to check the presence.

spec/factories/conferences.rb
divyanshumehta: ok will change it

divyanshumehta:
Help needed for `spec/lib/root_route_constraint.rb` . Only one test fails, I cannot figure it why.

spec/lib/root_route_constraint_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+describe RootRouteConstraint do
+  describe '#matches?' do
+
```
ChrisBr: Please move ``constraint = RootRouteConstraint.new`` to a ``let`` as you use it in every single test

spec/lib/root_route_constraint_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it 'returns false, when one conference is live but no splashpage' do
+      create(:conference)
+      constraint = RootRouteConstraint.new
+      expect(constraint.matches?).to eq false
+    end
+
+    it 'returns false, when one conference is live but no public splashpage' do
+      conference = create(:full_conference)
+      conference.splashpage.public = false
```
ChrisBr: You probably need to save the conference

divyanshumehta: I think there is some problem with the lines no 24-28

divyanshumehta: only the spec `return true, when one conference is live and has public splashpage` fails. When I test through the console `constraint.matches?` returns true. Still don't know why it fails.

divyanshumehta: Also I think there is some problem because all the spec in `root_route_constraint_spec.rb` are returning `flase` whereas the spec `return true, when one conference is live and has public splashpage`  must return `true`

divyanshumehta: Changes Made/

spec/lib/root_route_constraint_spec.rb
divyanshumehta: `let` initialize the `constraint` variable before the Factories are created, and hence the `constraint` variable will not be referring to the correct database configuration. `constraint = RootRouteConstraint.new` only needs to be run after the factories are created for a particular example and the creation for every example is different.

lib/root_route_constraint.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  ##
+  # Checks if only one conference is live and has a public splashpage
+  # If any conference is live it checks how many live conferences are there
+  # ====Returns
+  # * +true+ -> only one conferene is live AND the conference has public splashpage
+  # * +false+ -> no or more than one conferences are live or the only live conferece has no public splashpage
+  def matches?(*)
+    if @current.present? && @current.first.splashpage.present?
```
differentreality: So why isn't this 1 line?

app/controllers/conferences_controller.rb
```diff
@@ -11,6 +11,11 @@ def index
 
   def show; end
 
+  def current
+    current = Conference.where('start_date <= ? AND end_date >= ?', Date.current, Date.current).first
```
differentreality: What if the conference starts tomorrow? (And there is only 1 conference that hasn't ended yet).

divyanshumehta: if conference.start_date is tomorrow and one conference is live then the `current` variable will have the live conference and the next day both the conferences will be there in `current`. So the next day the home page will be the index action

differentreality: What if there is only 1 conference, which starts tomorrow?
We should redirect to the show action for this conference.

lib/root_route_constraint.rb
divyanshumehta: What should be in 1 line. I documented the code as you said https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L95

app/controllers/conferences_controller.rb
divyanshumehta: Ok. Making the change to include conference that start tomorrow.

lib/root_route_constraint.rb
differentreality: As you can see my comment is on line 13, so that's your code, not the comments.

The code can be:
`def matches?(*)

   @current.present? && @current.first.splashpage.present? && @current.count == 1 && @current.first.splashpage.public?

end
`

divyanshumehta:
@differentreality can you please look into why the test suite did not build on travis and review the changes that have been made

differentreality:
@divyanshumehta you should look at the log and try to understand why the failure occurs. You should also run rspec locally to check if all tests are passing, and to try and recreate the issue of Travis (after you investigate the logs and get an idea of what is going on). Once you do that, let's discuss.

divyanshumehta:
I checked the logs the line 
`current ||= Conference.where('start_date = ?', Date.current + 1).first`
does not allow the test suite to build.

differentreality:
This is inactive for a while now, and is also getting too complicated.

@divyanshumehta don't get discouraged by that. You already have another PR open, so if you still want to contribute, you can focus on that, and then in other junior issues.

PR #1293, opened 2017-02-12
intersecting events without overlapping in admin/schedule#show

differentreality:
Following up from #1168 does something like this make sense @Ana06 ? Does it solve all you needed about the overlapping issue and allow you to continue with #1106 ?

This introduces a new attribute for event_schedule, namely `enabled`.

When there is a selected schedule and an event is scheduled, if the event is withdrawn, then the event_schedule is **not** destroyed, but `event_schedule.enabled = false`

The EventSchedule model has a default scope to ignore records that are not enabled (ie.` enabled == false`).

We take the disabled event_schedules into account, only when we want to show intersecting events.

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -124,6 +124,15 @@ module Admin
 
     def cancel
       update_state(:cancel, 'Event canceled!')
+      selected_schedule = @event.program.selected_schedule
+      event_schedule = EventSchedule.unscoped.where(event: @event).find_by(schedule: selected_schedule) if selected_schedule
+      Rails.logger.debug "schedule: #{selected_schedule.inspect} and event_schedule #{event_schedule.inspect}"
+      if selected_schedule && event_schedule
+        event_schedule.enabled = false
+        event_schedule.save
+      else
+        @event.event_schedules.destroy_all
+      end
     end
 
     def reject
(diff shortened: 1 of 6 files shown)


Ana06:
@differentreality sorry I didn't take a look at this before.

Yes, I think something like this should work, but we should have to change things in more places right? As we need to set it to enabled again when scheduling it.

And I have one doubt. Why if there is not selected schedule and an event is withdrawn in one scheduled is it removed for the rest? Shouldn't we give the `withdrawn` method the schedule as well?

Ana06:
I even wonder if it would be easier for admin to be able to set a withdrawn attribute manually instead of doing it automatically in the schedule. It is quite confusing, as you could withdrawn an event by accident and there is no why to undo it, right? And also I doubt that this is currently working properly.

@ChrisBr @hennevogel ideas?

differentreality:
With the scope
`default_scope { where(enabled: true) }`
we shouldn't need more changes actually, but I will triple check anyway. Unless you are thinking of something specific?

Why would we need the withdrawn state in the eventschedule? The event has the state withdrawn. We can either remove it from all schedules (supposing there is no selected one), or show it with the label withdrawn. 

What do you mean set a withdrawn attribute manually? I am not following. 
Withdrawn state is triggered when a speaker withdraws the proposal (from proposals#index) or when an organizer cancels a proposal from admin/events#index. 
If an organizer cancels an event, it can mark it again as in 'New' state, and then accept it. 

Let me know if you meant something else.

I will work on this in the coming days.

differentreality:
@Ana06 shall we talk about this when you have time? 
Also are you thinking of finishing  #1106 and #1168?

PR #1294, opened 2017-02-13
Fixed #853 Made Sidebar collapsible for small screens.

### LINKED ISSUE #853: admin tickets view sidebar on small screens
admin's tickets view for conferences on small devices has sidebar uncollapsible
screen shot 2016-03-09 at 10 14 33 pm

nikhilgupta1211:
#853 

image

image

### DIFF
--- a/app/assets/stylesheets/osem.css.scss
+++ b/app/assets/stylesheets/osem.css.scss
@@ -94,3 +94,9 @@ p.comment-body {
 .box{
   height: 230px;
 }
+
+/* sidebar hamburger btn */
+.side-nav-btn{
+ margin-left: 10px;
+ float: left;
+}
--- a/app/views/layouts/_admin_sidebar.html.haml
+++ b/app/views/layouts/_admin_sidebar.html.haml
@@ -1,4 +1,4 @@
-%ul.nav.nav-stacked.nav-pills.mySidebar
+%ul.nav.nav-stacked.nav-pills.mySidebar.collapse.navbar-collapse#side-nav
   .btn-group
     %button{type:'button', class: 'btn btn-default btn-link dropdown-toggle', 'data-toggle'=>'dropdown'}
       %span.fa.fa-cog
--- a/app/views/layouts/_navigation.html.haml
+++ b/app/views/layouts/_navigation.html.haml
@@ -1,7 +1,13 @@
 .navbar.navbar-default.navbar-fixed-top.nav-osem{role: 'navigation'}
   .container
     .navbar-header
-      %button{"data-target"=>".navbar-collapse", "data-toggle"=>"collapse", class: 'navbar-toggle', type: 'button'}
+      - if @conference && @conference.short_title.present?
+        %button{ "data-target"=>"#side-nav", "data-toggle"=>"collapse", class: 'navbar-toggle side-nav-btn', type: 'button' }
+          %span.sr-only Toggle navigation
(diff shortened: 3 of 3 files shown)


nikhilgupta1211:
@Ana06 please can you have a look at this and tell if it is ok for the sidebar to come from the top or it should come from the side?

app/views/layouts/_admin.html.haml
```diff
@@ -1,14 +1,20 @@
 .container-fluid
+  - if @conference and not @conference.short_title.nil?
```
nikhilgupta1211: Actually i thought of this only but as we are using the condition `if @conference and not @conference.short_title.nil?` for the admin_sidebar  so i used it here also &#128517;. Should i push the changes ?

differentreality: Please commit the change @ChrisBr asked for!

nikhilgupta1211: Done 😊

nikhilgupta1211:
@ChrisBr i have made the changes

nikhilgupta1211:
@differentreality can you please have a look at this :)

differentreality:
@ChrisBr is already assigned on this issue, so he will take a look when he can. And frankly I prefer someone with more experience with frontend stuff to review this. This is a bit too much css for my taste.

differentreality:
@nikhilgupta1211 this should be possible using the bootstrap way, without adding the extra CSS. Wanna give it a go?

nikhilgupta1211:
@differentreality sure :), In this for collapse I am using bootstrap and the extra css is only for the animation of cross converting into the bars. Can you please explain a little more that how should I change it ?

differentreality:
I am not sure what you are asking. What does 'cross converting into the bars' mean?
What I am saying is look at the hamburger menu we already have, it is generated automatically by bootstrap without requiring any extra CSS. Why not just do the same for the second hamburger menu?

nikhilgupta1211:
Sorry, actually I meant this
image
Actually I didn't used the bootstrap ham their, because it would have been same as the navbar ham.
Should I change it to the bootstrap ham menu?

differentreality:
Yes, it makes sense to use the tools we already have, rather than introducing custom CSS.

nikhilgupta1211:
Ok, thanks :)

nikhilgupta1211:
image

image

@differentreality I have removed the extra css, please have a look at it :)

differentreality:
Why aren't you removing the rest of the CSS you have added? You can implement the same thing using bootstrap navbar.

nikhilgupta1211:
@differentreality, we are using
```
position: absolute; // for positioning otherwise If I would use it in a navbar-header class of bootstrap its position will not be correct, also it will shift the sidebar down
  z-index: 1000; // as we want the button to be on the navbar otherwise it will not be clickable when the bar is expanded
  right: -10px;
  top: 50px;
  margin-bottom: 10px;

  .icon-bar { // we have defined white color as default for the of navbars in osem so I changed the color here to black so it would be visible
    background-color: black;
  }
```
Also we are not using the .navbar.navbar-default classes for the sidebar.
Sorry, but can you please tell how should I do this with bootstrap?

differentreality:
Why not use navbar classes?

nikhilgupta1211:
@differentreality please have a look at this :)

image

Issue #1040, opened 2016-06-15
Show email in public option


ChrisBr:
At the moment we show the email of every speaker in
https://github.com/openSUSE/osem/blob/master/app/views/proposal/show.html.haml#L26

Some speakers don't want that. We should introduce an option to specify if a speaker wants that his email is public displayed or not. Or drop showing the email at all.

For now I hotfixed this on events.o.o

ChrisBr:
@hennevogel cc

differentreality:
That sounds to me a lot like 'I don't want to be video recorded while giving a talk'.
I think it is great that we can give a means of contact with the speaker.

I could, however, understand how speakers might not want the email they used to register to OSEM shown in public (and instead they would want to show another contact email).

markuskramerIgitt:
Thank you, @ChrisBr.
In the last weeks, I received spam, which is very unusual.
One reason for spam are crawlers.

This might be a conincidence, but I don't want my email shown in public. 
Of course I will put contact details into my presentation.

ChrisBr:
@differentreality I agree with @markuskramerIgitt, if you don't want your email in public displayed doesn't mean that you don't want to be recorded. On almost all social media platforms (linkedin, xing etc) the user can choose if he wants that the email is public and/or that it's possible to be contacted! That's nothing we should decide as organizers (or at least make speakers aware that we publish the emails).

PR #1296, opened 2017-02-14
#1040 Show email in public option

JewelSam:


JewelSam:
#1040 
But I don't know how I should update email_public of all existent users to true value. Do you usually do it using rake task?

ChrisBr:
@JewelSam Yes we do it via rake task. However, for a boolean value with a default it should set to true by default, shouldn't it?

JewelSam:
@ChrisBr I was suprised too, but no. 
I created task.

Ana06:
The problem is that for the object already created the default was `nil`, so this will only affect the new ones, that's why you need the task :wink:

ChrisBr:
LGTM

Issue #1286, opened 2017-02-09
Fix Travis failures in a better way


Ana06:
It used to happen something strange, after truncating the database and loading the seeds and before the first test is executed some data was introduced in the database. I have no idea of where this data comes from.

So this strange data kept there until a truncation was executed (after a feature). If a test query the information in the tables where strange data was introduced (for example User and Conference) before they were truncated it failed.

I added some puts here that illustrate what happened:

https://travis-ci.org/Ana06/osem/builds/191338450

This only happened in **Travis**, locally it worked.

To fix it the database was changed to truncate the database before each test in e4104ba2cdb84699ee2e83ee18b75a587b4c92ac

This is just a quick solution until we find out why was this happening. So, we should find a better solution to this problem and restore the database to it previous state, where the database was only truncate after a feature.

siddhantbajaj:
I think the data  introduced in the db is because  we are Linting factories before each feature.

Linting before(:suite) reduces the performance of our test.  

 As suggested by Factory girl documentation we should be doing it in a separate task and should ensure that we clear the database  after linting the factories. 

@differentreality @Ana06 thoughts?

PR #1297, opened 2017-02-15
fixed travis error

siddhantbajaj:
Issue #1286 

Since we are using webmock for one of our external request, we cannot add factory rake task to travis.
But we should clear the database after linting the factories.


Ana06:
Great! :+1: 

@siddhantbajaj can you please squash your commits into the first one? I mean just one commit _Fix Travis error_. In case that you don't know how to do it: https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one

spec/support/database_cleaner.rb
```diff
@@ -1,6 +1,16 @@
 RSpec.configure do |config|
-  config.before(:each) do
+  config.before(:suite) do
     DatabaseCleaner.clean_with(:truncation)
```
siddhantbajaj: @Ana06 I think we can skip this as we are already doing it after linting.

Ana06: but only if `ENV['OSEM_FACTORY_LINT'] != 'false'`

spec/support/factory_girl.rb
```diff
@@ -6,6 +6,7 @@
     if ENV['OSEM_FACTORY_LINT'] != 'false'
       mock_commercial_request
       FactoryGirl.lint
+      DatabaseCleaner.clean_with(:truncation)
```
Ana06: Just realised, we can not truncate the database here because this will delete the seeds, or we need to load the seeds after truncating.

spec/support/database_cleaner.rb
siddhantbajaj: But I think that should always remain true unless someone wants to manually run the factory_girl rake task which indeed won't be possible as one of our factory build depend on internet.

spec/support/factory_girl.rb
siddhantbajaj: So can we just just load and clean the database here only?

spec/support/database_cleaner.rb
Ana06: But you call want to change it to false to not lint the factories. In fact, the ideal solution would be running the linting code in a task as it is said here: https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#linting-factories

> Running it in a before(:suite), will negatively impact the performance of your tests when running single tests.

Then we can run this task in Travis and if it success then running the tests too.
And for the `mock_commercial_request` we already have this `before_each`: https://github.com/openSUSE/osem/blob/c48c8afa51b3072af27ae76bb2ef0523bf1c50a8/spec/support/external_request.rb#L6, so as long as we keep this code it should work. And to be honest I don't know why the `mock_commercial_request` call is inside the if `ENV['OSEM_FACTORY_LINT'] != 'false'` as this is completely unrelated to the lint stuff.

siddhantbajaj: > In fact, the ideal solution would be running the linting code in a task

Yes Indeed, my initial approach was the same but then I thought of doing it this way.
>And to be honest I don't know why the mock_commercial_request call is inside the if ENV['OSEM_FACTORY_LINT'] != 'false' as this is completely unrelated to the lint stuff.

I was in the same confusion but later find out that our factory is dependent on internet beacuse of this validation in which we check if the commercial is valid.
https://github.com/openSUSE/osem/blob/master/app/models/commercial.rb#L25
So we need to mock the request before linting the factories.

siddhantbajaj:
Now the thing we need to decide is whether we should keep factory_girl.rb file and set ENV['OSEM_FACTORY_LINT']  to false by default or we can just delete this file?

lib/tasks/factory_girl.rake
```diff
@@ -1,9 +1,11 @@
+require_relative '../../spec/support/external_request'
 namespace :factory_girl do
   desc "Verify that all FactoryGirl factories are valid"
   task lint: :environment do
     if Rails.env.test?
       begin
         DatabaseCleaner.start
+        mock_commercial_request
```
Ana06: :+1: Although I would put this in a separate commit as it is currently wrong and it is unrelated to the Travis error

Gemfile
```diff
@@ -211,6 +211,7 @@ end
 
 group :test do
   # as test framework
+  gem 'sqlite3'
```
Ana06: Why do you need to add this?

spec/support/database_cleaner.rb
Ana06: Right

spec/support/factory_girl.rb
```diff
@@ -6,6 +6,8 @@
     if ENV['OSEM_FACTORY_LINT'] != 'false'
       mock_commercial_request
       FactoryGirl.lint
```
Ana06: Now that we are doing it in a rake task I think that we can remove this file together with the `OSEM_FACTORY_LINT` as there is no point in linting the factories before each suite.

.travis.yml
```diff
@@ -24,5 +24,6 @@ before_script:
   - mysql -u root -e 'create database osem_test;'
   - RAILS_ENV=test bundle exec rake db:migrate --trace
 script:
+  - 'bundle exec rake factory_girl:lint'
```
Ana06: :1:

spec/support/database_cleaner.rb
Ana06: Right

Gemfile
siddhantbajaj: Because of a strange issue i faced. When I tried to run 'bundle exec rake factory_girl:lint'  it shows me an error to include this gem. But when i tried "bundle exec rake factory_girl:lint RAILS_ENV='test'" there was no errors.
Though we check the ENV in factory_girl task but still i don't know why i got that error.

spec/support/factory_girl.rb
siddhantbajaj: done.

.travis.yml
siddhantbajaj: I have changed this to "bundle exec rake factory_girl:lint RAILS_ENV='test'" so that we don't need to include that gem for test environment as well.

.travis.yml
```diff
@@ -24,5 +24,6 @@ before_script:
   - mysql -u root -e 'create database osem_test;'
   - RAILS_ENV=test bundle exec rake db:migrate --trace
 script:
+  - "bundle exec rake factory_girl:lint RAILS_ENV='test'"
```
Ana06: I would move this one line down to be executed just before the rspec tests

siddhantbajaj: done. :+1:

Ana06:
Just one more thing before we can merge it. Why did you write _temporary fix_ in the second commit message, why is it temporary?

I would just write something like: _Mock request to fix factory girl rake_

siddhantbajaj:
> Although I would put this in a separate commit as it is **currently wrong** and it is unrelated to the Travis error

I thought you were not really satisfied with this fix so thought of naming it as temporary. :flushed:

Ana06:
@siddhantbajaj 
> I thought you were not really satisfied with this fix so thought of naming it as temporary. :flushed:
I just meant that it doesn't work at the moment, but I am perfectly fine with the change :wink:

Ana06:
So let's merge it. Great job :tada:

Issue #1087, opened 2016-07-06
Update to Rails 5


ChrisBr:
[Rails 5 is out](http://weblog.rubyonrails.org/2016/7/2/this-week-in-rails-rails-5-is-out-with-new-guides-and-more/) so we should consider to update Rails in the next future.

kirantpatil:
[Turbolinks 5](https://github.com/turbolinks/turbolinks) will help OSEM a true hybrid application.

suprgyabhushan:
There is no error in upgrading to Rails 5 but is it not better to look up this matter at a later stage as many of the existing gems required for solving other bugs may not be compatible with Rails 5.

differentreality:
That's precisely the goal of this issue

AEtherC0r3:
I'm going​ to work on this

AEtherC0r3:
Rails 5 needs at least Ruby 2.2.2, the vagrant installation has 2.1 by default. A way to overcome this is to install rvm or rbenv and use that to install the required ruby version. Is there a simpler method available?

hennevogel:
@AEtherC0r3 sure. distro packages :-)

https://build.opensuse.org/project/show/devel:languages:ruby has everything you need

stephlocke:
Hi peeps,
I'm currently evaluating some conference management systems for use by some R conferences and my ruby guy pointed out the lack of Rails >=5 for osem right now. 

We're hoping to do a deployment of a prototype second week of July - do you think this issue is likely to get fixed and closed at least in the git repo (not necessarily a release) within the next 3-4 weeks?

Cheers for developing & maintaining this software!
Steph

differentreality:
Hi @stephlocke, we haven't done much progress with Rails 5 lately, as we are currently running 4 projects on GSoC, and we are focusing on those features. @AEtherC0r3 who started working on it is also a GSoC student. 
We also have the GSoC evaulations that are coming up, so we are quite busy on that end.

That said, the Rails 5 update is definitely something we want to implement - and we will - however I cannot say with certainty when. We are happily accepting contributions if you think you can help speed things up towards that direction. 

In the meantime feel free to try out our demo http://demo.osem.io/ and keep an eye out on this thread about any progress regarding Rails 5.

Thanks,
Stella

PR #1299, opened 2017-02-16
Update rolify to 5.1.0

shlok007:
https://github.com/RolifyCommunity/rolify/blob/master/CHANGELOG.rdoc

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -416,7 +416,7 @@ GEM
       http-cookie (>= 1.0.2, < 2.0)
       mime-types (>= 1.16, < 3.0)
       netrc (~> 0.7)
-    rolify (5.0.0)
+    rolify (5.1.0)
     rspec (3.0.0)
       rspec-core (~> 3.0.0)
       rspec-expectations (~> 3.0.0)


differentreality:
I think this should be included in https://github.com/openSUSE/osem/issues/1087

Also @shlok007 you have another 6 open PRs, at least 3 of them have recent comments and need to be updated. I would strongly suggest you first take care of your existing PRs, and work on new stuff after they are merged.

shlok007:
> I think this should be included in #1087

The reason I am updating dependencies to their rails 5 compatible version separately is because some gems ( e.g. [`devise`](https://github.com/plataformatec/devise) see ) have breaking changes with their rails 5 compatible version. Updating rails version and all dependencies, fixing their breaking changes in a single PR would be a little hard to review as well. 

Also, not all gems follow semantic versioning or have maintained a perfect changelog. This could help us find out which gems are no longer maintained or doesn't have a compatible version with rails 5 yet ( eg. [`rails-observers`](https://github.com/rails/rails-observers) see #1239 ) and we can look for alternatives before proceeding with #1087 .

>Also @shlok007 you have another 6 open PRs, at least 3 of them have recent comments and need to be updated. I would strongly suggest you first take care of your existing PRs, and work on new stuff after they are merged.

Thank you! I strongly agree, was just trying to cross a few minor things off the todo list. I'll keep this in mind. :slightly_smiling_face:

differentreality:
I'll merge this, but let's open separate PRs for updating gems only if needed - I agree that for devise there are some more essential changes to be made, so opening a new PR just for that might make sense (depending on how many other gems have the same problem). Other than that, I would leave gem updating for the overall Rails 5 update.

PR #1300, opened 2017-02-17
Added demo data

### LINKED ISSUE #1211: Add db seed
Include a db seed in order to minimize testing time.

siddhantbajaj:
Closes #1211 
Added demo data for:

- [x] Full conference happening in the future (with open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks)
- [x] Full conference in the past (with closed cfp, venue/rooms, submitted talks by multiple speakers, rejected talks, confirmed talks, multiple scheduled talks in various rooms)
- [x] Conference without venue/rooms, with submitted talks
- [x] Conference with closed registration, or with registrations that reached the limit
- [x] full range of events and  multiple schedules for all the above conferences

More suggestions needed!

### DIFF
--- a/lib/tasks/demo_data_for_development.rake
+++ b/lib/tasks/demo_data_for_development.rake
@@ -1,0 +1,77 @@
+namespace :data do
+  desc 'Create demo data for our local development'
+  task test: :environment do
+    include FactoryGirl::Syntax::Methods
+
+    def generate_program conference
+        program = conference.program
+        user1 = create(:user)
+        user2 = create(:user)
+
+        conference_rooms = conference.venue.rooms
+
+        selected_schedule = create(:schedule, program: program)
+        demo_schedule = create(:schedule, program: program)
+        program.update_attributes!(selected_schedule: selected_schedule)
+
+        create(:event, program: program, title: 'Demo Event', abstract: 'This is a demo event instance whose state not defined.')
+        create(:event, program: program, title: 'Demo Rejected Event', state: 'rejected', abstract: 'This is demo event instance in a rejected state.')
+        create(:event, program: program, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a demo event instance in unconfirmed state.')
+        create(:event, program: program, title: 'Demo Confirmed Unscheduled Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state.')
+
+        first_scheduled_event = create(:event, program: program, title: 'first_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        second_scheduled_event = create(:event, program: program, title: 'second_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance having multiple speakers.')
+
+        create(:event_user, event: multiple_speaker_event, user: user1, event_role: 'speaker')
+        create(:event_user, event: multiple_speaker_event, user: user2, event_role: 'speaker')
+
+        create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date + conference.start_hour.hours, room: conference_rooms.first)
+        create(:event_schedule, event: second_scheduled_event, schedule: selected_schedule, start_time: conference.start_date + conference.start_hour.hours + 15.minutes, room: conference_rooms.second)
+        create(:event_schedule, event: multiple_speaker_event, schedule: selected_schedule, start_time: conference.start_date + conference.start_hour.hours + 30.minutes, room: conference_rooms.third)
+        create(:event_schedule, event: first_scheduled_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours + 15.minutes, room: conference_rooms.third)
+        create(:event_schedule, event: second_scheduled_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours + 30.minutes, room: conference_rooms.third)
+        create(:event_schedule, event: multiple_speaker_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours, room: conference_rooms.first)
+
+        create(:registration, user: user1, conference: conference)
+        create(:registration, user: user2, conference: conference)
+    end
+
+    # This is a full conference demo instance that will happen in the future.
+    # By full conference it means all basic information about conference is already set.
+    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_date: 2.days.from_now, end_date: 6.days.from_now, start_hour: 8, end_hour: 20, description: 'This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.')
+    generate_program conference
+
+    # This is a full conference demo instance that has already happened.
+    # By full conference it means all basic information about conference is already set.
+    # Initially end date is set 6 days from now
+    # So that events can be created without any failure in validations.
+    conference = create(:full_conference, title: 'Jangouts Demo Conference', short_title: 'jangouts', start_date: 7.days.ago, end_date: 6.days.from_now, start_hour: 15, end_hour: 20, description: 'This is a full conference demo instance happened in the past. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.')
+    generate_program conference
+    conference.program.cfp.update_attributes!(start_date: 4.days.ago, end_date: 2.days.ago)
+    conference.update_attributes!(end_date: 1.day.ago)
+    conference.registration_period.update_attributes!(start_date: 9.days.ago, end_date: 8.days.ago)
+
+    # This is a conference that will happen in the future
+    # It only has a registration period and unscheduled events
+    registration_period = create(:registration_period)
+    conference = create(:conference, title: 'Portus Demo Conference', short_title: 'portus', registration_period: registration_period, start_date: 3.days.from_now, end_date: 7.days.from_now, start_hour: 10, end_hour: 15, description: 'This is a demo conference instance. No information about this conference is set by default.')
+    create(:event, program: conference.program, title: 'Demo Event', abstract: 'This is a demo event instance whose state not defined.')
+    create(:event, program: conference.program, title: 'Demo Rejected Event', state: 'rejected', abstract: 'This is demo event instance in a rejected state.')
+    create(:event, program: conference.program, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a demo event instance in unconfirmed state.')
+    create(:event, program: conference.program, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state.')
+
+    # This is a full conference demo instance that has already started.
+    # By full conference it means all basic information about conference is already set.
+    # Registration period of this conference is closed now.
+    registration_period = create(:registration_period, start_date: 4.days.ago, end_date: 2.days.ago)
+    conference = create(:full_conference, title: 'Yast Demo Conference', short_title: 'yast', registration_period: registration_period, start_date: 2.days.ago, end_date: 7.days.from_now, start_hour: 10, end_hour: 21, description: 'This is a full conference demo instance. Its registration period is closed.')
+    generate_program conference
+
+    # This is a full conference demo instance that will happen in the future.
+    # By full conference it means all basic information about conference is already set.
+    # Registration for this conference has reached its limit.
+    conference = create(:full_conference, title: 'Zypper Docker  Conference', short_title: 'zypper', registration_limit: 2, start_date: 3.days.from_now, end_date: 7.days.from_now, start_hour: 7, end_hour: 19, description: 'This is a full conference demo instance. Its registrations has reached the limit.')
+    generate_program conference
+    end
+end


lib/tasks/data_demo.rake
```diff
@@ -6,5 +6,13 @@ namespace :data do
...
     conference.contact.update_attributes(email: 'osemdemo@osem.io', sponsor_email: 'osemdemo@osem.io')
     create(:admin, email: 'admin@osem.io', username: 'admin', password: 'password123', password_confirmation: 'password123')
+    #Demo Events
+    program_id = conference.program.id
+    create(:event, program_id: program_id, title: 'Demo Event', abstract: 'This is a [Open Source Event Manager](http://osem.io/) demo event instance')
+    create(:event, program_id: program_id, title: 'Demo Rejected Event', state: 'rejected', abstract: 'This is a [Open Source Event Manager](http://osem.io/) demo event instance')
```
Ana06: Maybe having different abstracts for every event?

lib/tasks/data_demo.rake
```diff
@@ -6,5 +6,13 @@ namespace :data do
...
+    create(:event, program_id: program_id, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a [Open Source Event Manager](http://osem.io/) demo event instance')
+    create(:event, program_id: program_id, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a [Open Source Event Manager](http://osem.io/) demo event instance')
+    event = create(:event, program_id: program_id, title: 'Demo Scheduled Event', state: 'confirmed', abstract: 'This is a [Open Source Event Manager](http://osem.io/) demo event instance')
```
Ana06: It would be nice to have more than 1 scheduled event so you can see them in the schedule.

differentreality:
I'm glad to see someone took this up! We should consider which cases are important to us.
For example, 1 conference without venue/rooms. Another conference with multiple events scheduled, some events happening at the same time, in different rooms. This will help us have a nice variety of realistic options that also showcase OSEM's functionalities.

Ana06:
Also, an event with several speakers.

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
 
     conference = create(:full_conference, title: 'Open Source Event Manager Demo', short_title: 'osemdemo' ,description: "This is a [Open Source Event Manager](http://osem.io/) demo instance. You can log in as **admin** with the password **password123** or just you just [sign up](/accounts/sign_up) with your own user. We hope you enjoy checking out all the functionality, if you have questions don't hesitate to [contact us](http://osem.io/#contact)!\r\n\r\n## Data will be destroyed every thirty minutes or whenever someone updates the [OSEM source code on github](https://github.com/openSUSE/osem/commits/master).")
     conference.contact.update_attributes(email: 'osemdemo@osem.io', sponsor_email: 'osemdemo@osem.io')
-    create(:admin, email: 'admin@osem.io', username: 'admin', password: 'password123', password_confirmation: 'password123')
+    create(:conference, title: 'Demo Conference', short_title: 'Opensuse' ,description: "This is a demo conference instance. No information about this conference is set by default.")
```
Ana06: The coma before `description` should have the space after and not before:
`create(:conference, title: 'Demo Conference', short_title: 'Opensuse', description: "This is a demo conference instance. No information about this conference is set by default.")`
:wink:

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
...
+
+    admin = create(:admin, email: 'admin@osem.io', username: 'admin', password: 'password123', password_confirmation: 'password123')
+    user1 = create(:user, email: 'user1@osem.io', username: 'user1', password: 'password123', password_confirmation: 'password123')
+    user2 = create(:user, email: 'user2@osem.io', username: 'user2', password: 'password123', password_confirmation: 'password123')
+
+    selected_schedule = create(:schedule, program: program)
+    demo_schedule = create(:schedule, program: program)
+    program.update_attributes(selected_schedule: selected_schedule)
+    
+    create(:event, program: program, title: 'Demo Event', abstract: 'This is a demo event instance with its state not defined')
```
Ana06: This is a demo event instance **whose** state not defined

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
...
+    multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a [Open Source Event demo scheduled event instance having multiple speakers')
+
+    create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+    create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
+
+    create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.first)
+    create(:event_schedule, event: second_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.second)
+    create(:event_schedule, event: multiple_speaker_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.third)
+    create(:event_schedule, event: first_scheduled_event, schedule: demo_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.third)
```
Ana06: I would be better that `demo_schedule` is not the same as the `selected_schedule`. Just schedule the events in a different room and time. :wink:

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
...
+
+    first_scheduled_event = create(:event, program: program, title: 'first_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance')
+    second_scheduled_event = create(:event, program: program, title: 'second_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance')
+    multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a [Open Source Event demo scheduled event instance having multiple speakers')
+
+    create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+    create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
+
+    create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.first)
+    create(:event_schedule, event: second_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.second)
```
Ana06: It would be better if not all the events start at the same time. For example:

``` Ruby
create(:event_schedule, event: second_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours + 30.minutes, room: Room.second)
create(:event_schedule, event: multiple_speaker_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours + 15 minutes, room: Room.third)
```

lib/tasks/data_demo.rake
siddhantbajaj: Okay. :+1:

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
...
+    create(:event, program: program, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state')
+
+    first_scheduled_event = create(:event, program: program, title: 'first_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance')
+    second_scheduled_event = create(:event, program: program, title: 'second_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance')
+    multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance having multiple speakers')
+
+    create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+    create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
+
+    create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: Room.first)
```
differentreality: Room.first is too abstract. This should specifically say it is the first room of the conference where the event belongs to.

differentreality:
This is a nice start. I have made a comment with more information we can include. 
This is currently modifying the code for what we show on demo.osem.io, I'd rather we don't touch that and we create a separate task which we can use locally for development. What we want to show on our demo is probably much less than what we would like to have locally to play with our code and new features.

lib/tasks/data_demo.rake
```diff
@@ -5,6 +5,35 @@ namespace :data do
...
     conference.contact.update_attributes(email: 'osemdemo@osem.io', sponsor_email: 'osemdemo@osem.io')
-    create(:admin, email: 'admin@osem.io', username: 'admin', password: 'password123', password_confirmation: 'password123')
+    create(:conference, title: 'Demo Conference', short_title: 'Opensuse', description: "This is a demo conference instance. No information about this conference is set by default.")
+
+    program = conference.program
+
+    admin = create(:admin, email: 'admin@osem.io', username: 'admin', password: 'password123', password_confirmation: 'password123')
```
differentreality: Maybe there already is a user with username admin in our local instance, do you think we should provision for that?

siddhantbajaj: Actually I have to clean my db before running this task, while testing my code therefore I added it. But yes, we don't need this as we already have it in our local instance.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,70 @@
...
+
+    def generate_events program, conference
+        user1 = create(:user)
+        user2 = create(:user)
+
+        selected_schedule = create(:schedule, program: program)
+        demo_schedule = create(:schedule, program: program)
+        program.update_attributes(selected_schedule: selected_schedule)
+
+        build(:event, program: program, title: 'Demo Event', abstract: 'This is a demo event instance whose state not defined.').save(validate: false)
```
Ana06: Why do you do `save(validate: false)`? All the objects created here should be valid

Same for all the `events` and `scheduled_events`

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,70 @@
...
+        second_scheduled_event.save(validate: false)
+        multiple_speaker_event = build(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance having multiple speakers.')
+        multiple_speaker_event.save(validate: false)
+
+        create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+        create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
+
+        conference_rooms = conference.venue.rooms
+
+        create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: conference_rooms.first)
```
Ana06: Just realised, you can't not write `+ 20.hours` or + `15.hours` as the conference start at 9 and end at starts at 9 and end at 20. So you can do `+ 24.hours` to start the second day or without adding anything to start in the first day.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,70 @@
+namespace :data do
+  desc 'Create demo data for our local development'
+  task test: :environment do
+    include FactoryGirl::Syntax::Methods
+
+    def generate_events program, conference
```
Ana06: why did you introduce this method?

siddhantbajaj: Since we need the events for all conferences, so I thought it would be better to do it with a method rather than writing the same code to generate events for all the conferences.

lib/tasks/demo_data_for_development.rake
siddhantbajaj: Actually for our Jangouts Demo Conference( it has already happened ) some validations won't pass like this one:
https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L288
created_at attribute for our events would be after the end_date of our Jangouts Demo conference.

lib/tasks/demo_data_for_development.rake
siddhantbajaj: Suppose conference is created today that is on 22nd then conference.start_date.to_time will give time as 18:30 of 21st. So our first_scheduled_event gets scheduled at 14:30 of 22nd.

lib/tasks/demo_data_for_development.rake
Ana06: You can not introduce non valid records in our database because we are assuming everywhere that the records are valid, they should be valid. :wink:

If the validation you mention fails is because an event is scheduled after the conference end and that shouldn't happen. Is there any problem in scheduling it inside the conference period?

siddhantbajaj: No validation fails when event is scheduled,it fails when event is created because we are creating events of conference that has already happened.
Actually let's say we create a conference that has happened in the past.Our conference is like this:
> {start_date:22,end_date:24}

And now when we create its events following validation won't pass:

`errors.
        add(:created_at, "can't be after the conference end date!") if program.conference && program.conference.end_date &&
        (**Date.today** > program.conference.end_date)`

because this validation checks the current date(Date.today) and compares it with conference end_date.

siddhantbajaj: > You can not introduce non valid records in our database because we are assuming everywhere that the records are valid, they should be valid.

But we do have some conferences in our db that has already happened. I don't think these records will cause any such trouble though.

differentreality: Let's not negotiate data integrity. 
You should **not** introduce invalid records.
Think about the purpose of populating those records. It's to show them to a user so that the user can play around. What happens when the user tries to edit this Event?

Ana06: @siddhantbajaj
 > One possible solution is we change the validation and compare end_date with created_at attribute of the event instead of Date.today.

Do not change validation but the way you create the events to  make then valid. You could change the conference start time to today, create the events and then update the conference dates, right?

siddhantbajaj: @Ana06 yes we can definitely do this.:wink:

siddhantbajaj: @differentreality yes I got your point. I have made the changes.:thumbsup:

lib/tasks/demo_data_for_development.rake
differentreality: Where do you see that a newly created conference starts at 18:30 ?

siddhantbajaj: No conference does not start at 18:30. When I was doing something like this:

`create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time, room: conference_rooms.first)`

Start_time was set as 18:30. Since our conference start at 9:00 and end at 18:00 so I was not able to see the event in the schedule so I changed this to something like this:

`create(:event_schedule, event: first_scheduled_event, schedule: selected_schedule, start_time: conference.start_date.to_time + 20.hours, room: conference_rooms.first)`

differentreality: But if conference starts at 9am and event starts at 9am + 20 hours, that's 5am the next day, when the conference is not on?

siddhantbajaj: But **confefrence.start_date.to_time()** won't give you 9 am. Actually to_time() converts string to time, since our start_date does not have any information about  time so to_time() function will give you 00:00:00 in UTC or 05:30 in IST.
http://apidock.com/rails/String/to_time

Ana06: `conference.start_date` would give something like:

`Thu, 02 Mar 2017`

and `conference.start_date.to_time()` would give you:

`2017-03-02 00:00:00 +0100` or `2017-03-01 23:00:00 UTC` (we are using UTC for the times)

To avoid all this confusion I think it is much clear that you write it like that:

`conference.start_date + 9.hours`

this would give you:

`Thu, 02 Mar 2017 09:00:00 UTC +00:00`


Moreover, since https://github.com/openSUSE/osem/pull/1143 was merged you also need to take into account the `start_hour`. So you need to write something like:

`conference.start_date + conference.start.hour.hours`

or if you want for example to start one hour later than the conference start time:

`conference.start_date + conference.start.hour.hours + 1.hours`


You could also use that you now that for your conference the `start_hour` is 9, but I think it is more readable if you write the `start_hour` so you don't have to find out what that 9 is when reading this.

lib/tasks/demo_data_for_development.rake
Ana06: but do we want all the conference to have the same events? You could change the `start_hour` of the conference, so the schedule look a little bit different.

siddhantbajaj: Maybe I don't know. But look at it like this we have covered all possible cases for our conferences and we can check all possible type of events in each of this conference.:wink:

siddhantbajaj: > You could change the start_hour of the conference, so the schedule look a little bit different.

Yes we can do this. :+1:

lib/tasks/demo_data_for_development.rake
siddhantbajaj: Yes start_hour will certainly make code much more readable, will make these changes. :+1:

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,67 @@
...
+        create(:event_schedule, event: multiple_speaker_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours, room: conference_rooms.first)
+
+        create(:registration, user: user1, conference: conference)
+        create(:registration, user: user2, conference: conference)
+    end
+
+    conference = create(:full_conference, title: 'Jangouts Demo Conference', short_title: 'jangouts', start_hour: 15, description: "This is a full conference demo instance happened in the past. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.")
+    program = conference.program
+    generate_events program, conference
+    conference.update_attributes(start_date: 7.days.ago, end_date: 1.day.ago)
```
siddhantbajaj: > Do not change validation but the way you create the events to make then valid. You could change the conference start time to today, create the events and then update the conference dates, right?

Validations would fail if we do it before generating events.

Ana06: why?

Ana06: ah! because you want to create events for an conference already over.

The problem is that you are using the start_date of the conference for generating the events and then changing it.

What you can do it giving it the `start_date` in creation and update afterwards the `end_date`. This way the validations will not fail and you are not using a start_time that change afterwards.

I also think that it would be great that one of the conferences is not over yet, I mean that the end_date is not before the current day, but after.

Ana06:
@siddhantbajaj 
> Validations would fail if we do it before generating events.

Ah, yes, because you want to create events for an conference already over.The problem is that you are using the start_date of the conference for generating the events and then changing it.

What you can do it giving it the `start_date` in creation and update afterwards the `end_date`. This way the validations will not fail and you are not using a `start_time` that change afterwards.

I also think that it would be great that one of the conferences is not over yet, I mean that the `end_date` is not before the current day, but after.

lib/tasks/demo_data_for_development.rake
siddhantbajaj: > What you can do it giving it the start_date in creation and update afterwards the end_date. This way the validations will not fail and you are not using a start_time that change afterwards.

Yes, It seems to be the best fix to handle this. Thanks @Ana06

siddhantbajaj: >I also think that it would be great that one of the conferences is not over yet, I mean that the end_date is not before the current day, but after.

I have made yast conference like this. :wink:

siddhantbajaj:
@Ana06 I have made the changes. :+1:

Ana06:
great job! :tada: Please squash your commits so we can merge it.

@differentreality are you ok with it too?

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+        create(:event, program: program, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a demo event instance in unconfirmed state.')
+        create(:event, program: program, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state.')
+
+        first_scheduled_event = create(:event, program: program, title: 'first_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        second_scheduled_event = create(:event, program: program, title: 'second_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance having multiple speakers.')
+
+        create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+        create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
```
differentreality: Single quotes

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+        user2 = create(:user)
+
+        selected_schedule = create(:schedule, program: program)
+        demo_schedule = create(:schedule, program: program)
+        program.update_attributes(selected_schedule: selected_schedule)
+
+        create(:event, program: program, title: 'Demo Event', abstract: 'This is a demo event instance whose state not defined.')
+        create(:event, program: program, title: 'Demo Rejected Event', state: 'rejected', abstract: 'This is demo event instance in a rejected state.')
+        create(:event, program: program, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a demo event instance in unconfirmed state.')
+        create(:event, program: program, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state.')
```
differentreality: This is a confirmed but unscheduled event, I'd add the comment `Unscheduled` to make this clear.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+
+        first_scheduled_event = create(:event, program: program, title: 'first_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        second_scheduled_event = create(:event, program: program, title: 'second_scheduled_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance.')
+        multiple_speaker_event = create(:event, program: program, title: 'multiple_speaker_event', state: 'confirmed', abstract: 'This is a demo scheduled event instance having multiple speakers.')
+
+        create(:event_user, event: multiple_speaker_event, user: user1, event_role: "speaker")
+        create(:event_user, event: multiple_speaker_event, user: user2, event_role: "speaker")
+
+        conference_rooms = conference.venue.rooms
```
differentreality: I'd move this higher up before we start creating events.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+        create(:event_schedule, event: first_scheduled_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours + 15.minutes, room: conference_rooms.third)
+        create(:event_schedule, event: second_scheduled_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours + 30.minutes, room: conference_rooms.third)
+        create(:event_schedule, event: multiple_speaker_event, schedule: demo_schedule, start_time: conference.start_date + conference.start_hour.hours, room: conference_rooms.first)
+
+        create(:registration, user: user1, conference: conference)
+        create(:registration, user: user2, conference: conference)
+    end
+
+    conference = create(:full_conference, title: 'Jangouts Demo Conference', short_title: 'jangouts', start_hour: 15, start_date: 7.days.ago, description: "This is a full conference demo instance happened in the past. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.")
```
differentreality: Personally I'd prefer that we also define end_hour and end_date, now in order to read this, I have to go to the factory to check if those attributes are set and what's their value.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+    generate_program conference
+    conference.program.cfp.update_attributes(start_date: 4.days.ago, end_date: 2.days.ago)
+    conference.update_attributes(end_date: 1.day.ago)
```
differentreality: I would use `update_attributes!` instead in all instances.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+    conference.registration_period.update_attributes(start_date: 9.days.ago, end_date: 8.days.ago)
+
+    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_hour: 8, description: "This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.")
```
differentreality: I would add a short comment before every conference creation to say what this is supposed to do/have, and how it is different from other conferences, eg. empty conference, conference with schedule, etc.

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,63 @@
...
+    create(:event, program: conference.program, title: 'Demo Event', abstract: 'This is a demo event instance whose state not defined.')
+    create(:event, program: conference.program, title: 'Demo Rejected Event', state: 'rejected', abstract: 'This is demo event instance in a rejected state.')
+    create(:event, program: conference.program, title: 'Demo Unconfirmed Event', state: 'unconfirmed', abstract: 'This is a demo event instance in unconfirmed state.')
+    create(:event, program: conference.program, title: 'Demo Confirmed Event', state: 'confirmed', abstract: 'This is a demo event instance in a confirmed state.')
+
+    registration_period = create(:registration_period, start_date: 4.days.ago, end_date: 2.days.ago)
+    conference = create(:full_conference, title: 'Yast Demo Conference', short_title: 'yast' , registration_period: registration_period, end_hour: 23, start_date: 1.day.ago, description: "This is a full conference demo instance. Its registration period is closed.")
```
differentreality: Please use single quotes for the description as well, and anywhere else unless we have string interpolation.

siddhantbajaj:
@differentreality I have made the changes. :+1:

lib/tasks/demo_data_for_development.rake
```diff
@@ -37,27 +37,39 @@ namespace :data do
         create(:registration, user: user2, conference: conference)
     end
 
-    conference = create(:full_conference, title: 'Jangouts Demo Conference', short_title: 'jangouts', start_hour: 15, start_date: 7.days.ago, description: "This is a full conference demo instance happened in the past. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.")
+    # This is a full conference demo instance that has already happend.
```
differentreality: typo: happened

lib/tasks/demo_data_for_development.rake
```diff
@@ -37,27 +37,39 @@ namespace :data do
...
+    conference.registration_period.update_attributes!(start_date: 9.days.ago, end_date: 8.days.ago)
 
-    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_hour: 8, description: "This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.")
+    # This is a full conference demo instance that will happen in the future.
+    # By full conference it means all basic information about conference is already set.
+    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_date: 2.days.from_now, end_date: 6.days.from_now, start_hour: 8, end_hour: 20, description: 'This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.')
     generate_program conference
 
+    # This is a conference demo instance that will happen in the future.
+    # No information about this conference is set by default.
```
differentreality: I am confused about what that means. It seems we are setting a registration period for our conference, as well as some events.
Perhaps the conference has a registration period and unscheduled events?

lib/tasks/demo_data_for_development.rake
```diff
@@ -37,27 +37,39 @@ namespace :data do
...
     registration_period = create(:registration_period, start_date: 4.days.ago, end_date: 2.days.ago)
-    conference = create(:full_conference, title: 'Yast Demo Conference', short_title: 'yast' , registration_period: registration_period, end_hour: 23, start_date: 1.day.ago, description: "This is a full conference demo instance. Its registration period is closed.")
+    conference = create(:full_conference, title: 'Yast Demo Conference', short_title: 'yast', registration_period: registration_period, start_date: 2.days.ago, end_date: 7.days.from_now, start_hour: 10, end_hour: 21, description: 'This is a full conference demo instance. Its registration period is closed.')
     generate_program conference
 
-    conference = create(:full_conference, title: 'Zypper Docker  Conference', short_title: 'zypper' , registration_limit: 2, description: "This is a full conference demo instance. Its registrations has reached the limit.")
+    # This is a full conference demo instance that will happen in the future.
+    # By full conference it means all basic information about conference is already set.
+    # Registration for this conference has reached it's limit.
```
differentreality: typo: its

lib/tasks/demo_data_for_development.rake
```diff
@@ -0,0 +1,77 @@
...
+    conference.program.cfp.update_attributes!(start_date: 4.days.ago, end_date: 2.days.ago)
+    conference.update_attributes!(end_date: 1.day.ago)
+    conference.registration_period.update_attributes!(start_date: 9.days.ago, end_date: 8.days.ago)
+
+    # This is a full conference demo instance that will happen in the future.
+    # By full conference it means all basic information about conference is already set.
+    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_date: 2.days.from_now, end_date: 6.days.from_now, start_hour: 8, end_hour: 20, description: 'This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.')
+    generate_program conference
+
+    # This is a conference demo instance that will happen in the future.
```
differentreality: This seems to be the only conference that is to a full_conference. What about we move it either on the top or on the bottom?

lib/tasks/demo_data_for_development.rake
siddhantbajaj: I mean the basic information like venue,room etc.

siddhantbajaj: will change this to make it more clear

lib/tasks/demo_data_for_development.rake
siddhantbajaj: Yes we can do this.

lib/tasks/demo_data_for_development.rake
```diff
@@ -47,13 +52,8 @@ namespace :data do
...
     conference.registration_period.update_attributes!(start_date: 9.days.ago, end_date: 8.days.ago)
 
-    # This is a full conference demo instance that will happen in the future.
-    # By full conference it means all basic information about conference is already set.
-    conference = create(:full_conference, title: 'Open Source Event Manager Demo Conference', short_title: 'osemdemo', start_date: 2.days.from_now, end_date: 6.days.from_now, start_hour: 8, end_hour: 20, description: 'This is a full conference demo instance happening in the future. It contains open cfp, venue/rooms, submitted talks by multiple speakers, partly confirmed talks and multiple schedules.')
-    generate_program conference
-
     # This is a conference demo instance that will happen in the future.
-    # No information about this conference is set by default.
+    # Basic information (like venue,room,schedule,etc) is not set by default.
```
differentreality: What we care about is what is set, not what is not set. 
Also the 'by default' part is not necessary. True if you want rooms, you have to create them, they are not set by default. But how does that information help us here? We need to know that only X,Y things are set (and no other information). Think something along those lines:
```
This is a conference that will happen in the future
It only has a registration period and unscheduled events
```

differentreality:
@Ana06 looks good and useful :-) I will be definitely trying it out in the next few weeks.

Ana06:
@differentreality 

> @Ana06 looks good and useful :-) I will be definitely trying it out in the next few weeks.

Ok! So please @siddhantbajaj change the comment @differentreality says and squash your commits so we can merge it

siddhantbajaj:
@Ana06 done. :+1:

Issue #1284, opened 2017-02-09
Improve resources page


Ana06:
As discussed in https://github.com/openSUSE/osem/pull/1261 we want to improve the resources page to make it usable.

@differentreality:

> changing the quantity of resources should be simpler/easier. If you are in a hectic registration desk and doing registrations and giving out t-shirts, you don't have time to hit edit, change the quantity in a field, hit save. So perhaps there should be a show page with a big button or something. If we want this functionality in the app, we should give some love to the UI, so that it is not just a list of things, but has a functionality that gives some added value to the organizers.

It would be nice if screenshot with ideas are provided for this issue before starting programming it, as I would say that the perfect solution is not clear and there may be some discussion about it.

nikhilgupta1211:
@Ana06 @differentreality  as many conferences have various common resources we can give some default resources to be added like this
![image](https://cloud.githubusercontent.com/assets/14905919/23825878/4fbae798-06b8-11e7-9134-1fead27b7d0e.png)

Ana06:
> @Ana06 @differentreality as many conferences have various common resources we can give some default resources to be added like this

The speakers are not resources, are users and we already have then in our app. And I don't think that every conference what to record the chairs, so I would leave it open, so the admins can decide what they want to track.

differentreality:
I agree with @Ana06, we already have some stats about speakers in reports page, and I don't see why we would need to increase the number of speakers manually. While it is an interesting point, whether or not we are able to easily see how many of our speakers have already arrived (based on their registration attendance), this is not the place to do it (perhaps something to investigate and open another issue for?). 

I can see chairs being tracked for some smaller conferences/meetups but again I don't think it makes sense to have that available in advance.

Think of resources as swag (eg. tshirts, lanyards) you give to every attendee or against a donation, for which you want to know if you still have any (eg. if we offer tshirts to attendees, there is no way we can know how many size M tshirts we have left on the second day of the conference). That's a good example of what we are trying to offer the organizer in resources page. And for this case 'Tshirt M' would be 1 resource, 'Tshirt L' would be another issue, so that we can track how many we give / how many we have left every time we give one.

nikhilgupta1211:
> @nikhilgupta1211 the speakers are not resources, are users and we already have then in our app.

@Ana06 I meant speaker as a device

AndrewKvalheim:
Alternative proposal: #2452

PR #1301, opened 2017-02-17
Add increment and decrement functionality to left resource quantity 

BelieveC:
Addresses issue #1284 . Improve resource show page with an addition of increment and decrement functionality for Available quantity and Total left quantity using Ajax. 
increment

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -57,3 +57,5 @@ $(document).ready(function() {
         delegateSelector: 'a.smoothscroll'
     });
 });
+
+UnobtrusiveFlash.flashOptions['timeout'] = 3000;
--- a/app/assets/stylesheets/osem.css.scss
+++ b/app/assets/stylesheets/osem.css.scss
@@ -94,3 +94,8 @@ p.comment-body {
 .box{
   height: 230px;
 }
+
+.quantity_left{
+  height: 3em;
+  width: 9em;
+}
--- a/app/controllers/admin/resources_controller.rb
+++ b/app/controllers/admin/resources_controller.rb
@@ -2,6 +2,7 @@ module Admin
   class ResourcesController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :resource, only: [:show, :edit, :update, :destroy]
+    after_action :prepare_unobtrusive_flash, only: [:update]
 
     def index; end
 
@@ -23,12 +24,23 @@ module Admin
     end
 
     def update
-      if @resource.update_attributes(resource_params)
-        redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
-                    notice: 'Resource successfully updated.'
-      else
-        flash.now[:error] = "Resource update failed: #{@resource.errors.full_messages.join('. ')}."
-        render :edit
+      successful_update = check_successful_update params[:increment_used_resource_flag]
+      respond_to do |format|
+        if successful_update
+          format.html do
+            redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
+                        notice: 'Resource successfully updated.'
+          end
+          flash.now[:notice] = if params[:increment_used_resource_flag].to_i.zero?
+                                 "One #{@resource.name} freed."
+                               else
+                                 "One more #{@resource.name} used."
+                               end
+        else
+          flash.now[:error] = "Resource #{@resource.name}'s update failed: #{@resource.errors.full_messages.join('. ')}."
+          format.html{ render :edit }
+        end
+        format.js
       end
     end
 
@@ -47,6 +59,17 @@ module Admin
 
     def resource_params
       params.require(:resource).permit(:name, :description, :quantity, :used, :conference_id)
+    end
+
+    def check_successful_update(increment_used_resource_flag)
+      if increment_used_resource_flag.present?
+        # Increment/Decrement of Resource via Index Page
+        @resource.used = params[:increment_used_resource_flag].to_i == 1 ? (@resource.used + 1) : (@resource.used - 1)
+        @resource.save
+      else
+        # Update via Edit Page
+        @resource.update_attributes(resource_params)
+      end
     end
   end
 end
--- a/app/views/admin/resources/_update_links.html.haml
+++ b/app/views/admin/resources/_update_links.html.haml
@@ -1,0 +1,17 @@
+= link_to admin_conference_resource_path(conference.short_title,
+  resource, increment_used_resource_flag: 0), method: :put,
+  title: "Free 1 #{resource.name}",
+  class: "btn btn-primary #{'disabled' if resource.used <= 0}",
+  remote: true do
+  %span.fa.fa-minus.fa-2x.update_action
+%button.btn.btn-default.quantity_left.disabled.action-btn
+  %span{ id: "resource_left_#{resource.id}" }
+    = quantity_left_of(resource)
+= link_to admin_conference_resource_path(conference.short_title,
+  resource, increment_used_resource_flag: 1), method: :put,
+  title: "Use 1 #{resource.name}",
+  class: "btn btn-primary #{'disabled' if resource.used >= resource.quantity}",
+  remote: true do
+  %span.fa.fa-plus.fa-2x.update_action
+
+
--- a/app/views/admin/resources/index.html.haml
+++ b/app/views/admin/resources/index.html.haml
@@ -10,7 +10,7 @@
       %table.table.table-hover.datatable
         %thead
           %th Name
-          %th Left( Left/Total )
+          %th Remaining / Total
           %th Used
           %th Actions
         %tbody
@@ -20,9 +20,18 @@
                 = link_to(admin_conference_resource_path(@conference.short_title, resource.id)) do
                   = resource.name
                 %td
-                  = quantity_left_of(resource)
+                  .btn-group.hidden-sm.hidden-xs
+                    %span{ id: "group_large_update_actions_#{resource.id}" }
+                      = render partial: 'update_links',
+                          locals: { conference: @conference, resource: resource }
+
+                  .btn-group-vertical.visible-sm.visible-xs
+                    %span{ id: "group_small_update_actions_#{resource.id}" }
+                      = render partial: 'update_links',
+                          locals: { conference: @conference, resource: resource}
                 %td
-                  = resource.used
+                  %span{ id: "resource_used_#{resource.id}" }
+                    = resource.used
                 %td
                   .btn-group
                     = link_to 'Edit', edit_admin_conference_resource_path(@conference.short_title, resource.id),
@@ -33,4 +42,5 @@
 .row
   .col-md-12
     = link_to 'Add Resource', new_admin_conference_resource_path, class: 'btn btn-success pull-right', disabled: !(can? :create, Resource.new(conference_id: @conference_id))
+
 
--- a/app/views/admin/resources/update.js.erb
+++ b/app/views/admin/resources/update.js.erb
@@ -1,0 +1,5 @@
+$('<%= "#group_large_update_actions_#{@resource.id.to_s}" %>').html('<%= j render partial: "update_links", locals: { conference: @conference, resource: @resource} %>');
+$('<%= "#group_small_update_actions_#{@resource.id.to_s}" %>').html('<%= j render partial: "update_links", locals: { conference: @conference, resource: @resource} %>');
+$('<%= "#resource_left_#{@resource.id.to_s}" %>').text('<%= "#{@resource.quantity - @resource.used }/#{@resource.quantity}"%>').fadeIn();
+$('<%= "#resource_used_#{@resource.id.to_s}" %>').text(<%=@resource.used%>).fadeIn();
+
--- a/app/views/layouts/_admin.html.haml
+++ b/app/views/layouts/_admin.html.haml
@@ -8,7 +8,7 @@
         -# Index admin sidebar
         = render 'layouts/admin_sidebar_index'
     .col-md-10
-      #messages
+      .unobtrusive-flash-container#messages
         =render 'layouts/messages'
       #content
         = yield
--- a/spec/features/resource_spec.rb
+++ b/spec/features/resource_spec.rb
@@ -33,7 +33,7 @@ feature Resource do
       click_button 'Update Resource'
       resource.reload
 
-      expect(flash).to eq('Resource successfully updated.')
+      expect(page).to have_css('#messages'), text('Resource successfully updated.')
       expect(resource.name).to eq('changed_name')
     end
 


siddhantbajaj:
Great @BelieveC . There are some indentation issues in the changes you have made. You can have a look at them  here https://travis-ci.org/openSUSE/osem/builds/202780655 or by looking the details of your test.

BelieveC:
@siddhantbajaj, Fixed indentation issues.

app/controllers/admin/resources_controller.rb
```diff
@@ -39,7 +39,39 @@ def destroy
...
+
+    def increment_quantity
+      @resource.quantity = @resource.quantity + 1
+      @resource.save
+      respond_to do |format|
+        format.js{ render action: 'update_quantity' }
+      end
+    end
+
+    def decrement_quantity
```
Ana06: `increment_quantity`, `decrement_quantity` are exactly the same method except by the fact that one increment the variable and the other one decrements it. to avoid repeating code what about having only `increment_quantity` with a parameter, something like:


``` Ruby
  def increment_quantity(step = 1)
      @resource.quantity = @resource.quantity + step
      @resource.save
      respond_to do |format|
        format.js{ render action: 'update_quantity' }
      end
    end
```

and then

`increment_quantity` and `increment_quantity(-1)`

Moreover this would allow to change the increment step easily in case that is needed.

app/controllers/admin/resources_controller.rb
```diff
@@ -39,7 +39,39 @@ def destroy
       else
         redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
                     error: 'Resource was successfully destroyed.' \
-  	                "#{@resource.errors.full_messages.join('. ')}."
+                    "#{@resource.errors.full_messages.join('. ')}."
```
Ana06: why did you change this line? :confused:

differentreality:
The buttons are way too tiny. My finger is bigger than both of them, how can this be used with touch screens?

Also can't we use update to change the used attribute? Instead of introducing 2 new actions?

BelieveC:
@differentreality, We can do it using update action but I think it would be better to keep them as separate action because we can improve increment and decrement functionality with giving user option to increment or decrement it by more than 1 step if required. i.e Decrement by 10 or 100 and so on. What do you think @differentreality ?

differentreality:
But this is not what we are currently doing, nor do I see a use case for decreasing more than 1 per time.

BelieveC:
Sure, I would make requested changes.

BelieveC:
@differentreality, I have removed 2 new actions. Incrementing/Decrementing it using update action.

app/controllers/admin/resources_controller.rb
BelieveC: Sure! I will remove it.

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +23,19 @@ def create
...
-      else
-        flash[:error] = "Resource update failed: #{@resource.errors.full_messages.join('. ')}."
-        render :edit
+      respond_to do |format|
+        if @resource.update_attributes(resource_params)
+          format.html do
+            redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
+                        notice: 'Resource successfully updated.'
+          end
+          format.js{ render action: 'update_quantity' }
```
BelieveC: Adding an error handler for the else part.

BelieveC: Added flash message for else part
error

BelieveC:
- Index page view in Desktop resolution 
dindex

- Show page view in Desktop resolution 
dshow

- Index page view in Mobile resolution 
mindex

- Show page view in Mobile resolution
mshow

BelieveC:
@differentreality, How about this

1. 0c0080808579d9addc5c233718731bd1476e23d6

2.jquery-plugin-for-increment-decrement-number-spinner-stepperwidget

differentreality:
screenshot from 2017-02-28 11-53-35

Your +/- buttons are not aligned. Perhaps we should have another column just for the buttons? I think that would make them look nicer, what do you think?

This is showing that I have 153 tshirts left out of 150 tshirts that I had in total! That shouldn't be possible, both validations-wise and UI-wise, the button shouldn't be enabled unless I can click on it, otherwise the user will click whatever he or she wants. 

Used column does not change when I click on the +/- buttons.

app/views/admin/resources/_update_actions.html.haml
```diff
@@ -0,0 +1,5 @@
+= link_to admin_conference_resource_path(conference.short_title, resource,resource:{ used: resource.used-1 }), method: :put, remote: true do
+	%span.fa.fa-plus-circle.fa-2x.update_action
+
+= link_to admin_conference_resource_path(conference.short_title, resource,resource:{ used: resource.used + 1 }), method: :put, remote: true do
+	%span.fa.fa-minus-circle.fa-2x.update_action
```
differentreality: Empty line at the end of your file!

app/views/admin/resources/index.html.haml
```diff
@@ -16,14 +16,16 @@
...
-              %td
+              %td.col_vertical_align
                 = link_to(admin_conference_resource_path(@conference.short_title, resource.id)) do
                   = resource.name
                 %td
-                  = quantity_left_of(resource)
-                %td
-                  = resource.used
-                %td
+                  %span{ class: "quantity_left", id: "resource_left_#{resource.id}" }= quantity_left_of(resource)
```
differentreality: Please use single quotes unless there is string interpolation.

differentreality: Move the code inside your %span, not next to it.

app/views/admin/resources/_update_actions.html.haml
BelieveC: Sorry! Will fix it.

BelieveC:
@differentreality, I don't think adding a new column for increment/decrement button would look that good even though it would be properly aligned. I think it would be ambiguous to which column it tend to increment and decrement. Below are two different design, need some suggestion on them.

1. 
d1

2.
d2

differentreality:
Second design also looks nice, but please commit changes so that we can review it.

BelieveC:
Thanks, Will commit changes with the second design!

BelieveC:
@differentreality, Following are screenshot index and show page in mobile and desktop resolution

1. Desktop (Show page) 
dindex

2. Desktop (Index page) 
dshow

3. Mobile (Index page) 
mindex

4. Mobile(Show page) 
mshow

BelieveC:
@differentreality, Please can you review the changes?

differentreality:
For mobile devices, doesn't it make sense to have the two buttons further away? For example, minus button left or top, plus button right or bottom.

I am not sure we need this layout for the show page though.

app/views/admin/resources/_update_actions.html.haml
```diff
@@ -0,0 +1,10 @@
+.btn-group
+	%button.btn.btn-default.quantity_left.disabled
+		%span{ id: "resource_left_#{resource.id}" }
+			=quantity_left_of(resource)
+	%button{ class: "btn btn-default #{ 'disabled' if resource.used == resource.quantity }" }
+		=link_to admin_conference_resource_path(conference.short_title, resource,resource:{ used: resource.used + 1 }), method: :put, remote: true do
+			%span.fa.fa-minus.fa-2x.update_action
+	%button{ class: "btn btn-default #{ 'disabled' if resource.used == 0  }" }
+		=link_to admin_conference_resource_path(conference.short_title, resource,resource:{ used: resource.used-1 }), method: :put, remote: true do
+			%span.fa.fa-plus.fa-2x.update_action
```
differentreality: Please add empty line at the end of the file, it will make this red indication go away.

differentreality:
Can we really not live without adding all that new CSS?

app/views/admin/resources/index.html.haml
```diff
@@ -16,14 +16,16 @@
...
                 = link_to(admin_conference_resource_path(@conference.short_title, resource.id)) do
                   = resource.name
                 %td
-                  = quantity_left_of(resource)
-                %td
-                  = resource.used
-                %td
+                  %span{ id: "update_actions_#{resource.id}" }
+                    = render partial: 'update_actions', locals: { conference: @conference, resource: resource }
+                %td.vertical_align
```
differentreality: Align this column right, it will look better and it will be easier to read the numbers

differentreality:
What about showing some flash messages on quantity change so that we know what we just did?

app/views/admin/resources/_update_actions.html.haml
BelieveC: Thanks! Will add it.

app/views/admin/resources/index.html.haml
BelieveC: :+1:

BelieveC:
I will try my best to minimize used CSS. I am not sure about flash message similar to notification flash we use. We may try something else like toast notification or Some animation on increased or decreased quantity e.g Twitter like button. If not flash message would be a default way to do it. What do you think?

BelieveC:
For mobile devices, Design with Minus at top and Plus at the bottom makes column width too large for index page table. Keeping them on left and right respectively within possible width for a mobile resolution put a constraint on the size of a button and makes size too small. We try different layout for show page. Will get you some design :+1: Still will be happy to know your suggestion on a show page. :+1:

differentreality:
Let's stick to our regular flash messages. There are many things we can do, and I am sure everyone has a different idea on how to implement it, that doesn't mean we should do them all. Besides changing how we show messages is out of the scope of this PR. Use unobtrusive flash which we already have to show the messages please like the rest of the app does.

Goal should be for organizers to be able to use their mobile devices (touch screen) without pushing - when they want to push +, which is why I am suggesting we put those buttons as far away as possible.
Looking forward to your screenshots.

BelieveC:
I will make necessary changes :+1:

BelieveC:
@differentreality,
1. Desktop (Show page) 
dshow

2. Desktop (Index page) 
dindex

3. Mobile (Index page) 
mindex


4. Mobile(Show page) 
mshow

Looking forward for suggestion.

differentreality:
1. Do we truly need the extra CSS you have added in this PR?
2. The logic for the controller method to reduce resources is wrong. Try the following scenario: 
deviceA reduces resource from 5 to 4, deviceB (that had loaded the page before deviceA reduced the resource) reduces the same resource. Expected left amount of recources is 3, but we get 4.
3. I am open to out-of-the-box suggestions for a nicer index (?) page that would make this feature more functional to its true use (which is not just to see a list of our available resources - that's just a small part of it), sadly I am not very good with front-end ideas.

BelieveC:
@differentreality 
> Do we truly need the extra CSS you have added in this PR?

Yes, We require that CSS. I have tried to minimize the use of external CSS. 

> The logic for the controller method to reduce resources is wrong. Try the following scenario:
deviceA reduces resource from 5 to 4, deviceB (that had loaded the page before deviceA reduced the resource) reduces the same resource. Expected left amount of recources is 3, but we get 4.

Gotcha! I will fix it. Thanks.

> I am open to out-of-the-box suggestions for a nicer index (?) page that would make this feature more functional to its true use (which is not just to see a list of our available resources - that's just a small part of it), sadly I am not very good with front-end ideas.

I will add screenshot of few designs.

differentreality:
@BelieveC are you going to fix how resources are being updated?

BelieveC:
@differentreality, I have made the changes :+1:

BelieveC:
@differentreality I have made the changes :+1:

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +23,25 @@ def create
...
-        render :edit
+        flag = @resource.update_attributes(resource_params)
+      end
+      respond_to do |format|
+        if flag
+          format.html do
+            redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
+                        notice: 'Resource successfully updated.'
+          end
+          flash[:notice] = "One #{params[:inc].to_i == 1 ? 'resource freed' : 'more resource used'}"
```
differentreality: For this flash message, you need to think about what  the organizer needs to know, and give the organizer the right information to make his/her life easier. 
I, for example, would care to know **which** resource changed, eg. did I just click on T-shirt size M or size L?

I don't think that this message is very informative or helpful.

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +23,25 @@ def create
...
+          format.js{ render action: 'update_quantity' }
+        else
+          flash[:error] = "Resource update failed: #{@resource.errors.full_messages.join('. ')}."
```
differentreality: Again, I need to know which resource failed to update, otherwise this offers the organizer very little meaningful information.

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +23,25 @@ def create
     end
 
     def update
-      if @resource.update_attributes(resource_params)
-        redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
-                    notice: 'Resource successfully updated.'
+      if params[:inc].present?
```
differentreality: What does inc mean? Please rename it to something that I can understand when reading the code.

differentreality:
It is very confusing that I am clicking the + icon, but it does not mean that I just gave away 1 resource. The looks of the layout needs to be re-thought from an organizer's perspective. What does + mean for the organizer? When is the organizer going to use this interface and to do what?

app/controllers/admin/resources_controller.rb
BelieveC: Inc means increment. Alright, I will rename it.

app/controllers/admin/resources_controller.rb
BelieveC: Alright :+1:

app/controllers/admin/resources_controller.rb
BelieveC: I will make the changes.

BelieveC:
>What does + mean for the organizer? When is the organizer going to use this interface and to do what?

We can add tool pit for to give more information about what action would be performed when +/- is clicked. 

> The looks of the layout needs to be re-thought from an organizer's perspective

I agree with that. I'm trying to think about layout which would best suit for resource index. But nothing substantial came out. Will try for some more design. :+1:

BelieveC:
@differentreality 

Toolpits of increment and decrement actions:

- Decrease used resource button
increment

- Increase used resource button
decrement

differentreality:
We need to consider what + means for the organizer and provide an interface that does that is expected, not just add tooltips and messages. We need to make this easy for use. 
So what's our workflow here?
Organizer gives out an item -> clicks -  for deducting 1 thing (from the pile of items) that was given away (or sold in other cases), so in our words we decrease the remaining quantity. 

When implementing features like this you need to come up with a workflow first, so that we know what we want to do. 

You need to consider if the user (organizer) that uses OSEM will know what to do when presented with this page and sees - 23/50 + 
Is it obvious what needs to be done? Not immediately. What if we see the column's name, it says left( left/total). What it actually means is remaining and when we show the information we don't do
23( 23/50) as indicated on the column header, so perhaps the column header needs to be more accurate:
`Remaining / Total `

BelieveC:
> When implementing features like this you need to come up with a workflow first, so that we know what we want to do.

Thanks! I will that in mind :smile:

BelieveC:
@differentreality I don't think we make much change in the index page of resources with the constraint of displaying many resources on a single page. But How about adding Resource Image to resources. Following are benefits of Resource Image:

- Differentiate between two similar resources. e.g Chair1 and Chair2
- Easy to identify resource without reading the description and all.

What do you think?

app/views/admin/resources/update_quantity_error.js.erb
```diff
@@ -0,0 +1 @@
+$('#messages').html("<%= j render 'layouts/messages' %>");
```
differentreality: why not use unobtrusive flash? Then you shouldn't need this.
https://github.com/leonid-shevtsov/unobtrusive_flash

app/views/admin/resources/_update_actions.html.haml
```diff
@@ -0,0 +1,8 @@
+.btn-group
+  =link_to admin_conference_resource_path(conference.short_title, resource, used_increment_flag: 0), method: :put, title: "Use 1 #{resource.name}", class: "btn btn-primary action-btn #{ 'disabled' if resource.used == resource.quantity }", remote: true do
```
differentreality: space after =

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +23,29 @@ def create
     end
 
     def update
-      if @resource.update_attributes(resource_params)
-        redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
-                    notice: 'Resource successfully updated.'
+      if params[:used_increment_flag].present?
```
differentreality: We need to work a lot on this controller, it's impossible to read.

app/views/admin/resources/update_quantity_error.js.erb
BelieveC: Yeah! I will use it :+1:

app/controllers/admin/resources_controller.rb
BelieveC: I have refactored it now :+1:

app/views/admin/resources/_update_actions.html.haml
BelieveC: :+1:

BelieveC:
@differentreality I have made the changes!

differentreality:
I would try something like this for the index page, so that it responds better to screen size changes

screenshot from 2017-05-06 12-00-26
http://paste.opensuse.org/64030400

differentreality:
The flash messages don't disappear.

app/controllers/admin/resources_controller.rb
```diff
@@ -48,5 +56,22 @@ def destroy
...
+      else
+        @resource.update_attributes(resource_params)
+      end
+    end
+
+    def get_successful_update_flash(used_increment_flag)
+      if used_increment_flag.to_i == 1
+        "One #{@resource.name} freed."
+      else
+        "One more #{@resource.name} used."
```
differentreality: How/When do I access this case?

BelieveC: Update to the resource can be via Edit page or Increment/Decrement via Index page. `get_update_flag` action distinguish between above two updates in order to use proper update logic for a different scenario. `get_successful_update_flash` action is just used to generate flash for later update case. The sole purpose of the separate action was to make the code more readable.

BelieveC:
@differentreality Check whether you find controller readable? or else I will try something else!

BelieveC:
Fixed conflict in resources.rb

app/controllers/admin/resources_controller.rb
```diff
@@ -48,5 +56,24 @@ def destroy
...
+        # Increment/Decrement of Resource via Index Page
+        @resource.used = params[:increment_used_resource_flag].to_i == 1 ? (@resource.used + 1) : (@resource.used - 1)
+        @resource.save
+      else
+        # Update via Edit Page
+        @resource.update_attributes(resource_params)
+      end
+    end
+
+    def get_successful_update_flash(increment_used_resource_flag)
```
differentreality: This function is not needed

app/controllers/admin/resources_controller.rb
```diff
@@ -48,5 +56,24 @@ def destroy
     def resource_params
       params.require(:resource).permit(:name, :description, :quantity, :used, :conference_id)
     end
+
+    def check_successful_update(increment_used_resource_flag)
```
differentreality: This is not needed if you just have params[:used_change] which has values of -1 or +1

app/views/admin/resources/_update_links.html.haml
```diff
@@ -0,0 +1,7 @@
+= link_to admin_conference_resource_path(conference.short_title, resource, increment_used_resource_flag: 1), method: :put, title: "Use 1 #{resource.name}", class: "btn btn-primary action-btn #{ 'disabled' if resource.used== resource.quantity }", remote: true do
```
differentreality: I think it makes more sense that I give out something (which is my default action as an info desk volunteer) and I click +
(which should increase `used`)

app/assets/javascripts/application.js
```diff
@@ -57,3 +57,5 @@ $(document).ready(function() {
         delegateSelector: 'a.smoothscroll'
     });
 });
+
+UnobtrusiveFlash.flashOptions['timeout'] = 2000;
```
BelieveC: > That's 2 seconds, is that really enough time for someone to read a comment? 

It would vary person to person. Flash message have at max 3 words except for resource name. Apart from this, the reason I felt to keep it 2 second was just to assure the least number of flash message to appear on the page at any instant. More the number of messages, more resource content table would move towards bottom of page.

app/controllers/admin/resources_controller.rb
```diff
@@ -23,12 +24,19 @@ def create
...
-        flash.now[:error] = "Resource update failed: #{@resource.errors.full_messages.join('. ')}."
-        render :edit
+      successful_update = check_successful_update params[:increment_used_resource_flag]
+      respond_to do |format|
+        if successful_update
+          format.html do
+            redirect_to admin_conference_resources_path(conference_id: @conference.short_title),
+                        notice: 'Resource successfully updated.'
+          end
+          flash.now[:notice] = get_successful_update_flash params[:increment_used_resource_flag]
```
BelieveC: >Why do we need a function to define our flash message with a simple if statement?

I thought modularizing update action into smaller functions would make it more readable. I'm trying my best to make it readable by implementing different things. Will remove it.

app/views/admin/resources/index.html.haml
```diff
@@ -20,17 +22,16 @@
...
+                  %span{ id: "resource_used_#{resource.id}" }
+                    = resource.used
                 %td
                   .btn-group
                     = link_to 'Edit', edit_admin_conference_resource_path(@conference.short_title, resource.id),
                       method: :get, class: 'btn btn-primary', disabled: !(can? :update, resource)
                     = link_to 'Delete', admin_conference_resource_path(@conference.short_title, resource.id),
                       method: :delete, class: 'btn btn-danger', disabled: !(can? :destroy, resource),
                       data: { confirm: "Do you really want to delete #{resource.name}?" }
-.row
```
BelieveC: I asked for your opinion on this in the comment [link](https://github.com/openSUSE/osem/pull/1301#issuecomment-292151739). I will repeat what I proposed: 

- Unable to find it as there are paging 's next and previous action which occupies large space
- It is convenient to have important actions at the top rather than bottom e.g When we use google keep we have input to add a task at the top instead of the bottom.

BelieveC: > Doesn't the whole application have this in this position?

Just a proposal. I will revert it if don't find it appropriate.

app/views/admin/resources/_update_links.html.haml
BelieveC: Agree! i.e Our increment/decrement action should be shifted to `used` column right?

app/views/admin/resources/_update_actions.html.haml
```diff
@@ -0,0 +1,5 @@
+.btn-group.hidden-sm.hidden-xs
```
BelieveC: It can be! Will move it to index file :+1:

app/views/admin/resources/index.html.haml
differentreality: No, it is **not** appropriate if the 'New' button is sometimes on top, and sometimes on the bottom. This is not good for our design, not helpful to our users. You need to put where users can find it, based on their experience from other parts of the application. 
If you disagree with its position, that's a different issue you can bring up, and if decided so, it would change for the whole application, not just 1 page. Makes sense?

app/controllers/admin/resources_controller.rb
differentreality: Try coming back to your code 1-2 days after you have submitted your PR. Then go to the changed files and read your changes. Does it help you to jump from place to place to find what the controller does?

BelieveC: Yeah! You are right! I think it will grow on me with time. :smile:

app/views/admin/resources/index.html.haml
BelieveC: Alright! I would open a separate issue regarding above proposal +1:

app/assets/stylesheets/osem.css.scss
```diff
@@ -89,4 +89,15 @@ p.comment-body {
...
-}
\ No newline at end of file
+}
+
+.quantity_left{
+  height: 3em;
+  width: 9em;
+}
+
+@media (max-width: 767px){
```
BelieveC: Previously required to make buttoms responsive but as we have added `.visible-sm.visible-xs` class It is no longer required! Removed it.

app/controllers/admin/resources_controller.rb
BelieveC: Is [this](https://github.com/openSUSE/osem/pull/1301/commits/29eedd92cc5d2484d551a2625cd797961cff8664) what you want instead of function??

app/controllers/admin/resources_controller.rb
BelieveC: Are you suggesting to carry out update this way 
`@resource.update(name: params[:resource][:name],
                            description: params[:resource][:description],
                            used: params[:resource][:used].to_i + params[:used_count].to_i,
                            quantity: params[:resource][:quantity]
                           )`
Or something else?

BelieveC:
@differentreality, Can you review it? Let's finish it off within this month.

differentreality:
@BelieveC please squash your commits and rebase!

BelieveC:
@differentreality, Sorry It took long as I couldn't test it as environment was failing to build.

Rebased :+1:

differentreality:
@BelieveC nice, let's work on ideas that improve the UI.

BelieveC:
@differentreality Currently It looks like this

screenshot from 2017-08-09 14-41-31

I find it clean and simple. Any suggestion about design?

differentreality:
@BelieveC the issue was about refactoring how this looks. If you don't have any ideas for a significant change that can be beneficial to organizers, perhaps you should close this PR, until there are more ideas on the table, which we can discuss on the issue.

BelieveC:
@differentreality, Alright!

PR #1303, opened 2017-02-18
Include test for resources in features/ability_spec.rb

shlok007:


shlok007:
We are missing the other tests for volunteers coordinator here. I have added the one for resources and should probably add others in a separate PR. :slightly_smiling_face:

shlok007:
@Ana06 Please do tell me if I am missing something here. :slightly_smiling_face:

PR #1306, opened 2017-02-20
fixed Confirmed Scheduled Events issue

### LINKED ISSUE #1302: Issue with Confirmed events scheduled chart on event index page
Right now to check if the event is scheduled or not we check if its start time attribute is set or not. I think instead of this we should check if there is an entry for that respective confirmed event in EventSchedule table.

siddhantbajaj:
closes #1302

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -311,9 +311,10 @@ class Conference < ActiveRecord::Base
   # ====Returns
   # * +hash+ -> hash
   def scheduled_event_distribution
-    confirmed_events = program.events.where(state: 'confirmed')
-    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
-    unscheduled_value =  { 'value' => confirmed_events.where(start_time: nil).count, 'color' => 'red' }
+    confirmed_scheduled_events = program.events.confirmed.scheduled(program.selected_schedule.try(:id))
+    confirmed_unscheduled_events = program.events.confirmed - confirmed_scheduled_events
+    scheduled_value = { 'value' => confirmed_scheduled_events.count, 'color' => 'green' }
+    unscheduled_value = { 'value' => confirmed_unscheduled_events.count, 'color' => 'red' }
     { 'Scheduled' => scheduled_value, 'Unscheduled' => unscheduled_value }
   end
 
--- a/spec/models/conference_spec.rb
+++ b/spec/models/conference_spec.rb
@@ -778,6 +778,28 @@ describe Conference do
     end
   end
 
+  describe '#scheduled_event_distribution' do
+    let(:conference) { create(:conference) }
+    let(:confirmed_unscheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
+    let(:confirmed_scheduled_event) { create(:event_scheduled, program: conference.program) }
+
+    it '#scheduled_event_distribution does calculate correct values with events' do
+      confirmed_unscheduled_event
+      confirmed_scheduled_event
+      result = {}
+      result['Scheduled'] = { 'value' => 1, 'color' => 'green' }
+      result['Unscheduled'] = { 'value' => 1, 'color' => 'red' }
+      expect(conference.scheduled_event_distribution).to eq(result)
+    end
+
+    it '#scheduled_event_distribution does calculate correct values with no events' do
+      result = {}
+      result['Scheduled'] = { 'value' => 0, 'color' => 'green' }
+      result['Unscheduled'] = { 'value' => 0, 'color' => 'red' }
+      expect(conference.scheduled_event_distribution).to eq(result)
+    end
+  end
+
   describe '#event_distribution' do
 
     before(:each) do


app/models/conference.rb
```diff
@@ -329,8 +329,10 @@ def event_distribution
   # * +hash+ -> hash
   def scheduled_event_distribution
     confirmed_events = program.events.where(state: 'confirmed')
-    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
-    unscheduled_value =  { 'value' => confirmed_events.where(start_time: nil).count, 'color' => 'red' }
+    confirmed_scheduled_events = confirmed_events.reject{|a| a.event_schedules.empty?}
```
Ana06: what about introducing an scope in `Event` model for that?

siddhantbajaj: Sorry my bad, actually we already have a method to check if the event is scheduled.

app/models/conference.rb
```diff
@@ -329,8 +329,10 @@ def event_distribution
   # * +hash+ -> hash
   def scheduled_event_distribution
     confirmed_events = program.events.where(state: 'confirmed')
-    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
-    unscheduled_value =  { 'value' => confirmed_events.where(start_time: nil).count, 'color' => 'red' }
+    confirmed_scheduled_events = confirmed_events.select{|event| event.scheduled?}
```
Ana06: But even with the method I think it would be better to introduce an scope in the model for this, as we may want to use it somewhere else

Ana06: And you can write this with an argument instead of a block:

``` Ruby
confirmed_scheduled_events = confirmed_events.select(&:scheduled?)
```

siddhantbajaj: But how will we define the scope as the Event model does not have any attribute that defines if the event is scheduled or not?:confused:

Ana06: You can use methods as well in scopes

```  Ruby
scope : scheduled, -> { select(&:scheduled?) }
```

siddhantbajaj: Yes we should surely add this, but using it here might make the code a bit confusing.It will give all the scheduled events from all the conferences so we will have to select only the events of that particular conference.Ideal solution for me would be something like this:

`program.events.scheduled`

what do you think?

app/models/conference.rb
```diff
@@ -328,9 +328,10 @@ def event_distribution
   # ====Returns
   # * +hash+ -> hash
   def scheduled_event_distribution
-    confirmed_events = program.events.where(state: 'confirmed')
-    scheduled_value =  { 'value' => confirmed_events.where.not(start_time: nil).count, 'color' => 'green' }
-    unscheduled_value =  { 'value' => confirmed_events.where(start_time: nil).count, 'color' => 'red' }
+    confirmed_scheduled_events = Event.scheduled.select{|event| event.program == program}
```
Ana06: You can apply the scope to the events returned by program so you don't need to add the select part :wink:

`confirmed_scheduled_events = program.events.scheduled`

app/models/conference.rb
```diff
@@ -328,9 +328,10 @@ def event_distribution
...
+    confirmed_unscheduled_events = Event.unscheduled.select{|event| event.program == program}
```
Ana06: And here you could even do

`confirmed_unscheduled_events = program.events - confirmed_scheduled_event`

So you avoid calling the the reject(&:scheduled?) which would call the `scheduled?` method for all the objects. With that the `unscheduled` scope is not needed.

app/models/conference.rb
siddhantbajaj: @Ana06 Yes it worked :wink:.But since our Program model also has a method named scheduled so I just changed the name of our scope.

Ana06:
@siddhantbajaj 
> But since our Program model also has a method named scheduled so I just changed the name of our scope

or you can use the scope in program, as it is doing the same things and it is more directly, right? :wink:

siddhantbajaj:
@Ana06 yes we can do that as well. But that method takes schedule_id as argument so it can give schedule events of any schedule but with scope we can not do that.

Ana06:
@siddhantbajaj  I mean

`confirmed_scheduled_events = program.events.confirmed.scheduled(program.selected_schedule.id)`

The good thing of using this is that we use a join instead of applying the `scheduled?` method over every event.

siddhantbajaj:
@Ana06 yes you are right. :+1:

app/models/event.rb
```diff
@@ -41,6 +41,7 @@ class Event < ActiveRecord::Base
   scope :canceled, -> { where(state: 'canceled') }
   scope :withdrawn, -> { where(state: 'withdrawn') }
   scope :highlighted, -> { where(is_highlight: true) }
+  scope :scheduled_events, -> { confirmed.select(&:scheduled?) }
```
Ana06: Now you are not using this scope, so you can remove it. Sorry for the confusion, I realised that it was better to use the method in program afterwards. :confounded:

siddhantbajaj: No problem. I have made the changes.:wink:

siddhantbajaj:
@Ana06 I have made the changes, squashed commits and rebased.:wink:

siddhantbajaj:
@Ana06 we have a problem with this:
> confirmed_scheduled_events = program.events.confirmed.scheduled(program.selected_schedule.id)

conference will not have a selected schedule always. Therefore tests were failing.
I have made some changes to fix this.

differentreality:
This looks good to me, @Ana06?
We should definitely add tests though.

Ana06:
@differentreality looks great now! :wink: 

and yes, we need tests for it, to ensure that It doesn't break again. @siddhantbajaj can you please do this?

siddhantbajaj:
@Ana06 done :+1:

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,35 @@
     end
   end
 
+  describe '#scheduled_event_distribution' do
+
+    before(:each) do
+      @conference = create(:conference)
```
Ana06: there is already a conference variable, that you can use: https://github.com/siddhantbajaj/osem/blob/b8af1854a4a255d62a4507e7607f01ae89b79f02/spec/models/conference_spec.rb#L7

Although I have no idea why it is called `subjetc`. Maybe we should rename it (if you want to do so in this PR please add it in a different commit) :thinking:

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,35 @@
...
+
+      selected_schedule = create(:schedule, program: @conference.program)
```
Ana06: creation of variable should be in a `let`, lazy evaluation and you always know where to fine all the variables that you use in your test.

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,35 @@
...
+      @conference.program.update_attributes!(selected_schedule: selected_schedule) 
+
+      confirmed_unscheduled_event = create(:event, program: @conference.program, state: 'confirmed')
+      confirmed_scheduled_event = create(:event, program: @conference.program, state: 'confirmed')
+      create(:event_schedule, event: confirmed_scheduled_event, schedule: selected_schedule)
+
+      @result = {}
```
Ana06: we don't use `@` in test variables

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,35 @@
...
 
+  describe '#scheduled_event_distribution' do
+
+    before(:each) do
+      @conference = create(:conference)
+
+      selected_schedule = create(:schedule, program: @conference.program)
+      @conference.program.update_attributes!(selected_schedule: selected_schedule) 
+
+      confirmed_unscheduled_event = create(:event, program: @conference.program, state: 'confirmed')
```
Ana06: all this variables are only needed for the first test, so do not create them for both tests, the same with the assignment of the `result` variable :wink:

spec/models/conference_spec.rb
siddhantbajaj: OK, but should we use this as then we will have to change its start and end date so that the validations would pass and that's why I see most of the test are creating another conference variable instead of using it.

spec/models/conference_spec.rb
siddhantbajaj: done.

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,33 @@
     end
   end
 
+  describe '#scheduled_event_distribution' do
+    let!(:conference) { create(:conference) }
```
Ana06: why `let!` instead of `let`?

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,33 @@
...
+
+    it '#scheduled_event_distribution does calculate correct values with events' do
+      result = {} 
+      result['Scheduled'] = { 'value' => 1, 'color' => 'green' }
+      result['Unscheduled'] = { 'value' => 1, 'color' => 'red' }
+      expect(conference.scheduled_event_distribution).to eq(result)
+    end
+
+    it '#scheduled_event_distribution does calculate correct values with no events' do
+      conference.program.events.clear
```
Ana06: instead of clearing then, just do not create them here.

spec/models/conference_spec.rb
Ana06: Same for the others `let!`

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,33 @@
...
 
+  describe '#scheduled_event_distribution' do
+    let!(:conference) { create(:conference) }
+    let!(:selected_schedule) { create(:schedule, program: conference.program) }
+    let!(:confirmed_unscheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
+    let!(:confirmed_scheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
+    let!(:event_schedule) { create(:event_schedule, event: confirmed_scheduled_event, schedule: selected_schedule) }
+
+    before(:each) do 
+      conference.program.update_attributes!(selected_schedule: selected_schedule)
```
Ana06: if you don't give the `event_schedule` a schedule when creating it, the selected schedule will be automatically created (check [EventSchedule factory](https://github.com/openSUSE/osem/blob/master/spec/factories/event_schedule.rb)) and then you don't need this line, neither the `selected_schedule` creation (line 783).

spec/models/conference_spec.rb
siddhantbajaj: > why let! instead of let?

Just don't want to create conference in all the tests. 

> Same for the others let!

Yes we can use let at othere places.

spec/models/conference_spec.rb
siddhantbajaj: done. :+1:

spec/models/conference_spec.rb
siddhantbajaj: done. :+1:

spec/models/conference_spec.rb
Ana06: it will be created when calling the variable, `let!` is not needed. Please read: https://www.relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let :wink:

siddhantbajaj: >Just did not want to create conference in all the tests.

@Ana06 sorry by create here I mean call. Basically I was thinking that we should not call conference variable in both our test instead it should created immediately for both our test that's why I was using `let!` instead of `let` for conference. Since conference is needed in mostly all tests so even if we decide to add more test for scheduled_event_distribution we won't have to call conference variable in the new test.

siddhantbajaj:
Just did not want to create conference in all the tests.
@Ana06 sorry by create here I mean call. Basically I was thinking that we should not call conference variable in both our test instead it should created immediately for both our test that's why I was using let! instead of let for conference. Since conference is needed in mostly all tests so even if we decide to add more test for scheduled_event_distribution we won't have to call conference variable in the new test.

spec/models/conference_spec.rb
Ana06: @siddhantbajaj I understodd what you mean, but you don't need to call it in every test. If you know change `let!` by `let` it will work without changing  anything else.

As it is said in the [link](https://www.relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let), `let` is using lazy evaluation, and the variable will be created as soon as it is called the first time. That means that if you create another test that don't use it it won't be created. In your first test if would be created just before creating ` confirmed_unscheduled_event`. In the second would be created in the expect as it is called there.

Do you understand what I mean? :slightly_smiling_face:

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,30 @@
     end
   end
 
+  describe '#scheduled_event_distribution' do
+    let!(:conference) { create(:conference) }
+    let(:confirmed_unscheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
+    let(:confirmed_scheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
```
Ana06: you can create a `event_scheduled` instead of a normal event and then you don't need to create the event_schedule and to set the state. Look at: https://github.com/openSUSE/osem/blob/master/spec/factories/events.rb#L30

spec/models/conference_spec.rb
siddhantbajaj: @Ana06 yes sorry my bad, I was thinking that I will have to call conference specifically in both the test. Yes you are right it will be called automatically just before `confirmed_unscheduled_event` because we are using conference in it. I have made the changes. :slightly_smiling_face:

spec/models/conference_spec.rb
siddhantbajaj: done. :+1:

siddhantbajaj:
Squashed and rebased. :+1:

siddhantbajaj:
@differentreality @Ana06 Thanks a lot for your help. :wink:

spec/models/conference_spec.rb
```diff
@@ -778,6 +778,28 @@
...
 
+  describe '#scheduled_event_distribution' do
+    let(:conference) { create(:conference) }
+    let(:confirmed_unscheduled_event) { create(:event, program: conference.program, state: 'confirmed') }
+    let(:confirmed_scheduled_event) { create(:event_scheduled, program: conference.program) }
+
+    it '#scheduled_event_distribution does calculate correct values with events' do
+      confirmed_unscheduled_event
+      confirmed_scheduled_event
+      result = {}
```
Ana06: remove the whitespace in this line to make Rubocop happy and add it to the commit you already have :wink:

siddhantbajaj: done. :+1:

PR #1308, opened 2017-02-21
Implements configurable start and end hour of the dayused in schedule

davidsf:
This PR implements configurable start and end hour of the day to be used in the schedule (our event ends after 18:00h)


davidsf:
After the merge of PR #1143, this PR make no sense. Closing.

Issue #1298, opened 2017-02-16
Program.any_events_for_this_date? is possibly broken


eug3nix:
Hi everyone, please take a look at this
https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L153
this method rely on the events.start_time column which seems to be obsolete
as a result, the return value for this method is always FALSE
resulting in all dates on the public schedule events tab dropdown being greyed-out
I played with the code in IRB and it seems that this may be the correct code for the method:
`EventSchedule.where(schedule_id: selected_schedule.id).where(start_time: parsed_date..(parsed_date + 1)).any?`
Please advise

P.S. here is the screen which shows the issue from the user perspective(note the greyed-out dropdown items vs the event date)
![screenshot at 2017-02-22 16 13 44](https://cloud.githubusercontent.com/assets/4243476/23215406/efa37348-f91a-11e6-8a1e-d0b654980c52.png)

Thanks,
Eugene/CommandPrompt Inc.

differentreality:
https://github.com/openSUSE/osem/issues/1154
https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L246

Ana06:
This is already solved: #1309

PR #1309, opened 2017-02-22
fixes bug OSEM#1298

eug3nix:
Hi.
This pull request fixes the bug #1298 which I reported several days ago.
The bugfix incorporates the code which I proposed in the bug report.

Best regards,
Eugene/CommandPrompt Inc

### DIFF
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -150,7 +150,7 @@ class Program < ActiveRecord::Base
   # * +False+ -> If there is not any event for the given date
   def any_event_for_this_date?(date)
     parsed_date = DateTime.parse("#{date} 00:00").utc
-    events.where(start_time: parsed_date..(parsed_date + 1)).any?
+    EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1)).any?
   end
 
   private


app/models/program.rb
```diff
@@ -150,7 +150,7 @@ def languages_list
   # * +False+ -> If there is not any event for the given date
   def any_event_for_this_date?(date)
     parsed_date = DateTime.parse("#{date} 00:00").utc
-    events.where(start_time: parsed_date..(parsed_date + 1)).any?
+    EventSchedule.where(schedule_id: selected_schedule.id).where(start_time: parsed_date..(parsed_date + 1)).any?
```
Ana06: This is great! :+1: 

Just one thing, you can use the schedule directly without using the `id`, which is more readable and it is using the Rails relation instead of the `id` column. I mean: 

`EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1)).any?`

differentreality:
We should probably also add a test for this method.

differentreality:
@eug3nix would you please add relevant tests as well?

Ana06:
@eug3nix ping!

If you don't have time, maybe we can merge it and write the tests afterwards. It is currently wrong, anyway, so better merge it without tests than leaving it wrong.

eug3nix:
@Ana06 I agree, that would be fantastic if someone can handle that, since I cannot spend much time on it at the moment

PR #1311, opened 2017-02-23
Fixed #1310 removed duplicate methods

### LINKED ISSUE #1310: Duplicate methods in application_controller and conference.rb
method date_string in application_controller and method date_range_string in conference.rb produce the same result so we can keep one of them and remove the other .

nikhilgupta1211:
#1310

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,40 @@
 module ApplicationHelper
   ##
+  # Returns a string build from the start and end date of the given conference.
+  #
+  # If the conference is only one day long
+  # * %B %d %Y (January 17 2014)
+  # If the conference starts and ends in the same month and year
+  # * %B %d - %d, %Y (January 17 - 21 2014)
+  # If the conference ends in another month but in the same year
+  # * %B %d - %B %d, %Y (January 31 - February 02 2014)
+  # All other cases
+  # * %B %d, %Y - %B %d, %Y (December 30, 2013 - January 02, 2014)
+  def date_string(start_date, end_date)
+    startstr = 'Unknown - '
+    endstr = 'Unknown'
+    # When the conference is in the same month
+    if start_date.month == end_date.month && start_date.year == end_date.year
+      if start_date.day == end_date.day
+        startstr = start_date.strftime('%B %d')
+        endstr = end_date.strftime(' %Y')
+      else
+        startstr = start_date.strftime('%B %d - ')
+        endstr = end_date.strftime('%d, %Y')
+      end
+    elsif start_date.month != end_date.month && start_date.year == end_date.year
+      startstr = start_date.strftime('%B %d - ')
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,5 +1,40 @@
...
+  # If the conference starts and ends in the same month and year
+  # * %B %d - %d, %Y (January 17 - 21 2014)
+  # If the conference ends in another month but in the same year
+  # * %B %d - %B %d, %Y (January 31 - February 02 2014)
+  # All other cases
+  # * %B %d, %Y - %B %d, %Y (December 30, 2013 - January 02, 2014)
+  def date_string(start_date, end_date)
+    startstr = 'Unknown - '
+    endstr = 'Unknown'
+    # When the conference  in the same month
```
differentreality: You are missing and `is`

differentreality:
Nice @nikhilgupta1211 :) 
Where are the tests for the function though?

nikhilgupta1211:
@differentreality  added the tests for the function :)

app/helpers/application_helper.rb
nikhilgupta1211: Fixed it

nikhilgupta1211:
@differentreality can we merge this ? Actually a lot of other people have also been looking into this 😅

spec/helpers/application_helper_spec.rb
```diff
@@ -60,6 +60,24 @@
     end
   end
 
+  describe 'date_string' do
```
differentreality: #date_string

spec/helpers/application_helper_spec.rb
```diff
@@ -60,6 +60,24 @@
...
+    it 'If the conference is only one day long' do
```
differentreality: A better wording for your tests would be
```
context 'returns correct string'
  it 'when conference lasts 1 day'
  it 'when conference starts and ends in the same month and year'
  it 'when conference ends in another month, of the same year'
  it 'when conference ends in another month, of a different year'
```

Generally it is common practice to use **when** in it blocks, and if the test was to fail, it would read:

> #date_string returns correct string when conference lasts 1 day

differentreality:
If we are only using this function in the conference serializer, why doens't it make sense to leave it in the conference model instead of moving it to the application helper, which we then need to add in the conference serializer in order for the function to work?

nikhilgupta1211:
We are using the date_string method in these files also : -
app/views/conferences/_conference_details.html.haml 
app/views/conferences/show.html.haml 
app/views/conference_registrations/show.html.haml 

and as we are using this method for the views i thought to define it in the application_helper

differentreality:
@nikhilgupta1211 ah, yes, you are right :) 
This looks good! Let's rebase and squash commits - and wait for travis to succeed :-)

differentreality:
:tada: Good job! Thank you :-)

Issue #1315, opened 2017-03-01
Document openid implementation better


differentreality:
While there is some mention about implementing openid in our installation instructions, this is mainly intended for production.

It is very unclear to new contributors as to what needs to be done in order to implement openid in development - or that this is even possible!

AEtherC0r3:
Hello @differentreality, can you assign this issue to me?

differentreality:
@AEtherC0r3 go ahead!

AEtherC0r3:
This should have been closed when #1366 was merged

PR #1316, opened 2017-03-01
Fixes #1315, Add documentation for openID in development

BelieveC:
Fixes #1315
Add documentation to use openID in development.

PR #1317, opened 2017-03-02
Fixing issue #1315

Ramaraju-1996:
These 2 file changes tries to fix issue #1315(Document openid implementation better).


BelieveC:
@Ramaraju-1996  
PR #1316 already exist for the same issue.

Ramaraju-1996:
Thanks a lot @BelieveC 
I thought that PR#1316 was still open, so I thought may be I could also try to solve the issue.
Will look into the problem.

PR #1319, opened 2017-03-02
#1220 adding the cell size attribute to the database

JewelSam:
#1220

Currently we cannot have 5-minute slots for schedule of conf A, and 15-minute slots for conf B. So this PR adds cell size to the programs table.

### DIFF
--- a/app/controllers/admin/programs_controller.rb
+++ b/app/controllers/admin/programs_controller.rb
@@ -12,13 +12,15 @@ module Admin
       @program = @conference.program
       @program.assign_attributes(program_params)
       send_mail_on_schedule_public = @program.notify_on_schedule_public?
+      event_schedules_count_was = @program.event_schedules.count
 
       if @program.update_attributes(program_params)
         ConferenceScheduleUpdateMailJob.perform_later(@conference) if send_mail_on_schedule_public
         respond_to do |format|
           format.html do
-            redirect_to admin_conference_program_path(@conference.short_title),
-                        notice: 'The program was successfully updated.'
+            notice = 'The program was successfully updated.'
+            notice += ' You changed schedule interval and some events were unscheduled.' if @program.event_schedules.count != event_schedules_count_was
+            redirect_to admin_conference_program_path(@conference.short_title), notice: notice
           end
           format.js { render json: {} }
         end
@@ -36,7 +38,7 @@ module Admin
     private
 
     def program_params
-      params.require(:program).permit(:rating, :schedule_public, :schedule_fluid, :languages, :blind_voting, :voting_start_date, :voting_end_date, :selected_schedule_id)
+      params.require(:program).permit(:rating, :schedule_public, :schedule_interval, :schedule_fluid, :languages, :blind_voting, :voting_start_date, :voting_end_date, :selected_schedule_id)
     end
   end
 end
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -38,6 +38,7 @@ class Program < ActiveRecord::Base
       where(state: :confirmed, is_highlight: true)
     end
   end
+  has_many :event_schedules, through: :events
 
   has_many :event_users, through: :events
   has_many :speakers, -> { distinct }, through: :event_users, source: :user do
@@ -52,11 +53,15 @@ class Program < ActiveRecord::Base
 
 #   validates :conference_id, presence: true, uniqueness: true
   validates :rating, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 10, only_integer: true }
+  validates :schedule_interval, numericality: { greater_than_or_equal_to: 5, less_than_or_equal_to: 60 }, presence: true
+  validate :schedule_interval_divisor_60
   validate :voting_start_date_before_end_date
   validate :voting_dates_exist
 
   after_create :create_event_types
   after_create :create_difficulty_levels
+  after_save :unschedule_unfit_events, if: :schedule_interval_changed?
+  after_save :normalize_event_types_length, if: :schedule_interval_changed?
   validate :check_languages_format
 
   # Returns all event_schedules for the selected schedule ordered by start_time
@@ -198,5 +203,32 @@ class Program < ActiveRecord::Base
     errors.add(:languages, "can't be repeated") && return unless languages_array.uniq!.nil?
     # We check if every language is a valid ISO 639-1 language
     errors.add(:languages, 'must be ISO 639-1 valid codes') unless languages_array.select{ |x| ISO_639.find(x).nil? }.empty?
+  end
+
+  ##
+  # Check if schedule_interval is a divisor of 60 minutes
+  #
+  def schedule_interval_divisor_60
+    errors.add(:schedule_interval, 'must be a divisor of 60') if schedule_interval > 0 && 60 % schedule_interval > 0
+  end
+
+  ##
+  # Unschedule all the events which don't fit
+  #
+  def unschedule_unfit_events
+    unfit_schedules = event_schedules.select do |event_schedule|
+      event_schedule.start_time.min % schedule_interval > 0
+    end
+    EventSchedule.where(id: unfit_schedules.map(&:id)).destroy_all
+  end
+
+  ##
+  # Change event type length according schedule interval
+  #
+  def normalize_event_types_length
+    event_types.each do |event_type|
+      new_length = event_type.length > schedule_interval ? event_type.length - (event_type.length % schedule_interval) : schedule_interval
+      event_type.update_attributes length: new_length
+    end
   end
 end
--- a/app/views/admin/programs/_form.html.haml
+++ b/app/views/admin/programs/_form.html.haml
@@ -9,6 +9,7 @@
       = f.input :schedule_fluid, label: 'Allow submitters to change their event after it is scheduled'
       = f.input :rating, hint: 'Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals.'
       = f.input :languages, hint: "Enter the languages allowed for events as values of #{link_to('ISO 639-1', 'http://www.loc.gov/standards/iso639-2/php/code_list.php', target: "_blank")} language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages.".html_safe
+      = f.input :schedule_interval, hint: "It is the minimal time interval of your schedule. The value should be 5, 6, 10, 12, 15, 20, 30 or 60. Warning! Some events could be unscheduled when changing this value."
       = f.input :blind_voting, hint: 'Enable this feature if you do not want to show voting results and voters prior to user submitting a vote. For the feature to work you need to set the voting dates below as well'
       = f.input :voting_start_date, as: :string, input_html: { id: 'datetimepicker-voting_start_date', readonly: true, value: (f.object.voting_start_date.to_formatted_s(:db_without_seconds) unless f.object.voting_start_date.nil?) }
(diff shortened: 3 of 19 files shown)


Ana06:
This is not really about _made schedule cell size as conference option_ (that is already implemented) but about _adding the cell size attribute to the database_, right?

Ana06:
This is not really about _made schedule cell size as conference option_ (that is already implemented) but about _adding the cell size attribute to the database_, right?

@differentreality @eug3nix do you like the idea?

app/views/admin/programs/_form.html.haml
```diff
@@ -9,6 +9,7 @@
       = f.input :schedule_fluid, label: "Allow submitters to change their event after it is scheduled"
       = f.input :rating, hint: 'Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals.'
       = f.input :languages, hint: "Enter the languages allowed for events as values of #{link_to('ISO 639-1', 'http://www.loc.gov/standards/iso639-2/php/code_list.php', target: "_blank")} language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages.".html_safe
+      = f.input :schedule_cell_size, hint: "It is minimal time slots of your program. The value should be greater than 5, less than 60 and multiple of 60."
```
Ana06: I think the hint is not very clear. I would rewrite it, what about this:

_It is the minimal time interval of your schedule. The value should be 5, 6, 10, 12, 15, 20, 30 or 60_

spec/models/program_spec.rb
```diff
@@ -32,6 +32,18 @@
 
     it { is_expected.to validate_numericality_of(:rating).is_greater_than_or_equal_to(0).is_less_than_or_equal_to(10).only_integer }
 
+    it { is_expected.to validate_numericality_of(:schedule_cell_size).is_greater_than_or_equal_to(5).is_less_than_or_equal_to(60) }
+
+    describe 'schedule_cell_size_multiple_60' do
+      it 'is valid, when schedule_cell_size is multiple of 60' do
```
Ana06: _is valid, when schedule_cell_size is a **divisor** of 60_

spec/models/program_spec.rb
```diff
@@ -32,6 +32,18 @@
...
+        expect(build(:program, schedule_cell_size: 20)).to be_valid
+      end
+
+      it 'is not valid, when schedule_cell_size is not multiple of 60' do
```
Ana06: is not valid, when schedule_cell_size is not a **divisor** of 60

app/views/admin/programs/show.html.haml
```diff
@@ -49,6 +49,11 @@
             Yes
           - else
             No
+        %dt
+          Schedule cell size
```
Ana06: I don't like schedule cell size, what about _schedule interval_? :thinking:

app/models/program.rb
```diff
@@ -199,4 +201,11 @@ def check_languages_format
     # We check if every language is a valid ISO 639-1 language
     errors.add(:languages, 'must be ISO 639-1 valid codes') unless languages_array.select{ |x| ISO_639.find(x).nil? }.empty?
   end
+
+  ##
+  # Check if schedule_cell_size is multiple of 60 minutes
+  #
+  def schedule_cell_size_multiple_60
+    errors.add(:schedule_cell_size, 'must be multiple of 60') if schedule_cell_size > 0 && 60 % schedule_cell_size > 0
```
Ana06: must be a **divisor** of 60

app/models/program.rb
```diff
@@ -199,4 +201,11 @@ def check_languages_format
     # We check if every language is a valid ISO 639-1 language
     errors.add(:languages, 'must be ISO 639-1 valid codes') unless languages_array.select{ |x| ISO_639.find(x).nil? }.empty?
   end
+
+  ##
+  # Check if schedule_cell_size is multiple of 60 minutes
+  #
+  def schedule_cell_size_multiple_60
```
Ana06: It should be a divisor of 60, not a multiple. Please change the name of the method too. :wink:

eug3nix:
@Ana06 @differentreality Hey!
There is another thing to be considered while making this property easily accessible by the system users, imagine this use case:
I as a user set up a conference and the timeslot size of a conference
I add events and schedule these
Then I decide to change the timeslot size while having some older events in my schedule

There is a potential for some new bugs in that case since the LENGTH_STEP is used for the event_type duration validations and we may potentially end up with formally invalid event_types with incorrect lengths.
The schedule grids rendering will also be affected since the code is also using LENGTH_STEP in the event-brick size calculations.

So you may need to add another extra layer of "foolproofness" to either force re-validation of the already existing records OR maybe freeze this new functionality for non-pristine conferences (which is also not very elegant in my opinion)

Ana06:
@eug3nix 

> Then I decide to change the timeslot size while having some older events in my schedule


Right! When this attribute is changed we need to unscheduled all the events which doesn't fit.

JewelSam:
@Ana06 I fixed all requested changes. And feel free to fix my mistakes, I speak English not well :(
I have some questions:
1. I don't sure in my decision about event_types. But I couldn't find better. Any other ideas?
And if schedule interval was changed should we show some warning? About unscheduling events and changing event type's length. 
2. Could you help me, what should I do with Hakiri? I can't understand what I do wrong and how I can fix its warning "Attribute Restriction".

app/models/program.rb
```diff
@@ -199,4 +204,30 @@ def check_languages_format
...
+  # Check if schedule_interval is a divisor of 60 minutes
+  #
+  def schedule_interval_divisor_60
+    errors.add(:schedule_interval, 'must be a divisor of 60') if schedule_interval > 0 && 60 % schedule_interval > 0
+  end
+
+  ##
+  # Unschedule all the events which don't fit
+  #
+  def unschedule_unfit_events
```
Ana06: I am not pretty sure of what are unfit events for you? I mean, what events are you destroying in this method?


Because there are several options here. Deleting all the `event_schedules` for this conference would be one option. As I think it is not very clear what should happen with the events when you change the granularity of the schedule. @differentreality @eug3nix ideas?

Ana06:
@JewelSam 

> I don't sure in my decision about event_types. But I couldn't find better. Any other ideas?

What do you exactly mean?

> And if schedule interval was changed should we show some warning? About unscheduling events and changing event type's length.

Yes, if the events are deleted or the length of the events if changed we should render message afterwards. It would be great to have a warning in the schedule granularity to inform that changing it would eliminate/modify the current schedule and change the length of events.

> Could you help me, what should I do with Hakiri? I can't understand what I do wrong and how I can fix its warning "Attribute Restriction".

Hakiri says: _The project has a total of 1 security warning (0 in code and 1 in dependencies)_. So, the problem is not in your code, but in a gem. The error was solved in https://github.com/openSUSE/osem/pull/1323 so if you rebase it will be solved. :wink:

app/models/program.rb
JewelSam: I destroy all event schedules with start time is not suitable for new schedule interval. For example: 
schedule interval was 15 minutes
event1 start time = 9:15
event2 start_time = 9:30
new schedule interval = 10 minutes
Only event1 will be destroyed because 9:30 can be start time with new schedule interval 10 minutes

app/models/program.rb
```diff
@@ -199,4 +204,30 @@ def check_languages_format
...
+      diff % schedule_interval > 0
+    end
+    EventSchedule.where(id: unfit_schedules.map(&:id)).destroy_all
+  end
+
+  ##
+  # Change event type length according schedule interval
+  #
+  def normalize_event_types_length
+    event_types.each {|event_type| event_type.update_attributes length: schedule_interval * (event_type.length / schedule_interval_was) }
```
JewelSam: done

JewelSam:
@Ana06 I added warning but I'm not sure that it is clear enough :( 
Thank you for the explanation about Hakiri :)
I adjusted the length change in event types.

Ana06:
@JewelSam please read the [commits](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#commits) and [pull request](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#pull-requests-workflow) sections that we recently added to our [contributing guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md). :wink:

app/models/program.rb
```diff
@@ -199,4 +204,33 @@ def check_languages_format
...
+  def schedule_interval_divisor_60
+    errors.add(:schedule_interval, 'must be a divisor of 60') if schedule_interval > 0 && 60 % schedule_interval > 0
+  end
+
+  ##
+  # Unschedule all the events which don't fit
+  #
+  def unschedule_unfit_events
+    conference_start_hour = conference.start_hour
+    unfit_schedules = event_schedules.select do |event_schedule|
```
Ana06: as the conference always started in 00 and the `schedule_interval` is a divisor of 60, this can be written simpler: :wink: 

```
unfit_schedules = event_schedules.select do |event_schedule|
event_schedule.start_time.min % schedule_interval
end
```

app/controllers/admin/programs_controller.rb
```diff
@@ -17,8 +17,9 @@ def update
         ConferenceScheduleUpdateMailJob.perform_later(@conference) if send_mail_on_schedule_public
         respond_to do |format|
           format.html do
-            redirect_to admin_conference_program_path(@conference.short_title),
-                        notice: 'The program was successfully updated.'
+            notice = "The program was successfully updated."
+            notice += " You changed schedule interval and schedule can be changed. Please check." if @program.previous_changes.has_key?(:schedule_interval)
```
Ana06: what about showing this only if an event_schedule was destroyed?

And then the message can be:

_You changed schedule interval and some events were unscheduled_

app/views/admin/programs/_form.html.haml
```diff
@@ -9,6 +9,7 @@
       = f.input :schedule_fluid, label: "Allow submitters to change their event after it is scheduled"
       = f.input :rating, hint: 'Enter the number of different rating levels you want to have for voting on proposals. Enter 0 if you do not want to vote on proposals.'
       = f.input :languages, hint: "Enter the languages allowed for events as values of #{link_to('ISO 639-1', 'http://www.loc.gov/standards/iso639-2/php/code_list.php', target: "_blank")} language codes separated with commas. The first language would be the default language. Leave it blank if you do not want to specify languages.".html_safe
+      = f.input :schedule_interval, hint: "It is the minimal time interval of your schedule. The value should be 5, 6, 10, 12, 15, 20, 30 or 60. Warning! It would modify the current schedule and change the length of events."
```
Ana06: I think this message would be a little bit  clearer: :wink: 

_It is the minimal time interval of your schedule. The value should be 5, 6, 10, 12, 15, 20, 30 or 60. Warning! Some events could be unscheduled when changing this value._

app/models/program.rb
Ana06: ok, I like the idea! :wink:

Ana06:
@JewelSam you will also need to rebase:

`git rebase upstream master`

JewelSam:
@Ana06 I made the changes you requested, made rebase and added PR description, but I don't know how I can reword my commits, they are very small. Should I squash them?

differentreality:
When you squash your commits with
 `git rebase -i commit_hash_you_keep_before_the_ones_to_be_squashed`
you are then asked, in a text editor (by default vim), about which commit message you want to keep (you don't have to keep an existing one, you can write a new commit message instead).

Ana06:
@JewelSam yes please squash them. @differentreality already wrote how to do it :wink:

JewelSam:
@Ana06 done :)

PR #1325, opened 2017-03-03
Reverting #1297 as it broke our heroku deployment

ChrisBr:
The problem is the 

``
require_relative '../../spec/support/external_request' 
``
 lib/tasks/factory_girl.rake which requires webmock. However, webmock is only available in the test group in the Gemfile and Heroku runs in production environment.

@Ana06 @siddhantbajaj

### DIFF
--- a/spec/support/database_cleaner.rb
+++ b/spec/support/database_cleaner.rb
@@ -1,16 +1,6 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.before(:each) do
     DatabaseCleaner.clean_with(:truncation)
     Rails.application.load_seed
-  end
-
-  config.before(:each) do |example|
-    DatabaseCleaner.strategy = (example.metadata[:js] == true) ? :truncation : :transaction
-    DatabaseCleaner.start
-  end
-
-  config.after(:each) do |example|
-    DatabaseCleaner.clean
-    Rails.application.load_seed if example.metadata[:js] == true
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,5 +25,4 @@ before_script:
   - RAILS_ENV=test bundle exec rake db:migrate --trace
 script:
   - 'bundle exec rubocop -Dc .rubocop.yml'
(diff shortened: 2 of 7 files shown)


spec/support/database_cleaner.rb
```diff
@@ -1,16 +1,6 @@
 RSpec.configure do |config|
-  config.before(:suite) do
+  config.before(:each) do
     DatabaseCleaner.clean_with(:truncation)
     Rails.application.load_seed
```
lagartoflojo: Loading seeds before each test will be really slow, no?
I think we should get rid of seeds completely in tests.

Ana06: But you always expect the seed to be there. But we can implement the changes @siddhantbajaj did in #1297 without breaking heroku, so it would only be run before each test suite.

Ana06:
@ChrisBr @siddhantbajaj we could have two different rake tasks for this. The actual one that is used in Heroku and then a different one for the tests. This will solve all the problems, right? :thinking:

spec/support/database_cleaner.rb
lagartoflojo: > But you always expect the seed to be there.

Exactly the problem... each test should start with an empty database, and then only get filled with the necessary data needed by that test..

siddhantbajaj:
@Ana06 why are we using this task in Heroku as this task won't execute without that mock request.
Can we just add this task in .slugignore?

Ana06:
@siddhantbajaj 

> why are we using this task in Heroku as this task won't execute without that mock request.

We shouldn't use it. That is what we should have two tasks, one for Heroku and the other one for testing with the mock request.

> Can we just add this task in .slugignore?

what do you mean?

siddhantbajaj:
@Ana06 I mean why we even have to keep one for heroku,we can just mention this task in slugignore file so that heroku will just ignore this task.
https://devcenter.heroku.com/articles/slug-compiler#ignoring-files-with-slugignore

hennevogel:
This is not about using any task in Heroku. This is about requiring some gem that in all environments that is only available in some. You just can't do that.

PR #1328, opened 2017-03-05
Fixes #1188 set timezone to conference`s timezone

### LINKED ISSUE #1188: Wrong timezone information in schedule.xml export
The <date> element of the <event> elements in the export have an trailing Z indicating UTC while the datetime itself seems to be a local time.

nikhilgupta1211:
Fixes #1188 
This sets the timezone for all the requests to the conference`s timezone.
I have also removed the method set_timezone_for_this_request in this PR as we were not using it .

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -7,14 +7,6 @@ module Admin
 
     before_action :get_event, except: [:index, :create, :reports]
 
-    # FIXME: The timezome should only be applied on output, otherwise
-    # you get lost in timezone conversions...
-    # around_filter :set_timezone_for_this_request
-
-    def set_timezone_for_this_request(&block)
-      Time.use_zone(@conference.timezone, &block)
-    end
-
     def index
       @events = @program.events
       @tracks = @program.tracks
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -3,6 +3,14 @@ class SchedulesController < ApplicationController
   before_action :respond_to_options
   load_and_authorize_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
+  around_action :use_timezone_for_this_request
+
+  # FIXME: The timezome should only be applied on output, otherwise
+  # you get lost in timezone conversions...
+
+  def use_timezone_for_this_request(&block)
+    Time.use_zone(@conference.timezone, &block)
+  end
 
   def show
     @rooms = @conference.venue.rooms if @conference.venue


app/controllers/admin/events_controller.rb
```diff
@@ -7,14 +7,6 @@ class EventsController < Admin::BaseController
 
     before_action :get_event, except: [:index, :create, :reports]
 
-    # FIXME: The timezome should only be applied on output, otherwise
-    # you get lost in timezone conversions...
-    # around_filter :set_timezone_for_this_request
-
-    def set_timezone_for_this_request(&block)
```
Ana06: This method was not used, right?

app/controllers/schedules_controller.rb
```diff
@@ -3,6 +3,14 @@ class SchedulesController < ApplicationController
   before_action :respond_to_options
   load_and_authorize_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
+  around_action :use_timezone_for_this_request
+
+  # FIXME: The timezome should only be applied on output, otherwise
+  # you get lost in timezone conversions...
+
+  def use_timezone_for_this_request(&block)
```
Ana06: Have you tried this? Because I think it will break the whole schedule as it is implemented to work in UTC.

app/controllers/admin/events_controller.rb
nikhilgupta1211: Yes we were not using the method

app/controllers/schedules_controller.rb
nikhilgupta1211: @Ana06 i tried it and it was working 😅, actually by this we converting each request to the conference`s timezone.

Ana06: I have just tried to create a new conference with GMT-10:00 time zone and it doesn't work. I scheduled the event like this:
image
and the event time keep being utc in the show page:
image

and the time in the all events is wrong:
image

It could be that you tried it with a conference which timezone is UTC.

Ana06:
As I said https://github.com/openSUSE/osem/issues/1188 I think it is much easier and simpler to take the time zone into account when exporting to xml than storing the events in the conference timezone. 

With your approach you will have to change how the events are stored and with the other one only how the hour of the events is rendered. And not having to take the time zone into account for the schedule is really an advantage. :slightly_smiling_face: 

Also, @aki94 wrote in https://github.com/openSUSE/osem/issues/1188 that he wanted to work on this too. Not sure if he is working on it or not, but it is a good habit to write in the issue that you are working on it to avoid that several people work on the same thing at the same time. :wink:

nikhilgupta1211:
I am really sorry i didn`t saw that message ,actually i did this a day before when there was no one working on it and pushed it on the next day. I will keep this in mind for the next time .

At first i did this with the help of using` in_time_zone` in the `show.xml.haml` to convert the timezone in xml would that have been better ?

PR #1243, opened 2017-01-09
Fixed XSS in markdown() method

cniedzwiedz:
Fixed XSS resulting from markdown() method + tests

PR #1332, opened 2017-03-06
change escape_html: false to true

agrim123:
Previously due to https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb#L298 
was causing below problem
old
whereas expected is new


ChrisBr:
@agrim123 the html: false was by intention to fix XSS 
https://github.com/openSUSE/osem/pull/1243

agrim123:
@ChrisBr I am trying to fix it and the problem that i have stated :smile:

agrim123:
@ChrisBr Tthe previous issue is now resolved .
Now we need to pass markdown to the markdown function

ChrisBr:
Makes sense!
Thanks @agrim123

PR #1334, opened 2017-03-06
Fixed DeleteEventSchedules issue

### LINKED ISSUE #1333: Issue with DeleteEventSchedules method
I was playing around with conference start and end hour attribute when I found some issues related to DeleteEventSchedules method in conference controller. This method basically delete all EventSchedules that are not in the hours range of that particular conference. Issues that I found:
1. It deletes EventSchedules of all the conferences that are not in the hours range. Instead it should  delete EventSchedules of those events only that belong to that particular conference only.
2. If we set invalid start or end hour attribute of a conference then also EventSchedules gets deleted even though conference is not successfully updated.

For more information on DeleteEventSchedules method:
https://github.com/openSUSE/osem/pull/1143

siddhantbajaj:
closes #1333

### DIFF
--- a/app/controllers/admin/conferences_controller.rb
+++ b/app/controllers/admin/conferences_controller.rb
@@ -82,7 +82,6 @@ module Admin
       short_title = @conference.short_title
       @conference.assign_attributes(conference_params)
       send_mail_on_conf_update = @conference.notify_on_dates_changed?
-      delete_event_schedules if @conference.start_hour_changed? || @conference.end_hour_changed?
 
       if @conference.update_attributes(conference_params)
         ConferenceDateUpdateMailJob.perform_later(@conference) if send_mail_on_conf_update
@@ -188,15 +187,6 @@ module Admin
                                          :sponsorship_levels_attributes, :sponsors_attributes,
                                          :targets, :targets_attributes,
                                          :campaigns, :campaigns_attributes, :registration_limit)
-    end
-
-    def delete_event_schedules
-      event_schedules = EventSchedule.select do |e|
-        e.start_time.strftime('%H').to_i < @conference.start_hour ||
-        e.end_time.strftime('%H').to_i > @conference.end_hour ||
-        (e.end_time.strftime('%H').to_i == @conference.end_hour && e.end_time.strftime('%M').to_i > 0)
-      end
-      event_schedules.each(&:destroy)
     end
   end
 end
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -67,6 +67,7 @@ class Conference < ActiveRecord::Base
   before_create :create_email_settings
 
   after_create :create_free_ticket
+  after_update :delete_event_schedules
 
   ##
   # Checks if the user is registered to the conference
@@ -78,6 +79,20 @@ class Conference < ActiveRecord::Base
   # * +true+ - If the user isn't registered
   def user_registered? user
     user.present? && registrations.where(user_id: user.id).count > 0
+  end
+
+  ##
+  # Delete all EventSchedules that are not in the hours range
+  # After the conference has been successfully updated
+  def delete_event_schedules
+    if start_hour_changed? || end_hour_changed?
+      event_schedules = program.event_schedules.select do |event_schedule|
+        event_schedule.start_time.hour < start_hour ||
+        event_schedule.end_time.hour > end_hour ||
+        (event_schedule.end_time.hour == end_hour && event_schedule.end_time.minute > 0)
+      end
+      event_schedules.each(&:destroy)
+    end
   end
 
   ##
--- a/spec/models/conference_spec.rb
+++ b/spec/models/conference_spec.rb
@@ -1639,4 +1639,23 @@ describe Conference do
       expect(free_ticket.price_cents).to eq(0)
     end
   end
+
+  describe 'after_update' do
+    let(:conference) { create(:conference) }
+    let(:scheduled_event_before_conference) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.start_hour.hours) }
+    let(:scheduled_event_after_conference) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours - 1.hour) }
+    let!(:scheduled_event_during_conference) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.start_hour.hours + 3.hours) }
+
+    it 'delete event schedules that are not in hour ranges, when conference start hour is updated' do
+      scheduled_event_before_conference
+      conference.start_hour = conference.start_hour + 1
+      expect{ conference.save }.to change{ EventSchedule.count }.from(2).to(1)
+    end
+
+    it 'delete event schedules that are not in hour ranges, when conference end hour is updated' do
+      scheduled_event_after_conference
+      conference.end_hour = conference.end_hour - 2
+      expect{ conference.save }.to change{ EventSchedule.count }.from(2).to(1)
+    end
+  end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -120,6 +120,10 @@ Metrics/ClassLength:
   Exclude:
     - 'app/models/conference.rb'
 
+Metrics/BlockLength:
+  Exclude:
+    - 'spec/models/conference_spec.rb'
+
 #################### Lint ###############################
 
 # Wrap your assignment in condition if you mean it, otherwise it is most likely equality check
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -10,6 +10,7 @@ class Program < ActiveRecord::Base
   has_many :tracks, dependent: :destroy
   has_many :difficulty_levels, dependent: :destroy
   has_many :schedules, dependent: :destroy
+  has_many :event_schedules, through: :schedules
   belongs_to :selected_schedule, class_name: 'Schedule'
   has_many :events, dependent: :destroy do
     def require_registration


app/controllers/admin/conferences_controller.rb
```diff
@@ -82,7 +82,9 @@ def update
       short_title = @conference.short_title
       @conference.assign_attributes(conference_params)
       send_mail_on_conf_update = @conference.notify_on_dates_changed?
-      delete_event_schedules if @conference.start_hour_changed? || @conference.end_hour_changed?
+      if @conference.valid? && (@conference.start_hour_changed? || @conference.end_hour_changed?)
```
differentreality: You can still do that in one line

siddhantbajaj: yes. :+1:

app/controllers/admin/conferences_controller.rb
```diff
@@ -82,7 +82,7 @@ def update
       short_title = @conference.short_title
       @conference.assign_attributes(conference_params)
       send_mail_on_conf_update = @conference.notify_on_dates_changed?
-      delete_event_schedules if @conference.start_hour_changed? || @conference.end_hour_changed?
+      delete_event_schedules if @conference.valid? && (@conference.start_hour_changed? || @conference.end_hour_changed?)
```
Ana06: you could move this inside the

`if @conference.update_attributes(conference_params)`

and then you now the conference is valid and you can just write:

`delete_event_schedules if @conference.start_hour_changed? || @conference.end_hour_changed?`

I like it more as you are sure that the conference start/end time was changed.

siddhantbajaj: Yes I tried that first,but the problem with this approach is:
> @conference.start_hour_changed? || @conference.end_hour_changed?

will not give you true as the conference is now updated.
Then I thought of using previous_changes()  but that too did not give the desired results.

Ana06: what about moving it to a `after_update` callback? You should be able to use `changed?` there and it is much more clear.

siddhantbajaj:
@Ana06 made the changes. :+1:

app/models/conference.rb
```diff
@@ -81,6 +82,21 @@ def user_registered? user
   end
 
   ##
+  # Delete all EventSchedules that are not in the hours range
+  # After the conference has been successfully updated
+  def delete_event_schedules
+    if self.start_hour_changed? || self.end_hour_changed?
+      event_schedules = EventSchedule.select do |e|
```
lagartoflojo: `EventSchedule.select do |e|` will go over EVERY event schedule in the database. Not cool.
We should try to scope this somehow.
Maybe adding a `has_many :event_schedules` to the conference.

siddhantbajaj: > Maybe adding a has_many :event_schedules to the conference.

Yes looks like the ideal solution. But it requires a lot of changes so should it be a part of this PR or should we handle this in a different PR? @Ana06 @lagartoflojo

Ana06: @lagartoflojo I think it would be even cooler if we add a `has_many thought` (using schedules relation for example) relation in program, as it is easier. and with that you can just do here:

`conference.program.event_schedules.select do |event_schedule|`

So, it implies just modifying two lines :wink: Add it in a different commit in this PR please :slightly_smiling_face:

Ana06: And please use `event` instead `e`

siddhantbajaj: @Ana06 `has_many thought` Sorry I didn't understand this. Are you trying to say instead of conference, program should have one-to-many relation with event_schedules model?

Ana06: @siddhantbajaj I mean adding a `has_many thought` relation in program to be able to call it like 

``` Ruby
conference.program.event_schedules.select do |event_schedule|
```

http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association

siddhantbajaj: @Ana06 Okay so by `has_many thought` here you meant `has_many through`.I wasn't getting  what **thought** meant here and  was thinking that you might mean has many through association but then thought of confirming it first. But don't we generally use has many through associations  for many to many relation. Why do you want to use it for one to many relation between program and event schedule? Please correct me if am getting it all wrong.

Ana06: @siddhantbajaj take a look at this paragraph of the [link](http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association), as it exactly our case:


> The has_many :through association is also useful for setting up "shortcuts" through nested has_many associations. For example, if a document has many sections, and a section has many paragraphs, you may sometimes want to get a simple collection of all paragraphs in the document.

siddhantbajaj: Thanks @Ana06 and sorry I didn't knew this thing that's why didn't get it first time. I have made the changes. :wink:

app/models/conference.rb
```diff
@@ -86,11 +86,11 @@ def user_registered? user
   # After the conference has been successfully updated
   def delete_event_schedules
     if self.start_hour_changed? || self.end_hour_changed?
-      event_schedules = EventSchedule.select do |e|
-        e.event.program == program &&
-        (e.start_time.strftime('%H').to_i < start_hour ||
-        e.end_time.strftime('%H').to_i > end_hour ||
-        (e.end_time.strftime('%H').to_i == end_hour && e.end_time.strftime('%M').to_i > 0))
+      event_schedules = program.event_schedules.select do |event_schedule|
+        event_schedule.event.program == program &&
```
Ana06: you don't need that line now, right?

app/models/conference.rb
```diff
@@ -86,11 +86,11 @@ def user_registered? user
...
+        (event_schedule.start_time.strftime('%H').to_i < start_hour ||
```
Ana06: you can use here the `hour` method: :wink: 

`event_schedule.start_time.hour < start_hour`

siddhantbajaj: done.

app/models/conference.rb
siddhantbajaj: Yes you are right, :+1:

app/models/conference.rb
```diff
@@ -87,10 +87,9 @@ def user_registered? user
   def delete_event_schedules
     if self.start_hour_changed? || self.end_hour_changed?
       event_schedules = program.event_schedules.select do |event_schedule|
-        event_schedule.event.program == program &&
-        (event_schedule.start_time.strftime('%H').to_i < start_hour ||
-        event_schedule.end_time.strftime('%H').to_i > end_hour ||
-        (event_schedule.end_time.strftime('%H').to_i == end_hour && event_schedule.end_time.strftime('%M').to_i > 0))
+        event_schedule.start_time.hour < start_hour ||
+        event_schedule.end_time.hour > end_hour ||
+        (event_schedule.end_time.hour == end_hour && event_schedule.end_time.hour > 0)
```
Ana06: I think here you meant `minute` instead of `hour` in the second part: 

`(event_schedule.end_time.hour == end_hour && event_schedule.end_time.minute > 0)`

siddhantbajaj: Yes I have made the change.  :+1:

siddhantbajaj:
Okay. :wink:

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,27 @@
       expect(free_ticket.price_cents).to eq(0)
     end
   end
+
+  describe 'after_update' do
+    let(:conference) { create(:full_conference) }
```
Ana06: why do you need it to be a `full conference`?

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,27 @@
...
+    let(:first_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.start_hour.hours) }
+    let(:second_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours + 1.hour) }
+
+    it 'delete event schedules that are not in hour ranges, when conference start hour is updated' do
+      first_scheduled_event
+      conference.start_hour = conference.start_hour + 1.hour
+      conference.save
+      conference.run_callbacks :update
+      free_ticket = conference.tickets.first
```
Ana06: why do you need this for? :confused:

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,27 @@
...
+
+  describe 'after_update' do
+    let(:conference) { create(:full_conference) }
+    let(:first_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.start_hour.hours) }
+    let(:second_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours + 1.hour) }
+
+    it 'delete event schedules that are not in hour ranges, when conference start hour is updated' do
+      first_scheduled_event
+      conference.start_hour = conference.start_hour + 1.hour
+      conference.save
```
Ana06: why not just

`expect(conference).not_to be valid?`

siddhantbajaj: why would conference not be valid? :confused:

spec/models/conference_spec.rb
siddhantbajaj: No specific reason just prefer to use full_conference with all basic things set. But yes conference will do just fine here. :+1:

siddhantbajaj: No specific reason just prefer to use full_conference with all basic things set. But yes conference will do just fine here. :+1:

spec/models/conference_spec.rb
siddhantbajaj: Sorry my bad, not needed. :confounded:

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,26 @@
       expect(free_ticket.price_cents).to eq(0)
     end
   end
+
+  describe 'after_update' do
+    let(:conference) { create(:conference) }
+    let(:first_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.start_hour.hours) }
+    let(:second_scheduled_event) { create(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours + 1.hour) }
```
differentreality: I think it makes more sense to name the variables into something like scheduled_event_after_conference, scheduled_event_during_conference, etc.

differentreality: I would also add a scheduled event during the conference, so that we can be testing that we are only deleting the events outside conference hours, and not the other ones.

spec/models/conference_spec.rb
Ana06: oh, sorry. I read it wrong :rofl:

spec/models/conference_spec.rb
Ana06: then there is already a `conference` variable at the top: https://github.com/siddhantbajaj/osem/blob/007804f3ff4d43dd4ac2137d254cc6e1112839c2/spec/models/conference_spec.rb#L7

Although it is call subject, but we should use it here, so we don't have to define it twice. :wink:

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,26 @@
...
+
+    it 'delete event schedules that are not in hour ranges, when conference start hour is updated' do
+      first_scheduled_event
+      conference.start_hour = conference.start_hour + 1.hour
+      conference.save
+      conference.run_callbacks :update
```
Ana06: why do you need this?

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,26 @@
...
+      expect(EventSchedule.count).to eq(0)
```
Ana06: what about using `to_change` instead of to `eq`? Then we know that the EventSchedules where really deleted.

spec/models/conference_spec.rb
```diff
@@ -1639,4 +1639,26 @@
...
+    end
+
+    it 'delete event schedules that are not in hour ranges, when conference end hour is updated' do
+      second_scheduled_event
+      conference.end_hour = conference.end_hour - 2.hours
+      conference.save
+      conference.run_callbacks :update
+      expect(EventSchedule.count).to eq(0)
```
Ana06: same here, use `to_change` instead of to `eq`, then we know that the EventSchedules where really deleted.

spec/models/conference_spec.rb
siddhantbajaj: :+1:

spec/models/conference_spec.rb
siddhantbajaj: :+1:

spec/models/conference_spec.rb
siddhantbajaj: :+1:

spec/models/conference_spec.rb
siddhantbajaj: I was actually following the pattern given here:
https://github.com/siddhantbajaj/osem/blob/007804f3ff4d43dd4ac2137d254cc6e1112839c2/spec/models/conference_spec.rb#L1637
But yes works fine without it as well.

siddhantbajaj: > then there is already a conference variable at the top
But this is a already happened conference so I will have to change the start_date and end_date in the test so that I can successfully create my events. Therefore I didn't use it.

Ana06:
@siddhantbajaj there is still one offense, maybe we can exclude this file from the Block cop, as we are excluding Conference class anyway from the line cop.

siddhantbajaj:
done. :+1:

Issue #1264, opened 2017-02-01
Crowded ApplicationHelper [ Refactoring ]


shlok007:
Currently [ApplicationHelper](https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb) has 58 methods with 500+ lines of code. 
These methods can be divided into separate helper classes for separate views.

gunkiratk:
Hey,
I want to work on this issue

Ana06:
@gunkiratk go ahead! :wink:

gunkiratk:
@Ana06 thanks!
My Rgsoc partner @shaan15 would be working on it!

agrim123:
@shlok007 i m interested in working on this issue

shlok007:
@agrim123 Feel free to send a PR with your changes . :slightly_smiling_face:

PR #1335, opened 2017-03-06
Refactored Application helper into mutiple files

agrim123:
Separated the functions in [application_helper.rb](https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb) into 6 files :  
- users_helper.rb
- paths_helper.rb
- change_description_helper.rb
- format_helper.rb
- date_time_helper.rb
- events_helper.rb

Sensibly grouped the related functions into the above six individual files. 

Addresses #1264 .


ChrisBr:
@agrim123 please rebase!

Though @gunkiratk was already working on it and you and @lagartoflojo agreed in IRC that you work on some other issue?

lagartoflojo:
@agrim123 did you ask @shaan15 and @gunkiratk whether they were already working on this? It makes no sense to step on each other's work.

agrim123:
@lagartoflojo @ChrisBr I pinged on the issue thread itself and it has been around 1 day and I didn't receive any reply. I checked their forks, seen them up to date with upstream. I assumed that they might have not started it yet and so I went forward to make the PR.

Although if they submit a PR on the same issue in future I would be glad to close this.

lagartoflojo:
@agrim123 If I may be blunt, you are being very impatient, if you are only able to wait for a single day for someone to answer your messages. This is open source: people have other things to work on as well, they might not even have access to the internet every day. So let's not assume things just because someone did not reply within 24 hours.
On IRC I thought I asked you to look for another issue to work on, but you still worked on this one.
Please understand that this is a community effort, and it makes no sense to go stepping on each other.
I appreciate your help and your willingness to contribute to this project. Let's just please have more patience and work together.

agrim123:
@lagartoflojo Sorry for my behaviour, I will keep this in mind the next time.

app/helpers/application_helper.rb
```diff
@@ -1,39 +1,11 @@
...
-    result
-  end
-
+  # Including custom made helpers
+  include DateTimeHelper
+  include FormatHelper
+  include EventsHelper
+  include UsersHelper
+  include PathsHelper
+  include ChangeDescriptionHelper
```
lagartoflojo: Rails should load custom helpers automatically.. no need to include them here.

agrim123: @lagartoflojo i think only those helpers are automatically included which have a corresponding controller available and others have to included.
Still I m looking for a better solution.

lagartoflojo: I tested this. I created a CustomHelper with some random function in there, and it was accessible without any special "includes".

agrim123: Views donot require custom helpers to be included in application helper but if we are using a helper function in different controller then we have to include that helper ?

gunkiratk:
Hey @ChrisBr,
We are working on the issue. Since we are beginners, we are taking slightly longer time than expected. Do you still want us to continue working on the issue, now that someone else completed it before us?

app/helpers/application_helper.rb
agrim123: @lagartoflojo Gentle reminder 
How shall I proceed with this PR?

ChrisBr:
@gunkiratk hm, no sending a PR doesn't really makes sense! Let's wait till this one is merged and let's see if something is left.
Otherwise feel free to pick up another junior job!

app/helpers/admin/volunteers_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+	module VolunteersHelper
+		def can_manage_volunteers(conference)
+			if (current_user.has_role? :organizer, conference) || (current_user.has_role? :volunteers_coordinator, conference)
+				true
+			else
+				false
+			end
+		end
+	end
+end
```
lagartoflojo: It seems you are using tabs for indentation.
You should use 2 spaces.

app/helpers/application_helper.rb
```diff
@@ -145,4 +145,26 @@ def concurrent_events(event)
...
+    last = all[-1]
+    ets = ''
+    if all.length > 1
+      ets << first.join(', ')
+      ets << " and #{last}"
+    else
+      ets = all.join
+    end
+    ets
+  end
```
lagartoflojo: We can replace this whole thing with:

```ruby
def event_types(conference)
  conference.program.event_types.map { |et| et.title.pluralize }.to_sentence
end
```

app/helpers/admin/volunteers_helper.rb
lagartoflojo: Also, let's put a question mark at the end of that method name: `can_manage_volunteers?`

app/helpers/application_helper.rb
lagartoflojo: Let's keep working on this PR :)

app/helpers/admin/volunteers_helper.rb
agrim123: Also we can do something like this 
`!!(current_user.has_role? :organizer, conference) || (current_user.has_role? :volunteers_coordinator, conference)`
to `can_manage_volunteers?`
if we are returning boolean

app/helpers/application_helper.rb
agrim123: We can do this but this results in
`['one', 'two', 'three'].to_sentence # => "one, two, and three"` (if we need the comma)
or a better way 
`['one', 'two', 'three'].to_sentence(last_word_connector: ' and ') # => "one, two and three"`
http://apidock.com/rails/Array/to_sentence

app/helpers/admin/volunteers_helper.rb
agrim123: Also we can do something like this 
`!!((current_user.has_role? :organizer, conference) || (current_user.has_role? :volunteers_coordinator, conference))`
to `can_manage_volunteers?`
if we are returning boolean

app/helpers/application_helper.rb
lagartoflojo: I don't see a problem with that additional comma. Let's just use plain `to_sentence` and keep it simple.

agrim123:
@lagartoflojo I have done the requested changes. Please review.

app/helpers/admin/volunteers_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Admin
+  module VolunteersHelper
+    def can_manage_volunteers?(conference)
+      if (current_user.has_role? :organizer, conference) || (current_user.has_role? :volunteers_coordinator, conference)
+        true
+      else
+        false
+      end
```
lagartoflojo: It seems I wrote a comment a few days, but never actually sent it.
We can rewrite this as:
```ruby
def can_manage_volunteers?(conference)
  current_user.has_role?(:organizer, conference) || current_user.has_role?(:volunteers_coordinator, conference)
end
```

agrim123: on it :+1:

app/helpers/change_description_helper.rb
```diff
@@ -0,0 +1,88 @@
+module ChangeDescriptionHelper
```
differentreality: Change description doesn't make a lot of sense when you read it.. This file contains functions used for versions. What about we call it versions_helper instead?

agrim123: sure @differentreality I will change the name of the helper and arrange functions accordingly.

agrim123:
@differentreality I have done the requested changes. Please review.

agrim123:
@lagartoflojo Please review.

differentreality:
Please rebase

differentreality:
@lagartoflojo what do you think?

differentreality:
@agrim123 please rebase

PR #1337, opened 2017-03-06
Added validation in EventSchedule model

siddhantbajaj:
Since #1143 is merged. EventSchedule start time should in hours range of the conference.

### DIFF
--- a/app/models/event_schedule.rb
+++ b/app/models/event_schedule.rb
@@ -10,6 +10,8 @@ class EventSchedule < ActiveRecord::Base
   validates :room, presence: true
   validates :start_time, presence: true
   validates :event, uniqueness: { scope: :schedule }
+  validate :start_after_end_hour
+  validate :start_before_start_hour
 
   scope :confirmed, -> { joins(:event).where('state = ?', 'confirmed') }
   scope :canceled, -> { joins(:event).where('state = ?', 'canceled') }
@@ -36,6 +38,16 @@ class EventSchedule < ActiveRecord::Base
   end
 
   private
+
+  def start_after_end_hour
+    return unless event && start_time && event.program && event.program.conference && event.program.conference.end_hour
+    errors.add(:start_time, "can't be after the conference end hour (#{event.program.conference.end_hour})") if start_time.hour >= event.program.conference.end_hour
+  end
+
+  def start_before_start_hour
+    return unless event && start_time && event.program && event.program.conference && event.program.conference.start_hour
+    errors.add(:start_time, "can't be before the conference start hour (#{event.program.conference.start_hour})") if start_time.hour < event.program.conference.start_hour
+  end
 
   def conference_id
     schedule.program.conference_id
--- a/spec/models/event_schedule_spec.rb
+++ b/spec/models/event_schedule_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper'
 
 describe EventSchedule do
+  let(:conference) { create(:conference) }
 
   describe 'association' do
     it { should belong_to(:schedule) }
@@ -17,5 +18,32 @@ describe EventSchedule do
     it { is_expected.to validate_presence_of(:event) }
     it { is_expected.to validate_presence_of(:room) }
     it { is_expected.to validate_presence_of(:start_time) }
+
+    describe '#start_after_end_hour' do
+      context 'is invalid' do
+        it 'when event schedule start_time is after the conference end_hour, and returns an error message' do
+          new_scheduled_event = build(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours + 1.hour)
+          expect(new_scheduled_event.valid?).to eq false
+          expect(new_scheduled_event.event_schedules.first.errors[:start_time]).to eq ["can't be after the conference end hour (#{conference.end_hour})"]
+        end
+      end
+
+      context 'is valid' do
+        it 'when event schedule start_time is between the conference end_hour and start_hour' do
+          new_scheduled_event = build(:event_scheduled, program: conference.program, hour: conference.start_date + conference.end_hour.hours - 1.hour)
+          expect(new_scheduled_event.valid?).to eq true
+        end
+      end
+    end
+
+    describe '#start_before_start_hour' do
+      context 'is invalid' do
+        it 'when event schedule start_time is before the conference start_hour, and returns an error message' do
+          new_scheduled_event = build(:event_scheduled, program: conference.program, hour: conference.start_date)
+          expect(new_scheduled_event.valid?).to eq false
+          expect(new_scheduled_event.event_schedules.first.errors[:start_time]).to eq ["can't be before the conference start hour (#{conference.start_hour})"]
+        end
+      end
+    end
   end
 end
--- a/spec/controllers/admin/event_schedules_controller_spec.rb
+++ b/spec/controllers/admin/event_schedules_controller_spec.rb
@@ -23,7 +23,7 @@ describe Admin::EventSchedulesController do
                               schedule_id: schedule.id,
                               event_id: create(:event, program: conference.program).id,
                               room_id: create(:room, venue: venue).id,
-                              start_time: conference.start_date)
+                              start_time: conference.start_date + conference.start_hour.hours)
         end
 
         it 'saves the event schedule to the database' do
@@ -66,7 +66,7 @@ describe Admin::EventSchedulesController do
                                 schedule_id: schedule.id,
                                 event_id: create(:event, program: conference.program).id,
                                 room_id: room.id,
-                                start_time: conference.start_date)
+                                start_time: conference.start_date + conference.start_hour.hours)
           event_schedule.reload
         end
 
@@ -75,7 +75,7 @@ describe Admin::EventSchedulesController do
         end
 
         it 'updates the start_time' do
-          expect(event_schedule.start_time).to eq(conference.start_date)
+          expect(event_schedule.start_time).to eq(conference.start_date + conference.start_hour.hours)
         end
 
         it 'has 200 status code' do
--- a/spec/factories/event_schedule.rb
+++ b/spec/factories/event_schedule.rb
@@ -9,7 +9,7 @@ FactoryGirl.define do
         venue = create(:venue, conference: program.conference)
       end
       (event_schedule.room = create(:room, venue: venue)) unless event_schedule.room.present?
-      (event_schedule.start_time = program.conference.start_date.to_time) unless event_schedule.start_time.present?
+      (event_schedule.start_time = program.conference.start_date + program.conference.start_hour.hours) unless event_schedule.start_time.present?
       unless event_schedule.schedule.present?
         unless program.selected_schedule.present?
           schedule = create(:schedule, program: program)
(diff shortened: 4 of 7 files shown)


app/models/event_schedule.rb
```diff
@@ -37,6 +39,16 @@ def replacement?
 
   private
 
+  def start_after_end_hour
+    errors.
+    add(:start_time, "can't be after the conference end hour (#{event.program.conference.end_hour})") if event.program && event.program.conference && event.program.conference.end_hour && start_time && start_time.strftime('%H').to_i > event.program.conference.end_hour
```
ChrisBr: For this complicated conditional I don't like using the conditional at the end of the line... I would go with
```
if ...
...
end
```
Easier to read in my opinion...

app/models/event_schedule.rb
```diff
@@ -37,6 +39,16 @@ def replacement?
...
+  end
+
+  def start_before_start_hour
+    errors.
+    add(:start_time, "can't be before the conference start hour (#{event.program.conference.start_hour})") if event.program && event.program.conference && event.program.conference.start_hour && start_time && start_time.strftime('%H').to_i < event.program.conference.start_hour
```
ChrisBr: Same here!

app/models/event_schedule.rb
lagartoflojo: Don't we have a max-link-length rubocop rule? haha

siddhantbajaj: @lagartoflojo @ChrisBr Yes we do have it. Sorry my bad,I was focusing on the test cases. I have made the changes. :+1:

differentreality:
What about adding tests for new functions you introduced?

differentreality:
And please take care of the failing tests.

siddhantbajaj:
> What about adding tests for new functions you introduced?

Done. :+1:

siddhantbajaj:
> And please take care of the failing tests.

Sorry I didn't rebase.

differentreality:
While these tests are good, we are still missing a case I think.
conf.end_hour = 20:00
event.start_time = 19:00
But event has an event_type with 2h duration.

How shall we address this?

siddhantbajaj:
>How shall we address this?

We can add a validation for this as well but then again it might be putting too much restrictions for the organisers. What do you think? @differentreality

differentreality:
How will this add too many restrictions for organizers? Can you please explain?

siddhantbajaj:
So in our schedules we show all the slots from conference start hour to end hour available. But if we create this validation then organiser won't be allowed to schedule events whose duration exceeds the conference end hour. Even though it is technically right but may be we should allow the organiser to manage this case on their own. But its complete fine with me to add validation for this as well. :+1: 
Screenshot:
 
screenshot from 2017-03-26 20 09 36

app/models/event_schedule.rb
```diff
@@ -37,6 +39,20 @@ def replacement?
 
   private
 
+  def start_after_end_hour
+    if event && event.program && event.program.conference && event.program.conference.end_hour && start_time && start_time.strftime('%H').to_i > event.program.conference.end_hour
```
differentreality: This is long and gets confusing, what about we re-write it as:

return unless event && start_time && event.program && event.program.conference && event.program.conference.end_hour

and then have a single line with the 
errors.add() if start_time.strftime('%H').to_i > event.program.conference.end_hour

differentreality:
OK let's leave this as is.

app/models/event_schedule.rb
siddhantbajaj: done. :+1:

app/models/event_schedule.rb
```diff
@@ -37,6 +39,16 @@ def replacement?
 
   private
 
+  def start_after_end_hour
+    return unless event && start_time && event.program && event.program.conference && event.program.conference.end_hour
```
Ana06: there is a validation that ensure that `event ` and `start_time` are present, so it is not needed to check it here. And the same for `program` in Event. So:

``` Ruby
return unless event.program.conference.try(:end_hour)

File: app/models/event_schedule.rb [hunk 5]
```diff
@@ -37,6 +39,16 @@ def replacement?
...
+    errors.add(:start_time, "can't be after the conference end hour (#{event.program.conference.end_hour})") if start_time.strftime('%H').to_i > event.program.conference.end_hour
```
Ana06: you can use `hour` method :wink: 

``` Ruby
errors.add(:start_time, "can't be after the conference end hour (#{event.program.conference.end_hour})") if start_time.hour > event.program.conference.end_hour

File: app/models/event_schedule.rb [hunk 6]
```diff
@@ -37,6 +39,16 @@ def replacement?
...
+  end
+
+  def start_before_start_hour
+    return unless event && start_time && event.program && event.program.conference && event.program.conference.start_hour
```
Ana06: Same here:

``` Ruby
return unless event.program.conference.try(:start_hour)

File: app/models/event_schedule.rb [hunk 7]
```diff
@@ -37,6 +39,16 @@ def replacement?
...
+    errors.add(:start_time, "can't be before the conference start hour (#{event.program.conference.start_hour})") if start_time.strftime('%H').to_i < event.program.conference.start_hour
```
Ana06: Same here, use `hour` method :wink:

app/models/event_schedule.rb
Ana06: and it should be `>=` as if the conference ends at 19:00, yo don't want any event scheduled at 19:00

spec/models/event_schedule_spec.rb
```diff
@@ -17,5 +17,47 @@
     it { is_expected.to validate_presence_of(:event) }
     it { is_expected.to validate_presence_of(:room) }
     it { is_expected.to validate_presence_of(:start_time) }
+
+    describe '#start_after_end_hour' do
+      context 'is invalid' do
+        it 'when event schedule start_time is after the conference end_hour, and returns an error message' do
+          conference = create(:conference)
```
Ana06: move the variables to a `let`, then you don't need to create then same for every test

spec/models/event_schedule_spec.rb
```diff
@@ -17,5 +17,47 @@
...
+          event = create(:event, program: conference.program)
```
Ana06: create an `event_scheduled` instead an `event` and `event_schedule`: https://github.com/openSUSE/osem/blob/master/spec/factories/events.rb#L31

spec/models/event_schedule_spec.rb
```diff
@@ -17,5 +17,47 @@
...
+          new_event_schedule = build(:event_schedule, event: event, start_time: conference.start_date + conference.end_hour.hours + 1.hour)
+          expect(new_event_schedule.valid?).to eq false
+          expect(new_event_schedule.errors[:start_time]).to eq ["can't be after the conference end hour (#{conference.end_hour})"]
+        end
+      end
+
+      context 'is valid' do
```
Ana06: the two valid contexts can be merged as to be valid it has to be created before the conference end_date and after the start_date, and basically you have exactly the same text twice.

Ana06:
@siddhantbajaj also please rebase! :wink:

app/models/event_schedule.rb
siddhantbajaj: :+1:

app/models/event_schedule.rb
siddhantbajaj: :+1:

app/models/event_schedule.rb
siddhantbajaj: :+1:

app/models/event_schedule.rb
siddhantbajaj: :+1:

spec/models/event_schedule_spec.rb
siddhantbajaj: :+1:

spec/models/event_schedule_spec.rb
siddhantbajaj: :+1:

spec/models/event_schedule_spec.rb
siddhantbajaj: But here I don't want to use default start_time attribute of event_schedule.

siddhantbajaj:
done. :wink:

Ana06:
@siddhantbajaj also please check why are the tests failing :wink:

app/models/event_schedule.rb
```diff
@@ -37,6 +39,16 @@ def replacement?
 
   private
 
+  def start_after_end_hour
+    return unless errors.blank? && event.program.conference.try(:end_hour)
```
Ana06: why did you add `unless errors.blank?`?

siddhantbajaj: We do have validations to check presence of event, start_time and everything else. If event is `nil` `validates :event, presence: true` will fail, but `event.program.conference.try(:end_hour)` will still be executed and it will raise error and because of this our test were failing. So to avoid this `event.program.conference.try(:end_hour) `should be executed only when all our previous validations have passed therefore I added `unless errors.blank?`.

Ana06: ok, to be honest I am not sure if this is better than to check if they are present as you did before. Maybe using `try` for everything like `event.try(:program).try(:conference).try(:end_hour)`.

To be honest I don't have a strong opinion here, so write it with `errors`, `try` or checking if they are present as you had done before, as you think is better. But what about at least moving this to the validation definition? I mean:

`validate :start_after_end_hour if event.try(:program).try(:conference).try(:end_hour)`
`validate :start_before_start_hour if event.try(:program).try(:conference).try(:start_hour)`

Because really when the time is not present you can not validate it. :thinking:

siddhantbajaj: Yes we can do that but still then we would have to check the presence of `start_time` in the validation body so basically we will have if condition in validation definition as well as in validation body so I don't know if that's okay, may be we should have the if condition only in the validation body like earlier. But yes I think `errors` approach is something we should not go with. Actually I myself wasn't convinced with that approach as it requires all previous validations to pass successfully which is not desired.
So I think we should switch back to our previous approach what do you think?

Ana06: @siddhantbajaj 

>  But yes I think errors approach is something we should not go with.

:+1: 

> Yes we can do that but still then we would have to check the presence of start_time in the validation body so basically we will have if condition in validation definition as well as in validation body so I don't know if that's okay, may be we should have the if condition only in the validation body like earlier.

But if you have it in validation definition why do you need it in the body too?

siddhantbajaj: @Ana06 
But we are not checking the presence start_time here.
`validate :start_after_end_hour if event.try(:program).try(:conference).try(:end_hour)`
We can do something like this
`validate :start_after_end_hour if (event.try(:program).try(:conference).try(:end_hour) && start_time)`
But then it will become too long for single line if. Therefore I was thinking to do this in validation body.

Ana06: It is true that it is quite long. ok, fine then, do it as you say :+1:

siddhantbajaj: okay. :wink:

Issue #1340, opened 2017-03-08
Make Splashpage components selected by default


Ana06:
When creating the Splashpage it ask you what components you want to include on it. I wold say that more likely you will select all or most of them, so I would show all of them already selected. Like this:

![image](https://cloud.githubusercontent.com/assets/16052290/23695603/595610a0-03e0-11e7-9288-a0ffad337da9.png)

nikhilgupta1211:
@Ana06 I am on it :)

PR #1341, opened 2017-03-08
Issue #1340 Make Splashpage components selected by default

nikhilgupta1211:
#1340 
Default state of components of splashpage changed to checked .

### DIFF
--- a/app/views/admin/splashpages/_form.html.haml
+++ b/app/views/admin/splashpages/_form.html.haml
@@ -6,15 +6,15 @@
   .col-md-8
     = semantic_form_for(@splashpage, url: admin_conference_splashpage_path(@conference.short_title)) do |f|
       = f.inputs name: 'Components' do
-        = f.input :include_tracks, label: 'Display tracks on the splashpage?'
-        = f.input :include_program, label: 'Display program on the splashpage?'
-        = f.input :include_cfp, label: 'Display call for papers information on splashpage, while cfp is open?'
-        = f.input :include_venue, label: 'Display venue on the splashpage?'
-        = f.input :include_registrations, label: 'Display the registration period on the splashpage?'
-        = f.input :include_tickets, label: 'Display tickets on the splashpage?'
-        = f.input :include_lodgings, label: 'Display the lodgings on the splashpage?'
-        = f.input :include_sponsors, label: 'Display sponsors on the splashpage?'
-        = f.input :include_social_media, label: 'Display social media on the splashpage?'
+        = f.input :include_tracks, label: 'Display tracks', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_tracks) }
+        = f.input :include_program, label: 'Display program', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_program) }
+        = f.input :include_cfp, label: 'Display call for papers information on splashpage, while cfp is open', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_cfp) }
+        = f.input :include_venue, label: 'Display venue', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_venue) }
+        = f.input :include_registrations, label: 'Display the registration period', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_registrations) }
+        = f.input :include_tickets, label: 'Display tickets', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_tickets) }
+        = f.input :include_lodgings, label: 'Display the lodgings', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_lodgings) }
+        = f.input :include_sponsors, label: 'Display sponsors', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_sponsors) }
+        = f.input :include_social_media, label: 'Display social media', input_html: { checked: params[:action] == 'new' || @splashpage.try(:include_social_media) }
       = f.inputs name: 'Access' do
         = f.input :public, label: 'Make splash page public?'
       %p.text-right
--- a/spec/features/versions_spec.rb
+++ b/spec/features/versions_spec.rb
@@ -221,15 +221,21 @@ feature 'Version' do
     click_button 'Save Splashpage'
 
     click_link 'Edit'
-    check('Make splash page public')
-    check('Display tracks on the splashpage?')
-    check('Display the registration period on the splashpage?')
+    uncheck('Display program')
+    uncheck('Display call for papers information on splashpage, while cfp is open')
+    uncheck('Display venue')
+    uncheck('Display tickets')
+    uncheck('Display the lodgings')
+    uncheck('Display sponsors')
+    uncheck('Display social media')
+    check('Make splash page public?')
     click_button 'Save Splashpage'
 
     click_link 'Delete'
     visit admin_revision_history_path
     expect(page).to have_text("#{organizer.name} created new splashpage in conference #{conference.short_title}")
-    expect(page).to have_text("#{organizer.name} updated public, include tracks and include registrations of splashpage in conference #{conference.short_title}")
+    expect(page).to have_text("#{organizer.name} updated public, include program, include cfp, include venue, include tickets, include lodgings,
+      include sponsors and include social media of splashpage in conference #{conference.short_title}")
     expect(page).to have_text("#{organizer.name} deleted splashpage in conference #{conference.short_title}")
   end
 


nikhilgupta1211:
@Ana06 according to the test some components are only selected but as we are making all the components selected by default the test `./spec/features/versions_spec.rb:218` is failing . Should i modify the test ?

Ana06:
@nikhilgupta1211 sure!

[In the tests](https://github.com/openSUSE/osem/blob/master/spec/features/versions_spec.rb#L224) we select 3 options. You should change it to unselect the others so the test keep testing the same thing. :wink:

nikhilgupta1211:
Thanks :) I have implemented the changes .

nikhilgupta1211:
@Ana06  i think there is problem in this that as we are using a form partial for both edit and new action, so when a user create a splashpage with some unchecked options they will appear to be checked in the edit view and the user have to uncheck them again . So should i try to fix this or as there will not be much unchecked options as mentioned in the issue we can use this 😉?

Ana06:
@nikhilgupta1211 

> @Ana06 i think there is problem in this that as we are using a form partial for both edit and new action, so when a user create a splashpage with some unchecked options they will appear to be checked in the edit view and the user have to uncheck them again . So should i try to fix this or as there will not be much unchecked options as mentioned in the issue we can use this :wink:?

It should only by checked by default when creating it, when editing it should show the saved ones.

nikhilgupta1211:
@Ana06 can you please tell me that should i implement it using javascript or ruby?

nikhilgupta1211:
We can add a check-all option to check and uncheck all the checkboxes

Ana06:
@nikhilgupta1211 

> @Ana06 can you please tell me that should i implement it using javascript or ruby?

Ruby!

nikhilgupta1211:
@Ana06 i have made the changes :)

spec/features/versions_spec.rb
```diff
@@ -218,6 +218,15 @@
   scenario 'display changes in splashpages', feature: true, versioning: true, js: true do
     visit admin_conference_splashpage_path(conference.short_title)
     click_link 'Create Splashpage'
+    uncheck('Display tracks on the splashpage?')
```
sayalilunkad: I think it would look better if we leave the 'on the splashpage?' bit. It seems too repetitive. @Ana06 What do you think?

nikhilgupta1211:
I have made the changes

Ana06:
@nikhilgupta1211 please read the [commits](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#commits) and [pull request](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#pull-requests-workflow) sections that we recently added to our [contributing guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md). :wink:

nikhilgupta1211:
@Ana06 should i squash the commits ?

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,11 @@
 module ApplicationHelper
+  # Set the value of checkbox to checked for the splashpages#new action
+  def chk
```
differentreality: What does chk mean ?!

nikhilgupta1211: sorry I forgot to change the name of the method

nikhilgupta1211:
I have also added the Venue and Lodgings to the list 
image

app/views/admin/splashpages/_form.html.haml
```diff
@@ -6,15 +6,15 @@
...
-        = f.input :include_tracks, label: 'Display tracks on the splashpage?'
-        = f.input :include_program, label: 'Display program on the splashpage?'
-        = f.input :include_cfp, label: 'Display call for papers information on splashpage, while cfp is open?'
-        = f.input :include_venue, label: 'Display venue on the splashpage?'
-        = f.input :include_registrations, label: 'Display the registration period on the splashpage?'
-        = f.input :include_tickets, label: 'Display tickets on the splashpage?'
-        = f.input :include_lodgings, label: 'Display the lodgings on the splashpage?'
-        = f.input :include_sponsors, label: 'Display sponsors on the splashpage?'
-        = f.input :include_social_media, label: 'Display social media on the splashpage?'
+        = f.input :include_tracks, label: 'Display tracks', input_html: { checked: checked }
```
differentreality: What if we just have
`input_html: { params[:action] == 'new' ? 'checked' : false }`

nikhilgupta1211: done :)

differentreality:
:+1: Please squash and rebase!

nikhilgupta1211:
Done :)

app/views/admin/splashpages/_form.html.haml
```diff
@@ -6,15 +6,15 @@
...
-        = f.input :include_tracks, label: 'Display tracks on the splashpage?'
-        = f.input :include_program, label: 'Display program on the splashpage?'
-        = f.input :include_cfp, label: 'Display call for papers information on splashpage, while cfp is open?'
-        = f.input :include_venue, label: 'Display venue on the splashpage?'
-        = f.input :include_registrations, label: 'Display the registration period on the splashpage?'
-        = f.input :include_tickets, label: 'Display tickets on the splashpage?'
-        = f.input :include_lodgings, label: 'Display the lodgings on the splashpage?'
-        = f.input :include_sponsors, label: 'Display sponsors on the splashpage?'
-        = f.input :include_social_media, label: 'Display social media on the splashpage?'
+        = f.input :include_tracks, label: 'Display tracks', input_html: { checked: params[:action] == 'new' ? 'checked' : false }
```
Ana06: You can not set it to false, you have to take the value from conference splashpage:

`@conference.splashpage.try(:include_tracks)` returns `false` or `true` if it is set, or `nil` in case that it is not

app/views/admin/splashpages/show.html.haml
```diff
@@ -60,6 +60,20 @@
           - else
             No
         %dt
+          Include Venue:
```
Ana06: Those are things that where missed in the show page, right?

They should be in a different commit as they are unrelated to making Splashpage components selected by default :wink:

spec/features/versions_spec.rb
```diff
@@ -218,12 +218,21 @@
   scenario 'display changes in splashpages', feature: true, versioning: true, js: true do
     visit admin_conference_splashpage_path(conference.short_title)
     click_link 'Create Splashpage'
+    uncheck('Display tracks')
```
Ana06: instead of unchecking everything when creating it, what about leaving by default here and after clicking the edit link unchecking everything except _Make splash page public_, _Display tracks on the splashpage?_ and _Display the registration period on the splashpage?_

app/views/admin/splashpages/show.html.haml
nikhilgupta1211: Yes, ok :)

spec/features/versions_spec.rb
nikhilgupta1211: @Ana06 that does not pass the test I think for that we have to also update the expect(page).to

Ana06: why?

nikhilgupta1211: Currently when the elements were not selected by default we were updating the _Make splash page public, Display tracks on the splashpage? and Display the registration period on the splashpage?_  but now as all the components are selected by default so by unchecking them in the edit link we are updating all of them except for the _tracks and registration component_ so we have to change the version_spec.rb:237 expect(page).to

nikhilgupta1211:
I have made the changes in the tests :)

app/views/admin/splashpages/_form.html.haml
```diff
@@ -6,15 +6,15 @@
...
-        = f.input :include_tracks, label: 'Display tracks on the splashpage?'
-        = f.input :include_program, label: 'Display program on the splashpage?'
-        = f.input :include_cfp, label: 'Display call for papers information on splashpage, while cfp is open?'
-        = f.input :include_venue, label: 'Display venue on the splashpage?'
-        = f.input :include_registrations, label: 'Display the registration period on the splashpage?'
-        = f.input :include_tickets, label: 'Display tickets on the splashpage?'
-        = f.input :include_lodgings, label: 'Display the lodgings on the splashpage?'
-        = f.input :include_sponsors, label: 'Display sponsors on the splashpage?'
-        = f.input :include_social_media, label: 'Display social media on the splashpage?'
+        = f.input :include_tracks, label: 'Display tracks', input_html: { checked: params[:action] == 'new' ? 'checked' : @conference.splashpage.try(:include_tracks) }
```
Ana06: This you tried this? I would say it doesn't work :confounded: 

If `checked` attribute is present it will be checked even if it is set to false (`checked="false"`), so the value doesn't mind)

Ana06: why about writing this like:

``` Ruby
f.input :include_tracks, label: 'Display tracks', input_html: { checked: params[:action] == 'new' || @conference.splashpage.try(:include_tracks) }
```

As `true` and ''checked` both work :wink:

Ana06: And the same for the rest

nikhilgupta1211: @Ana06 I guess we don't need to do `@conference.splashpage.try(:include_tracks)` we can do `@splashpage.try(:include_tracks)`

nikhilgupta1211:
I have made the changes :)

PR #1344, opened 2017-03-08
Added validations in  Registration model

siddhantbajaj:
I think registration should not created after registration period end date. Therefore I have added a validation in registration model for the same.

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -27,6 +27,7 @@ class Registration < ActiveRecord::Base
   validates_uniqueness_of :user_id, scope: :conference_id, message: 'already Registered!'
   validate :registration_limit_not_exceed, on: :create
   validate :registration_to_events_only_if_present
+  validate :before_end_of_registration_period, on: :create
 
   after_create :set_week, :subscribe_to_conference, :send_registration_mail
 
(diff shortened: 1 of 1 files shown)


differentreality:
What about a conference without cfp when organizers will simply submit talks?

siddhantbajaj:
@differentreality Yes you are right it, It won't be applicable then. I have removed that validation.

siddhantbajaj:
@sayalilunkad I have made the changes. :+1:

differentreality:
I have some doubts whether or not we should have this. Technically what you wrote is accurate. However an admin/organizer etc are authorized to perform certain actions, including :new and :create for Registration.
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L147

If we add this, our codebase and our ability.rb file will not match in what they allow and what not. In my opinion we should not add this.

siddhantbajaj:
@differentreality Yes maybe you are right but aren't we already doing this in our app like organiser/admin can also manage events.
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L159
But still we have this validation:
https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L31

differentreality:
Yes, but we have an open PR to introduce the feature of events created by admin

differentreality:
Part of what this PR is trying to accomplish is covered with the tests of #new action in https://github.com/openSUSE/osem/blob/master/spec/controllers/conference_registration_controller_spec.rb

I suggest we add tests for #create action.

PR #1346, opened 2017-03-09
Fixes #1081, Add more profile info on user Profile

### LINKED ISSUE #1081: Add more info on user profile

BelieveC:
Fixes #1081 
Adds info for following social media and social code

- Twitter
- Linkedin
- Github
- Gna
- Gnu
- Savannah
- Google plus
- Diaspora
- Website url

screenshot from 2017-03-09 14-46-17

Working on Test for same.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -8,15 +8,25 @@ class UsersController < ApplicationController
 
   # GET /users/1/edit
   def edit
+    @tab_select = params[:tab]
+    if @tab_select.present?
+      respond_to do |format|
+        format.js { render action: 'field_select' }
+      end
+    else
+      render :edit
+    end
   end
 
   # PATCH/PUT /users/1
   def update
-    if @user.update(user_params)
-      redirect_to @user, notice: 'User was successfully updated.'
-    else
-      flash[:error] = "An error prohibited your Profile from being saved: #{@user.errors.full_messages.join('. ')}."
-      render :edit
+    respond_to do |format|
+      if @user.update(user_params)
+        flash[:notice] = 'User was successfully updated.'
+      else
+        flash[:error] = "An error prohibited your Profile from being saved: #{@user.errors.full_messages.join('. ')}."
+      end
+      format.js
     end
   end
 
@@ -24,6 +34,6 @@ class UsersController < ApplicationController
 
     # Only allow a trusted parameter "white list" through.
     def user_params
-      params.require(:user).permit(:name, :biography, :nickname, :affiliation)
+      params.require(:user).permit(:name, :biography, :nickname, :affiliation, :googleplus, :twitter, :gna, :gnu, :github, :gitlab, :website_url, :linkedin, :diaspora, :savannah)
     end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,6 +63,9 @@ class User < ActiveRecord::Base
 
   validate :biography_limit
 
+  validates :website_url, :linkedin, :googleplus, :gnu, :twitter, :gitlab, :github, :gna, :savannah, :diaspora,
+            format: { with: URI.regexp(%w(http https)), message: 'This is an invalid URL. It should start with http or https'}, allow_blank: true
+
   ##
   # Checkes if the user attended the event
   # This is used for events that require registration
--- a/app/views/users/_personal_info_fields.html.haml
+++ b/app/views/users/_personal_info_fields.html.haml
@@ -1,0 +1,18 @@
+= semantic_form_for(user, url: user_path(user.id), remote: true) do |f|
+	= f.input :name, as: :string
+	= f.input :nickname, as: :string
+	.control-label
+	= "Avatar"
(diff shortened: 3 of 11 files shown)


app/controllers/users_controller.rb
```diff
@@ -8,22 +8,32 @@ def show
...
-    else
-      flash[:error] = "An error prohibited your Profile from being saved: #{@user.errors.full_messages.join('. ')}."
-      render :edit
+    respond_to do |format|
+      if @user.update(user_params)
+        flash[:notice] = 'User was successfully updated.'
+      else
+        flash[:error] = "An error prohibited your Profile from being saved: #{@user.errors.full_messages.join('. ')}."
+      end
+      format.js
```
ChrisBr: Why do you not render the erros in js?

ChrisBr:
I don't really see the point doing everything with AJAX except making the code more hard to understand IMHO

app/controllers/users_controller.rb
```diff
@@ -8,22 +8,32 @@ def show
 
   # GET /users/1/edit
   def edit
+    @tab_select = params[:tab]
+    if @tab_select.present?
```
ChrisBr: Please don't do that with AJAX. Just use the bootstrap tabs to hide and show the different partials.

app/controllers/users_controller.rb
ChrisBr: Why is the format.js necessary at all as you didn't change the form to use AJAX?

ChrisBr:
Please also provide screenshots of the tabs and also of the user show page.

lagartoflojo:
Do we really need all of this?
I mean, who uses all these random social media sites?
Why not just add a single field, "Social media profile link", and as help text for that single field, put the Twitter and Facebook examples.
Maybe include "personal website link" as well.
But anything else is complete overkill in my opinion.

👎

app/controllers/users_controller.rb
BelieveC: @ChrisBr Initially even I did the same. I used bootstrap tab to render different partial for sake of simplicity. But the problem that I faced with Bootstrap tabs are as follows:

- It just add `.active` class to a current tab. i.e  Hides the partials of inactive tab
- When I tried submitting the form with this approach. Inactive tab fields acted as hidden fields to the form and also got submitted. 
This is the issue which @differentreality was talking about in previous PR #1180 
>Perhaps. The user, however, will be saving data that are not visible (ie data on the other tab will be saved as well).
 
To solve above issues, Option I had were as follows:
- Implement separate form in a separate tab to detach it from rest of the inactive tab fields. I really didn't like this approach
- Use Ajax to add and remove necessary and unnecessary fields respectively. 

Looking forward to suggestion.

app/controllers/users_controller.rb
BelieveC: > Why is the format.js necessary at all as you didn't change the form to use AJAX?
 
I have used it to render different form partial as discusses in above comments.

BelieveC:
Screenshots
- Personal Tab
screenshot from 2017-03-17 09-33-00
- Social Media Tab
screenshot from 2017-03-17 09-33-05
- Social Code Tab
screenshot from 2017-03-17 09-33-08
User show page:
screenshot from 2017-03-17 09-34-36

differentreality:
I am not against giving options (for social media to be added), but I would like to limit the chaos a bit. Does it make sense to put a dropdown with available options of social media? User selects from dropdown, enters url in an input box next to the drop down. 
We only show what the user has.

lagartoflojo:
What would be the drop-down's function?

lagartoflojo:
Also, what's the difference between "social media" and "social code"? Maybe I'm not "social" enough to understand 😝

differentreality:
drop down -> List of social media -> User chooses one -> enters profile link (in the input box) -> then save I guess.

My understanding is that social code is social media where you share your code, eg. github.

BelieveC:
@lagartoflojo 
> Do we really need all of this?
I mean, who uses all these random social media sites?

I agree with your opinion. Actually, these social media sites were already considered in the #1180. This pull request was indeed to implement tab as suggested by @ChrisBr 

> Maybe a second tab with "Contact Information" / "Social Media"?

BelieveC:
@differentreality  and @lagartoflojo 

I think we should consider following social sites:

- Twitter
- Github
- Linkedin( Maybe)

Personally, I would also like to add

- Website Url

What do you think?

lagartoflojo:
I'm not sure then that the drop-down is useful for much other than making people click more.
Why not just a add "social media link" and "personal website link" fields, that's it? Most people have one favorite social media site, and if anything they can link to another one from the "personal website link".
What's the use case of doing anything more?

BelieveC:
@differentreality @lagartoflojo 

I agree with @lagartoflojo 

> just add "social media link" and "personal website link" fields

Finally what should I do?

differentreality:
Closing due to inactivity. We should probably discuss on the issue what it is that we want to do exactly (which fields to add, how to show them, etc)

PR #1348, opened 2017-03-09
Fixed registration limit issue

### LINKED ISSUE #1347: Issue with registration limit
When we create a conference its registration limit is by default set as 0 (which means no registration limit).
Now suppose 2 users register for the conference so no of registrations for the conference becomes 2. If we now try to update the registration limit of the conference to 1, conference will get successfully updated which I think should not happen.

siddhantbajaj:
closes #1347

### DIFF
--- a/app/views/admin/conferences/edit.html.haml
+++ b/app/views/admin/conferences/edit.html.haml
@@ -24,5 +24,5 @@
         = f.input :start_hour, :input_html => {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, :input_html => {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: "Registrations" do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit). You currently have " + pluralize(@conference.registrations.count,'registration')
       = f.action :submit, :as => :button, :button_html => {:class => "btn btn-primary"}


siddhantbajaj:
@differentreality  I have made the changes. :+1:

app/views/admin/conferences/edit.html.haml
```diff
@@ -24,5 +24,5 @@
         = f.input :start_hour, :input_html => {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, :input_html => {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: "Registrations" do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit). You already have #{@conference.registrations.count} registrations."
```
differentreality: I'd rather say 'you currently have...' instead of you 'you already have' because it sounds right even if it is 0 registrations, and we are not checking if there are actually any registrations. 

Also if there is 1 registration, it should read 'you have 1 registration', not '1 registrations'.

siddhantbajaj:
@differentreality I have made the changes. :+1:

app/views/admin/conferences/edit.html.haml
```diff
@@ -24,5 +24,8 @@
         = f.input :start_hour, :input_html => {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, :input_html => {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: "Registrations" do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+        - if @conference.registrations.count == 1
```
differentreality: What about http://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize ?

differentreality:
Please push new commits to your PR, instead of making changes to the existing commit!

app/views/admin/conferences/edit.html.haml
siddhantbajaj: done. :+1:

app/views/admin/conferences/edit.html.haml
```diff
@@ -24,5 +24,5 @@
         = f.input :start_hour, :input_html => {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, :input_html => {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: "Registrations" do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit)"
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: "Limit the number of registrations to the conference (0 no limit). You currently have " + pluralize(@conference.registrations.count,'registration','registrations')
```
differentreality: You don't need to add the word in plural. Pluralize can show the proper plural word on its own. You only have to add both singular and plural words, when you want to use something different, eg. 1 person, 2 users.

differentreality:
Nice work :-) 
Please squash your commits, and rebase for merging.

siddhantbajaj:
@differentreality Thank you for helping me out. I have made the changes. :+1:

differentreality:
:+1:

PR #1352, opened 2017-03-10
Fixes #1343, Allow viewing public schedule without login

### LINKED ISSUE #1343: Log in is needed to see the schedule
It is not allowed to see the schedule without log in.

BelieveC:
Fixes #1343 

Allow viewing public schedule without login.

schedule

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -1,7 +1,8 @@
 class SchedulesController < ApplicationController
+  load_and_authorize_resource
   protect_from_forgery with: :null_session
   before_action :respond_to_options
-  load_and_authorize_resource :conference, find_by: :short_title
+  load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
 
   def show
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -62,6 +62,10 @@ class Ability
       can [:new, :create], Event do |event|
         event.program.cfp_open? && event.new_record?
       end
+
+      can [:show, :events], Schedule do |schedule|
+        schedule.program.schedule_public
+      end
     end
   end
 


app/controllers/schedules_controller.rb
```diff
@@ -1,8 +1,12 @@
 class SchedulesController < ApplicationController
   protect_from_forgery with: :null_session
   before_action :respond_to_options
-  load_and_authorize_resource :conference, find_by: :short_title
+  skip_authorization_check
```
Ana06: you shouldn't skip the authorisation and handle it here. You should do it in the ability file, see: https://github.com/BelieveC/osem/blob/edb11d195dffac16597da6bf1f53d5678d208d46/app/models/ability.rb#L37

BelieveC: Thanks! Will make necessary changes :+1:

app/models/ability.rb
```diff
@@ -34,8 +34,8 @@ def not_signed_in
       conference.splashpage && conference.splashpage.public == true
     end
     # Can view the schedule
-    can [:schedule, :events], Conference do |conference|
-      conference.program.cfp && conference.program.schedule_public
+    can [:show, :schedule, :events], Conference do |conference|
```
differentreality: How is this related to the issue? 
:show gives you access to the show action of the Conference. How is this relevant with the schedule?

BelieveC: `schedules_controller ` load_authorize conference object for use in schedules show action. To make schedule show action accessible to user who is not logged in, conference show action is added to can in ability.rb

Ana06: previously `schedule` was a method in `conferences_controller`, but now it is a different controller

BelieveC:
@Ana06 Any more suggestion and changes?

app/models/ability.rb
differentreality: @Ana06 you don't think we should be fixing how we authorize, now that we have a different model for Schedule?

Ana06: yes, my comment was just the reason why you are right, this is unrelated.

BelieveC:
Alright, I will make the changes :+1:

differentreality:
@BelieveC ping, what's the status of this? Do you need help with anything?

BelieveC:
@differentreality I have made the changes. Sorry for the delay. I'm having exams in my college. So was not able to dig deeper.

app/controllers/schedules_controller.rb
```diff
@@ -1,7 +1,8 @@
 class SchedulesController < ApplicationController
+  load_and_authorize_resource
   protect_from_forgery with: :null_session
   before_action :respond_to_options
-  load_and_authorize_resource :conference, find_by: :short_title
```
Ana06: why did you changed this?

BelieveC: Following are the reasons to change
`load_and_authorize_resource :conference, find_by: :short_title` to `load_resource :conference, find_by: :short_title`

- Keeping `load_and_authorize_resource :conference, find_by: :short_title` doesn't allow user to access schedule without login because we don't have can ability for show action of conference for `not_signed_in` user 
- In order to allow unsigned user to view schedule which indirectly requires conference, I have removed `authorize_resource` part and added authorization for `schedules_controller` to make up for it.

BelieveC: @Ana06, Do you want any changes?

Ana06: no, it is ok! :wink:

PR #1353, opened 2017-03-10
Improve room creation using ajax

BelieveC:
Creating multiple rooms is painful for following reasons

- Create action redirect to index page every time you create a new room.
- Need to again browse to new room page to create room.

Improvement

- Create multiple rooms on index page itself without refreshing index page.
- An Increase in work efficiency.

screenshot from 2017-03-10 20-47-09

**Updated Screenshot**
- Room's Index page

screenshot from 2017-03-14 13-25-24

- Room's New Page

screenshot from 2017-03-14 13-25-39

@differentreality, Screenshots Updated

### DIFF
--- a/app/controllers/admin/rooms_controller.rb
+++ b/app/controllers/admin/rooms_controller.rb
@@ -3,6 +3,7 @@ module Admin
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :venue, through: :conference, singleton: true
     load_and_authorize_resource through: :venue
+    after_action :prepare_unobtrusive_flash, only: [:create]
 
     def index; end
 
@@ -14,12 +15,13 @@ module Admin
 
     def create
       @room = @venue.rooms.new(room_params)
-      if @room.save
-        redirect_to admin_conference_venue_rooms_path(conference_id: @conference.short_title),
-                    notice: 'Room successfully created.'
-      else
-        flash.now[:error] = "Creating Room failed: #{@room.errors.full_messages.join('. ')}."
-        render :new
+      respond_to do |format|
+        if @room.save
+          flash.now[:notice] = 'Room successfully created.'
+        else
+          flash.now[:error] = "Creating Room failed: #{@room.errors.full_messages.join('. ')}."
+        end
+        format.js
       end
     end
 
--- a/app/views/admin/rooms/_form.html.haml
+++ b/app/views/admin/rooms/_form.html.haml
@@ -8,8 +8,8 @@
         = @room.name
 .row
   .col-md-8
-    = semantic_form_for(@room, url: (@room.new_record? ? admin_conference_venue_rooms_path : admin_conference_venue_room_path(@conference.short_title, @room))) do |f|
-      = f.input :name, input_html: { autofocus: true}
-      = f.input :size, input_html: {size: 5}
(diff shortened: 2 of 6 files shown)


differentreality:
While I am not against the idea, I disagree with the implementation and the existence of a form in the index page.

BelieveC:
Okay, I will try something else :+1:

BelieveC:
@differentreality, Form is removed from the index page. Now user can add multiple rooms from view of new action itself.

differentreality:
Please update the screenshot in the first comment.

app/controllers/admin/rooms_controller.rb
```diff
@@ -3,6 +3,7 @@ class RoomsController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource :venue, through: :conference, singleton: true
     load_and_authorize_resource through: :venue
+    after_action :prepare_unobtrusive_flash
```
ChrisBr: where does this filter come from?

app/controllers/admin/rooms_controller.rb
```diff
@@ -14,12 +15,13 @@ def new
...
-      else
-        flash[:error] = "Creating Room failed: #{@room.errors.full_messages.join('. ')}."
-        render :new
+      respond_to do |format|
+        if @room.save
+          flash[:notice] = 'Room successfully created.'
+        else
+          flash[:error] = "Creating Room failed: #{@room.errors.full_messages.join('. ')}."
+        end
+        format.js{ render 'message' }
```
ChrisBr: Why message?

app/controllers/admin/rooms_controller.rb
BelieveC: Filter is used from `gem 'unobtrusive_flash'` documentation. It is used to create a flash message after an ajax call. 
flash_message

app/controllers/admin/rooms_controller.rb
BelieveC: Previously, I used it to render flash message without unobtrusive flash. I will remove message partial and explicit render from controller. :+1:

differentreality:
I don't feel this is adding much to our app.

BelieveC:
> I don't feel this is adding much to our app.

@differentreality I would ask to consider following things it add to our app

- Saves a lot of time when you are creating a large number of rooms at once.
- Increase work efficiency.

If you are still not happy with implementation, I could add [modal](http://getbootstrap.com/javascript/#modals-grid-systeml) on index page to add new room instead of new page.

differentreality:
I wonder if it makes sense to use cocoon and have dynamic associations to create multiple rooms in the new form.

BelieveC:
@differentreality 
> I wonder if it makes sense to use cocoon and have dynamic associations to create multiple rooms in the new form.

I don't think it would make sense to use cocoon to and have dynamic associations to create multiple rooms in the new form of venue due to following reasons:

- Venue form is already pretty long.
- Room have two attributes i.e name and size. It would create both fields each time we add new dynamic association. I think single attribute would be great for a dynamic association in a nested form. 
- Think of a venue with 100 different rooms. Then while editing either any room or venue all hundred rooms will be there in the venue form. It would end up being too large.

BelieveC:
@differentreality Any changes or suggestions?

differentreality:
@BelieveC while we appreciate the effort and idea, this PR has been ongoing for too long and we don't seem to have decided how it would make sense to implement it. At the same time it is not super urgent to get this in.

As a result, I will be closing it at this point. Please don't let this discourage you for your other PRs and future contribution!

PR #1359, opened 2017-03-12
Fix #1324 remove view tests

### LINKED ISSUE #1324: Remove view tests
The test views/conferences/show.html.haml_spec.rb fails 2 out of 10 times .

nikhilgupta1211:
@Ana06 for now i have removed the view tests which were failing most of the time. Should  i also remove the other view tests too ?


Ana06:
@nikhilgupta1211 

> @Ana06 for now i have removed the view tests which were failing most of the time. Should i also remove the other view tests too ?

yes! I think it is better no to test view, and we really don't have many view tests.

nikhilgupta1211:
@Ana06 i have removed all the view tests.

PR #1369, opened 2017-03-14
Add some tips to installation guide

siaw23:
For some juniors, having Ruby 2.4 will cause a lot of compatibility errors involving `json` and a couple of other gems. I added information just in case someone runs into such errors. To avoid wasting time "fixing" what doesn't need fixing.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,8 +1,22 @@
 # Install Open Source Event Manager
-All the information that you need to install OSEM. If you have any problems with installing don't hesitate to [contact us](https://github.com/openSUSE/osem#contact)
+This file contains all the information you need to install OSEM. If you run into any problems, do not hesitate to [contact us](https://github.com/openSUSE/osem#contact).
+
+If you encounter
+```
+An error occurred while installing yajl-ruby (1.2.0), and Bundler cannot continue.
+Make sure that `gem install yajl-ruby -v '1.2.0'` succeeds before bundling.
+```
+or
+
+```
+An error occurred while installing json (1.8.3), and Bundler cannot continue.
+Make sure that `gem install json -v '1.8.3'` succeeds before bundling.
+```
+
+check that your Ruby version is not greater than `ruby-2.3.3`.
 
 ## Versions
-OSEM is an [semantic versioned](http://semver.org/) app. That means given a version number MAJOR.MINOR.PATCH we increment the:
+OSEM is a [semantic versioned](http://semver.org/) app. That means given a version number MAJOR.MINOR.PATCH we increment the:
 
 1. MAJOR version when we make incompatible changes,
 2. MINOR version when we add functionality in a backwards-compatible manner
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -12,13 +26,13 @@ OSEM is an [semantic versioned](http://semver.org/) app. That means given a vers
 You can find the latest OSEM releases on our [release page](https://github.com/openSUSE/osem/releases/latest) ([older release here](https://github.com/openSUSE/osem/releases))
 
 ## Deploy
-OSEM is a *Ruby on Rails* application. We recommend to run OSEM in production with [mod_passenger](https://www.phusionpassenger.com/download/#open_source)
+OSEM is a *Ruby on Rails* application. We recommend you run OSEM in production with [mod_passenger](https://www.phusionpassenger.com/download/#open_source)
 and the [apache web-server](https://www.apache.org/). There are tons of guides on how to deploy rails apps on various
 base operating systems. [Check Google](https://encrypted.google.com/search?hl=en&q=ruby%20on%20rails%20apache%20passenger) ;-)
 
-For more information about rails and what it can do, see the [rails guides.](http://guides.rubyonrails.org/getting_started.html)
+For more information about Fails and what it can do, see the [Fails Guides.](http://guides.rubyonrails.org/getting_started.html)
```
differentreality: I assume that's from an auto spell checker or something?

INSTALL.md
```diff
@@ -1,8 +1,22 @@
 # Install Open Source Event Manager
-All the information that you need to install OSEM. If you have any problems with installing don't hesitate to [contact us](https://github.com/openSUSE/osem#contact)
+This file contains all the information you need to install OSEM. If you run into any problems, do not hesitate to [contact us](https://github.com/openSUSE/osem#contact).
+
+If you encounter
+```
+An error occurred while installing yajl-ruby (1.2.0), and Bundler cannot continue.
+Make sure that `gem install yajl-ruby -v '1.2.0'` succeeds before bundling.
```
Ana06: This is the standard error that is shown when `bundle install` fails, right? And it could happen with any other gem too.

INSTALL.md
```diff
@@ -1,8 +1,22 @@
...
+```
+or
+
+```
+An error occurred while installing json (1.8.3), and Bundler cannot continue.
+Make sure that `gem install json -v '1.8.3'` succeeds before bundling.
```
Ana06: Same here

INSTALL.md
```diff
@@ -1,8 +1,22 @@
...
+```
+
+check that your Ruby version is not greater than `ruby-2.3.3`.
```
differentreality: We require Ruby 2.1.2, right @hennevogel @ChrisBr ?
But still I don't think we should list all possible/known issues in our installation guide. Perhaps the README and/or INSTALLATION file(s) should be mentioning the ruby and the rails versions.

INSTALL.md
```diff
@@ -12,13 +26,13 @@ OSEM is an [semantic versioned](http://semver.org/) app. That means given a vers
 You can find the latest OSEM releases on our [release page](https://github.com/openSUSE/osem/releases/latest) ([older release here](https://github.com/openSUSE/osem/releases))
 
 ## Deploy
-OSEM is a *Ruby on Rails* application. We recommend to run OSEM in production with [mod_passenger](https://www.phusionpassenger.com/download/#open_source)
+OSEM is a *Ruby on Rails* application. We recommend you run OSEM in production with [mod_passenger](https://www.phusionpassenger.com/download/#open_source)
```
Ana06: You can't not remove the *to*:

_We recommend you *to* run_

INSTALL.md
siaw23: Oh right. My editor is not tech savvy 😂

INSTALL.md
siaw23: yes of course. in that case i setting the "required" ruby version in the Gemfile will be a smarter approach. yes? so new users can get started faster easier without running into errors.

INSTALL.md
siaw23: yes. when bundle install fail because of compatibility issues between ruby and the json gem.

Ana06: but then it will show this error, so I don't see the point in including it here. :thinking:

INSTALL.md
Ana06: I guess is 2.2.3, as that is what we are using for Travis: https://github.com/openSUSE/osem/blob/61930f8572d2239943c5bf135734cd7df0017c54/.travis.yml#L6

hennevogel: we are ruby version agnostic. we just require what gems require.

INSTALL.md
siaw23: well it happens for just 2 gems.

hennevogel: On your system. On any other system this can happen with any other gem. This is why it makes not so much sense to add this here...

siaw23: @Ana06 @hennevogel got it :). do we close this PR then?

hennevogel: @siaw23 nope just make it about the type fixes :-)

PR #1370, opened 2017-03-14
Move proposal edit to admin

alator21:
Changed proposal links of edit path to redirect in admin namespace from reports page.


ChrisBr:
Yes please squash your commits!

PR #1371, opened 2017-03-15
Increase test coverage for user model. Fix small bug.

amiedes:
This PR does the following things:

* Wrote specs for `User#attended_event?`
* Wrote specs for `User#registered_to_event?`
* Fix exception being thrown in `User#registered_to_event?` when the user
is not registered in the corresponding event conference.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -87,7 +87,7 @@ class User < ActiveRecord::Base
   # ====Returns
   # * +true+ or +false+
   def registered_to_event? event
-    event.registrations.pluck(:id).include? self.registrations.find_by(conference_id: event.program.conference.id).id
+    event.registrations.include? registrations.find_by(conference: event.program.conference)
   end
 
   def subscribed? conference
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -14,6 +14,8 @@ describe User do
   let(:event1) { create(:event, program: conference.program) }
   let(:another_conference) { create(:conference) }
   let(:event2) { create(:event, program: another_conference.program) }
+  let(:registration) { create(:registration, user: user, conference: conference) }
+  let(:events_registration) { create(:events_registration, event: event1, registration: registration) }
 
   describe 'validation' do
     it 'has a valid factory' do
@@ -87,10 +89,56 @@ describe User do
   end
 
   describe 'methods' do
+    describe '#attended_event?' do
+      context 'user has attended to the event' do
+        before do
+          events_registration.update_attributes(attended: true)
+        end
+
+        it 'returns true' do
+          expect(user.attended_event?(event1)).to be true
+        end
+      end
+
+      context 'user did not register for the event' do
+        it 'returns false' do
+          expect(user.attended_event?(event1)).to be false
+        end
+      end
+
+      context 'user registered for the event, but did not attend' do
+        before do
+          events_registration
+        end
+
+        it 'returns false' do
+          expect(user.attended_event?(event1)).to be false
+        end
+      end
+    end
+
     describe '#name' do
       it 'returns the username as name if there is not name' do
         user = create(:user, name: nil)
         expect(user.name).to eq(user.username)
+      end
+    end
+
+    describe '#registered_to_event?' do
+      context 'user has registered to event' do
+        before do
+          events_registration
+        end
+
+        it 'returns true' do
+          expect(user.registered_to_event?(event1)).to be true
+        end
+      end
+
+      context 'user has not registered to event' do
+        it 'returns false' do
+          expect(user.registered_to_event?(event1)).to be false
+        end
       end
     end
 


Ana06:
@amiedes when working on two unrelated things (writing tests and fixing a bug in this case) you should open two different PR as it make much easier to review it, or at least several commits in the PR.

So, please separate for changes in two commits. :wink:

It would also be helpful if you add a description to your commits, for example for the bug one:

```
Fix bug in User#registered_to_event?

Fix exception being thrown in `User#registered_to_event?` when the user is not registered in the corresponding event conference.
```

spec/models/user_spec.rb
```diff
@@ -87,13 +87,62 @@
...
       it 'returns the username as name if there is not name' do
         user = create(:user, name: nil)
         expect(user.name).to eq(user.username)
       end
     end
 
+    describe '#registered_to_event?' do
+      context 'user has registered to event' do
+        before do
+          registration = create(:registration, user: user, conference: conference)
```
Ana06: do this in a let, then it is created when needed, can be reused and then you have all the object created for your tests together and at the beginning. 

And as you are using it for both `#attended_event?` and `#registered_to_event?` you can move it up (just after `describe 'methods' do`) so you only need to define it once. :slightly_smiling_face:

spec/models/user_spec.rb
```diff
@@ -87,13 +87,62 @@
...
+          create(:events_registration, event: event1, registration: registration)
```
Ana06: same here, move it to a let!

spec/models/user_spec.rb
```diff
@@ -87,13 +87,62 @@
   end
 
   describe 'methods' do
+    describe '#attended_event?' do
+      context 'user has attended to the event' do
+        before do
+          registration = create(:registration, user: user, conference: conference)
```
Ana06: do this in a let, then it is created when needed, can be reused and then you have all the object created for your tests together and at the beginning. 

And as you are using it for both `#attended_event?` and `#registered_to_event?` you can move it up (just after `describe 'methods' do`) so you only need to define it once. :slightly_smiling_face:

spec/models/user_spec.rb
```diff
@@ -87,13 +87,62 @@
...
+          create(:events_registration, event: event1, registration: registration, attended: true)
```
Ana06: Same here, although this one needs to be called in the before, to create it. So:

````Ruby
 describe '#attended_event?' do
  let(:registration) { :create(:registration, user: user, conference: conference) }
  let(:events_registration) { create(:events_registration, event: event1, registration: registration, attended: true) }
  
  context 'user has attended to the event' do
    before do
      events_registration
    end

File: spec/models/user_spec.rb [hunk 5]
```diff
@@ -87,13 +87,62 @@
...
+
+      context 'user did not register for the event' do
+        it 'returns false' do
+          expect(user.attended_event?(event1)).to be false
+        end
+      end
+
+      context 'user registered for the event, but did not attend' do
+        before do
+          registration = create(:registration, user: user, conference: conference)
```
Ana06: same here

spec/models/user_spec.rb
```diff
@@ -87,13 +87,62 @@
...
+          create(:events_registration, event: event1, registration: registration)
```
Ana06: same here

app/models/user.rb
```diff
@@ -87,7 +87,8 @@ def name
   # ====Returns
   # * +true+ or +false+
   def registered_to_event? event
-    event.registrations.pluck(:id).include? self.registrations.find_by(conference_id: event.program.conference.id).id
+    conference_registration = self.registrations.find_by(conference_id: event.program.conference.id)
```
amiedes: I did not thought about that 😮. It sure looks much better that way! Thanks 🙂

amiedes:
Hey @Ana06, thanks a lot for your review!

I moved the `let(:registration) { create(:registration, user: user, conference: conference) }` after the `describe '#attended_event?' do` and `describe '#registered_to_event?' do` lines because, even thought this means this line it's duplicated, since the only two methods that need the creation of a `:registration` are these I thought it was cleared to keep this declaration there than right under the `describe 'methods' do`.

If you think it's better to keep it right under `describe 'methods'` I can move it there, no problem 😉

About the case of `create(:events_registration)`, what do you think about using `let!(:events_registration)` instead of `before { events_registration }`?

Thanks a lot for your toughts! 🙂

Ana06:
@amiedes 
> I moved the `let(:registration) { create(:registration, user: user, conference: conference) }` after the `describe '#attended_event?' do` and `describe '#registered_to_event?' do` lines because, even thought this means this line it's duplicated, since the only two methods that need the creation of a `:registration` are these I thought it was clearer to keep this declaration there than right under the `describe 'methods' do`.

One of the principles of Rails is DRY (Don't Repeat Yourself), so there should not be a repeated line. You can move it after `describe 'methods'` or with the other lets at the beginning of file (https://github.com/openSUSE/osem/pull/1371/files#diff-12b107c16792b9ecba685e51b51826f1R17) if you think this is clearer. :wink: 

> About the case of `create(:events_registration)`, what do you think about using `let!(:events_registration)` instead of `before { events_registration }`?

The problem of using `let!` is that you have to repeat the line. :confounded:

amiedes:
Hmmm... I just noticed `create(:events_registration, ...)` isn't really duplicated since sometimes it has `attended: true` and others it doesn't.

BTW, do you want me to squash the latest little commits?

Ana06:
@amiedes 

> Hmmm... I just noticed `create(:events_registration, ...)` isn't really duplicated since sometimes it has `attended: true` and others it doesn't.

True, but one of them is twice. And you can always set the attendance attribute in the before. So you have the let only once :wink:

Ana06:
@amiedes 

> BTW, do you want me to squash the latest little commits? 

Normally, the commits are squashed when the PR is ready to merge, but I am also fine if you want to squash them already, as it is almost ready. :slightly_smiling_face:

amiedes:
@Ana06 Done!

spec/models/user_spec.rb
```diff
@@ -87,13 +89,53 @@
   end
 
   describe 'methods' do
+    describe '#attended_event?' do
+      context 'user has attended to the event' do
+        before { events_registration.update_attributes(attended: true) }
```
Ana06: I would prefer writing this with `do end` instead `{ }`, as it is how it appears in the rspec documentation

spec/models/user_spec.rb
```diff
@@ -87,13 +89,53 @@
...
+      end
+
+      context 'user did not register for the event' do
+        it 'returns false' do
+          expect(user.attended_event?(event1)).to be false
+        end
+      end
+
+      context 'user registered for the event, but did not attend' do
+        before { events_registration }
```
Ana06: same here

spec/models/user_spec.rb
```diff
@@ -87,13 +89,53 @@
...
     describe '#name' do
       it 'returns the username as name if there is not name' do
         user = create(:user, name: nil)
         expect(user.name).to eq(user.username)
       end
     end
 
+    describe '#registered_to_event?' do
+      context 'user has registered to event' do
+        before { events_registration.update_attributes(attended: true) }
```
Ana06: same here

Ana06: why do you need to update the attribute here?

spec/models/user_spec.rb
amiedes: @Ana06 But the preferred Ruby syntax for one-liner blocks is with `{ }` 😕 . And this same syntax is used several times across the file in lines I didn't write.

spec/models/user_spec.rb
amiedes: That was a mistake. I just fixed it.

spec/models/user_spec.rb
Ana06: I know, but if you look at RSpec documentation it is written the other way around: https://www.relishapp.com/rspec/rspec-core/v/2-2/docs/hooks/before-and-after-hooks So when someone read the code you wrote he will probably go to the documentation, so following it would make much easier for him to understand your code.


And you also find places where it is done as I said: https://github.com/amiedes/osem/blob/83f132f84c4063091283066afc5bc00353006237/spec/models/user_spec.rb#L262

amiedes: @Ana06 Ok I think it's ready now. I'll make my best to stick to project conventions for the following commits 🙃

Ana06:
@amiedes  There are merge conflicts, please rebase to master, so we can merge it

PR #1372, opened 2017-03-16
Fixes #1230. Makes comment counter more efficient

### LINKED ISSUE #1230: Making comment counter more efficient

siaw23:
Reduce calls to `COUNT(*)` by 2. Need a review on this then I'll add tests.

### DIFF
--- a/0
+++ b/0
@@ -1,0 +1,2 @@
+      invoke  active_record
+      create    db/migrate/20170316042351_add_comment_count_to_events.rb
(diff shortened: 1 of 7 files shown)


differentreality:
Please follow our contributing guide https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md

It seems your PR contains unrelated changes, and that it is doing the very same changes as in https://github.com/openSUSE/osem/pull/1235

Please choose an issue that has not already been implemented!

PR #1375, opened 2017-03-16
Fix jQuery version

TheAssassin:
Draggable.js is not compatible with jQuery >= 3 which means that several UI elements are broken, but jquery-rails ships with jQuery >= 3 by default now. This patch makes OSEM use jQuery 2 again, so that Draggable.js works again.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -10,7 +10,7 @@
 // WARNING: THE FIRST BLANK LINE MARKS THE END OF WHAT'S TO BE PROCESSED, ANY BLANK LINE SHOULD
 // GO AFTER THE REQUIRES BELOW.
 //
-//= require jquery
+//= require jquery2
 //= require jquery_ujs
 //= require jquery.mobile.custom.min
 //= require jquery.ui.draggable


differentreality:
You don't seem to have included all the necessary changes for this. Please update the PR.

TheAssassin:
See https://travis-ci.org/openSUSE/osem/builds/211680051#L501. This PR is not failing because I have not included necessary changes, but because Travis uses a very old (cached?) version of the gem, which results in my fix being incompatible to the old version.

Travis build log:
```Using rails-assets-jquery 2.2.1```

Docker container build log:
```Installing rails-assets-jquery 3.2.1```

New approach: rather than fixing it in a way that works for that newer version of the gem, I would suggest pinning that older version in the Gemfile. I'll create a new commit that does that instead, and will hopefully pass the Travis checks.

General tip: version pinning is really important for (web) projects like this. It can be somewhat generalized (in this case pinning to 2.0.0 <= version < 3.0.0 would probably work fine). But it has to be done to avoid such version conflicts. Some dev should take the time to check the dependency versions that work fine (Travis log should help with that) and pin them accordingly.

TheAssassin:
Oops...

I just have seen that it is kinda pinned in the `Gemfile.lock` and I forgot to add this to my Dockerfile before I call bundler. But somehow I wonder why that `Gemfile.lock` is in the repository at all, back in the days I learned that this was a local file only, which had been created on my system for the test installation. I've put all my version pinning in the Gemfile itself in some smaller projects I created a few years ago.

PR #1384, opened 2017-03-16
Fixed venue link in todo list for organizer

### LINKED ISSUE #1383: "Add venue" link in todo list does not work properly for organiser.
"Add venue" link in todo list is not active for the organiser of that conference. Organiser of a conference can update/create the venue. So it should it be active in the todo list for organiser as well.
Screenshot of above issue:
screenshot from 2017-03-17 02 45 50

siddhantbajaj:
closes #1383

### DIFF
--- a/app/views/admin/conferences/_todo_list.html.haml
+++ b/app/views/admin/conferences/_todo_list.html.haml
@@ -21,14 +21,16 @@
       = link_to 'Set up call for papers', admin_conference_program_cfp_path(conference_progress['short_title'])
     - else
       Set up call for papers
-  %li{ 'class' => "list-group-item #{class_for_todo(conference_progress['venue'])}" }
-    %span{ 'class' => icon_for_todo(conference_progress['venue']) }
-    - if can? :update, @conference.venue
+  %li{'class'=>"list-group-item #{class_for_todo(conference_progress['venue'])}"}
+    %span{'class'=>icon_for_todo(conference_progress['venue'])}
+    - if can? :update, Venue.new(conference: @conference)
+      - @conference.reload
       - if conference.venue
         = link_to 'Add venue', edit_admin_conference_venue_path(conference_progress['short_title'])
       - else
         = link_to 'Add venue', new_admin_conference_venue_path(conference_progress['short_title'])
     - else
+      - @conference.reload
       Add venue
   %li{ 'class' => "list-group-item #{class_for_todo(conference_progress['rooms'])}" }
     %span{ 'class' => icon_for_todo(conference_progress['rooms']) }
--- a/spec/features/ability_spec.rb
+++ b/spec/features/ability_spec.rb
@@ -7,14 +7,16 @@ feature 'Has correct abilities' do
   let(:conference3) { create(:full_conference) } # user is info_desk
   let(:conference4) { create(:full_conference) } # user is volunteer coordinator
   let(:conference5) { create(:full_conference) } # user has no role
+  let(:conference6) { create(:conference) } # user is organizer, venue is not set by default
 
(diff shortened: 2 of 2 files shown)


lagartoflojo:
Can you add a test for the previously failing case, and now should be passing?

siddhantbajaj:
@lagartoflojo Added test. :+1:

app/views/admin/conferences/_todo_list.html.haml
```diff
@@ -23,7 +23,7 @@
       Set up call for papers
   %li{'class'=>"list-group-item #{class_for_todo(conference_progress['venue'])}"}
     %span{'class'=>icon_for_todo(conference_progress['venue'])}
-    - if can? :update, @conference.venue
+    - if can? :update, @conference
```
differentreality: Technically this should be checking if the user can create or update a venue. 
If we do that (due to the create check where we will initialize a venue object associated with conference), we would need to reload the conference, so that it shows the right link. While we have a venue object, the link will redirect to edit, even though that object is not saved, and does not contain the data we need.

Ana06:
@siddhantbajaj please read the [commits](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#commits) and [pull request](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md#pull-requests-workflow) sections that we recently added to our [contributing guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md). :wink:

app/views/admin/conferences/_todo_list.html.haml
siddhantbajaj: Yes @differentreality do you suggest something like this:
`-if can? :update, @conference.build_venue`
and then reloading the conference?
I tried this approach first, but the problem I faced with this approach is suppose conference already has a venue then 
`-if can? :update, @conference.build_venue` 
this will delete the venue stored in database and create a new temporary venue for this conference.

spec/features/ability_spec.rb
```diff
@@ -7,14 +7,16 @@
   let(:conference3) { create(:full_conference) } # user is info_desk
   let(:conference4) { create(:full_conference) } # user is volunteer coordinator
   let(:conference5) { create(:full_conference) } # user has no role
+  let(:conference6) { create(:conference) } # user is organizer, venue is not set by default
 
-  let(:role_organizer) { Role.find_by(name: 'organizer', resource: conference1) }
+  let(:role_organizer_first) { Role.find_by(name: 'organizer', resource: conference1) }
```
differentreality: role_organizer_conf1 and role_organizer_conf6 seems to make more sense, no?

app/views/admin/conferences/_todo_list.html.haml
differentreality: What if you do Venue.new(conference: @conference) instead?

spec/features/ability_spec.rb
siddhantbajaj: Yes, I have made the changes. :+1:

app/views/admin/conferences/_todo_list.html.haml
siddhantbajaj: Thank you it worked. :+1:

differentreality:
@siddhantbajaj please squash and rebase!

siddhantbajaj:
@differentreality :+1:

differentreality:
Nice, please squash again the commits and rebase. 
It is best if your commit title says what this change is doing, not starting with the issue reference, because that doesn't provide any substantial information to the reader of the commit. You can add the the issue you are solving at the end of the commit title or even better at the end of the commit body.

siddhantbajaj:
done :+1:

differentreality:
:+1:

PR #1389, opened 2017-03-17
Update commit section in CONTRIBUTING.md

ChrisBr:
to improve the commit messages of future contributors and follow best practices.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,19 +85,22 @@ You can access the app [localhost:3000](http://localhost:3000). Whatever you cha
   * If not, please first leave a comment on how you want to implement it and wait for contributors' feedback
 
 ### Commits
-* Commit title should be short and descriptive
-  * title (or summary line) is the first line of the commit message
-  * that says what the commit is doing
-  * in no more than 50 characters
-  * starting with a word like 'Fix' or 'Add' or 'Change'
-  * **without** a period (.) at the end
-  * followed by a blank line
-* Commit messages are
-  * up to 72 characters
-  * with break lines
-* Reference the issue(s) the commit closes
-  * https://help.github.com/articles/closing-issues-via-commit-messages
-  * If you haven't done so since the beginning, you should reference the issue when you squash your commits
+* The commit title should be short and descriptive, and it should follow these rules:
+  * It should be the first line of the commit message
+  * It explains what the commit does in the imperative mood (Fix foobar)
+  * It's no longer than 50 characters
+  * It starts with a word like 'Fix' or 'Add' or 'Change'
+  * It does not end with a period (.)
+  * It is followed by a blank line
+  * It should contain more than only the reference to the issue
+
+Read this [article](https://chris.beams.io/posts/git-commit/) for more information
+* The commit message should follow these rules:
+  * It should explain why the change is made
+  * It should not describe how the change is made, code is generally self-explanatory
+  * It's not longer than 72 characters
+  * It should contain line breaks
+  * It should contain a [reference](https://help.github.com/articles/closing-issues-via-commit-messages) to the issue 
 
 ### Pull Requests workflow
 Please open a pull request (PR) only when you have finished coding, and your changes are ready to be reviewed for merging.


lagartoflojo:
The text is really kind of weird if you read the first word of each point:

```markdown
* Commit title should be short and descriptive
  * Title is...
  * That says...
  * In no more...
  * Starting with...
  * **without** a...
  * Followed by...
  * Read this...
```
I find it quite hard to follow.

Instead, try to make the first word always the same type of word (verb, noun or pronoun, pick one).

```markdown
* The commit title should be short and descriptive, and it should follow these rules:
  * It should be the first line of the commit message
  * It explains what the commit does in the imperative mood (Fix foobar)
  * It's no longer than 50 characters
  * It starts with a word like 'Fix' or 'Add' or 'Change'
  * It does not end with a period (.)
  * It is followed by a blank line

Read this [article](https://chris.beams.io/posts/git-commit/) for more information
```

AEtherC0r3:
At the end of the commit section it says:
> * Reference the issue(s) the commit closes
>   * https://help.github.com/articles/closing-issues-via-commit-messages
>   * If you haven't done so since the beginning, you should reference the issue when you squash your commits

Referencing the issue is also requested in the pr workflow section:
>  * Reference the issue(s) you are solving

Maybe the commit section should be changed to read "You should reference the issue **only** when you squash your commits", in order to avoid unnecessarily spamming  the issue's thread

CONTRIBUTING.md
```diff
@@ -86,13 +86,16 @@ You can access the app [localhost:3000](http://localhost:3000). Whatever you cha
...
 ### Commits
 * Commit title should be short and descriptive
-  * title (or summary line) is the first line of the commit message
-  * that says what the commit is doing
-  * in no more than 50 characters
-  * starting with a word like 'Fix' or 'Add' or 'Change'
+  * Title (or summary line) is the first line of the commit message
+  * That says what the commit is doing in imperative mood (Fix foobar)
+  * In no more than 50 characters
+  * Starting with a word like 'Fix' or 'Add' or 'Change'
```
Ana06: Can we also add _Enable_ here? So:

_Starting with a word like 'Fix' or 'Add', 'Change' or 'Enable'_

ChrisBr:
@lagartoflojo @Ana06 good points, updated!

CONTRIBUTING.md
```diff
@@ -85,19 +85,22 @@ You can access the app [localhost:3000](http://localhost:3000). Whatever you cha
...
+  * It starts with a word like 'Fix' or 'Add' or 'Change'
+  * It does not end with a period (.)
+  * It is followed by a blank line
+  * It should contain more than only the reference to the issue
+
+Read this [article](https://chris.beams.io/posts/git-commit/) for more information
+* The commit message should follow these rules:
+  * It should explain why the change is made
+  * It should not describe how the change is made, code is generally self-explanatory
+  * It's not longer than 72 characters
```
lagartoflojo: This rule is badly written. According to that link referenced above, the rule is "6. Wrap the body at 72 characters".
That means that each individual line should not be longer than 72 chars, but the message sure can be longer!

CONTRIBUTING.md
```diff
@@ -85,19 +85,22 @@ You can access the app [localhost:3000](http://localhost:3000). Whatever you cha
...
+  * It should contain line breaks
```
lagartoflojo: Kind of a weird requirement to "force" line breaks in the message...

CONTRIBUTING.md
ChrisBr: Yeah, you're right!

CONTRIBUTING.md
ChrisBr: hm! Suggest dropping?

CONTRIBUTING.md
```diff
@@ -85,19 +85,22 @@ You can access the app [localhost:3000](http://localhost:3000). Whatever you cha
...
-  * https://help.github.com/articles/closing-issues-via-commit-messages
-  * If you haven't done so since the beginning, you should reference the issue when you squash your commits
+* The commit title should be short and descriptive, and it should follow these rules:
+  * It should be the first line of the commit message
+  * It explains what the commit does in the imperative mood (Fix foobar)
+  * It's no longer than 50 characters
+  * It starts with a word like 'Fix' or 'Add' or 'Change'
+  * It does not end with a period (.)
+  * It is followed by a blank line
+  * It should contain more than only the reference to the issue
```
Ana06: Do we want the reference to the issue in the message? I would say:

It shouldn't contain the reference to the issue

CONTRIBUTING.md
lagartoflojo: Maybe I just don't understand the rule... Care to explain why it's necessary?

CONTRIBUTING.md
differentreality: I think in the future, it may be helpful to reference the issue. In the description though, not the title

CONTRIBUTING.md
differentreality: So that the 1st case does not occur, and we end up with commit messages like the 2nd commit.

commit 21052b2639e471628b0db52affd61198a306890a
Author: Stella Rouzi <differentreality@gmail.com>
Date:   Tue Mar 21 11:49:55 2017 +0200

    This is a very very big commit message that I am writing here to test something I need, because I want to test what happens with a very very long commit message, and if I have to scroll horizontally in order to view it, or not. and if I do, I want to include this into our styling guide

commit 50f097811c88f3220701e958446b53a96a7ce9c5
Author: Stella Rouzi <differentreality@gmail.com>
Date:   Tue Jun 28 18:16:06 2016 +0300

    Add surveys; custom generated by admins
    
    Available during registration or after conference



screenshot from 2017-03-21 11-51-36

lagartoflojo: The first case is already solved by "wrap the body at 72 characters" thought.

CONTRIBUTING.md
Ana06: yes, I agreem but it is already written in the description section

differentreality: Ah right, I think we should drop this line all together, or re-write it:
Do not reference the issue in the commit title

CONTRIBUTING.md
differentreality: True ;)

lagartoflojo: Ok, let's drop it in favor of "wrap at 72 chars" :)

hennevogel:
You really want to make it extra hard for new contributors? :-)

ChrisBr:
@hennevogel we want to make it extra easy ;)

differentreality:
So how about we include the new sections (commits & PR guidelines, label meaning) in our wiki and keep the CONTRIBUTING.md file shorter with a reference to https://github.com/openSUSE/osem/wiki#contributing ?

lagartoflojo:
What would be the purpose of moving it away to the wiki? I mean, contributors would then need to read two pages instead of just one.
If our `contributing.md` file is too long, then. maybe we can think about making it shorter?

differentreality:
My thinking is that it will become shorter, by having less information (basically stay as short as it was). It will make it easier for new contributors by having less information. Then there will be more - and more detailed - information on the wiki. For example, we don't need to have the label explanation in the contributing guide (primary source of info), but it is useful to have it, for both new contributors, as well as maintainers, and it can be found in the wiki.

Issue #1388, opened 2017-03-17
Write test for program#any_event_for_this_date?(date)


Ana06:
Write test/s for program#any_event_for_this_date?(date) as they were skipped in 
https://github.com/openSUSE/osem/pull/1309

nikhilgupta1211:
on it :)

namangupta01:
@nikhilgupta1211 are you working on it?

PR #1391, opened 2017-03-18
Implements tests for program#any_event_for_this_date?(date)

GabrielSVinha:
Add tests for [#1388](https://github.com/openSUSE/osem/issues/1388) .


differentreality:
Hey @GabrielSVinha ,

just a few tips that will help us all: 

You should be referencing the issues in the description of the PR (ie first comment), not in the title, so that the issues get a reference (now they don't).

Instead the title of the PR should give a bit more information about what this PR is about, ie add tests for X Y Z thing. That will helps a lot when reviewing!

In the same concept the commit message 'Add tests for # 1388', doesn't really say much about what this commit is about. Of course you can change that later, when the PR is ready to be merged.

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
     end
   end
 
+  describe '#any_event_for_this_date?' do
+    it 'returns type should be boolean' do
```
differentreality: Please re-write this into a sentence that makes sense.

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
...
+      expect(program.any_event_for_this_date?('').in?([true, false])).to eq(true)
```
differentreality: Please remove redundant parenthesis.

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
...
+    end
+
+    it 'returns todays events if date is nil' do
```
differentreality: **today's** events**,** if...

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
...
+      date = DateTime.parse(' 00:00').utc
```
differentreality: You are testing that the function is returning events for **today**, then the date needs to be today, and it can go directly into the next statement.

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
...
+      events = EventSchedule.where(schedule: program.selected_schedule).where(start_time: date..(date + 1))
+      expect(program.any_event_for_this_date?('')).to eq(events.any?)
```
differentreality: You need to test if the function is returning true, when there are events, and if it returns false, when there aren't any events. Unless you make sure that there are events, or there are not any events, you can't test this.

spec/models/program_spec.rb
```diff
@@ -199,4 +199,21 @@
...
+    end
+
+    it 'returns todays events if date is empty' do
+      date = DateTime.parse(' 00:00').utc
+      events = EventSchedule.where(schedule: program.selected_schedule).where(start_time: date)
+      expect(program.any_event_for_this_date?('')).to eq(events.any?)
+    end
+  end
 end
```
differentreality: You need to test what happens when 
* date is nil
* date is a valid date

and when 
* there is no selected_schedule
* when there are event schedule, but not event starts at desired time
* when there are event schedules, but there is a desired time

GabrielSVinha:
@differentreality I changed the title and description, on my way to fix all your requests (sorry bout the english errors, i'm not a native speaker)

spec/models/program_spec.rb
GabrielSVinha: This fails rubocup, should I do time anyway?

differentreality: I don't know what it is that makes rubocop fail. You need to show me the code and the failure.

spec/models/program_spec.rb
differentreality: How about writing it simply as 'returns boolean'

differentreality:
@GabrielSVinha do you need more help with this?

GabrielSVinha:
@differentreality I am implementing all the other tests. Today I will commit it.

spec/models/program_spec.rb
```diff
@@ -199,4 +201,46 @@
...
+      expect(event_program.any_event_for_this_date? '' ).to eq true
+    end
+
+    it "returns true if the date passed is valid and there are events in it"do
+      event = create(:event_schedule, start_time: Time.utc(2017, "jan", 01))
+      schedule.event_schedules = [event]
+      expect(event_program.any_event_for_this_date? "2017-01-01").to eq true
+    end
+
+    it "returns false if there is no selected_schedule for the event on the date"do
```
GabrielSVinha: I believe it is the best approach

spec/models/program_spec.rb
```diff
@@ -4,6 +4,8 @@
   subject { create(:program) }
   let!(:conference) { create(:conference, end_date: Date.today + 3) }
   let!(:program) { conference.program }
+  let!(:event_program) { create(:program, selected_schedule: schedule)}
```
GabrielSVinha: We don't really need an event_program just for testing the events. I will use program

spec/factories/event_schedule.rb
```diff
@@ -2,6 +2,7 @@
 
 FactoryGirl.define do
   factory :event_schedule do
+    start_time DateTime.parse("#{Date.today.to_s} 00:00").utc
```
Ana06: You can also write it like that, more readable in my opinion:

`DateTime.now.utc.beginning_of_day`

But this is wrong, conference start_hour is 9, so you can not schedule it at 00.

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
     end
   end
 
+  describe '#any_event_for_this_date?' do
+    context 'without selected_schedule' do
+      it 'returns false if there are events on the date' do
+        event = create(:event_schedule, start_time: Date.today.to_time)
```
Ana06: better add this to a `let`.

start_time is expected to be UTC, and Date.today.to_time is not.

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+        program.selected_schedule.event_schedules = [event]
```
Ana06: there is not selected schedule, so you can't do that. That's why the test is failling

spec/factories/event_schedule.rb
GabrielSVinha: IS there any other way to set the hour? I know that that DateTime.parse(string) works, it would be like this:

`DateTime.parse("#{Date.today} 10:00").utc`

spec/models/program_spec.rb
GabrielSVinha: Sure.

spec/models/program_spec.rb
GabrielSVinha: Yeah, I noticed it. It was removed and next commit it will no longer be there.

spec/factories/event_schedule.rb
Ana06: The problem with parse is that after you change it to utc, so the time would also change, and won't be 10:00. And it could happens that it is outside of the conference time too

spec/models/program_spec.rb
```diff
@@ -4,6 +4,7 @@
   subject { create(:program) }
   let!(:conference) { create(:conference, end_date: Date.today + 3) }
   let!(:program) { conference.program }
+  let!(:schedule) { create(:schedule)}
```
Ana06: why do you need the `!` here?

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+    end
+
+    context 'with selected schedule' do
+      before :each do
+        program.selected_schedule = schedule
+        program.save!
+      end
+
+      it 'returns boolean' do
+        expect(program.any_event_for_this_date?('').in? [true, false] ).to eq true
```
Ana06: `program.any_event_for_this_date?('')` will return false, with this you are not testing that the value is `true` or `false`.

I think you should remove this test case. :slightly_smiling_face:

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+      end
+
+      it 'returns false if there are not events today and date is nil' do
+        expect(program.any_event_for_this_date? nil).to eq false
+      end
+
+      it 'returns true if there are events today and date is nil' do
+        event = create(:event_schedule)
```
Ana06: this won't be needed when to move `event` to a `let`

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+        program.selected_schedule.event_schedules = [event]
+        program.save!
```
Ana06: I would prefer this two lines to be in a `before` section and leave in the test only the expect. :slightly_smiling_face:

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+      before :each do
+        program.selected_schedule = schedule
+        program.save!
+      end
+
+      it 'returns boolean' do
+        expect(program.any_event_for_this_date?('').in? [true, false] ).to eq true
+      end
+
+      it 'returns false if there are not events today and date is nil' do
```
Ana06: does it makes any difference if there are events of not? :wink:

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+      end
+
+      it 'returns true if there are events today and date is nil' do
+        event = create(:event_schedule)
+        program.selected_schedule.event_schedules = [event]
+        program.save!
+        expect(program.any_event_for_this_date? nil).to eq true
+      end
+
+      it 'returns false if there are not events today and date is empty' do
```
Ana06: it doesn't make any difference if there are events or not when the date is empty/nil. I would say that it is enough to test it when date is empty and there is an event for that time which is in the selected schedule. Although it is enough with having in the title 'returns false if date is empty'

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+
+      it 'returns true if there are events today if date passed is empty' do
+        event = create(:event_schedule, schedule: schedule)
+        program.selected_schedule.event_schedules = [event]
+        program.save!
+        expect(program.any_event_for_this_date? '').to eq true
+      end
+
+      it 'returns true, when there are events in supplied date' do
+        event = create(:event_schedule, start_time: Date.today.to_time)
```
Ana06: this won't be needed when moving this to a let

spec/models/program_spec.rb
```diff
@@ -199,4 +200,61 @@
...
+        program.selected_schedule.event_schedules = [event]
```
Ana06: you could also add the event to the selected schedule when creating it.

differentreality:
What's the progress on this @GabrielSVinha ? Do you need any help finishing it?

GabrielSVinha:
@differentreality I have been caught in hard weeks on my college, lots of tests, etc. I will focus on  it now.

differentreality:
@GabrielSVinha would you be able to finish this in the next few days?

spec/models/program_spec.rb
GabrielSVinha: Not really nedded, removing it.

spec/factories/event_schedule.rb
GabrielSVinha: But, without the utc it returns the same time (`+0000`)

GabrielSVinha:
Working on it right now @differentreality

spec/models/program_spec.rb
GabrielSVinha: OK

differentreality:
Oops I missed this being updated :( Sorry! 
It seems that there was a comment form @Ana06 about `Date.parse()` which was not really resolved? @Ana06 ?

spec/factories/event_schedule.rb
Ana06: ok, fine

bear454:
@GabrielSVinha get those tests passing ;-)

Ana06:
@GabrielSVinha do you need help to finish this? Please if so just ask! :wink:

differentreality:
Closing due to lack of response. https://github.com/openSUSE/osem/pull/1799

PR #1392, opened 2017-03-19
Adds tests for Answer Model

BelieveC:
Adds 
- model spec file `spec/models/answer_spec.rb` 
-  validation tests
- association tests


differentreality:
Please don't put time in anything related to volunteers (we are not currently using it, and it needs to be refactored on its whole) or questions (it is being refactored)

BelieveC:
@differentreality Alright!

PR #1393, opened 2017-03-20
Fixes #1188 show the conference timezone in show.xml.haml

### LINKED ISSUE #1188: Wrong timezone information in schedule.xml export

nikhilgupta1211:
#1188 
Removed the trailing Z from event.time and added a timezone tag in the xml file

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,6 +34,11 @@ module ApplicationHelper
     result
   end
 
+  # Returns time with conference timezone
+  def time_with_timezone(time)
+    time.strftime('%F %R') + ' ' + @conference.timezone.to_s
+  end
+
   ##
   # Checks if the voting has already started, or if it has already ended
   #
--- a/app/views/admin/events/_proposal.html.haml
+++ b/app/views/admin/events/_proposal.html.haml
@@ -117,7 +117,7 @@
           %td
             %b Scheduled time
           %td
-            = @event.time
+            = time_with_timezone(@event.time)
       %tr
         %td
           %b Submitter
--- a/app/views/schedules/show.xml.haml
+++ b/app/views/schedules/show.xml.haml
@@ -16,7 +16,7 @@
           %room{ name: room.name }
             - events_in_rooms[room].each do |event|
               %event{ guid: event.guid, id: event.id }
-                %date= event.time.iso8601
+                %date= time_with_timezone(event.time)
                 %start= event.time.strftime('%H:%M')
                 %duration= length_timestamp(event.event_type.length)
                 %room= event.room.name


app/views/schedules/show.xml.haml
```diff
@@ -3,6 +3,7 @@
   %conference
     %acronym= @conference.short_title
     %title= @conference.title
+    %timezone= @conference.timezone
```
differentreality: Don't remove this please.

nikhilgupta1211: sorry i didn't understand you

app/views/schedules/show.xml.haml
```diff
@@ -16,7 +17,7 @@
           %room{ name: room.name }
             - events_in_rooms[room].each do |event|
               %event{ guid: event.guid, id: event.id }
-                %date= event.time.iso8601
+                %date= event.time.iso8601.tr('Z', '')
```
differentreality: I wondering if the time should include the timezone (conference's timezone)

nikhilgupta1211: @differentreality like this https://github.com/openSUSE/osem/issues/1188#issuecomment-287754919 ?

differentreality: Yeap

nikhilgupta1211:
Done :)

differentreality:
@Ana06 what do you think?

app/helpers/application_helper.rb
```diff
@@ -34,6 +34,13 @@ def date_string(start_date, end_date)
     result
   end
 
+  # Returns the time with the conference's timezone
+  def set_offset(time, timezone)
```
Ana06: you don't need to calculate offsets. I think it is quite difficult to understand.

You can just do, getting the time in UTC and adding the conference timezone:

``` Ruby
time.strftime("%F %R") + " " + timezone.to_s
```

This would get something like:

`2017-03-27 09:30 (GMT-10:00) Hawaii`

And if you like that it render the name of the time zone (I personally like it), yo can also use the `parse` method, basically remove the time zone, which is always UTC, and get the time time in this zone. 

``` Ruby
timezone.parse(time.strftime("%F %R"))
```

In this case you'll get something like:

`Mon, 27 Mar 2017 09:30:00 HST -10:00`


@nikhilgupta1211 @differentreality what do you think? I like the first alternative I give much more than calculate offset and the given string is also more readable in my opinion.

app/helpers/application_helper.rb
```diff
@@ -34,6 +34,13 @@ def date_string(start_date, end_date)
...
+    time = time.in_time_zone(timezone)
+    offset = time.utc_offset
+    time + -offset
```
Ana06: adding a negative value is the same as subtracting the value itself!!!!

`time - offset`

app/helpers/application_helper.rb
nikhilgupta1211: @Ana06  We can use the parse method but as we will be converting the time into the ISO 8601 format displaying the timezone name isn't required like in the first method.

Ana06: why is it compulsory to convert it into the ISO 8601 format?

nikhilgupta1211: I think it is not compulsory to use it, So should I use the first approach or the second ?

nikhilgupta1211:
@Ana06 I have made the changes :)

app/views/admin/events/_proposal.html.haml
```diff
@@ -117,7 +117,7 @@
           %td
             %b Scheduled time
           %td
-            = @event.time
+            = @event.time.strftime("%F %R") + " " + @conference.timezone.to_s
```
Ana06: You could have this in a helper, so you don't repeat the code :wink:

Ana06:
@differentreality are you fine with this approach?

app/views/admin/events/_proposal.html.haml
nikhilgupta1211: Done :)

nikhilgupta1211:
Done :)

Ana06:
@nikhilgupta1211 as I said in my last comment:
> read the Contribution guide. Do not add the issue this is solving in the commit title, but at the end of the description.

nikhilgupta1211:
@Ana06 fixed it :)

Ana06:
It is still wrong :confounded: 

Do **not** add the issue this is solving in the commit title, but at the end of the description.

nikhilgupta1211:
Sorry, is this one right ?

Ana06:
yes :+1:

Issue #1394, opened 2017-03-21
Create resource with empty used/total_quantity throws an exception


shlok007:
**Issue**: Creating/Updating resource with empty `quantity` or `used` returns the following.

![error](https://cloud.githubusercontent.com/assets/14155445/24133851/41cc4378-0e26-11e7-897d-249b45864cc8.png)

**Possible Solution**: Add a validation for presence of `used` and `quantity` while creating/updating resource.

agrim123:
@shlok007 I think we need to add check apart from adding validations for `used` and `quantity`. Something like this can be done
```
def used_less_than_quantity
  if !used.nil? && !quantity.nil?
    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
  end
end
```

Ana06:
@shlok007 
> Possible Solution: Add a validation for presence of used and quantity while creating/updating resource.

and what if you don't know the total resources you have? I wouldn't make this compulsory.

@agrim123 I like your idea, feel free to open a PR. Although you can use `present?` instead of `! nil?` which is more readable and have `used <= quantity` in the same condition.

agrim123:
@Ana06 I think the `quantity` input field should have a default value of 0 as `used` has. It is causing problems while rendering index page for resources, giving error 
`undefined method '-' for nil:NilClass`
because `quantity` is being assigned a nil value !

shlok007:
Adding a default value wouldn't solve the problem of resources whose quantity are not yet know. Maybe we we could rewrite the helper function and in the views render something else ( like a  dash(`-`) ) when the `quantity` is not yet known? This could be a temporary solution until we come up with https://github.com/openSUSE/osem/issues/1284

Something in terms of:
![resource_index](https://cloud.githubusercontent.com/assets/14155445/24146021/d18010ca-0e59-11e7-8284-28587f7475fd.png)

differentreality:
Maybe we should not be adding resources without a total number

agrim123:
But `quantity_left_of` helper function is performing a arithmetic calculation so we need to give it an integer value.

agrim123:
Maybe we can do something like this 
```
def used_less_than_quantity
  return (errors.add(:used, 'can not be higher than total quantity') unless used <= quantity) if used.present? && quantity.present?
end
```
@shlok007 @differentreality what say?

Ana06:
@differentreality 

> Maybe we should not be adding resources without a total number

[...]

PR #1397, opened 2017-03-21
Fix issue while creating or updating resources

agrim123:
Previously on creating a new resource with empty form raised error
`undefined method '<=' for nil::NilClass`
Tried to fix it by adding an extra check for presence of fields, simultaneously set a default value of `quantity` in form to 0.

Still following problem prevails : 

- [x] If a user forcefully submits an empty form we need to check to avoid raising an exception 
`undefined method '-' for nil:NilClass`.
- [x] We can also add a default value 0 to `quantity`

Addresses #1394

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -578,6 +578,7 @@ module ApplicationHelper
   end
 
   def quantity_left_of(resource)
+    return '-/-' if resource.quantity.blank?
     "#{resource.quantity - resource.used}/#{resource.quantity}"
   end
 
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -1,10 +1,12 @@
 class Resource < ActiveRecord::Base
   belongs_to :conference
-  validate :used_less_than_quantity
+  validates :name, :used, :quantity, presence: true
+  validates :used, :quantity, numericality: { greater_than_or_equal_to: 0, only_integer: true }
+  validate :used_no_more_than_quantity
 
   private
 
-  def used_less_than_quantity
-    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
+  def used_no_more_than_quantity
+    errors.add(:used, 'cannot be higher than total quantity') if used.present? && quantity.present? && used > quantity
   end
 end
--- a/app/views/admin/resources/_form.html.haml
+++ b/app/views/admin/resources/_form.html.haml
@@ -9,7 +9,7 @@
   .col-md-8
     = semantic_form_for(@resource, :url => (@resource.new_record? ? admin_conference_resources_path : admin_conference_resource_path(@conference.short_title, @resource))) do |f|
       = f.input :name
-      = f.input :description, input_html: { rows: 5, data: { provide: "markdown-editable" } }
+      = f.input :description, input_html: { rows: 5, data: { provide: 'markdown-editable' } }
       = f.input :used
       = f.input :quantity
       %p.text-right
--- a/lib/tasks/normalize_resources.rake
+++ b/lib/tasks/normalize_resources.rake
@@ -1,0 +1,12 @@
+namespace :data do
+  desc 'Update resources with nil quantity/used fields'
+  task normalize_resources: :environment do
+    Resource.where('used is ? or quantity is ?', nil, nil).each do |resource|
+      resource.used = 0 if resource.used.blank?
+      resource.quantity = resource.used if resource.quantity.blank?
+      unless resource.save
+        puts "Failed to update resource #{resource.name} (ID #{resource.id})"
+      end
+    end
+  end
+end
--- a/spec/models/resource_spec.rb
+++ b/spec/models/resource_spec.rb
@@ -4,6 +4,24 @@ describe Resource do
   let(:conference) { create(:conference) }
   let(:resource) { create :resource }
 
+  it { is_expected.to validate_presence_of(:name) }
+
+  it { is_expected.to validate_presence_of(:used) }
+
+  it { is_expected.to validate_presence_of(:quantity) }
+
+  it { is_expected.to validate_numericality_of(:used) }
+
+  it { is_expected.to validate_numericality_of(:quantity) }
+
+  it { is_expected.not_to allow_value(-1).for(:used) }
+
+  it { is_expected.to allow_value(0).for(:used) }
+
+  it { is_expected.not_to allow_value(-1).for(:quantity) }
+
+  it { is_expected.to allow_value(0).for(:quantity) }
+
   it 'has a valid factory' do
     expect(build(:resource)).to be_valid
   end


shlok007:
The solution to the problem that prevails would depend on whether we want to validate presence of `quantity` or not which is being discussed in #1394 .

app/models/resource.rb
```diff
@@ -5,6 +5,6 @@ class Resource < ActiveRecord::Base
   private
 
   def used_less_than_quantity
-    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
+    return errors.add(:used, 'can not be higher than total quantity') if used.present? && quantity.present? && !(used <= quantity)
```
shlok007: Maybe we could get rid of `return` as well ?

app/views/admin/resources/_form.html.haml
```diff
@@ -11,6 +11,6 @@
       = f.input :name
       = f.input :description, input_html: { rows: 5, data: { provide: "markdown-editable" } }
       = f.input :used
-      = f.input :quantity
+      = f.input :quantity, :input_html => { :value => 0 }
```
shlok007: You should consider using the new ruby syntax instead. :v:

app/models/resource.rb
```diff
@@ -5,6 +5,6 @@ class Resource < ActiveRecord::Base
   private
 
   def used_less_than_quantity
-    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
+    errors.add(:used, 'can not be higher than total quantity') if used.present? && quantity.present? && !(used <= quantity)
```
differentreality: Why do we want to throw this error if  used equals quantity?
&& used > quantity

differentreality:
A resource is identified by a name and a quantity. We need to validate for the existence of those two things.

agrim123:
Then shall we add a validation in model itself?

differentreality:
Yes, exactly.

differentreality:
If we had previously added resources without quantity attribute filled, this is still not going to work. We need to provide for a solution for that.

differentreality:
@agrim123 ping

agrim123:
@differentreality How shall I proceed with this PR?

differentreality:
@agrim123 please read my comment from 5 days ago and make your changes work accordingly.

agrim123:
@differentreality I think https://github.com/openSUSE/osem/issues/1394#issuecomment-288054311 solution is good.

shlok007:
>A resource is identified by a name and a quantity. We need to validate for the existence of those two things.

@agrim123 That solution was before the above decision. In my opinion if a resource is identified by name and quantity then they have to be present now. We shouldn't consider otherwise.

shlok007:
Also, please add relevant tests for the validations introduced.

shlok007:
@agrim123 consider the scenario:
One bright sunny day, someone is going to look at our code and see these two conflicting lines[ [1](https://github.com/openSUSE/osem/pull/1397/files#diff-127a850b7d7b20173ad3fba88f8a25e3R576) [2](https://github.com/openSUSE/osem/pull/1397/files#diff-2248e5266fb5961b5fbc8d18e969e716R3) ]. If we have provided a validation for _quantity_ then checking if _quantity_ is present or not inside the helper doesn't make sense. The condition will be removed from the helper method and the PR might get merged, breaking apps that still have quantity set to _nil_ ( considering the fact that people have actually started using resources )

@agrim123 @differentreality  Is a task for checking the old data for `nil` quantity and updating it to say `0` sounds somewhat a better solution or maybe I am thinking too much? :slightly_smiling_face:

agrim123:
@shlok007 Regarding duplicate validations I will fix them.
Instead of updating `nil` quantity to zero displaying `-` is better so that at least it can be updated. `0` somewhat doesn't sound good to me!

differentreality:
If we have a validation for quantity to exist, that means that all our records will have that. As such, we should introduce a task that brings our records to the state they should be.

agrim123:
So shall we add a function to check previously added resources and prompt users to update?

differentreality:
We need to create a rake task that will take care of this, I guess that means that we put as quantity the amount of used, or 0.

agrim123:
@differentreality Can you please brief me on how to apply this?

differentreality:
There are plenty of tasks in OSEM, why don't you take a look at how things are done?

agrim123:
@differentreality Please review the rake task I added.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
+namespace :resources do
```
differentreality: this can be part of data namespace

app/models/resource.rb
```diff
@@ -1,10 +1,11 @@
 class Resource < ActiveRecord::Base
   belongs_to :conference
+  validates :name, :quantity, presence: true
   validate :used_less_than_quantity
 
   private
 
   def used_less_than_quantity
-    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
+    errors.add(:used, 'can not be higher than total quantity') if used.present? && quantity.present? && used > quantity
```
differentreality: `if used > quantity` is enough once the proper validations are added

app/models/resource.rb
```diff
@@ -1,10 +1,11 @@
 class Resource < ActiveRecord::Base
   belongs_to :conference
+  validates :name, :quantity, presence: true
```
differentreality: your task implies that `used` should also be present, so I would validate this as well. 

And I would also like to update our DB accordingly, so that null is **not** allowed for `used` and `quantity`.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
...
+  desc "Update resources with nil quantity"
```
differentreality: Single quotes.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
...
+  task update_resource_quantity: :environment do
+    Resource.all.each do |r|
```
differentreality: When I read `r`, I have to think about what it is referring to. Please use meaningful words, ie `resource`, that helps a lot to understand the code, when you read it later on.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
...
+      if r.quantity.nil? || r.used.nil?
+        if r.used.nil?
+          Resource.update(r.id, quantity: 0, used: 0)
```
differentreality: You can re-write this in a more compact and elegant way:

```
resource.used = 0 unless resource.used.present?
resource.quantity = resource.used unless resource.quantity present?
```

Then you can do `resource.save` with an if statement to make sure the resource is properly saved, and if not, inform the user about it at the end.

app/helpers/application_helper.rb
```diff
@@ -573,7 +573,11 @@ def replacement_event_notice(event_schedule)
   end
 
   def quantity_left_of(resource)
-    "#{resource.quantity - resource.used}/#{resource.quantity}"
+    if resource.quantity.nil?
+      '-/-'
```
differentreality: This can be a single line:
`return '-/-' unless resource.quantity.present?`

It is generally a good practice to exit a function in the beginning, unless the necessary variables/values are set.
In our case, if there is no quantity, we should not continue. 
In other cases we may need a couple of values to be set, and if they aren't we just return nil or false or another default value, and don't continue further on. That way, particularly in a more complex method, you don't have to add a ton of if statements just to check the presence of a value.

app/models/resource.rb
```diff
@@ -1,10 +1,11 @@
...
   validate :used_less_than_quantity
```
differentreality: Just to be in sync with what we are actually validating, we can perhaps rephrase this into something like `used_no_more_than_quantity`

lib/tasks/update_resource_quantity.rake
differentreality: And to make sure this works, you should create resources in master, then go into your branch and execute this task to make sure things become the way they are supposed to, and nothing else is missing.

app/models/resource.rb
differentreality: numerical fields should probably be validated for their numericality as well - should I be able to enter -5 as quantity or used?

app/models/resource.rb
agrim123: When user leaves `used` blank then error is generated
```
undefined method `>' for nil:NilClass
```

differentreality: Did you check all my other comments? As I said after we add validations.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
+namespace :data do
+  desc 'Update resources with nil quantity'
+  task update_resource_quantity: :environment do
+    Resource.all.each do |resource|
+      resource.used = 0 unless resource.used.present?
+      resource.quantity = resource.used unless resource.quantity.present?
+      if resource.save
+        puts 'Success for resource with id ' + resource.id.to_s + ' and name ' + resource.name
+      else
+        puts 'Failed for resource with id ' + resource.id.to_s + ' and name ' + resource.name
```
differentreality: If this was successful, I don't really need to know the id and name of the resource, right?
A successful message once at the end would be enough.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
...
+      end
+    end
+  end
+end
```
differentreality: Whenever you see this in the code you push in github, please fix it. You can configure your text editor to do this automatically, so that you never have to care for this again. Ping me on IRC to help you enable it.

spec/models/resource_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:conference) { create(:conference) }
   let(:resource) { create :resource }
 
+  it { is_expected.to validate_presence_of(:name) }
+
+  it { is_expected.to validate_presence_of(:quantity) }
+
```
differentreality: You are missing tests for your model

app/models/resource.rb
```diff
@@ -1,10 +1,12 @@
 class Resource < ActiveRecord::Base
   belongs_to :conference
-  validate :used_less_than_quantity
+  validates :name, presence: true
+  validates :used, :quantity, presence: true, numericality: { greater_than_or_equal_to: 0, only_integer: true }
```
differentreality: I'd rather you add all the attributes that need to be present in one validation, meaning move the presence validation in the above line, instead of validating for presence in more than one place.
That way we can easily see in 1 validation which are the attributes that need to be present.

app/models/resource.rb
differentreality: :+1:  Yes OK let's leave it like that, otherwise we would need to alter the validation statement

differentreality:
Very good :) I have merely made some styling/usability comments that I think we should address.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
+namespace :data do
+  desc 'Update resources with nil quantity'
+  task update_resource_quantity: :environment do
```
differentreality: This is not very accurate

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
+namespace :data do
+  desc 'Update resources with nil quantity'
```
differentreality: This is not very accurate, as we are not really doing that (updating the resources that have nil quantity).
It is also not very descriptive, as it doesn't say how we are updating them.

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,14 @@
...
+  task update_resource_quantity: :environment do
+    Resource.all.each do |resource|
```
differentreality: It makes sense to iterate only through the records that matter to us, instead of all the records.

agrim123: How do we choose those records?

lib/tasks/update_resource_quantity.rake
agrim123: How about `Fix resources with empty quantity or used values`?

lib/tasks/update_resource_quantity.rake
shlok007: https://apidock.com/rails/ActiveRecord/Calculations/pluck

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,10 @@
+namespace :data do
+  desc 'Update resources with nil quantity'
+  task update_resource_quantity: :environment do
+    Resource.where('used is null or quantity is null').each do |resource|
```
differentreality: http://guides.rubyonrails.org/active_record_querying.html#array-conditions

lib/tasks/update_resource_quantity.rake
differentreality: normalize_resources

lib/tasks/update_resource_quantity.rake
differentreality: Update resources with nil quantity/used fields

lib/tasks/update_resource_quantity.rake
```diff
@@ -0,0 +1,10 @@
...
+      resource.used = 0 unless resource.used.present?
+      resource.quantity = resource.used unless resource.quantity.present?
+      puts 'Failed for resource with id ' + resource.id.to_s + ' and name ' + resource.name if !resource.save
```
differentreality: How about the following? It is a bit more readable as this line is getting too long.

```
unless resource.save
  puts "Failed to update resource #{resource.name} (ID #{resource.id})"
end
```

lib/tasks/update_resource_quantity.rake
differentreality: You seem to have ignored this comment, why?

lib/tasks/update_resource_quantity.rake
differentreality: There is still no message at the end indicating that the task has fully run, how can the user running the task know that it has finished?

app/models/resource.rb
```diff
@@ -1,10 +1,12 @@
...
+  validates :name, :used, :quantity, presence: true
+  validates :used, :quantity, numericality: { greater_than_or_equal_to: 0, only_integer: true }
+  validate :used_no_more_than_quantity
 
   private
 
-  def used_less_than_quantity
-    errors.add(:used, 'can not be higher than total quantity') unless used <= quantity
+  def used_no_more_than_quantity
+    errors.add(:used, 'can not be higher than total quantity') if used.present? && quantity.present? && used > quantity
```
differentreality: `cannot` in one word is most commonly used

differentreality:
There are a couple of more comments to refine your code. Please make the changes, squash your commits and rebase so we can merge this! :tada:

differentreality:
@agrim123 can you also squash your commits please? I don't think we need all 4 commits

lib/tasks/normalize_resources.rake
```diff
@@ -0,0 +1,12 @@
+namespace :data do
+  desc 'Update resources with nil quantity/used fields'
+  task normalize_resources: :environment do
+    Resource.where('used is null or quantity is null').each do |resource|
```
differentreality: Please fix this where statement according to the guide:
http://guides.rubyonrails.org/active_record_querying.html#array-conditions

agrim123: Not able to find any other option
`Resource.where('used is null or quantity is null').each do |resource|` 
Reason: http://stackoverflow.com/questions/4340517/how-to-compare-null-in-rails-active-record-query

differentreality: What about `Resource.where('used is ? or quantity is ?', nil, nil)` ?

differentreality:
:+1: good job!

PR #1407, opened 2017-03-25
Docker support for production use

### LINKED ISSUE #1377: Production grade Dockerfile
I am working on a Dockerfile for deploying OSEM in production at the moment. As @hennevogel mentioned in #1213, this might be considerable to add to this project.

I think that the OSEM might benefit quite much from having this included upstream, because that simplifies deployment greatly (especially since there's no installation documentation, the referral to

> tons of guides on how to deploy rails apps on various base operating systems

are not really useful to the end user, especially unexperienced people. I have struggled a bit, too, during the creation of my Dockerfile.

I would be willing to send in a PR for this, although I would not pin it down to MySQL/MariaDB, as PostgreSQL works just as fine as SQLite3 (latter is useful for people who just want to try it out, it can work smaller scale deployments, too).

TheAssassin:
This commit adds a Docker infrastructure that is ready for production use. It is meant to simplify the deployment of OSEM for everyone who wants to host their own instances.

It includes many features like data persistence, automatic secret key generation and persistence and automatic database initialization and upgrading. This should make updating the Docker container as easy as
possible.

This is the first version, and although it is very stable (as far as I can tell from about two days of extensive testing), this might not yet fit OSEM's needs completely, so I regard it as a draft at the moment.

Feel free to comment and make suggestions, I'd like to make this as good and easy to use as possible, as I am one of the potential end users who are going to profit from this commit a lot.

This PR would close #1377 when being merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ pickle-email-*.html
 .env.development
 .env.test
 .env.local
+docker-compose.{env,yml}
--- a/.gitignore
+++ b/.gitignore
@@ -36,4 +36,5 @@ pickle-email-*.html
 .env.development
 .env.test
 .env.local
-docker-compose.{env,yml}
+docker-compose.env
+docker-compose.yml
--- a/docker-compose.env.example
+++ b/docker-compose.env.example
@@ -32,6 +32,7 @@ SECRET_KEY_BASE=changemechangemechangeme
 # you should comment out or remove the mailhog service from docker-compose.yml,
 # too
 OSEM_EMAIL_ADDRESS=osem@mailhog
+OSEM_SMTP_AUTHENTICATION=login
 OSEM_SMTP_ADDRESS=mailhog
 OSEM_SMTP_PORT=1025
 OSEM_SMTP_USERNAME=mailhog
--- a/docker-compose.env.example
+++ b/docker-compose.env.example
@@ -1,0 +1,38 @@
+## database related variables ##
+
+# variables prefixed with MYSQL_ are used by both database and web containers
+# variables prefixed with DATABASE_ are used exlusively by the web container
+
+MYSQL_DATABASE=osem
+MYSQL_USER=osem
+MYSQL_PASSWORD=changemeimmediately
+MYSQL_ROOT_PASSWORD=changemeevenmoreimmediately
+
+# the following settings should not be modified unless the database service
+# is renamed in docker-compose.yml or you plan to use an external database
+DATABASE_HOST=database
+DATABASE_PORT=3306
+
+
+## OSEM options ##
+# you can configure any option described in this document here instead of
+# having to create a .env file:
+# https://github.com/openSUSE/osem/blob/master/dotenv.example
+
+OSEM_NAME=Dockerized OSEM
+OSEM_HOSTNAME=http://localhost:9292
+OSEM_ERRBIT_HOST=localhost
+SECRET_KEY_BASE=changemechangemechangeme
+
+# these settings work for the MailHog server that is enabled by default in
+# docker-compose.yml
+# if you do not plan to use MailHog (you most likely don't want to), you need
+# to change these settings to use an external working mailserver, otherwise
+# your users are going to see the HTTP status 500 page
+# you should comment out or remove the mailhog service from docker-compose.yml,
+# too
+OSEM_EMAIL_ADDRESS=osem@mailhog
+OSEM_SMTP_ADDRESS=mailhog
+OSEM_SMTP_PORT=1025
+OSEM_SMTP_USERNAME=mailhog
+OSEM_SMTP_PASSWORD=mailhog
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -31,13 +31,16 @@ user=$MYSQL_USER
 password=$MYSQL_PASSWORD
 ABC
 
-if [ $(echo "show tables;" | dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
+echo ">>> Waiting for database to get ready to connect..."
+dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s true
+
+if [ $(echo "show tables;" | mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
     echo ">>> Initializing database..."
-    dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:schema:load
+    bundle exec rake db:schema:load
 fi
 
 echo ">>> Upgrading database..."
-dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:migrate
+bundle exec rake db:migrate
 
 rm .my.cnf
 
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -1,0 +1,45 @@
+#! /bin/bash
+
+set -e
+
+# data directory is required for caching the secret key in a file
+if [ "$DATA_DIR" == "" ]; then
+    echo -n "Error: DATA_DIR environment variable not set!"
+    echo "Are you sure you are running this script in a Docker container?"
+    exit 1
+fi
+
+SECRET_KEY_FILE="$DATA_DIR/secret_key"
+
+if [ ! -f "$SECRET_KEY_FILE" ]; then
+    install -m 0600 /dev/null "$SECRET_KEY_FILE"
+    SECRET_KEY=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 100 | head -n 1)
+    echo "$key" > "$SECRET_KEY_FILE"
+    chmod -w "$SECRET_KEY_FILE"
+else
+    SECRET_KEY=$(cat "$SECRET_KEY_FILE")
+fi
+
+export SECRET_KEY
+export RAILS_ENV=production
+
+install -m 0600 /dev/null .my.cnf
+cat > .my.cnf <<ABC
+[client]
+user=$MYSQL_USER
+password=$MYSQL_PASSWORD
+ABC
+
+if [ $(echo "show tables;" | mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
+    echo ">>> Initializing database..."
+    dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:schema:load
+fi
+
+echo ">>> Upgrading database..."
+dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:migrate
+
+echo ">>> Precompiling assets..."
+bundle exec rake assets:precompile
+
+echo ">>> Starting application server..."
+exec puma -e production
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -38,6 +38,8 @@ fi
 echo ">>> Upgrading database..."
 dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:migrate
 
+rm .my.cnf
+
 echo ">>> Precompiling assets..."
 bundle exec rake assets:precompile
 
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -42,4 +42,4 @@ echo ">>> Precompiling assets..."
 bundle exec rake assets:precompile
 
 echo ">>> Starting application server..."
-exec puma -e production
+exec bundle exec rails server -e production -p 9292
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -45,4 +45,4 @@ echo ">>> Precompiling assets..."
 bundle exec rake assets:precompile
 
 echo ">>> Starting application server..."
-exec bundle exec rails server -e production -p 9292
+exec bundle exec rails server -e production -b 0.0.0.0 -p 9292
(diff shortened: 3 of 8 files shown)


ChrisBr:
@lagartoflojo mind to have a look? I have not much experience with docker ...

TheAssassin:
I've updated the PR to include some documentation.

docker/init.sh
```diff
@@ -0,0 +1,45 @@
...
+# data directory is required for caching the secret key in a file
+if [ "$DATA_DIR" == "" ]; then
+    echo -n "Error: DATA_DIR environment variable not set!"
+    echo "Are you sure you are running this script in a Docker container?"
+    exit 1
+fi
+
+SECRET_KEY_FILE="$DATA_DIR/secret_key"
+
+if [ ! -f "$SECRET_KEY_FILE" ]; then
```
hennevogel: Why do you reinvent `bundle exec rails secret`? :-)

docker/init.sh
```diff
@@ -0,0 +1,45 @@
...
+user=$MYSQL_USER
+password=$MYSQL_PASSWORD
+ABC
+
+if [ $(echo "show tables;" | mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
+    echo ">>> Initializing database..."
+    dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:schema:load
+fi
+
+echo ">>> Upgrading database..."
```
hennevogel: There is no need for this. The schema always contains the latest schema :-) There will never be updates...

docker/init.sh
```diff
@@ -0,0 +1,45 @@
...
+dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:migrate
+
+echo ">>> Precompiling assets..."
+bundle exec rake assets:precompile
+
+echo ">>> Starting application server..."
+exec puma -e production
```
hennevogel: We are neither using nor shipping puma in the bundle. I guess you want to just run `rails server -e production`.

TheAssassin: Thanks for the tip. I have used puma in the past with Redmine with great success, it's really efficient and fast. But I understand that another dependency is unnecessary, I'll change it to use the existing server.

docker/init.sh
TheAssassin: Well, upgrading the database makes sure that whenever the OSEM repository is updated and a new container is built, changes to the schema will be applied to the database, too.
The only time the upgrades are applied unnecessarily is during the first run. This could be easily "fixed" by putting the upgrading part into an else clause of the preceding if block, but I think that's not really necessary. But if you insisted on it, I'd "fix" it.

docker/init.sh
TheAssassin: As mentioned before, I don't have too much experience with either Ruby or Rails. This method worked quite well, and the key is cached in a named volume.

This could also obsolete having to set `SECRET_KEY_BASE` in `docker-compose.env`. And it brings up the question: Should the Docker container still autogenerate the key(s) and store them in a volume?

At the moment, there's the following options:

  * Autogenerate the keys (and optionally store them in a volume, otherwise they had to be regenerated on updates). Later, the user could use `docker-compose run bundle exec rails secrets`to regenerate them. The only problem would be how to put the `secrets.yml` in a container, but I guess I could figure that out (a symlink would probably work).
  * Have the user create a `secrets.yml` on the host that is mounted as a volume. This is something I'd really like to avoid, because it requires Ruby and all the deps on the host, too, which eliminates many of Docker's advantages. This would also create a bunch of security issues because a user would have to make sure the file is kept secret.
  * Make the user set `SECRET_KEY_BASE` and `SECRET_KEY` in the `docker-compose.env` file. This might work well, too, but doesn't sound like the best option either.

I would go for the first approach.

docker/init.sh
TheAssassin: Fun fact: I just noticed that `rails server` is supposed to launch Puma anyway (http://guides.rubyonrails.org/command_line.html).

TheAssassin:
JFYI: I've implemented every requested change except the schema related one.

AEtherC0r3:
When deploying a rails app in production it's a good practice to use a reverse proxy instead of exposing the rails server directly. Therefore, it might be appropriate to rename the "web" container to "app" and add a new "web" container that will act as a reverse proxy (maybe nginx?). What do you think?

TheAssassin:
I already documented this in the [guide I wrote](https://github.com/openSUSE/osem/pull/1407/commits/cdd3735c0994957cf1c654f1001a31d4c77ef078#diff-7d442b7eb49f5fc377f51e74b291cfc1R71). If you wondered why I expose the OSEM port to the local host only in `docker-compose.yml`, you've found the reason: OSEM should be deployed behind a reverse proxy. (Experienced users would deploy it behind a reverse proxy anyway.)

I wouldn't add a separate NGINX container because most Docker deployments (especially with compose) are done on single host Docker servers with a reverse proxy running on the host and serving anything that is run via Docker. Especially for smaller scale setups and when there's more than one app running on that host, it's a must anyway, because it is the only way to get multiple apps on the ports 80 or 443 (on separate domains for example) without the need for multiple IP addresses. For such setups, running another NGINX is just redundant and should be avoided. An interesting reason to deploy a separate NGINX container would only be to automatically serve static files directly from NGINX, bypassing the rails server. But that's again a more complex situation (with roblems like how to get the static files into the NGINX container), and since Rails is able to set HTTP headers like `ETag` or `Expire`, NGINX and other reverse proxy are even able to cache these files if configured accordingly (browsers cache them anyway). This could be added to the documentation, sure, but I don't want to include something this complex because it's actually not in the scope of such an example Docker configuration to add such functionality. It has to be maintained, the configs need to be updated regulary etc., and noone really wants to do that. And only high-load servers would need to optimize these factors, small scale deployments don't have any problems with that usually.

About renaming the container: That might be a good idea, I usually call the container with the actual web app `web` in all my configs, but `app` (or e.g. `osem`) work just as well.

lguerard:
Hi,

I don't know if here's the correct place to ask for help on your docker-OSEM so let me know if that's not the case.

I have tried installing your version for evaluation but localhost:9292 shows me nothing afterwards.. The port is set to listen and is open, I have looked the whole afternoon without success...

If you have any ideas please let me know!
Thanks.

TheAssassin:
I was able to rebase the master branch. I'll build a container and perform some tests to see if it still works.

TheAssassin:
@lguerard I'm sorry I cannot provide single user support for this. I am just too busy. You can join #osem on Freenode if you want and ask your questions there, maybe I'll find a few minutes to review your problem.

A few general tips:
  - check the Docker log, it contains the output of the initialization and server processes
  - if you cannot find the problem, post the logs when asking for help (make sure you strip all private information before you share it with a pastebin service and/or third parties in general)

TheAssassin:
@lguerard lucky for you I had the same issue and fixed it in the upstream. Please update, rebuild the container and check if it works now.

TheAssassin:
@ChrisBr could you please trigger a review from @hennevogel or @lagartoflojo? That PR is getting quite old now, and I'd really like to see that merged soon, so I can use it in production and spread the spirit of OSEM.

ChrisBr:
@TheAssassin you already pinged them, right :) If they don't reply, they're probably busy :(

docker/init.sh
```diff
@@ -0,0 +1,48 @@
...
+export RAILS_ENV=production
+
+install -m 0600 /dev/null .my.cnf
+cat > .my.cnf <<ABC
+[client]
+user=$MYSQL_USER
+password=$MYSQL_PASSWORD
+ABC
+
+if [ $(echo "show tables;" | mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
```
lagartoflojo: Also, the command could just be:

```
mysql -e "show tables;" --host $DATABASE_HOST --port $DATABASE_PORT --password=$MYSQL_PASSWORD $MYSQL_DATABASE
```

TheAssassin: > Missing `--password=$MYSQL_PASSWORD`

No, this is explicitly not required. I've implemented a more secure way a few lines above, writing the password to the MySQL client config.

And whether to use echo or `-e` doesn't really matter. If you insist, I can change it.

docker/init.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo ">>> Initializing database..."
+    dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s bundle exec rake db:schema:load
```
TheAssassin: Right. I'm not sure why I didn't notice earlier. I'll implement a change.

.gitignore
```diff
@@ -36,3 +36,4 @@ pickle-email-*.html
 .env.development
 .env.test
 .env.local
+docker-compose.{env,yml}
```
TheAssassin: Just because I'm curious, what version of Git do you have and what OS do you run?

docker/init.sh
```diff
@@ -0,0 +1,48 @@
...
+    echo "Are you sure you are running this script in a Docker container?"
+    exit 1
+fi
+
+SECRET_KEY_FILE="$DATA_DIR/secret_key"
+
+if [ ! -f "$SECRET_KEY_FILE" ]; then
+    echo ">>> Creating a new secret key file..."
+    install -m 0600 /dev/null "$SECRET_KEY_FILE"
+    SECRET_KEY=$(bundle exec rails secret)
```
TheAssassin: As you should can see, the file already _uses_ the `bundle` command to generate the secret. But the secret should be cached across multiple containers. In Docker world, containers are disposable things, they should not contain any permanent data, so one has to save it to volumes. Please see the Docker docs for more information. So, just calling the command is not enough.

docker-compose.env.example
```diff
@@ -0,0 +1,38 @@
...
+# if you do not plan to use MailHog (you most likely don't want to), you need
+# to change these settings to use an external working mailserver, otherwise
+# your users are going to see the HTTP status 500 page
+# you should comment out or remove the mailhog service from docker-compose.yml,
+# too
+OSEM_EMAIL_ADDRESS=osem@mailhog
+OSEM_SMTP_ADDRESS=mailhog
+OSEM_SMTP_PORT=1025
+OSEM_SMTP_USERNAME=mailhog
+OSEM_SMTP_PASSWORD=mailhog
```
TheAssassin: Well, not necessarily. The mailhog test system provided by the default configuration for example works without it. I am not necessarily sure whether it picks the right authentication method automatically, or its default value works. I'd suggest to add it as a commented line, as it is an optional value.

lagartoflojo: Without that option, Rails errors out:

```
$ docker exec web rails c production
> ActionMailer::Base.mail(from: "me@example.com", to: "you@example.com", subject: "test", body: "test").deliver
ArgumentError: wrong authentication type
```

docker/init.sh
lagartoflojo: What I mean is that it should call **rake** secret, not **rails** secret.

**rails** secret does not work:

```
$ docker-compose exec web bundle exec rails secret
Error: Command 'secret' not recognized
Did you mean: `$ rake secret` ?
```

docker/init.sh
lagartoflojo: Hmm, strange. When I tried before, consecutive runs of `docker-compose up` would re-run `rake db:schema:load`. I just tried again, and it works fine, the schema is only loaded on the first run. Sorry for the noise.

docker/init.sh
lagartoflojo: (Only in Rails 5 did they make `rake` and `rails` are interchangeable, which was always a source of confusion... OSEM is still on Rails 4)

.gitignore
lagartoflojo: I tried this on MacOS Sierra, git version 2.6.3.

docker/init.sh
TheAssassin: I see. That obviously has to be fixed. I have to admit I'm not too experienced with Rails, and OSEM doesn't have any real documentation, it expects a certain degree with Ruby and Rails.

docker-compose.env.example
TheAssassin: Interesting. I'm kind of too lazy to try to reproduce that now. I'll implement it, and check whether mailhog still works.

TheAssassin:
All open points should be fixed.

docker/init.sh
```diff
@@ -0,0 +1,48 @@
...
+export RAILS_ENV=production
+
+install -m 0600 /dev/null .my.cnf
+cat > .my.cnf <<ABC
+[client]
+user=$MYSQL_USER
+password=$MYSQL_PASSWORD
+ABC
+
+if [ $(echo "show tables;" | dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s mysql --host $DATABASE_HOST --port $DATABASE_PORT $MYSQL_DATABASE | wc -l) -le 1 ]; then
```
lagartoflojo: Just wondering... wouldn't it be easier to put a single `dockerize` call before line 34 and be done with it? Any particular reason it's better to duplicate it?

TheAssassin: `dockerize` is like a launcher for another command that makes sure a host/port combination is reachable, or renders the environment variables into a config file. It guess it should work, if used like `dockerize -wait ... true`.

I just used it twice before because I didn't think of that possibility, and the if clause didn't evaluate every time, and as the command below it requires a database connection, I inserted it there, too. I'll have to test that, and will push a commit.

lagartoflojo: ok! Yeah, I asked because if I [read the docs](https://github.com/jwilder/dockerize#waiting-for-other-dependencies) correctly, it doesn't require a command, just `dockerize -wait tcp://$DATABASE_HOST:$DATABASE_PORT -timeout 60s` should suffice.

lagartoflojo:
Merged! Thanks a lot @TheAssassin!

TheAssassin:
Thanks!

TheAssassin:
P.S.: Let me know if I can help you with further adjustments.

lagartoflojo:
I would actually like to adapt (or altogether add) Docker support for development purposes. I actually already use Docker to work on OSEM, using my own docker-compose.yml file, but it would be great if we offered this out of the box.

Is that something you'd be interested in tackling?

PR #1408, opened 2017-03-25
Fix #1406. Registration form consistency with User model

### LINKED ISSUE #1406: in registration form

msdundar:
There is an inconsistent situation with User model and new registration form.

-  "Name" field stands for the real name of users, however it's not clear if a full name, first name or last name expected.

The issue (#1406) is because of lack of necessary validations on password and password_confirmation. Beside this, there is no validation about "name" on the User model and it seems like name is an optional field:

```
  def name
    self[:name].blank? ? username : self[:name]
  end
```

So it is probably best to remove { required: true } tag for 'name', and, add necessary validations for password and password_confirmation.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -53,7 +53,7 @@ class User < ActiveRecord::Base
 
   scope :admin, -> { where(is_admin: true) }
 
-  validates :email, presence: true
+  validates :email, :password, :password_confirmation, presence: true
 
   validates :username,
             uniqueness: {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -53,7 +53,8 @@ class User < ActiveRecord::Base
 
   scope :admin, -> { where(is_admin: true) }
 
-  validates :email, :password, :password_confirmation, presence: true
+  validates :email, presence: true
+  validates :password, :password_confirmation, presence: true, on: :create
 
   validates :username,
             uniqueness: {
--- a/app/views/devise/registrations/edit.html.haml
+++ b/app/views/devise/registrations/edit.html.haml
@@ -2,21 +2,22 @@
   .row
     .col-md-12
       .page-header
-        %h1 Edit your Account
+        %h1 Edit Your Account
   .row
     .col-md-12
       = semantic_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
         = f.input :username, required: false, input_html: { autocomplete: 'off' }
         = f.input :email, required: false, input_html: { autocomplete: 'off' }
-        = f.input :email_public
+        = f.input :email_public, label: "Make my email public"
         = f.input :password, hint: "(Leave blank if you don't want to change it)", input_html: { autocomplete: 'off' }
         = f.input :password_confirmation, input_html: { autocomplete: 'off' }
         = f.inputs name: 'OpenID' do
-          %h4
-            Currently the following openIDs are associated with your account
-          - @openids.each do |openid|
-            %li= "#{openid.provider}:#{openid.email}"
-            %br
+          - if @openids.present?
+            %h4
+              Currently the following openIDs are associated with your account
+              - @openids.each do |openid|
+                %li= "#{openid.provider}:#{openid.email}"
+                %br
           %h4
             To add an openID with a different email address to your account, sign in with your
             openID while logged in to OSEM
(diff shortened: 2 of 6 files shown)


app/models/user.rb
```diff
@@ -53,7 +53,8 @@ class User < ActiveRecord::Base
 
   scope :admin, -> { where(is_admin: true) }
 
-  validates :email, :password, :password_confirmation, presence: true
+  validates :email, presence: true
+  validates :password, :password_confirmation, presence: true, on: :create
```
msdundar: This separation is needed. Otherwise, 'edit profile' will not work, since both edit account and edit profile are working on the same model (User).

app/views/devise/sessions/new.html.haml
```diff
@@ -7,7 +7,7 @@
             Sign In
         .panel-body
           = semantic_form_for(resource, as: resource_name, url: session_path(resource_name)) do |f|
-            = f.input :login
+            = f.input :login, label: "Username / Email"
```
msdundar: To be consistent with the dropdown login form in the navbar.

app/views/users/edit.html.haml
```diff
@@ -11,7 +11,7 @@
         .control-label
           = "Avatar"
         = image_tag(@user.gravatar_url(size: '48'), title: "Yo #{@user.name}!", alt: '')
-        = link_to 'Change your avatar here', 'https://gravatar.com'
+        = link_to 'Change your avatar here', 'https://gravatar.com', target: :_blank
```
msdundar: Better to open links in a new tab.

app/models/user.rb
```diff
@@ -53,7 +53,7 @@ class User < ActiveRecord::Base
 
   scope :admin, -> { where(is_admin: true) }
 
-  validates :email, presence: true
+  validates :email, :password, :password_confirmation, presence: true
```
Ana06: This should be done like:

``` Ruby
validates :email, presence: true

validates :password, confirmation: true
validates :password_confirmation, presence: true
```

You can read: http://guides.rubyonrails.org/active_record_validations.html#confirmation


with that you shouldn't need your last commit (acbc7df96a96807e5c893fe52a7573525f27ef55)

app/models/user.rb
Ana06: if you edit the password if should also be validated

msdundar: If you remove on: :create, the profile/edit page will not work. Because they both operate on the same User model.

Ana06: this doesn't work neither. If you want to change your password you are not creating the user, but updating it and it should also be validated. Have you read https://github.com/openSUSE/osem/pull/1408#discussion_r108110859?

As you can read in the [Rails guide](http://guides.rubyonrails.org/active_record_validations.html#confirmation), this should be done like:

``` Ruby
validates :email, presence: true

validates :password, confirmation: true
validates :password_confirmation, presence: true
```

Ana06:
@msdundar you will also need to rebase:

`git rebase upstream master`

Ana06:
@msdundar do you need help to implement the requested changes? :slightly_smiling_face:

msdundar:
@Ana06 I am gonna push it soon :)

app/views/devise/registrations/edit.html.haml
```diff
@@ -2,21 +2,22 @@
...
       .page-header
-        %h1 Edit your Account
+        %h1 Edit Your Account
   .row
     .col-md-12
       = semantic_form_for(resource, as: resource_name, url: registration_path(resource_name), html: { method: :put }) do |f|
         = f.input :username, required: false, input_html: { autocomplete: 'off' }
         = f.input :email, required: false, input_html: { autocomplete: 'off' }
-        = f.input :email_public
+        = f.input :email_public, label: "Make my email public"
```
differentreality: Perhaps `Show my email` makes more sense?

app/views/devise/sessions/new.html.haml
differentreality: I would expect you to have commented that in the description of the PR, as one of the things you are doing in this PR.

differentreality:
@msdundar any progress on this? Do you need more help finalizing this PR so that it can be merged? You will need to rebase against latest changes on master branch.

differentreality:
@msdundar we would like to get this in as soon as possible. Do you plan on working on this, and do you think we can finish it within the coming week?

differentreality:
Closing this due to inactivity and lack of response

PR #1421, opened 2017-03-29
Enable Style/NegatedIf Cop and Fix Alphabetical Ordering

### LINKED ISSUE #1415: Enable Style/NegatedIf Rubocop cop
Enable Style/NegatedIf Rubocop cop: https://rubocop.readthedocs.io/en/latest/cops_style/#stylenegatedif

sunny-b:
Checks for the use of if in the negated condition. It supports auto correct, which was run to make the changes.

One of the cops was also out of alphabetical order, so that was changed as well.

Closes #1415

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -63,14 +63,14 @@ Style/EmptyLiteral:
 Style/For:
   Enabled: true
 
-# Checks that operators have space around them, except for ** which should not have surrounding space.
-Style/SpaceAroundOperators:
-  Enabled: true
-
 # Use the new Ruby 1.9 hash syntax
 Style/HashSyntax:
   Enabled: true
   EnforcedStyle: ruby19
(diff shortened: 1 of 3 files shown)


sunny-b:
I think they should be squashed now. I looked up how to do it, but if I didn't do it correctly, let me know.

Ana06:
@sunny-b there is something strange as your first commit author is `Iris Sprague`, witouht a Github account associated. I guess you hadn't correctly  configured git when you did the first commit, could that be? If that is the case you shouldn't worry, as the second was already correct. After you squashed them it looks strange as there are two authors. But I think this is not a problem and we can merge it :wink:

sunny-b:
Yeah I was on my wife's computer and git was configured under her name. I changed it after I noticed it was the wrong author

PR #1425, opened 2017-03-30
Fix indentation issue in notification dropdown layout

siddhantbajaj:
screenshot from 2017-03-31 02 30 03
There is an issue in notification drop down menu (I have highlighted it with black colour).

### DIFF
--- a/app/views/layouts/_navigation.html.haml
+++ b/app/views/layouts/_navigation.html.haml
@@ -28,19 +28,18 @@
           - if can? :index, Comment
             %ul.nav.navbar-nav.navbar-right
               %li.dropdown
-                %a.dropdown-toggle{"data-toggle" => "dropdown", href: '#'}
-                  - if unread_notifications(current_user)
-                    Notifications (#{unread_notifications(current_user).length})
-                    %span.fa.fa-comment
-                    %b.caret
-                    %ul.dropdown-menu
-                      - if unread_notifications(current_user).length > 0
-                        %li.dropdown-header Last 5 Comments for:
-                        - unread_notifications(current_user).limit(5).group_by{ |comment| comment.commentable}.each do |event, comments|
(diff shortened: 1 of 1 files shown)


differentreality:
Then do we really need this `if`:
https://github.com/openSUSE/osem/blame/master/app/views/layouts/_navigation.html.haml#L32

siddhantbajaj:
No actually I don't think this if condition has any significance here. So we can actually remove this.

Ana06:
@differentreality 
> Then do we really need this if:
https://github.com/openSUSE/osem/blame/master/app/views/layouts/_navigation.html.haml#L32

That depends on if we want to show the notification panel when there are not notifications, right?

siddhantbajaj:
> That depends on if we want to show the notification panel when there are not notifications, right?

@Ana06 Yes, but right now this if condition evaluates to true every time even if there are no notifications (unread_notifications will return empty array). So I think we should change this if condition if we don't want notification panel when there are no notifications to something like this:
`unless unread_notifications(current_user).empty?`

Ana06:
@siddhantbajaj 

> @Ana06 Yes, but right now this if condition evaluates to true every time even if there are no notifications (unread_notifications will return empty array). So I think we should change this if condition if we don't want notification panel when there are no notifications to something like this:
unless unread_notifications(current_user).empty?

Yes, or just always show the panel with 0 notifications when it is empty. Both solutions are fine for me.

Ana06:
@siddhantbajaj @differentreality what do we want to do then?

siddhantbajaj:
> just always show the panel with 0 notifications when it is empty

@Ana06 I think this is much better fix because there might be a case when there are no unread notifications but user wants to see the previous seen notifications. So I think we should always show the notification panel.

differentreality:
I think this looks good. Notifications drop down shows always, and it at least has the 'show all comments' link, right?

If so, I'd say let's prepare it for merging.

differentreality:
:+1:

PR #1429, opened 2017-04-03
use flash.now instead of flash before calling render

eug3nix:
the flash messages are used incorrectly in OSEM controllers in those cases where controller method just renders a template after setting the flash message, accordingly to the Rails official documentation, one should use flash.now in such cases:
http://guides.rubyonrails.org/action_controller_overview.html#the-flash

the code in its current state can produce a misleading flash message output on the pages where the message is not relevant to the page being displayed, here is the scenario to reproduce the issue:
1 - go to any form which may produce flash messages on template render 
2 - provoke the error (e.g. by providing invalid form data)
3 - submit the form
4 - observe the error in the re-rendered form
5 - open some other OSEM page in the different tab - observe the duplicate flash error message

this bug was discussed with @differentreality in the project IRC chat and we agreed that this should be fixed, this PR has the necessary changes.

Sincerely
Eugene/CommandPrompt Inc.

### DIFF
--- a/app/controllers/admin/campaigns_controller.rb
+++ b/app/controllers/admin/campaigns_controller.rb
@@ -15,7 +15,7 @@ module Admin
         redirect_to admin_conference_campaigns_path(conference_id: @conference.short_title),
                     notice: 'Campaign successfully created.'
       else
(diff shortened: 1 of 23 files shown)


differentreality:
:+1:

Issue #1427, opened 2017-04-03
admin/resources error if there is no current_user


differentreality:
If you try to access admin/resources#index directly, and you are not signed_in, it throws an error instead of asking you to sign in.

shlok007:
Oh! I'll work on it.

nasia-sam:
i would be interested to work on it :)

differentreality:
Go ahead @nasia-sam

PR #1434, opened 2017-04-04
Fix admin/resources inherit

nasia-sam:
Issue  #1427 admin/resources_controller.rb inherited ApplicationController instead of Admin::BaseController


ChrisBr:
@nasia-sam great :+1:

PR #1436, opened 2017-04-04
Add description to xml. fixes #1435

### LINKED ISSUE #1435: Missing description field in schedule xml
The app giggity complains about the xml returned by our osem installation. 

The problem is that the xml doesn't have a description field.

davidsf:
Add field description with the content of abstract to schedule show xml.  Fix #1435

### DIFF
--- a/app/views/schedules/show.xml.haml
+++ b/app/views/schedules/show.xml.haml
@@ -27,6 +27,7 @@
                 %subtitle= event.subtitle
                 %track= event.track.name if event.track
                 %abstract= event.abstract
+                %description= event.abstract
                 %recording
                   %license/
                   %optout=false #FIXME


ChrisBr:
Why is abstract not enough?

differentreality:
@davidsf can you please provide some feedback (there is more discussion in the issue) so that we can close this?
Also please reference the issue in your PR description.

differentreality:
@davidsf ping

davidsf:
@differentreality Adding the description field with the abstract information works well  for us. Giggity works with both fields, so I think the content of this PR should be merged.

Thanks

Ana06:
@davidsf I think it would be useful to add a description in the commit message explaining that we add this field to make this work with Giggity

differentreality:
OK let's get this in!

Issue #1450, opened 2017-04-12
Update Rubocop


Ana06:
Update Rubocop to the last version 0.48.1.

`.rubocop_todo.yml` file should be generated again after the update.

Ana06:
@agrim123 do you want to work on this? :wink:

agrim123:
@Ana06 I think I can give it a shot.

PR #1396, opened 2017-03-21
Add Haml lint

agrim123:
- Added haml_lint gem
- Added haml-lint.yml
- Changed app/views/resources/ files accordingly to reflect the haml-lint working

Addresses #1395

PR #1451, opened 2017-04-12
update rubocop to 0.48.1

agrim123:
Addresses #1450


agrim123:
@Ana06 On running `bundle update rubocop` I got the error for `nokogiri`. So I ran `bundle update nokogiri` but it downgraded the version to `1.6.8.1` and that's why hakiri is generating warnings.

differentreality:
Don't you think we should keep nokogiri at a version (>=1.7.1 according to Hakiri) for which hakiri doesn't complain, even if that means we update more gems?

agrim123:
But then I have to manually modify Gemfile.lock cause each time I run bundle update nokogiri it downgrades the gem! And without it, I am unable to update rubocop!

shlok007:
Have a look at https://github.com/openSUSE/osem/pull/1396#issuecomment-293786809

>But then I have to manually modify Gemfile.lock 'cause each time I run bundle update nokogiri it downgrades the gem! And without it, I am unable to update rubocop!

Have a look at these http://bundler.io/man/bundle-update.1.html#PATCH-LEVEL-OPTIONS  and see if any one of that helps.

Ana06:
@agrim123 the way to update a gem is:

1. Change the version in the [Gemfile](https://github.com/openSUSE/osem/blob/master/Gemfile#L202), in this case:
`gem 'rubocop', '~> 0.48.1', require: false`

2. Run `bundle install`
This way the dependencies will be automatically solved.

And again **you shouldn't manually modify the Gemfile.lock**

Ana06:
@agrim123 remember you need to generate the `.rubocop_todo.yml` file :wink:

agrim123:
@Ana06 I already generated it using
`rubocop --auto-gen-config`

.rubocop_todo.yml
```diff
@@ -323,6 +323,11 @@ Style/ConditionalAssignment:
     - 'app/helpers/application_helper.rb'
     - 'db/migrate/20140610165551_migrate_data_person_to_user.rb'
     - 'db/migrate/20140820124117_undo_wrong_migration20140801080705_add_users_to_events.rb'
+    - 'app/models/ticket_purchase.rb'
```
agrim123: done :+1:

differentreality:
Please rebase

agrim123:
@Ana06 
This is the original file that exists in `upstream` 
https://github.com/openSUSE/osem/blob/master/.rubocop_todo.yml#L199-L214

And in my second [commit](https://github.com/openSUSE/osem/pull/1451/commits/5a184c513950915a5e5c2c00d42cd7fe575b2dfa), I have regenerated the file and those deletions are because of that.
In my third [commit](https://github.com/openSUSE/osem/pull/1451/commits/16adc447cad6c65ee466e36790ad512dca0b8234),
 I added the conflict files manually.

Ana06:
@agrim123 
> In my third commit, I added the conflict files manually.

When rebasing you had to regenerate the todo file again, and those changes should be in the second commit, but hey are in the third. The only cop affected by the Rubocop bug is `ConditionalAssignment`, and you are changing more things in the third commit. Those things shouldn't be there.

PR #1452, opened 2017-04-14
Normalize openid mock accounts' credentials and add tests

### LINKED ISSUE #1365: Grooming of openids in development
* The mocks use emails with domain `email.com`, we should change that
* We should use similar name formats for the users created
* And we should also add relevant tests to prevent #1362

AEtherC0r3:
Issue:
1. The emails of the mock accounts use email.com which is neither reserved nor controlled by osem
2. The credentials of the mock accounts are arbitrary, thus they don't indicate to the user/developer the provider with which they are associated (in the web interface)
3. The available specs don't test signup for every supported provider

Solution:
1.  Change email.com to example.com
2. Change the credentials to comply with the following patterns:
    * name: 'provider user'
    * email: 'user-provider@example.com'
    * username: 'userProvider'
3. Add relevant tests and a "trap" to catch changes to the supported providers (addition, renaming, removal)

Closes #1365

### DIFF
--- a/spec/features/omniauth_spec.rb
+++ b/spec/features/omniauth_spec.rb
@@ -18,13 +18,13 @@ feature Openid do
       within('#openidlinks') do
         click_link 'omniauth-google'
       end
-      expect(flash).to eq('test-1@gmail.com signed in successfully with google')
+      expect(flash).to eq('test-1@example.com signed in successfully with google')
       expect(Openid.count).to eq(expected_count_openid)
       expect(User.count).to eq(expected_count_user)
     end
 
     scenario 'signs in an existing user' do
-      create(:user, email: 'test-participant-1@google.com')
+      create(:user, email: 'test-participant-1@example.com')
       expected_count_openid = Openid.count + 1
       expected_count_user = User.count
       visit '/accounts/sign_in'
@@ -33,7 +33,7 @@ feature Openid do
       within('#openidlinks') do
         click_link 'omniauth-google'
       end
-      expect(flash).to eq('test-participant-1@google.com signed in successfully with google')
+      expect(flash).to eq('test-participant-1@example.com signed in successfully with google')
       expect(Openid.count).to eq(expected_count_openid)
       expect(User.count).to eq(expected_count_user)
     end
@@ -51,7 +51,7 @@ feature Openid do
 
     scenario 'adds openid to existing user' do
       # Sign in user
-      user = create(:user, email: 'test-participant-1@google.com')
+      user = create(:user, email: 'test-participant-1@example.com')
       sign_in user
 
       # Add openID to current user
@@ -63,15 +63,15 @@ feature Openid do
       within('#openidlinks') do
         click_link 'omniauth-google'
       end
-      expect(flash).to eq('test-participant-1@google.com signed in successfully with google')
+      expect(flash).to eq('test-participant-1@example.com signed in successfully with google')
       expect(Openid.count).to eq(expected_count_openid)
       expect(User.count).to eq(expected_count_user)
-      expect(Openid.where(email: 'test-1@gmail.com').first.nil?).to eq(false)
+      expect(Openid.where(email: 'test-1@example.com').first.nil?).to eq(false)
     end
 
     scenario 'signs in with openID using the same email as another associated openid' do
       # Sign in user
-      create(:user, email: 'test-participant-1@google.com')
+      create(:user, email: 'test-participant-1@example.com')
       expected_count_openid = Openid.count + 1
       expected_count_user = User.count
       visit '/accounts/sign_in'
@@ -80,7 +80,7 @@ feature Openid do
       within('#openidlinks') do
         click_link 'omniauth-google'
       end
-      expect(flash).to eq('test-participant-1@google.com signed in successfully with google')
+      expect(flash).to eq('test-participant-1@example.com signed in successfully with google')
       expect(Openid.count).to eq(expected_count_openid)
       expect(User.count).to eq(expected_count_user)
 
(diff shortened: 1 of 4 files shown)


spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,43 @@
     end
   end
 
+  shared_examples 'sign up with openid provider' do |provider|
+    scenario "#{provider} has option to log in" do
+      visit '/accounts/sign_in'
+      expect(page.has_content?('or sign in using')).to be true
```
differentreality: You can use `eq` instead of `be` here as well.

spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,43 @@
...
   describe 'omniauth' do
     if User.omniauth_providers.present?
       it_behaves_like 'sign in with openid'
     end
   end
+
+  describe 'omniauth provider' do
+    providers = %w(facebook google suse github)
+
+    it 'verifies presence' do
```
differentreality: This looks more like a model test to me, rather than a feature test.

spec/support/omniauth_macros.rb
```diff
@@ -6,6 +6,10 @@ module OmniauthMacros
   Rails.application.secrets.google_secret = 'test secret google'
   Rails.application.secrets.facebook_key = 'test key facebook'
   Rails.application.secrets.facebook_secret = 'test secret facebook'
+  Rails.application.secrets.suse_key = 'test key suse'
+  Rails.application.secrets.suse_secret = 'test secret suse'
+  Rails.application.secrets.github_key = 'test key github'
+  Rails.application.secrets.github_secret = 'test secret github'
```
differentreality: Should we move these into env file?

spec/support/omniauth_macros.rb
```diff
@@ -74,4 +78,66 @@ def mock_auth_existing_user_admin
                               }
                             )
   end
+
+  def mock_auth_accounts
```
differentreality: Can we remove the previous mocks?

spec/features/omniauth_spec.rb
AEtherC0r3: If it fails then someone will, naturally, edit the 'providers' variable to make it pass.
It is designed to cause the 'sign up with openid provider' test to fail if the contributor doesn't make the appropriate changes to 'omniauth_macros.rb' too.

I must leave a comment to remind future contributors to also edit 'development.rb'.

Shall I move it to a model test and count on future contributors and reviewers to make and check that the right changes have been made to the rest of the files, instead of enforcing it via technical means? (at least for 'omniauth_macros.rb')

spec/support/omniauth_macros.rb
AEtherC0r3: There is no '.env.test' yet, if we want to add it then we can move these in there.

Are we aware of any reason why a '.env.test' file hasn't been used so far?

spec/support/omniauth_macros.rb
AEtherC0r3: With some modifications we can remove up to one of the previous mock accounts. We can't replace more than one because of the unique username each of these mock accounts has.
If we removed the usernames then we could replace more of the previous mock accounts, but then the mock accounts defined in mock_auth_accounts would not be identical to the ones in development.rb

spec/models/user_spec.rb
```diff
@@ -426,4 +426,10 @@
       expect(user.events_registrations).to eq [@events_registration1, @events_registration2]
     end
   end
+
+  describe '.omniauth_providers' do
+    it 'contains providers' do
+      expect(User.omniauth_providers).to eq [:suse, :google, :facebook, :github]
```
AEtherC0r3: This is a partial test since it doesn't take into account the 'OSEM_ICHAIN_ENABLED' environment variable

spec/support/omniauth_macros.rb
```diff
@@ -74,4 +78,67 @@ def mock_auth_existing_user_admin
                               }
                             )
   end
+
+  # These must be identical to the ones in /config/environments/development.rb
```
differentreality: Why? Let's add that to the comment as well.

spec/support/omniauth_macros.rb
```diff
@@ -74,4 +78,67 @@ def mock_auth_existing_user_admin
...
+  def mock_auth_accounts
+    OmniAuth.config.mock_auth[:facebook] =
+      OmniAuth::AuthHash.new(
+                              provider: 'facebook',
+                              uid: 'facebook-test-uid-1',
```
differentreality: This uid occurs twice in the file, is that on purpose?

spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,37 @@
...
   end
 
+  shared_examples 'sign up with openid provider' do |provider|
+    scenario "#{provider} has option to log in" do
+      visit '/accounts/sign_in'
+      expect(page.has_content?('or sign in using')).to eq true
+      expect(page.has_link?("omniauth-#{provider}")).to eq true
+    end
+
+    scenario "#{provider} signs up new user" do
```
differentreality: sign up with #{provider}

spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,37 @@
     end
   end
 
+  shared_examples 'sign up with openid provider' do |provider|
+    scenario "#{provider} has option to log in" do
```
differentreality: log in with #{provider}

spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,37 @@
...
+  end
+
   describe 'omniauth' do
     if User.omniauth_providers.present?
       it_behaves_like 'sign in with openid'
     end
   end
+
+  describe 'omniauth provider' do
+    User.omniauth_providers.each do |provider|
```
differentreality: This can be moved in the previous describe 'omniauth' and inside the if statement.

spec/features/omniauth_spec.rb
```diff
@@ -118,9 +118,37 @@
...
+      end
+      expect(flash).to eq("user-#{provider}@example.com signed in successfully with #{provider}")
+      expect(Openid.count).to eq(expected_count_openid)
+      expect(User.count).to eq(expected_count_user)
+    end
+  end
+
   describe 'omniauth' do
     if User.omniauth_providers.present?
       it_behaves_like 'sign in with openid'
```
differentreality: If the other one says 'sign up with openid provider' this one should also say 'sign in with openid provider'

differentreality:
Let's change all openid emails to example.com as well.

spec/support/omniauth_macros.rb
AEtherC0r3: No, it is coincidental

differentreality:
Nice work @AEtherC0r3 
Please fix the latest comment & squash/rebase for merging. Please keep different commits for the different things you were doing.

differentreality:
Nice, let's get it in :+1:

PR #1461, opened 2017-04-19
Don't allow not_signed_in user to view private schedule

BelieveC:
Fix bug i.e private schedule is viewable to `not_signed_in` user in the PR #1352  by fixing load_and_authorize of schedule in `schedules_controller`

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -1,9 +1,9 @@
 class SchedulesController < ApplicationController
-  load_and_authorize_resource
   protect_from_forgery with: :null_session
   before_action :respond_to_options
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
+  load_and_authorize_resource :selected_schedule, through: :program, singleton: true
 
   def show
     @rooms = @conference.venue.rooms if @conference.venue


differentreality:
The term private schedule is not very suitable. What do you mean exactly? And how can your bug be reproduced?

BelieveC:
> The term private schedule is not very suitable. What do you mean exactly?

In the Program model, we have an attribute `schedule_public`. When It is false we don't a

BelieveC:
> The term private schedule is not very suitable. What do you mean exactly?

In the Program model, we have an attribute `schedule_public`. When It is false we don't allow not_signed_in user to see it. I'm referring it as private schedule.

> And how can your bug be reproduced?

Actually in the PR #1352, when the schedule is not public we didn't show schedule link on the app's root page. Upto this was fine. But when I tried entering the URL to schedule, it allows to see schedule because `load_and_authorize_resource` is not able to load the resource as we don't pass `param[:id]` in the path to see the schedule. Selected schedule is loaded using program's `selected_schedule` attribute. To fix described bug, I have made above changes.

differentreality:
I will re-write this as following, because it isn't immediately clear what you mean:
Users are allowed to view the schedule even if the schedule is not marked as public in program.

The explanation you offered in the previous comment as to why this is happening is absolutely correct.

differentreality:
#1462 is relevant so please make everything work properly in 1 single PR. Keep whichever PR you want, and close the other one.

BelieveC:
I will close this one and add a commit to #1462 fixing above bug.

PR #1462, opened 2017-04-19
Refactor authorization for schedule and event action of Conferences Controller

### LINKED ISSUE #1457: Refactor authorization for schedule
There is no longer a `schedule` action for conferences controller
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L37

Program doesn't really need to be authorized for schedule, we only need to load it, and only authorize schedule resource
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/schedules_controller.rb#L6

BelieveC:
Remove can ability for `schedule` and `events` action in `ability.rb` as both the actions are no longer present in `conferences_controller`. Replace `load_and_authorize_resource` with `load_resource` for `program` in the `schedules_controller`.

Closes #1457

### DIFF
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -3,7 +3,7 @@ module Admin
     # By authorizing 'conference' resource, we can ensure there will be no unauthorized access to
     # the schedule of a conference, which should not be accessed in the first place
     load_and_authorize_resource :conference, find_by: :short_title
-    load_and_authorize_resource :program, through: :conference, singleton: true
+    load_resource :program, through: :conference, singleton: true
     load_and_authorize_resource :schedule, through: :program
     load_resource :event_schedules, through: :schedule
     load_resource :selected_schedule, through: :program, singleton: true
--- a/app/controllers/admin/schedules_controller.rb
+++ b/app/controllers/admin/schedules_controller.rb
@@ -3,7 +3,7 @@ module Admin
     # By authorizing 'conference' resource, we can ensure there will be no unauthorized access to
     # the schedule of a conference, which should not be accessed in the first place
     load_and_authorize_resource :conference, find_by: :short_title
-    load_resource :program, through: :conference, singleton: true
+    load_and_authorize_resource :program, through: :conference, singleton: true
     load_and_authorize_resource :schedule, through: :program
     load_resource :event_schedules, through: :schedule
     load_resource :selected_schedule, through: :program, singleton: true
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -3,15 +3,12 @@ class SchedulesController < ApplicationController
   before_action :respond_to_options
   load_resource :conference, find_by: :short_title
   load_resource :program, through: :conference, singleton: true, except: :index
+  before_action :presence_of_selected_schedule
   load_and_authorize_resource :selected_schedule, through: :program, singleton: true
 
   def show
     @rooms = @conference.venue.rooms if @conference.venue
     schedules = @program.selected_event_schedules
-    unless schedules
-      redirect_to events_conference_schedule_path(@conference.short_title)
-    end
-
     @events_xml = schedules.map(&:event).group_by{ |event| event.time.to_date } if schedules
     @dates = @conference.start_date..@conference.end_date
     @step_minutes = @program.schedule_interval.minutes
@@ -48,5 +45,10 @@ class SchedulesController < ApplicationController
     respond_to do |format|
       format.html { head :ok }
     end if request.options?
+  end
+
+  def presence_of_selected_schedule
+    return if @program.selected_event_schedules
+    redirect_to root_path, notice: 'Program is yet to be scheduled.'
   end
 end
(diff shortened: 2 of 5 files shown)


BelieveC:
@differentreality 

As you suggested to replace `load_and_authorize_resource` with `load_resource` for program in `Admin::SchedulesController`, this breaks ability dependencies for schedules_controller as follow:

- We allow `:show` action for `conference` as well as `schedule` for `not_signed_in` user under minor constraints of public_splashpage and public_schedule. But what happens next is troublesome. These abilities get inherited to `signed_in` user then to user  `signed_in_with_roles`. User `signed_in_with_roles` have access to `Admin` namespace. Action `admin::schedules_controller#show` become accessible to each of the user `signed_in_with_roles`. 
- `load_and_authorize_resource` for the program in the `admin:schedules_controller` helps us out as it stops above behavior as we are not authorizing `programs_controller#show` to any of ancestors in the hierarchy. If we replace it with `load_resource` this breaks above ability dependency and gives undesired behavior.

BelieveC:
@differentreality 

Above [comment](https://github.com/openSUSE/osem/pull/1462#issuecomment-296081613l) is a workaround to actual problem which I think needs to resolved separately.
The problem is as follows:

- If we remove 'load_and_authorize_resource' for program. It breaks the desired abilities as follows:
      Expected Behavior:  Anyone whether signed or unsigned should have an access to public schedule and public splash page.
     Undesired Behavior: As schedule uses `show` action in both `schedules_controller` and `admin::schedules_controller` we, in turn, give access to both show action to users `signed_in_with_roles`.

Ref for possible solution: [Link](https://github.com/ryanb/cancan/wiki/Admin-Namespace).

differentreality:
As we discussed in IRC yesterday, I would expect you to list the cases when the schedule can be viewed, and when it can't. 

Currently you are only testing this for the not_signed_in user, but what if the user is signed in? Can the user then see a schedule which is not public? What if the user has the role organizer or cfp for  the conference? What if the user has the role info_desk?

BelieveC:
On it :+1:

BelieveC:
@differentreality Access to view schedules for different roles i.e Example [http://demo.osem.io/conferences/osemdemo/schedule](http://demo.osem.io/conferences/osemdemo/schedule)

| User Type/Role   |      Public Schedule      | Not Public Schedule |
|----------|:-------------:|------:|
| not_signed_in | can | cannot |
| Admin | can | can |
| Organizer| can | can |
| CFP | can | can |
| Info Desk | can | cannot |
| Volunteers Coordinator | can | cannot |

differentreality:
Table looks nice! 

You must try the code you wrote locally to check if it works, before sending a PR, and before pushing new changes. Please do so, and fix the error that occurs, as this is unfortunately not not working at the moment.

BelieveC:
Yeah! Will fix the errors :smile:

BelieveC:
> You must try the code you wrote locally to check if it works, before sending a PR, and before pushing new changes.

Demo conference on my local system has initialized schedule so above error didn't appear when I tested my code before pushing the changes. Missed the corner case. Will make sure to thoroughly test code before pushing changes in PR or creating a PR.

differentreality:
Can you please briefly list the cases you are trying to solve/avoid?

BelieveC:
# Cases Covered

1. Don't allow not signed in user to view private schedule
2. Refactor authorization for schedule and event action of Conferences Controller

# Avoided 

1. Removal of `load_and_authroize_resource` in `schedules_controller` to avoid access to `admin::schedules_controller' show action

differentreality:
When/How do we access the events action of the controller with the changes you have made?

BelieveC:
We access `schedules_controller`'s show action when we try to look for a public schedule from conference index page or start page of osem. It provides link to schedule if it is public and scheduled.

differentreality:
I asked about action events, not show. Can you please clarify about that?

BelieveC:
@differentreality, Sorry I read it incorrectly. You access `events` action when a schedule is public and selected for not signed in user. Different roles can access same as per updated table in this [comment](https://github.com/openSUSE/osem/pull/1462#issuecomment-296828834).

differentreality:
@BelieveC I am not sure if you understand the question, or I am not seeing what you mean in that table. I cannot extract that information from the table. 
The controller has an action named `events`. While our current master access that action when the schedule is public, but there is no selected schedule, in your implementation I am not able to do this any more. So are we using events action after your changes? And if yes, how? Please show me in https://osem-demo-pr-1462.herokuapp.com/ once it is deployed.

BelieveC:
@differentreality 

>  So are we using events action after your changes? And if yes, how?

We are using events action for public and selected schedule as per current deploy. 
Example link: https://osem-demo-pr-1462.herokuapp.com/conferences/osemdemo/schedule/events

differentreality:
@BelieveC Do you plan on working on this more? 

As I said before, right now I never get redirected to schedules#events, I can only go to show action. 
When there is no schedule, and there are confirmed events, we should be seeing the events action. That's why we have it.

differentreality:
I'll close this due to inactivity

PR #1465, opened 2017-04-22
Add helper function for speaker selection

### LINKED ISSUE #1455: Move users collection to a helper
Comment https://github.com/openSUSE/osem/pull/1278#issuecomment-283924951 in #1278 has not been implemented yet.

The code in the view should look like this https://github.com/openSUSE/osem/pull/1278#issuecomment-287769708

AEtherC0r3:
The line '@users = User.all.order(:name)' was replicated a lot of times in EventsController and ProposalsController. So, it has been removed and a helper function was added ('speaker_selector_input') that generates the field where the @users variable was used.
Also, the query has been made more specific by keeping only the required fields and excluding disabled users.

The 'select' method was used instead of 'pluck' because the latter returns an array of arrays, therefore it would make the code less readable, since array indeces would have to be used instead of the more human friendly '.name, .email, .id'

Fix #1455

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -602,4 +602,11 @@ module ApplicationHelper
   def speaker_links(event)
     event.speakers.map{ |speaker| link_to speaker.name, admin_user_path(speaker) }.join(', ').html_safe
   end
+
+  def speaker_selector_input(form)
+    users = User.active.pluck(:id, :name, :username, :email).map { |user| [user[0], user[1].blank? ? user[2] : user[1], user[2], user[3]] }.sort_by { |user| user[1].downcase }
+    form.input :speakers, as: :select,
+                          collection: options_for_select(users.map {|user| ["#{user[1]} (#{user[2]}) #{user[3]}", user[0]]}, @event.speakers.map(&:id)),
+                          include_blank: false, label: 'Speakers', input_html: { class: 'select-help-toggle', multiple: 'true' }
+  end
 end
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -29,6 +29,7 @@ FactoryGirl.define do
       Quisque cursus facilisis consequat. Etiam volutpat ligula turpis, at
       gravida.
     EOS
+    is_disabled false
 
     after(:create) do |user|
       user.is_admin = false
@@ -43,6 +44,10 @@ FactoryGirl.define do
         user.is_admin = true
         user.save!
       end
+    end
+
+    trait :disabled do
+      is_disabled true
     end
   end
 
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -62,7 +62,6 @@ module Admin
       @comments = @event.root_comments
       @comment_count = @event.comment_threads.count
       @user = @event.submitter
-      @users = User.all.order(:name)
       @url = admin_conference_program_event_path(@conference.short_title, @event)
       @languages = @program.languages_list
     end
@@ -80,7 +79,6 @@ module Admin
     end
 
     def update
-      @users = User.all.order(:name)
       @languages = @program.languages_list
       if @event.update_attributes(event_params)
 
@@ -99,7 +97,6 @@ module Admin
 
     def create
       @url = admin_conference_program_events_path(@conference.short_title, @event)
-      @users = User.all.order(:name)
       @languages = @program.languages_list
       @event.submitter = current_user
 
@@ -115,7 +112,6 @@ module Admin
     def new
       @url = admin_conference_program_events_path(@conference.short_title, @event)
       @languages = @program.languages_list
-      @users = User.all.order(:name)
     end
 
     def accept
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -25,7 +25,6 @@ class ProposalsController < ApplicationController
 
   def edit
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
-    @users = User.all.order(:name)
     @languages = @program.languages_list
   end
 
@@ -61,7 +60,6 @@ class ProposalsController < ApplicationController
 
   def update
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
-    @users = User.all.order(:name)
 
     if @event.update(event_params)
       redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -53,6 +53,7 @@ class User < ActiveRecord::Base
   accepts_nested_attributes_for :roles
 
   scope :admin, -> { where(is_admin: true) }
+  scope :active, -> { where(is_disabled: false) }
 
   validates :email, presence: true
 
--- a/app/views/proposals/_proposal_form.html.haml
+++ b/app/views/proposals/_proposal_form.html.haml
@@ -4,9 +4,7 @@
 
     = f.input :subtitle, as: :string
 
-    = f.input :speakers, as: :select,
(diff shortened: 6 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -597,4 +597,11 @@ def concurrent_events(event)
     end
     concurrent_events
   end
+
+  def speaker_selector_input(form)
+    users = User.select(:id, :name, :email).where(is_disabled: false).order(:name)
```
differentreality: This does not show the records ordered. 
We order by name, but some of the records do not have a name (so name is nil). When we do `users.map` in the collection and call `user.name` it either shows the name or the username of the user (due to https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L82)

As such the first records showing up could be a username that begins with z, which doesn't make a lot of sense when you want your records to be ordered.

If you run this in rails console you will see an error caused by this.

lagartoflojo: Also, because `User#name` looks up `username` when `name` is `nil`, you must also `select :username`.

lagartoflojo: Using `.select` will instantiate `User` objects. When we have a lot of users in the DB, that will eat up a lot of memory. Use `.pluck` instead. That will give you an array of arrays, which is lighter.

app/helpers/application_helper.rb
```diff
@@ -597,4 +597,11 @@ def concurrent_events(event)
...
+    form.input :speakers, as: :select,
+                          collection: options_for_select(users.map {|user| ["#{user.name} (#{user.email})", user.id]}, @event.speakers.map(&:id)),
```
lagartoflojo: Not including `username` in the options for select means that the auto-complete widget won't be able to search by username. Is that something we might want?

AEtherC0r3: A lot of people prefer to refer to other by their username, so i will add that as well

app/helpers/application_helper.rb
AEtherC0r3: Do we really want to sacrifice code readability for a reduced memory footprint? (by using .pluck)

lagartoflojo: We could possibly be loading thousands of users into memory, so unfortunately I would say yes.

app/helpers/application_helper.rb
```diff
@@ -597,4 +597,11 @@ def concurrent_events(event)
     end
     concurrent_events
   end
+
+  def speaker_selector_input(form)
+    users = User.select(:id, :name, :username, :email).where(is_disabled: false).sort_by {|user| user.name}
```
lagartoflojo: Could you move the `.where(is_disabled: false)` into a named scope?
Then this line can read: `User.active.select(...` which would be nicer.

AEtherC0r3: yup

app/helpers/application_helper.rb
AEtherC0r3: ok then

app/helpers/application_helper.rb
```diff
@@ -599,9 +599,9 @@ def concurrent_events(event)
   end
 
   def speaker_selector_input(form)
-    users = User.active.select(:id, :name, :username, :email).sort_by {|user| user.name}
+    users = User.active.pluck(:id, :name, :username, :email).map { |user| [user[0], user[1].blank? ? user[2] : user[1], user[2], user[3]] }.sort_by { |user| user[1] }
```
differentreality: Are you sure` sort_by { |user| user[1] }` works when the user.name is nil?

AEtherC0r3: user[1] always has a value because of the `user[1].blank? ? user[2] : user[1]` in .map, which is the equivalent of User#name (https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L83)

AEtherC0r3:
The cop that failed on travis is in .rubocop_todo.yml (https://github.com/openSUSE/osem/blob/master/.rubocop_todo.yml#L85). Any idea why that happened?

differentreality:
Exactly what it says:
`app/helpers/application_helper.rb:1:1: C: Metrics/ModuleLength: Module has too many lines. [478/472]`

You added code in application_helper file, so it exceeded the maximum lines we have set. This is being addressed in #1335 so I wouldn't take any action for the failure at the moment.

AEtherC0r3:
I referenced that the cop is in .rubocop_todo.yml because travis uses the .rubocop.yml file. Therefore, from what I understand that cop shouldn't have run at all. (BTW in vagrant rubocop reports no offences) Am I missing something?

differentreality:
Sorry I missed that, this was changed in https://github.com/openSUSE/osem/pull/1451, it shouldn't have. The todo file should be excluding the application_helper. Worth opening an issue for.

spec/factories/users.rb
```diff
@@ -44,6 +45,13 @@
         user.save!
       end
     end
+
+    factory :disabled_user do
+      after(:create) do |user|
+        user.is_disabled = true
+        user.save!
+      end
+    end
```
lagartoflojo: instead of a factory, you can use a trait here (not tested):

```ruby
trait :disabled do
  is_disabled true
end
```

And then in your test: `let(:disabled_user) { create(:user, :disabled) }`

I'm not sure why the `after :create` block is needed? Can't the `is_disabled` attribute be set like in a normal factory, as I show above?

spec/factories/users.rb
```diff
@@ -32,6 +32,7 @@
 
     after(:create) do |user|
       user.is_admin = false
+      user.is_disabled = false
```
lagartoflojo: Why are these not "normal" attributes as in lines 17-31?

spec/factories/users.rb
```diff
@@ -32,6 +32,7 @@
...
       # save with bang cause we want change in DB and not just in object instance
       user.save!
```
lagartoflojo: > save with bang cause we want change in DB and not just in object instance

that's not true...
from the [Rails docs](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/Persistence.html#method-i-save):

> By default, `save` always run validations. If any of them fail the action is cancelled and **save returns false**.

and:

> With `save!` validations always run. If any of them fail **ActiveRecord::RecordInvalid gets raised.**

(emphasis mine)

lagartoflojo: Ok, maybe I misread the comment. It's saying that we want to persist the change in attribute. The fact that it's "with bang" is irrelevant.

Still raises the question of why we don't set these attributes like normal in a factory (ie (`is_disabled true`)

spec/factories/users.rb
AEtherC0r3: Because of https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L18 and https://github.com/openSUSE/osem/blob/master/app/models/user.rb#L200-L204 we need the after block for the is_admin attribute, the is_disabled attribute can be set normally

spec/factories/users.rb
AEtherC0r3: sorry about that, the after block it isn't necessary

app/helpers/application_helper.rb
```diff
@@ -597,4 +597,11 @@ def concurrent_events(event)
     end
     concurrent_events
   end
+
+  def speaker_selector_input(form)
+    users = User.active.pluck(:id, :name, :username, :email).map { |user| [user[0], user[1].blank? ? user[2] : user[1], user[2], user[3]] }.sort_by { |user| user[1] }
```
differentreality: At least for me, this is very difficult to read. @lagartoflojo do you think we should keep it like this?

differentreality: Also this sorts capitalized words first, and non-capitalized words later, which does not produce a truly sorted list

differentreality:
Ping

differentreality:
Looks nice, let's squash it!

differentreality:
:+1:

PR #1471, opened 2017-04-26
Add Style/WordArray at rubocop.yml #1463

### LINKED ISSUE #1463: Enable Style/WordArray Rubocop cop
Enable Style/WordArray Rubocop cop: https://rubocop.readthedocs.io/en/latest/cops_style/#stylewordarray

nasia-sam:
Add Style/WordArray at rubocop.yml 
Resolves #1463

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -104,6 +104,10 @@ Style/TrailingBlankLines:
 Style/TrailingWhitespace:
   Enabled: true
 
+# Check for array literals made up of word-like strings, that are not using the %w() syntax
+Style/WordArray:
+  Enabled: true
+
 # This cop checks for numeric comparisons that can be replaced by a predicate method.
 Style/ZeroLengthPredicate:
   Enabled: true
--- a/app/models/event_user.rb
+++ b/app/models/event_user.rb
@@ -1,6 +1,6 @@
 class EventUser < ActiveRecord::Base
   # TODO Do we need these roles?
-  ROLES = [['Speaker', 'speaker'], ['Submitter', 'submitter'], ['Moderator', 'moderator']]
+  ROLES = [%w[Speaker speaker], %w[Submitter submitter], %w[Moderator moderator]]
 
   belongs_to :event
   belongs_to :user
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -928,11 +928,3 @@ Style/UnneededInterpolation:
 Style/VariableNumber:
   Exclude:
     - 'spec/models/ticket_purchase_spec.rb'
-
-# Offense count: 6
-# Cop supports --auto-correct.
-# Configuration parameters: SupportedStyles, WordRegex.
-# SupportedStyles: percent, brackets
-Style/WordArray:
-  EnforcedStyle: percent
-  MinSize: 3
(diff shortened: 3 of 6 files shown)


.rubocop.yml
```diff
@@ -108,6 +108,11 @@ Style/TrailingWhitespace:
 Style/ZeroLengthPredicate:
   Enabled: true
 
+# Check for array literals made up of word-like strings, that are not using the %w() syntax.
+Style/VariableNumber:
```
Ana06: You wanted to enable WordArray, this is other cop.

app/models/event_user.rb
```diff
@@ -1,6 +1,6 @@
 class EventUser < ActiveRecord::Base
   # TODO Do we need these roles?
-  ROLES = [['Speaker', 'speaker'], ['Submitter', 'submitter'], ['Moderator', 'moderator']]
+  ROLES = [%w[Speaker speaker], %w[Submitter submitter], %w[Moderator moderator]]
```
Ana06: Isn't this with `( )` instead `[ ]`? :thinking:

differentreality: Yeah, it is confusing for me too. I think we should not enable this cop at all, and let people write things either way.

nasia-sam: I followed the example of https://rubocop.readthedocs.io/en/latest/cops_style/#stylewordarray -I'm fixing it now

nasia-sam: offenses

differentreality: @Ana06 what do you think? Shall we drop this cop?

Ana06: @differentreality why do you want to drop it?

What I meant with my commit is that is `%w()` and not `%w[]`

But even if you don't like the `%w()` syntax we can also enable the cop to enforce using an array with the strings. I would prefer to use `%w()`, because it is the default, but the other options is also fine. Enabling it we will ensure that we are consistent and we always use the same syntax.

https://rubocop.readthedocs.io/en/latest/cops_style/#stylewordarray

differentreality: In the documentation of the cop I didn't find an option to enable %w(), can you also check please? 
That said I think that we can be flexible about this,  and use whichever format we prefer each time.

Ana06: @differentreality `%w()` is the default for this cop. Or I don't understand what you mean. :thinking:

differentreality:
@nasia-sam please rebase this, so that it can be reviewed

.rubocop.yml
```diff
@@ -108,6 +108,14 @@ Style/TrailingWhitespace:
 Style/ZeroLengthPredicate:
   Enabled: true
 
+
+# Check for array literals made up of word-like strings, that are not using the %w() syntax
+Style/WordArray:
+  EnforcedStyle: percent
+  MinSize: 3
```
Ana06: Why setting `MinSize` to 3? Was before like that, but we can change it now to 0, or remove it, as 0 is the default. :wink:

.rubocop.yml
```diff
@@ -108,6 +108,14 @@ Style/TrailingWhitespace:
 Style/ZeroLengthPredicate:
   Enabled: true
 
+
+# Check for array literals made up of word-like strings, that are not using the %w() syntax
+Style/WordArray:
+  EnforcedStyle: percent
```
Ana06: this is the default, so you don't need it

Ana06:
Merged!! :+1:

Issue #829, opened 2016-03-05
Allow confirmed speakers to register even after registration period is over


sonalkr132:
Taken from [this comment](https://github.com/openSUSE/osem/pull/795#issuecomment-190727264).

> We should probably provision for confirmed speakers to be able to register even after the registration period (they will be there anyway).
> All that might collide with registration_limit, we should address it.
> @differentreality 

We need to be more explicit with authorization of registration, `new` and `create` actions should be allowed if user is confirmed speaker.

AEtherC0r3:
I'm going to work on this.
About the registration_limit, do we want to take into account the speakers of new, unconfirmed and confirmed events in order not to exceed it or just bypass the registration_limit for speakers of confirmed events in case it has been reached or exceeded

bear454:
This creates other problems, when call for papers _precedes_ registration.

How about we allow registration by speakers _anytime after registration starts_ ?

namangupta01:
@bear454 Yes you are right! Confirmed speaker is allowed to register when registration period is set. Done in pr #1629

PR #1476, opened 2017-05-04
Allow confirmed speakers to register even when the registration period is closed

AEtherC0r3:
Speakers of confirmed events should be able to register regardless of the registration period and the registration limit.

This pr:
* Gives to confirmed speakers the ability to register even when normal users can't
* Provisions for registrations of speakers, so as not to exceed the registration limit

Related issues: #829, #802

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -594,8 +594,15 @@ class Conference < ActiveRecord::Base
     (email_settings.conference_registration_dates_updated_subject.present? && email_settings.conference_registration_dates_updated_body.present?)
   end
 
+  ##
+  # Checks if the registration limit has been exceeded
+  # Additionally, it takes into account the confirmed speakers that haven't registered yet
+  #
+  # ====Returns
+  # * +True+ -> If the registration limit has been reached or exceeded
+  # * +False+ -> If the registration limit hasn't been exceeded
   def registration_limit_exceeded?
-    registration_limit > 0 && registrations.count >= registration_limit
+    registration_limit > 0 && registrations.count + program.speakers.confirmed.count - program.speakers.confirmed.registered(program.conference).count >= registration_limit
   end
 
   # Returns an hexadecimal color given a collection. The returned color changed
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -46,6 +46,10 @@ class Program < ActiveRecord::Base
     def confirmed
       joins(:events).where(events: { state: :confirmed })
     end
+
+    def registered(conference)
+      joins(:registrations).where('registrations.conference_id = ?', conference.id)
+    end
   end
 
   accepts_nested_attributes_for :event_types, allow_destroy: true
--- a/app/views/admin/conferences/edit.html.haml
+++ b/app/views/admin/conferences/edit.html.haml
@@ -24,5 +24,5 @@
         = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: 'Registrations' do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). You currently have ' + pluralize(@conference.registrations.count, 'registration')
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events (they will still be able to register even if it has been reached). You currently have ' + pluralize(@conference.registrations.count, 'registration')
       = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}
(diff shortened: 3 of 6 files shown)


app/models/conference.rb
```diff
@@ -595,7 +614,20 @@ def notify_on_registration_dates_changed?
...
   def registration_limit_exceeded?
-    registration_limit > 0 && registrations.count >= registration_limit
+    registration_limit > 0 && registrations.count + unregistered_speakers >= registration_limit
+  end
+
+  # Returns the number of speakers of new, unconfirmed and confirmed events that are not registered
+  def unregistered_speakers
+    pending_speakers = []
+    program.events.each do |event|
+      if %w(new unconfirmed confirmed).include? event.state
```
differentreality: This means we will restrict the number of registrations, based on submitted proposals that may or may not happen. I don't agree with that.

When implementing features, it is best that we first agree on what we are trying to achieve / how do we want things to work.

app/models/conference.rb
```diff
@@ -595,7 +595,16 @@ def notify_on_registration_dates_changed?
   end
 
   def registration_limit_exceeded?
-    registration_limit > 0 && registrations.count >= registration_limit
+    registration_limit > 0 && registrations.count + unregistered_speakers >= registration_limit
+  end
+
+  # Returns the number of speakers of new, unconfirmed and confirmed events that are not registered
```
differentreality: Let's fix this comment as well and call the function unregistered_confirmed_speakers

app/views/conferences/_conference_details.html.haml
```diff
@@ -27,8 +27,8 @@
               - if conference.user_registered?(current_user)
                 = link_to "My Registration", conference_conference_registration_path(conference.short_title), class: 'btn btn-default'
               - else
-                = link_to "Register", new_conference_conference_registration_path(conference.short_title), class: "btn btn-default", disabled: conference.registration_limit_exceeded?
-                - if conference.registration_limit_exceeded?
+                = link_to "Register", new_conference_conference_registration_path(conference.short_title), class: "btn btn-default", disabled: cannot?(:new, Registration.new(conference_id: conference.id))
+                - if cannot?(:new, Registration.new(conference_id: conference.id)) && conference.registration_limit_exceeded?
```
differentreality: Why do we still need `&& conference.registration_limit_exceeded?` ?

app/views/admin/conferences/edit.html.haml
```diff
@@ -24,5 +24,5 @@
         = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}
         = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: 'Registrations' do
-        = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). You currently have ' + pluralize(@conference.registrations.count, 'registration')
+        = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events. You currently have ' + pluralize(@conference.registrations.count, 'registration')
```
differentreality: Let's clarify what we mean:

Please note that the registration limit doesn\'t apply to speakers of confirmed events **, and they will still be able to register.**

app/views/conferences/_conference_details.html.haml
AEtherC0r3: Because we want the message "Sorry, no places left" to appear only when the reason the user doesn't have the ability to register is that the registration limit had been exceeded

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -23,12 +23,33 @@
     end
   end
 
+  shared_examples 'create registration' do |user, ichain|
```
differentreality: This set of tests is more about 'cannot access #new action', as you are not testing the creation of a registration, but whether or not a user can access the new action for a registration. So I would change the description of the shared examples to reflect what they are doing.

spec/controllers/conference_registration_controller_spec.rb
```diff
@@ -345,18 +398,7 @@
...
-            end
-
-            it 'user variable exists' do
-              expect(assigns(:user)).not_to be_nil
-            end
-
-            it 'renders the new template' do
-              expect(response).to render_template('new')
-            end
+            it_behaves_like 'create registration', nil, false
```
differentreality: 'false'

app/models/conference.rb
```diff
@@ -594,8 +594,15 @@ def notify_on_registration_dates_changed?
...
+  ##
+  # Checks if the registration limit has been exceeded
+  # Additionally, it takes into account the confirmed speakers that haven't registered yet
+  #
+  # ====Returns
+  # * +True+ -> If the registration limit has been reached or exceeded
+  # * +False+ -> If the registration limit hasn't been exceede
   def registration_limit_exceeded?
-    registration_limit > 0 && registrations.count >= registration_limit
+    registration_limit > 0 && registrations.count + program.speakers.confirmed - program.speakers.confirmed.registered(program.conference) >= registration_limit
```
differentreality: You forgot the .count part

app/models/program.rb
```diff
@@ -46,6 +46,9 @@ def highlights
     def confirmed
       joins(:events).where(events: { state: :confirmed })
     end
+    def registered(conference)
```
differentreality: Empty line above

differentreality:
Please fix the comments/failing tests and rebase/squash for merge!

differentreality:
:+1:

PR #1481, opened 2017-05-10
Changed lodging cards to same height w.r.t. to its row.

### LINKED ISSUE #1456: Make each Lodging Card in the Splashpage of equal height w.r.t to its row
Lodging Cards in the Splashpage have different height according to its content.

screenshot from 2017-04-14 19-14-26

We can improve user experience by making Height of each card in the single row of the same height as anyway we are starting next lodging card on next row.

This can be easily done using Flexbox. Ref: [Link](https://clearleft.com/posts/270).

divyanshumehta:
Fixes #1456 . The lodging cards are now of the same height w.r.t. to its row
screenshot from 2017-05-10 16-17-19

### DIFF
--- a/app/views/conferences/_lodging.html.haml
+++ b/app/views/conferences/_lodging.html.haml
@@ -12,22 +12,26 @@
           = @conference.venue.city
       %p.lead
         We recommend these affordable lodging accommodations for your visit.
-  - @conference.lodgings.each_slice(3) do |slice|
-    .row.row-centered
-      - slice.each do |lodging|
-        .col-md-4.col-sm-4.ticket.col-centered.col-top
-          .thumbnail
-            - unless lodging.picture?
-              %p.text-center
-                %i.fa.fa-home.fa-5x
+
+  .row.row-centered{ style:"display: flex; flex-wrap: wrap" }
+    - @conference.lodgings.each do |lodging|
+      .col-md-4.col-sm-4.col-centered.col-top{ style:"display:flex;" }
+        .thumbnail
+          - if lodging.picture?
+            -if lodging.website_link.present?
+              = link_to(lodging.website_link, class: 'thumbnail') do
+                = image_tag lodging.picture.large.url, class: 'img-responsive img-lodging'
             - else
+              = image_tag lodging.picture.large.url, class: 'img-responsive img-lodging'
+          - else
+            %p.text-center
               -if lodging.website_link.present?
                 = link_to(lodging.website_link, class: 'thumbnail') do
-                  = image_tag lodging.picture.large.url, class: 'img-responsive img-lodging'
+                  %i.fa.fa-home.fa-5x
               - else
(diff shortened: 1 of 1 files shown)


divyanshumehta:
The CI test fails due to the error in files `app/helpers/application_helper.rb` and `spec/controllers/admin/users_controller_spec.rb`. The failures are in rubocop within these two files. They have not been caused by the commits made by me.

differentreality:
If we are going to do this, I think we should also align the 'cards' properly, so that they have the same width, and it still looks nice when there is more than 1 line of cards.

divyanshumehta:
@differentreality ok got it

differentreality:
@divyanshumehta ping are you going to finish this?

divyanshumehta:
yes actually i had lost my laptop and could not work for some time, going to finish this soon

divyanshumehta:
How it this layout
screenshot from 2017-07-24 13-37-52

app/views/conferences/_lodging.html.haml
```diff
@@ -12,22 +12,22 @@
...
-              -if lodging.website_link.present?
-                = link_to(lodging.website_link, class: 'thumbnail') do
-                  = image_tag lodging.picture.large.url, class: 'img-responsive img-lodging'
-              - else
+
+  .row.row-centered{ style:"display: flex; flex-wrap: wrap" }
+    - @conference.lodgings.each do |lodging|
+      .col-md-4.col-sm-4.ticket.col-centered.col-top{ style:"display:flex;" }
+        .thumbnail
+          - unless lodging.picture?
```
differentreality: Don't use unless-else, use if-else instead

divyanshumehta: OK :+1:

divyanshumehta:
@differentreality how does this look the links are working even if picture is not there
screenshot from 2017-08-12 00-49-06

app/views/conferences/_lodging.html.haml
```diff
@@ -12,22 +12,26 @@
...
-            - unless lodging.picture?
-              %p.text-center
-                %i.fa.fa-home.fa-5x
-            - else
+
+  .row.row-centered{ style:"display: flex; flex-wrap: wrap" }
+    - @conference.lodgings.each do |lodging|
+      .col-md-4.col-sm-4.col-centered.col-top{ style:"display:flex;" }
+        .thumbnail
+          - if not lodging.picture?
```
differentreality: Here you should use ! instead of not, but I don't understand why your workflow is
`if not sth ... else...` instead of 
`if sth ... else ...`

divyanshumehta: ok will make the changes

divyanshumehta:
All done i guess

differentreality:
:tada: Yes @divyanshumehta nice :+1:

divyanshumehta:
Happy to help :smile:

Issue #1470, opened 2017-04-26
.rubocop_todo.yml should exclude the application helper


AEtherC0r3:
In #1451 the line that excludes the application helper was removed (https://github.com/openSUSE/osem/pull/1451/files#diff-74423563ec7c74f4cb7f74689b49e7c1L45)

This change causes rubocop on travis to fail with `app/helpers/application_helper.rb:1:1: C: Metrics/ModuleLength: Module has too many lines. [478/472]`

Ana06:
Rubocop in master is working, see the last build: https://travis-ci.org/openSUSE/osem/builds/225553409

In #1451 the `todo` file was generated again after update Rubocop. Application helper got removed but the number of lines increased.

AEtherC0r3:
The number 472 was deliberately chosen in order for the rule to cop without the application_helper excluded. This implies that if someone adds even a single line of code then the cop will fail. As far as I know the purpose of rubocop is to enforce best coding practices. Therefore, I don't think that changing the rules to fit our case is the "proper" thing to do. So I think we should leave the exclusion of application helper file in rubocop_todo.yml and revert the maximum module length to 256.

differentreality:
#1335 will solve this. Indeed the file should have stayed excluded in rubocop.

PR #1483, opened 2017-05-10
Fix Travis-Ci build error

BelieveC:
Rubocop offenses lead to failure of Travis-ci build. Offenses files are added as excluded files to respective offense in rubocop_todo.yml

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,7 +14,7 @@ Bundler/OrderedGems:
   Exclude:
     - 'Gemfile'
 
-# Offense count: 28
+# Offense count: 29
 Lint/AmbiguousBlockAssociation:
   Exclude:
     - 'app/models/comment.rb'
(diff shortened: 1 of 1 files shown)


.rubocop_todo.yml
```diff
@@ -80,10 +81,12 @@ Metrics/LineLength:
 Metrics/MethodLength:
   Max: 56
 
-# Offense count: 1
+# Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
   Max: 472
```
Ana06: I think it would be better to increase this to 475, than adding an exclude

BelieveC: Alright!

BelieveC: Done :+1:

differentreality: @Ana06 but then we just propagate the same problem as https://github.com/openSUSE/osem/issues/1470#issuecomment-298690605

Ana06: ok @differentreality , then @BelieveC please change it again. And sorry for the mess :confounded:

BelieveC: Sure! I will make the changes :+1:

BelieveC: Done!

differentreality:
:+1:

PR #1469, opened 2017-04-25
Add Haml lint

agrim123:
- Added haml_lint gem
- Added haml-lint.yml

In continuation with #1396 
Addresses #1395

PR #1486, opened 2017-05-11
Fix haml-lint error

Ana06:
The error was introduced when merging https://github.com/openSUSE/osem/pull/1469, as a offense was introduced since the tests there were run and the PR got merged. :bowtie:


Ana06:
I'll merge it, as it is a simple haml-lint offense and it is breaking the test suite :wink:

PR #1487, opened 2017-05-15
Change too bright color to matte

SergioAtSUSE:
In the screenshot cannot be seen very well, but in my screen the color makes the label unreadable

image

image


Ana06:
@SergioAtSUSE I agree it is difficult to read. I like the new green color, but I don't like the color you chose for _Medium_ as it is like a dirty yellow. What about an orange? And please add an screenshot, as you are changing two colors and only one screenshot :wink: 

And also please add a description to the commit message, to make more clear that this is to make more readable the text in the difficulty label. :slightly_smiling_face:

SergioAtSUSE:
Nice.
You are welcome.

PR #1494, opened 2017-05-27
Fix registration sorting

AEtherC0r3:
When clicking on 'Actions' in order to change the sorting of the lines nothing happens.
Now, the lines are sorted by the value of registration.attended (if the attendee is absent or present)


differentreality:
Nice, this is useful :+1:

PR #1506, opened 2017-05-30
Organiser cannot remove admins #1459

### LINKED ISSUE #1459: Organiser of a conference can remove the admin of OSEM app
Since organiser of the conference has access to revision history therefore organiser can revert is_admin attribute of admin. I don't think organiser should be allowed to revert this change.

screenshot from 2017-04-14 21 06 04

nasia-sam:
Organisers of  conferences have access to revision history so they are able revert the is_admin attribute of admins. Fixes #1459

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -189,6 +189,10 @@ class Ability
     can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
       version.item_type == 'User' || (conf_ids_for_organizer.include? version.conference_id)
     end
+
+    cannot [:revert_object, :revert_attribute], PaperTrail::Version do |version|
+      version.item_type == 'User' && version.item.is_admin?
+    end
   end
 
   def signed_in_with_cfp_role(user)


differentreality:
@nasia-sam your latest change introduces errors in the test suite. Do you have a fix for it?

nasia-sam:
@differentreality not yet. Do you want me to find a way to fix it or just rollback so it's fixed once pull request#1519 is merged?

differentreality:
I merged #1519 
You should rollback please, besides the change is irrelevant to the current pr

differentreality:
I see 2 cases regarding this issue. 

We should not allow the organizer to:
1. edit any attribute of an admin user
2. create an admin, which might happen indirectly through versions by reverting `is_admin` attribute
  a. true -> false (this is covered by the 1st case, where the user is an admin already)
  b. false -> true (this is not covered, and revert is still possible)

Please check changeset option of paper_trail
https://github.com/airblade/paper_trail#3c-diffing-versions

differentreality:
@nasia-sam please update this, so that we can merge it

shlok007:
We plan to allow only site administrators to have access to revision history where `item_type == 'User'` . Refer: https://github.com/openSUSE/osem/issues/1459#issuecomment-324702009 
This discussion doesn't make much sense now and the PR can be safely closed.

bear454:
@nasia-sam can we close this?

PR #1513, opened 2017-06-01
New opensuse deployment

hennevogel:


differentreality:
We mentioned on IRC that https://events-test.opensuse.org might not have ran the task for schedules

PR #1516, opened 2017-06-03
Created PhysicalTicket model

siddhantbajaj:
This model represents a ticket of conference. So now we have three models to handle tickets:

1. Ticket: It represent different types of ticket available for a conference.
2. TicketPuchase: It represent all the Physical tickets bought in single purchase by a user.
3. PhysicalTicket: It will represent a single purchased ticket of a conference.

### DIFF
--- a/app/controllers/payments_controller.rb
+++ b/app/controllers/payments_controller.rb
@@ -18,7 +18,7 @@ class PaymentsController < ApplicationController
 
     if @payment.purchase && @payment.save
       update_purchased_ticket_purchases
-      redirect_to conference_conference_registration_path(@conference.short_title),
+      redirect_to conference_physical_ticket_index_path,
                   notice: 'Thanks! Your ticket is booked successfully.'
     else
       @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
@@ -38,6 +38,8 @@ class PaymentsController < ApplicationController
   end
 
   def update_purchased_ticket_purchases
-    current_user.ticket_purchases.by_conference(@conference).unpaid.update_all(paid: true, payment_id: @payment.id)
+    current_user.ticket_purchases.by_conference(@conference).unpaid.each do |ticket_purchase|
+      ticket_purchase.pay(@payment)
+    end
   end
 end
--- a/app/controllers/physical_ticket_controller.rb
+++ b/app/controllers/physical_ticket_controller.rb
@@ -1,0 +1,12 @@
+class PhysicalTicketController < ApplicationController
+  before_action :authenticate_user!
+  load_resource :conference, find_by: :short_title
+  load_and_authorize_resource
+  authorize_resource :conference_registrations, class: Registration
+
+  def index
+    @physical_tickets = current_user.physical_tickets.by_conference(@conference)
+  end
+
+  def show; end
+end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -87,6 +87,7 @@ class Ability
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
     can [:new, :create], Payment, user_id: user.id
+    can [:index, :show], PhysicalTicket, user_id: user.id
 
     can [:create, :destroy], Subscription, user_id: user.id
 
--- a/app/models/physical_ticket.rb
+++ b/app/models/physical_ticket.rb
@@ -1,0 +1,6 @@
+class PhysicalTicket < ActiveRecord::Base
+  belongs_to :ticket_purchase
+  has_one :ticket, through: :ticket_purchase
+  has_one :conference, through: :ticket_purchase
+  has_one :user, through: :ticket_purchase
+end
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -13,6 +13,8 @@ class TicketPurchase < ActiveRecord::Base
   delegate :price_cents, to: :ticket
   delegate :price_currency, to: :ticket
 
+  has_many :physical_tickets
+
   scope :paid, -> { where(paid: true) }
   scope :unpaid, -> { where(paid: false) }
   scope :by_conference, ->(conference) { where(conference_id: conference.id) }
@@ -45,8 +47,8 @@ class TicketPurchase < ActiveRecord::Base
       purchase = new(ticket_id: ticket.id,
                      conference_id: conference.id,
                      user_id: user.id,
-                     quantity: quantity,
-                     paid: ticket.price_cents.zero?)
+                     quantity: quantity)
+      purchase.pay(nil) if ticket.price_cents.zero?
     end
     purchase
   end
@@ -59,6 +61,13 @@ class TicketPurchase < ActiveRecord::Base
 
     purchase.quantity = quantity if quantity > 0
     purchase
+  end
+
+  def pay(payment)
+    update_attributes(paid: true, payment: payment)
+    PhysicalTicket.transaction do
+      quantity.times { physical_tickets.create }
+    end
   end
 end
 
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -2,6 +2,7 @@ class TicketPurchase < ActiveRecord::Base
   belongs_to :ticket
   belongs_to :user
   belongs_to :conference
+  belongs_to :payment
 
   validates :ticket_id, :user_id, :conference_id, :quantity, presence: true
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,6 +6,11 @@ end
 
 class User < ActiveRecord::Base
   rolify
+  has_many :physical_tickets, through: :ticket_purchases do
+    def by_conference(conference)
+      where('ticket_purchases.conference_id = ?', conference)
+    end
+  end
   has_many :users_roles
   has_many :roles, through: :users_roles, dependent: :destroy
 
--- a/app/views/conferences/_conference_details.html.haml
+++ b/app/views/conferences/_conference_details.html.haml
@@ -38,3 +38,5 @@
               = link_to 'Subscribe', conference_subscriptions_path(conference.short_title), method: :post, class: 'btn btn-default'
             - else
               = link_to 'Unsubscribe', conference_subscriptions_path(conference.short_title), method: :delete, class: 'btn btn-default'
+            - if current_user && current_user.physical_tickets.by_conference(conference).any?
+              = link_to "My Tickets", conference_physical_ticket_index_path(conference.short_title), class: 'btn btn-default'
--- a/app/views/physical_ticket/index.html.haml
+++ b/app/views/physical_ticket/index.html.haml
@@ -1,0 +1,35 @@
+.container
+  .row
+    .col-md-12.page-header
+      %h2
+        Tickets
+      .text-muted
+        Your tickets for the conference
+
+    .col-md-12
+      - if @physical_tickets.present?
+        %table.table.table-bordered.table-striped.table-hover#roles
+          %thead
+            %th ID
+            %th Type
+            %th User
+            %th Actions
+          %tbody
+            - @physical_tickets.each do |physical_ticket|
+              %tr
+                %td= physical_ticket.id
+                %td= physical_ticket.ticket.title
+                %td= physical_ticket.user.name
+                %td
+                  .btn-group
+                    = link_to 'Show',
+                    conference_physical_ticket_path(@conference.short_title,
+                                                    physical_ticket.id),
+                    class: 'btn btn-primary'
+                    = link_to 'Generate PDF',
+                    conference_physical_ticket_path(@conference.short_title,
+                                                    physical_ticket.id,
+                                                    format: :pdf),
+                    class: 'button btn btn-default btn-info'
+      - else
+        %h5 No Tickets found!
--- a/db/migrate/20170603095900_create_physical_tickets.rb
+++ b/db/migrate/20170603095900_create_physical_tickets.rb
@@ -1,0 +1,9 @@
+class CreatePhysicalTickets < ActiveRecord::Migration
+  def change
+    create_table :physical_tickets do |t|
+      t.integer :ticket_purchase_id, null: false
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170531094819) do
+ActiveRecord::Schema.define(version: 20170603095900) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -287,6 +287,12 @@ ActiveRecord::Schema.define(version: 20170531094819) do
     t.integer  "conference_id",                  null: false
     t.datetime "created_at",                     null: false
     t.datetime "updated_at",                     null: false
+  end
+
+  create_table "physical_tickets", force: :cascade do |t|
+    t.integer  "ticket_purchase_id", null: false
+    t.datetime "created_at",         null: false
+    t.datetime "updated_at",         null: false
   end
 
   create_table "programs", force: :cascade do |t|
--- a/spec/factories/ticket_purchases.rb
+++ b/spec/factories/ticket_purchases.rb
@@ -4,5 +4,11 @@ FactoryGirl.define do
     conference
     ticket
     quantity 10
+    factory :paid_ticket_purchase do
+      after(:build) do |ticket_purchase|
+        payment = create(:payment)
+        ticket_purchase.pay(payment)
+      end
+    end
   end
 end
--- a/spec/models/ticket_purchase_spec.rb
+++ b/spec/models/ticket_purchase_spec.rb
@@ -34,7 +34,6 @@ describe TicketPurchase do
     it 'is valid with a quantity greater than zero' do
       should allow_value(1).for(:quantity)
     end
-
   end
 
   describe 'self#purchase' do
@@ -54,7 +53,6 @@ describe TicketPurchase do
       expect(TicketPurchase.count).to eq(1)
       expect(purchase.quantity).to eq(10)
       expect(message.blank?).to be true
-
     end
 
     it 'creates a purchase for one ticket' do
@@ -114,6 +112,24 @@ describe TicketPurchase do
       expect(TicketPurchase.count).to eq(1)
       expect(purchase.quantity).to eq(10)
       expect(message.blank?).to be true
+    end
+  end
+
+  describe 'after_create' do
+    let(:ticket_purchase) { create(:ticket_purchase, quantity: 4, paid: true) }
+
+    it 'creates physical tickets equal to the quantity of purchase' do
+      expect(ticket_purchase.physical_tickets.count).to eq(4)
+    end
+  end
+
+  describe 'after_update' do
+    let(:ticket_purchase) { create(:ticket_purchase, quantity: 5) }
+
+    it 'creates physical tickets if the payment is made successfully' do
+      ticket_purchase
+      ticket_purchase.paid = true
+      expect{ ticket_purchase.save }.to change{ ticket_purchase.physical_tickets.count }.from(0).to(5)
     end
   end
 end
--- a/spec/models/ticket_purchase_spec.rb
+++ b/spec/models/ticket_purchase_spec.rb
@@ -114,22 +114,4 @@ describe TicketPurchase do
       expect(message.blank?).to be true
     end
   end
-
-  describe 'after_create' do
-    let(:ticket_purchase) { create(:ticket_purchase, quantity: 4, paid: true) }
-
-    it 'creates physical tickets equal to the quantity of purchase' do
-      expect(ticket_purchase.physical_tickets.count).to eq(4)
-    end
-  end
-
-  describe 'after_update' do
-    let(:ticket_purchase) { create(:ticket_purchase, quantity: 5) }
-
-    it 'creates physical tickets if the payment is made successfully' do
-      ticket_purchase
-      ticket_purchase.paid = true
-      expect{ ticket_purchase.save }.to change{ ticket_purchase.physical_tickets.count }.from(0).to(5)
-    end
-  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'responders', '~> 2.0'
 # as the database for Active Record
 # choose only one
 gem 'mysql2'
-#gem 'pg'
+# gem 'pg'
 
 # for observing records
 gem 'rails-observers'
--- a/app/controllers/ticket_purchases_controller.rb
+++ b/app/controllers/ticket_purchases_controller.rb
@@ -11,7 +11,7 @@ class TicketPurchasesController < ApplicationController
         redirect_to new_conference_payment_path,
                     notice: 'Please pay here to get tickets.'
       elsif current_user.ticket_purchases.by_conference(@conference).paid.any?
-        redirect_to conference_conference_registration_path(@conference.short_title),
+        redirect_to conference_physical_ticket_index_path,
                     notice: 'You have free tickets for the conference.'
       else
         redirect_to conference_tickets_path(@conference.short_title),
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -19,6 +19,7 @@ class Conference < ActiveRecord::Base
   has_one :email_settings, dependent: :destroy
   has_one :program, dependent: :destroy
   has_one :venue, dependent: :destroy
+  has_many :physical_tickets, through: :ticket_purchases
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
   has_many :supporters, through: :ticket_purchases, source: :user
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -126,6 +126,7 @@ Osem::Application.routes.draw do
     resources :tickets, only: [:index]
     resources :ticket_purchases, only: [:create, :destroy]
     resources :payments, only: [:index, :new, :create]
+    resources :physical_ticket, only: [:index, :show]
     resource :subscriptions, only: [:create, :destroy]
     resource :schedule, only: [:show] do
       member do
--- a/spec/factories/physical_ticket.rb
+++ b/spec/factories/physical_ticket.rb
@@ -1,0 +1,5 @@
+FactoryGirl.define do
+  factory :physical_ticket do
+    ticket_purchase
+  end
+end
--- a/spec/features/ticket_purchases_spec.rb
+++ b/spec/features/ticket_purchases_spec.rb
@@ -101,12 +101,10 @@ feature Registration do
 
         click_button 'Continue'
 
-        expect(current_path).to eq(conference_conference_registration_path(conference.short_title))
+        expect(current_path).to eq(conference_physical_ticket_index_path(conference.short_title))
         purchase = TicketPurchase.where(user_id: participant.id, ticket_id: free_ticket.id).first
         expect(purchase.quantity).to eq(5)
         expect(purchase.paid).to be true
-
-        expect(page.has_content?("5 #{free_ticket.title} Tickets for $ 0")).to be true
       end
     end
 
--- a/spec/models/physical_ticket_spec.rb
+++ b/spec/models/physical_ticket_spec.rb
@@ -1,0 +1,14 @@
+require 'spec_helper'
+
+describe PhysicalTicket do
+
+  describe 'association' do
+    it { is_expected.to belong_to :ticket_purchase }
+  end
+
+  describe 'validations' do
+    it 'has a valid factory' do
+      expect(build(:physical_ticket)).to be_valid
+    end
+  end
+end


app/helpers/physical_ticket_helper.rb
```diff
@@ -0,0 +1,2 @@
+module PhysicalTicketHelper
```
Ana06: I guess there are aome classes that were automatically generated, but if by now they are not use, like it is the case with this helper, I would better remove them 🤔

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,6 @@
+class PhysicalTicket < ActiveRecord::Base
+	belongs_to :conference
+	belongs_to :user
+	belongs_to :ticket_purchases
+	belongs_to :ticket
```
Ana06: A ticket purchase already belongs to ticket, right? So you don't need this relation with ticket. In case that at some point the ticket is needed we can introduce a thought ticket_purchase relation that do not need extra fields in the database. :wink:

app/models/ticket.rb
```diff
@@ -1,6 +1,7 @@
 class Ticket < ActiveRecord::Base
   belongs_to :conference
   has_many :ticket_purchases, dependent: :destroy
+  has_many :physical_tickets, dependent: :destroy
```
Ana06: Same here, this is not really needed. :wink:

app/assets/javascripts/physical_ticket.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
Ana06: This file is not needed

app/assets/stylesheets/physical_ticket.css.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the PhysicalTicket controller here.
```
Ana06: This file is not needed

app/controllers/physical_ticket_controller.rb
```diff
@@ -0,0 +1,10 @@
+class PhysicalTicketController < ApplicationController
+  before_filter :authenticate_user!, except: [:new, :create]
```
Ana06: There is only the `new` action for this controller 😕

app/models/ability.rb
```diff
@@ -86,6 +86,7 @@ def signed_in(user)
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
     can [:new, :create], Payment, user_id: user.id
+    can [:new, :create], PhysicalTicket, user_id: user.id
```
Ana06: There is not :create: action in the controller 😕

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,6 @@
+class PhysicalTicket < ActiveRecord::Base
+	belongs_to :conference
```
Ana06: A ticket purchase already belongs to a conference so this is also not needed.

app/controllers/physical_ticket_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+  load_resource :conference, find_by: :short_title
+  load_and_authorize_resource
+  authorize_resource :conference_registrations, class: Registration
+
+  def new
+  	@unassigned_ticket_purchases = current_user.ticket_purchases.unassigned.by_conference(@conference)
```
Ana06: I am not sure what is the idea of this. Create all the available icket in the database as unassigned?  : thinking:

app/models/ticket_purchase.rb
```diff
@@ -13,8 +13,12 @@ class TicketPurchase < ActiveRecord::Base
   delegate :price_cents, to: :ticket
   delegate :price_currency, to: :ticket
 
+  has_many :physical_tickets
+
   scope :paid, -> { where(paid: true) }
   scope :unpaid, -> { where(paid: false) }
+  scope :assigned, -> { where(assigned: true) }
```
Ana06: The assigbed scope is not currently used, right? So I would prefer introducing it when we need it, it as some point this happens

test/controllers/physical_ticket_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class PhysicalTicketControllerTest < ActionController::TestCase
```
Ana06: We are not using rails tests, but RSpec instead!

app/helpers/physical_ticket_helper.rb
siddhantbajaj: okay. :+1:

app/models/physical_ticket.rb
siddhantbajaj: I'll remove it. :+1:

app/models/physical_ticket.rb
siddhantbajaj: :+1:

app/models/ticket.rb
siddhantbajaj: :+1:

app/assets/javascripts/physical_ticket.js
siddhantbajaj: Sorry from now I'll make sure I don't commit unused files. :slightly_smiling_face:

app/assets/stylesheets/physical_ticket.css.scss
siddhantbajaj: :+1:

app/controllers/physical_ticket_controller.rb
siddhantbajaj: It was something related to assignment of users to tickets. But I'll remove it now.

app/models/ticket_purchase.rb
siddhantbajaj: :+1:

siddhantbajaj:
Workflow of Ticket Purchase feature:

1. Number of tickets required

screenshot from 2017-06-07 03 54 53

2. Generate Tickets

screenshot from 2017-06-07 03 55 13

3. Generate PDF/show

screenshot from 2017-06-07 03 55 36

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,14 @@
+class PhysicalTicket < ActiveRecord::Base
+	belongs_to :user
```
Ana06: this relation should be through `ticket_purchase`

app/controllers/physical_ticket_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    @current_ticket_purchases = current_user.current_ticket_purchases(@conference)
+  end
+
+  def show
+  end
+
+  def create
+    @ticket_purchases = current_user.ticket_purchases.by_conference(@conference)
+    @ticket_purchases.each do |ticket_purchase|
+      if !ticket_purchase.contains_physical_tickets?
```
Ana06: `if !` == `unless` and it is more readable :wink:

Ana06: and why not creating only the ones of the last `ticket_purchase`? 

I mean, this should be done in the ticket_purchase model. When quantity is set to 3, you should ensure that there are 3 physical tickets. :wink:

app/controllers/physical_ticket_controller.rb
```diff
@@ -0,0 +1,29 @@
+class PhysicalTicketController < ApplicationController
```
Ana06: I doubth that you need in this controller something else except index and show.

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,14 @@
...
+	belongs_to :ticket_purchase
+
+	scope :by_user, ->(user) { where(user_id: user.id) }
+
+	def ticket_type
+		ticket_purchase.ticket
+	end
+
+	def conference
```
Ana06: this can be done with a through association, but I would say it is not needed yet.

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,14 @@
+class PhysicalTicket < ActiveRecord::Base
+	belongs_to :user
+	belongs_to :ticket_purchase
+
+	scope :by_user, ->(user) { where(user_id: user.id) }
+
+	def ticket_type
```
Ana06: this can be done with a through association, but I would say it is not needed yet.

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,14 @@
+class PhysicalTicket < ActiveRecord::Base
+	belongs_to :user
+	belongs_to :ticket_purchase
+
+	scope :by_user, ->(user) { where(user_id: user.id) }
```
Ana06: you could create a through association in user for this.

app/models/ticket_purchase.rb
```diff
@@ -58,4 +60,10 @@ def self.update_quantity(conference, quantity, ticket, user)
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  ##
+  # Return true if tickets in ticket purchase are already created.
+  def contains_physical_tickets?
+    return PhysicalTicket.exists?(ticket_purchase_id: self)
```
Ana06: `return` is not needed. And this can be written as:

`physical_tickets.any?`

app/views/physical_ticket/_tickets.html.haml
```diff
@@ -0,0 +1,21 @@
+.page-header
```
Ana06: how all this work with all conferences where physical tickets didn't exist?

db/migrate/20170603095900_create_physical_tickets.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePhysicalTickets < ActiveRecord::Migration
+  def change
+    create_table :physical_tickets do |t|
+      t.integer :user_id, null: false
```
Ana06: this is not needed

app/models/physical_ticket.rb
siddhantbajaj: Yes I'll make the changes. :wink:

app/controllers/physical_ticket_controller.rb
siddhantbajaj: if we create Physical Tickets in ticket purchase controller then basically we are creating tickets before their payment is made through stripe.

app/controllers/physical_ticket_controller.rb
siddhantbajaj: if we create Physical Tickets in ticket purchase controller then basically we are creating tickets before their payment is made through stripe.

app/models/physical_ticket.rb
siddhantbajaj: I'll change this. :wink:

app/controllers/physical_ticket_controller.rb
siddhantbajaj: Sorry my bad did you mean we should create tickets in after_create and after_update callbacks of ticket purchase model?

Ana06: exactly! :wink: after the payment is done, we set the purchase to paid right? in that moment you also should create the physical tickets :slightly_smiling_face:

app/models/user.rb
```diff
@@ -100,6 +101,10 @@ def supports? conference
     ticket_purchases.find_by(conference_id: conference.id).present?
   end
 
+  def tickets conference
+    conference.physical_tickets.select { |t| t.user == self }
```
Ana06: `physical_tickets.where(conference: conference)`

app/views/physical_ticket/_tickets.html.haml
Ana06: ^^

Ana06:
@siddhantbajaj 

One important detail. There is something currently "wrong" in the payment. Check this:
https://github.com/Ana06/osem/blob/7a7fcced487de914c9bd2bfc9e4e51839413aa8e/app/controllers/payments_controller.rb#L41

`update_all` skips all validations and callbacks, so your `after_update` in the TicketPurchase model doesn't work.

So, the first thing should be correcting this, because we do not want to use method that skip validations.

What I would do is to create a pay method in the TicketPurchase model and call it from `payment_controller `method. In the pay method you will update the paid attribute and create the physical tickets.

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,15 @@
+class PhysicalTicket < ActiveRecord::Base
+  belongs_to :ticket_purchase
+
+  def ticket_type
```
mdeniz: This should be a :through relationship

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,15 @@
...
+	ticket_purchase.ticket
+  end
+
+  def conference
+	ticket_purchase.conference
```
mdeniz: This should be a :through relationship

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def user
+	ticket_purchase.user
```
mdeniz: This should be a :through relationship

app/models/user.rb
```diff
@@ -100,6 +101,10 @@ def supports? conference
     ticket_purchases.find_by(conference_id: conference.id).present?
   end
 
+  def tickets conference
```
mdeniz: Use parenthesis for params 😄

spec/factories/ticket_purchases.rb
```diff
@@ -4,5 +4,6 @@
     conference
     ticket
     quantity 10
+    paid false
```
mdeniz: Does it worth to have a 'sub' factory for already paid ticket_purchases??

spec/models/ticket_purchase_spec.rb
```diff
@@ -2,118 +2,136 @@
 
 describe TicketPurchase do
 
-  describe 'validations' do
-    it 'has a valid factory' do
-      expect(build(:ticket_purchase)).to be_valid
+  # describe 'validations' do
```
mdeniz: Why those commented specs?

siddhantbajaj: Sorry my bad I was testing something here so I commented all other specs and forgot to un-comment them.

spec/factories/ticket_purchases.rb
siddhantbajaj: Yes I think it would be useful to have a sub factory here for paid ticket_purchases.

app/models/ticket_purchase.rb
```diff
@@ -58,4 +63,24 @@ def self.update_quantity(conference, quantity, ticket, user)
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  def pay(payment)
+    self.update_attributes(paid: true, payment_id: payment.id)
```
Ana06: you can write this as:

``` Ruby
update_attributes(paid: true, payment: payment)
```

`self` is not needed. Please read: https://hackhands.com/three-golden-rules-understand-self-ruby :wink: And Rubocop can also help you, as the `Style/RedundantSelf` cop is enabled. 

And as we have a relation, you can use it instead of the `id` :slightly_smiling_face:

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,6 @@
+class PhysicalTicket < ActiveRecord::Base
+  belongs_to :ticket_purchase
+  delegate :ticket, to: :ticket_purchase
```
Ana06: `delegate` is for expose contained objects’ public methods as your own. Ref: https://apidock.com/rails/Module/delegate

for this you should use `:through` associations. See http://guides.rubyonrails.org/association_basics.html#the-has-one-through-association

spec/factories/ticket_purchases.rb
```diff
@@ -4,5 +4,11 @@
...
+    factory :paid_ticket_purchase do
+    	after(:build) do |ticket_purchase|
```
Ana06: you also need to create physical tickets here

spec/factories/ticket_purchases.rb
Ana06: It is false as default in the database, so this shouldn't be needed, right? :thinking:

spec/models/ticket_purchase_spec.rb
```diff
@@ -116,4 +116,22 @@
       expect(message.blank?).to be true
     end
   end
+
```
Ana06: This is in the wrong commit!

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170606001722) do
+ActiveRecord::Schema.define(version: 20170603095900) do
```
Ana06: This is in the wrong commit!

app/models/ticket_purchase.rb
```diff
@@ -13,11 +13,16 @@ class TicketPurchase < ActiveRecord::Base
...
   delegate :price_currency, to: :ticket
 
+  has_many :physical_tickets
+
   scope :paid, -> { where(paid: true) }
   scope :unpaid, -> { where(paid: false) }
   scope :by_conference, ->(conference) { where(conference_id: conference.id) }
   scope :by_user, ->(user) { where(user_id: user.id) }
 
+  after_create :create_physical_tickets
```
mdeniz: This 2 methods are unneeded if you create the tickets at the pay method

app/models/user.rb
```diff
@@ -6,6 +6,7 @@ class UserDisabled < StandardError
 
 class User < ActiveRecord::Base
   rolify
+  has_many :physical_tickets, through: :ticket_purchases, dependent: :destroy
```
mdeniz: I guess that we are not going to use this relationship without scoping it by conference... so, better if you do something like this:

```ruby
has_many :physical_tickets, -> (conference) { where(conference: conference) }, through: :ticket_purchases, dependent: :destroy
```

If you do so, then the **tickets** method that you created (line 104 here) is not needed.

app/models/user.rb
```diff
@@ -100,6 +101,10 @@ def supports? conference
     ticket_purchases.find_by(conference_id: conference.id).present?
   end
 
+  def tickets(conference)
```
mdeniz: This method will be unneeded if you scope the relationship by conference

app/views/conferences/_conference_details.html.haml
```diff
@@ -38,3 +38,5 @@
               = link_to 'Subscribe', conference_subscriptions_path(conference.short_title), method: :post, class: 'btn btn-default'
             - else
               = link_to 'Unsubscribe', conference_subscriptions_path(conference.short_title), method: :delete, class: 'btn btn-default'
+            - if !current_user.nil? && current_user.tickets(conference).any?
```
mdeniz: Would be better to use **present?** here? Also if you use the scoped relationship for **physical_tickets** it will look like this:

```ruby
if current_user.present? && current_user.physical_tickets(conference).any?
```
Or just using **try**:

```ruby
if current_user.try(:physical_tickets, conference).try(:any?)
```

spec/factories/ticket_purchases.rb
mdeniz: Instead of marking the object as paid and creating physical_tickets by hand, you can use the **pay** method, so maybe you need a payment object too.

spec/factories/ticket_purchases.rb
```diff
@@ -4,5 +4,11 @@
...
+    		ticket_purchase.paid = true
```
mdeniz: And BTW, normally overriding just an attribute is better to put this directly on the subfactory, not in a **after(:build)**:

```ruby
factory :paid_ticket_purchase do
  paid true
end
```

spec/models/ticket_purchase_spec.rb
mdeniz: Those spec will be unneeded if you remove the methods that are tested

app/models/ticket_purchase.rb
```diff
@@ -58,4 +63,24 @@ def self.update_quantity(conference, quantity, ticket, user)
...
     purchase
   end
+
+  def pay(payment)
+    self.update_attributes(paid: true, payment_id: payment.id)
+  end
+
+  protected
+
+  def create_physical_tickets
```
siddhantbajaj: But then what about Free tickets they does not have any payment and paid attribute is already true.:thinking:

siddhantbajaj: But then what about Free ticket_purchases they does not have any payment and their paid attribute is already true.

spec/factories/ticket_purchases.rb
siddhantbajaj: I think we should create tickets in after_create method because there are free ticket purchases as well that does not have any payment and if we do that we don't have to create tickets here.

spec/factories/ticket_purchases.rb
siddhantbajaj: Sure I'll make the change. :+1:

app/models/ticket_purchase.rb
Ana06: for example,

instead of setting paid to `true` in https://github.com/Ana06/osem/blob/7a7fcced487de914c9bd2bfc9e4e51839413aa8e/app/models/ticket_purchase.rb#L47

in the next line you can just do:

`purchase.pay if  ticket.price_cgents.zero?`

mdeniz: I agree with Ana, call the **pay** method in the purchase_ticket after creating the **TicketPurchase** object if it is free. That way you will have all the logic in one place, the **pay** method.

But rather than just doing **purchas.pay** you will need to do **purchase.pay(nil)** in the case of free tickets.

siddhantbajaj: > That way you will have all the logic in one place, the pay method.

This makes sense, I'll make the changes. :+1:

app/models/user.rb
siddhantbajaj: I tried this but it's not working, it gives a SQL exception because physical ticket does not have conference_id column. So I am looking for a work around for this.:thinking:

app/models/physical_ticket.rb
siddhantbajaj: Thanks a lot for this. I've made the changes. :+1:

app/controllers/payments_controller.rb
```diff
@@ -18,7 +18,7 @@ def create
 
     if @payment.purchase && @payment.save
       update_purchased_ticket_purchases
-      redirect_to conference_conference_registration_path(@conference.short_title),
+      redirect_to conference_physical_ticket_index_path,
```
siddhantbajaj: So like `update_purchased_ticket_purchases` is still excuting in the background and the user is redirected to index page? I am not sure if redirect_to works synchronously.

app/views/conferences/_conference_details.html.haml
siddhantbajaj: :+1:

app/models/user.rb
siddhantbajaj: :+1:

app/models/ticket_purchase.rb
siddhantbajaj: :+1:

app/models/user.rb
Ana06: You can add a through relation for conference also in ticket :wink:

app/models/physical_ticket.rb
```diff
@@ -0,0 +1,9 @@
+class PhysicalTicket < ActiveRecord::Base
+  belongs_to :ticket_purchase
+  has_one :ticket, through: :ticket_purchase
+  has_one :conference, through: :ticket_purchase
+  has_one :user, through: :ticket_purchase
+
+  scope :by_conference, ->(conference) { joins(:ticket_purchase).where('ticket_purchases.conference_id = ?', conference.id) }
```
mdeniz: Why don't you use the relationship instead of doing a join by hand?

```ruby
scope :by_conference, ->(conference) { where(conference: conference) }
```

app/models/ticket_purchase.rb
```diff
@@ -58,4 +60,11 @@ def self.update_quantity(conference, quantity, ticket, user)
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  def pay(payment)
+    update_attributes(paid: true, payment_id: payment)
```
mdeniz: I guess you can assign it to **payment** instead of to  **payment_id**

```ruby
 update_attributes(paid: true, payment: payment)
```

app/models/ticket_purchase.rb
```diff
@@ -58,4 +60,11 @@ def self.update_quantity(conference, quantity, ticket, user)
...
+    quantity.times do
+      PhysicalTicket.create(ticket_purchase_id: id)
```
mdeniz: For creating related objects is better to use the relationship methods it will assign the TicketPurchase object without having to explicitly write it there. 

Also I would prefer to surround this with a transaction. And BTW even better using a one line block style for the iterating part:

```ruby
PhysicalTicket.transaction do
  quantity.times { physical_tickets.create }
end
```

app/models/user.rb
mdeniz: Does this works like Ana said using the through relationship?

app/models/physical_ticket.rb
siddhantbajaj: Yes I tried that, but its not working. Same SQL exception no such column conference_id in physical_tickets table. 🤔

mdeniz: Ok then 👍

app/models/user.rb
siddhantbajaj: But I already have a through relation for conference in the physical ticket and still it didn't work. :confused:

app/models/ticket_purchase.rb
siddhantbajaj: No I think we have to specify the exact attribute name in update_attributes. I tried this and it gives an error unknown attribute payment.

app/models/ticket_purchase.rb
siddhantbajaj: Sure I'll make the change. 🙂

app/models/physical_ticket.rb
Ana06: That's probably the reason why there is a relation with conference in most of our models, to avoid joins. I am not sure what is better. :thinking:

mdeniz: Less dangerous to do the joins, and at the same time less performance. But adding a relationship directly to the conference can end in having wrong data because the duplication.

Ana06:
@siddhantbajaj please remember to squash the commits :wink:

app/models/ticket_purchase.rb
Ana06: Please check it again. You should be able to use `payment` directly as @mdeniz says.

app/models/user.rb
Ana06: you need the `join`. But anyway the by_conference method is always used like:

`user.physical_tickets.by_conference`

so you can just add the `by_conference` scope here, and even remove it from the other model, as would only be used here.

app/views/conferences/_conference_details.html.haml
```diff
@@ -38,3 +38,5 @@
               = link_to 'Subscribe', conference_subscriptions_path(conference.short_title), method: :post, class: 'btn btn-default'
             - else
               = link_to 'Unsubscribe', conference_subscriptions_path(conference.short_title), method: :delete, class: 'btn btn-default'
+            - if current_user.present? && current_user.physical_tickets.by_conference(conference).any?
```
Ana06: you don't need the `present?`, just:

``` Ruby
if current_use && current_user.physical_tickets.by_conference(conference).any?
```

app/models/user.rb
siddhantbajaj: > so you can just add the by_conference scope here, and even remove it from the other model, as would only be used here. 

Sorry I didn't understand what you mean here. we already have by_conference scope in PhysicalTicket model why do we have to remove it from there and it in this user model. 😕

app/models/ticket_purchase.rb
siddhantbajaj: I found the mistake actually one to many association between ticket purchase and payment wasn't set up. There was no `belongs_to:payment` in ticket purchase model.

app/models/ticket_purchase.rb
```diff
@@ -60,6 +63,13 @@ def self.update_quantity(conference, quantity, ticket, user)
     purchase.quantity = quantity if quantity > 0
     purchase
   end
+
+  def pay(payment)
+    update_attributes(paid: true, payment: payment)
+    PhysicalTicket.transaction do
```
Ana06: Why do we need to wrap this into a transaction? What the difference between creating one of them or any of them?

I am not pretty sure how we should handle this. @mdeniz opinions/ideas?

app/models/ticket_purchase.rb
```diff
@@ -2,6 +2,7 @@ class TicketPurchase < ActiveRecord::Base
   belongs_to :ticket
   belongs_to :user
   belongs_to :conference
+  belongs_to :payment
```
Ana06: @siddhantbajaj why did you need to add  this?

app/models/ticket_purchase.rb
Ana06: @siddhantbajaj I spoke with @mdeniz and it is okey like that :wink:

app/models/ticket_purchase.rb
siddhantbajaj: > Please check it again. You should be able to use payment directly as @mdeniz says.

It was not working because one to many association between Payment and Ticket Purchase wasn't properly set up.

Ana06: Then add this is in a different commit, as it to solve something that was already wrong and not really related with this feature please. And after that I would say we are ready to merge it. @mdeniz do you want to take a last look? :wink:

siddhantbajaj: Okay 🙂

app/views/physical_ticket/index.html.haml
```diff
@@ -0,0 +1,32 @@
...
+                %td= physical_ticket.id
+                %td= physical_ticket.ticket.title
+                %td= physical_ticket.user.name
+                %td
+                  .btn-group
+                    = link_to 'Show',
+                    conference_physical_ticket_path(@conference.short_title, physical_ticket.id),
+                    class: 'btn btn-primary'
+                    = link_to 'Generate PDF',
+                    conference_physical_ticket_path(@conference.short_title, physical_ticket.id,format: :pdf),
```
Ana06: you could write this as

```
conference_physical_ticket_path(@conference.short_title, 
                                physical_ticket.id,format: :pdf),
```

to avoid it too be that long :wink:

Ana06:
Merged! :tada: Great job! :+1:

PR #1517, opened 2017-06-04
Introduce organizations to OSEM

### LINKED ISSUE #1502: Modify schema to introduce organizations
The part of modified schema is as follows:

organization_table

Organizations will have a belongs_to has_many relationship with conference.

### LINKED ISSUE #1503: Introduce views to manage organizations
Views to be added can include the following:

- admin/organization#index
admin organization index

- organization#new
screenshot_20170529_160157

### LINKED ISSUE #1521: Create a migration for organizations
The migration should include the following:

- [ ] Add organization as reference to conferences
- [ ] Create a first default organization.
- [ ] Fill out the organization_id column on all existing conferences with the default organization's id.
- [ ] Create a foreign key so that the conference-organization relation is always valid.

shlok007:
closes #1521, closes #1503, closes #1502, 

This PR includes the appropriate controllers, models and views required for organizations in OSEM.

The new views includes the following:

organizations#index:
screenshot_20170604_141915

organization#new:
screenshot_20170604_141941

admin/organizations#index:
screenshot_20170604_142346

user dropdown menu to include organizations:
screenshot_20170604_142522


db/migrate/20170531094817_add_organizaton_id_to_conference.rb
```diff
@@ -0,0 +1,5 @@
+class AddOrganizatonIdToConference < ActiveRecord::Migration
+  def change
+    add_column :conferences, :organization_id, :integer
```
lagartoflojo: Since `organization_id` is a required field in conference, we probably want to:
1. First create the field. You can take advantage of the [`add_reference`](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference) method instead of adding an integer. Plus, you probably want to use the `index: true` option, since we'll be filtering conferences by organization.
1. Create a first default organization.
1. Fill out the `organization_id` column on all existing conferences with the default organization's id.
1. Create a [foreign key](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_foreign_key) so that the conference-organization relation is always valid. 

This should all be done in migrations (not rake tasks), since we are "migrating" the data from the old no-organization format to the new one.

lagartoflojo:
Hi @shlok007, the PR is looking good!
I was thinking: what if we just limit the organization-creation capabilities to site admins for now? I think this will be easier to implement going forward and is also good enough. We can then think about allowing anyone to create new organizations later.

app/views/layouts/_admin_sidebar_index.html.haml
```diff
@@ -32,3 +32,7 @@
       = link_to(admin_revision_history_path) do
         %span.fa.fa-history
         Revision History
+    %li
+      = link_to(admin_organizations_path) do
+        %span.fa.fa-group
+        Organizations
```
shlok007: Are we planning to merge this PR after the organizations are introduced or wait until they are actually useful ( custom domain and rest of the features ) ? If we are not merging it until then, then there shouldn't be any need to hide these, right?

lagartoflojo: The idea is to merge the PRs as soon as they are ready. If we just create a huge PR at the end of GSoC, it will be impossible to merge.

shlok007:
Thank you @lagartoflojo !
And as you said, it would be better to limit creation to the site admins for now. We can work on the appropriate permissions while working on introducing 'organization admins' later on. :slightly_smiling_face:

db/migrate/20170531094817_add_organizaton_id_to_conference.rb
shlok007: Thank you for this! :slightly_smiling_face:

app/views/admin/organizations/index.html.haml
```diff
@@ -0,0 +1,31 @@
...
+          %th Actions
+        %tbody
+          - @organizations.each do |organization|
+            %tr
+              %td
+                = organization.name
+              %td
+                = organization.conferences.count
+              %td
+                = organization.conferences.count
```
shlok007: That was supposed to be temporary. I'll create an issue for it and fix this once the migrations are ready.

app/views/admin/organizations/index.html.haml
```diff
@@ -0,0 +1,31 @@
...
+              %td
+                .btn-group
+                  = link_to 'Edit', edit_organization_path(organization),
+                    method: :get, class: 'btn btn-primary'
+                  = link_to 'Delete', admin_organization_path(organization),
+                    method: :delete, class: 'btn btn-danger'
```
shlok007: Sure.

app/views/admin/organizations/_form.html.haml
```diff
@@ -0,0 +1,13 @@
+= semantic_form_for(@organization) do |f|
+  = f.inputs name: 'Organization details' do
+    = f.input :name, as: :string, required: true
+    = f.input :description, input_html: { rows: 5 }, placeholder: 'Decribe about your organization..'
```
shlok007: Yes of course. Will change this as well.

app/views/layouts/_admin_sidebar_index.html.haml
shlok007: Okay. That makes much sense. 
By feature flag, do you mean something like https://github.com/balvig/chili ?

lagartoflojo: While that gem looks interesting, I don't think we need anything so "fancy", since it's just temporary. Just reading an environment variable should be enough :)

app/views/admin/organizations/index.html.haml
```diff
@@ -0,0 +1,31 @@
+.row
+  .col-md-12
+    .page-header
+      %h1 Organizations
+      .btn-group.pull-right
+        = link_to 'Add Organization', new_admin_organization_path, class: 'btn btn-success pull-right'
```
lagartoflojo: I'd change Add -> Create

app/views/admin/organizations/_form.html.haml
```diff
@@ -0,0 +1,13 @@
+= semantic_form_for(@organization, url: (@organization.new_record? ? admin_organizations_path : admin_organization_path(@organization))) do |f|
+  = f.inputs name: 'Organization details' do
+    = f.input :name, as: :string, required: true
+    = f.input :description, as: :text, input_html: { rows: 10 }, placeholder: 'Decribe about your organization..'
```
lagartoflojo: I think this is better as a `textarea`

app/views/admin/organizations/index.html.haml
lagartoflojo: ok!

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'spec_helper'
+
+describe OrganizationsController do
```
lagartoflojo: `OrganizationsController` only has the index action.
`Admin::OrganizationsController` has the other actions as well.

spec/controllers/admin/organizations_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe Admin::OrganizationsController do
```
lagartoflojo: Missing the tests for the other actions.

config/routes.rb
```diff
@@ -102,7 +103,7 @@
     get '/revision_history/:id/revert_object' => 'versions#revert_object', as: 'revision_history_revert_object'
     get '/revision_history/:id/revert_attribute' => 'versions#revert_attribute', as: 'revision_history_revert_attribute'
   end
-
+  resources :organizations
```
lagartoflojo: Since we only have the `index` action, it's better to say:

`resources :organizations, only: [:index]`

db/migrate/20170531094819_move_conferences_to_organizations.rb
```diff
@@ -0,0 +1,21 @@
...
+  class TempOrganization < ActiveRecord::Base
+    self.table_name = 'organizations'
+  end
+
+  def change
+    add_reference :conferences, :organization, index: true
+    add_foreign_key :conferences, :organizations, dependent: :delete
+
+    TempConference.reset_column_information
+    organization = TempOrganization.create(name: 'organization', description: 'Default organization to migrate old conferences to the new version of OSEM')
```
lagartoflojo: I think that a description of "Default organization" would be enough :)

db/migrate/20170531094819_move_conferences_to_organizations.rb
```diff
@@ -0,0 +1,21 @@
...
+    self.table_name = 'conferences'
+  end
+
+  class TempOrganization < ActiveRecord::Base
+    self.table_name = 'organizations'
+  end
+
+  def change
+    add_reference :conferences, :organization, index: true
+    add_foreign_key :conferences, :organizations, dependent: :delete
```
lagartoflojo: If the `conferences` table already has some conferences with NULL `organization_id`, will `add_foreign_key` work?

db/migrate/20170529215453_create_organizations.rb
```diff
@@ -0,0 +1,9 @@
+class CreateOrganizations < ActiveRecord::Migration
+  def change
+    create_table :organizations do |t|
+      t.string :name
```
lagartoflojo: Since `name` is a required field, it also makes sense to put it here:
`t.string :name, null: false`

db/migrate/20170531094819_move_conferences_to_organizations.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    TempConference.reset_column_information
+    organization = TempOrganization.create(name: 'organization', description: 'Default organization to migrate old conferences to the new version of OSEM')
+    TempConference.all.each do |conference|
+      conference.organization_id = organization.id
+      conference.save!
+    end
```
lagartoflojo: Take a look at [`update_all`](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/Relation.html#method-i-update_all)

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  let!(:organization) { create(:organization) }
+  let!(:admin) { create(:admin, is_admin: true) }
```
lagartoflojo: Since you are using the `:admin` factory, the `is_admin: true` parameter is not necessary.
Either way, I think here we should be testing with a regular user, right?

spec/controllers/admin/organizations_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  let(:admin) { create(:admin) }
```
lagartoflojo: I think we're missing some tests to make sure that only admins have access to these routes.

spec/models/organization_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+describe Organization do
+  let(:organization) { create(:organization) }
+
+  describe 'validation' do
+    it 'has a valid factory' do
```
lagartoflojo: This test is not necessary, since we have https://github.com/openSUSE/osem/blob/master/lib/tasks/factory_girl.rake

app/views/admin/organizations/_form.html.haml
shlok007: https://github.com/justinfrench/formtastic#the-available-inputs It says `:text` is used for a textarea in formtastic.

spec/controllers/organizations_controller_spec.rb
shlok007: I forgot to change the tests as well after making suggested changes. I'll fix this.

spec/controllers/organizations_controller_spec.rb
shlok007: Yes, I'll use a regular user here and admin in tests for `Admin::OrganizationsController`

db/migrate/20170531094819_move_conferences_to_organizations.rb
shlok007: I am not sure how to check that.
Shouldn't this work the same:
`add_reference :conferences, :organization, foreign_key: true, index: true`

Looks like passing the `foreign_key: true` option to `add_reference` execute `add_foreign_key` statement as well:
https://github.com/rails/rails/blob/d09a9bc07ddaf3c570aced434ab64cf06860678c/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L668

db/migrate/20170531094819_move_conferences_to_organizations.rb
shlok007: Rubocop warns not to use `update_all` because it skips model based validations.

https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/rails/skips_model_validations.rb#L34

app/views/admin/organizations/_form.html.haml
lagartoflojo: Ahhh magic gems 😛

db/migrate/20170531094819_move_conferences_to_organizations.rb
lagartoflojo: Yup, that makes sense!

db/migrate/20170531094819_move_conferences_to_organizations.rb
lagartoflojo: Yes, the `foreign_key` option should work. I was actually thinking about the `null: false` option. We should have it, but I don't think it will work if the DB already has conferences.
I'm thinking that if the database already contains conferences, upon creating the `organization_id` field, it will be NULL, but we should not allow NULL values (hence we should use `null: false`). But then the field won't be created. So we need to create it first without the `null: false` option, fill it with the default organization's id, and then add the `null:  false` option.

shlok007: Something in terms of:
```ruby
add_reference :conferences, :organization, index: true  
# migrate old data..  
add_foreign_key :conferences, :organizations, null: false, on_delete: :cascade
```

app/controllers/admin/conferences_controller.rb
```diff
@@ -61,11 +61,19 @@ def index
...
 
     def create
-      @conference = Conference.new(conference_params)
-
+      conference_params_copy = conference_params
+      if ENV['ORGANIZATIONS_ENABLED'] != 'true'
+        org = Organization.exists?(name: 'default') ? Organization.find_by(name: 'default') : Organization.create(name: 'default')
+        conference_params_copy[:organization_id] = org.id
+      end
+      @conference = Conference.new(conference_params_copy)
```
lagartoflojo: I was thinking we could do something simpler.
How about just finding the default organization and assigning it to the conference? We don't need to check any env variables.
Also, we know that the default organization exists, since it's [created in the migration](https://github.com/openSUSE/osem/pull/1517/files#diff-44200c9e4f9c94556ad974cc5653a0bbR15) (btw, you called it "organization", not "default").

app/controllers/admin/conferences_controller.rb
```diff
@@ -61,11 +61,19 @@ def index
...
+      @organizations = {}
+      Organization.all.each do |organization|
+        @organizations.store(organization.name, organization.id) if can? :create, Conference.new(organization: organization)
+      end
     end
 
     def create
-      @conference = Conference.new(conference_params)
-
+      conference_params_copy = conference_params
```
lagartoflojo: This will not really "copy" the parameters. Though, I'm not clear why you would need to do it at all.

app/views/admin/conferences/new.html.haml
```diff
@@ -6,6 +6,8 @@
                           input_html: { required: 'required' }
         = f.input :short_title, hint: "A short and unique handle for your conference, using only letters, numbers, underscores, and dashes. This will be used to identify your conference in URLs etc. Example: 'froscon2011'",
                                 input_html: { required: 'required', pattern: '[a-zA-Z0-9_-]+', title: 'Only letters, numbers, underscores, and dashes.' }, prepend: conferences_url + '/'
+        - if ENV['ORGANIZATIONS_ENABLED'] == 'true'
+          = f.input :organization, as: :select, collection: @organizations
```
lagartoflojo: Let's just assign the default organization in the controller for now. No need to display it in the UI.

spec/models/conference_spec.rb
```diff
@@ -1630,7 +1630,7 @@
   end
 
   describe 'after_create' do
-    let(:conference) { Conference.new(title: 'ABC', short_title: 'XYZ', start_date: Date.today, end_date: Date.today + 10, timezone: 'GMT') }
+    let(:conference) { create(:conference) }
```
lagartoflojo: nice

app/controllers/admin/conferences_controller.rb
```diff
@@ -61,11 +61,19 @@ def index
...
+      if ENV['ORGANIZATIONS_ENABLED'] != 'true'
+        org = Organization.exists?(name: 'default') ? Organization.find_by(name: 'default') : Organization.create(name: 'default')
```
lagartoflojo: You can just use [`find_or_create_by`](http://api.rubyonrails.org/v4.2.5/classes/ActiveRecord/Relation.html#method-i-find_or_create_by).

app/controllers/admin/conferences_controller.rb
shlok007: That's better. I have done exactly that now.

app/controllers/admin/conferences_controller.rb
shlok007: Thank you!

shlok007:
I have commented the failing tests here. I'll make a follow up PR as soon as this gets merged with work related to introducing `organization admins` and appropriate permissions. Also, fixing the commented failing tests..

@lagartoflojo Please do let me know if there are any more changes to be made here before it can be merged. :slightly_smiling_face:

lagartoflojo:
@shlok007 I'm confused, why do we want to comment the failing tests instead of fixing the tests/code...?

shlok007:
I thought it would be better to work on appropriate permissions for the failing `destroy` action first. I am confused as to why this is failing and not giving appropriate permission *might* be a cause. Since, we are not working on  permissions here in this PR I thought it would be better to let it be and fix it in the PR after introducing the correct permissions which is really more important at the moment and this could be fixed along with it.

lagartoflojo:
But you don't know if permissions is the real reason why this is failing... We have to understand why this is failing, and then think about how to tackle it.

shlok007:
Okay sure, I probably should look into it more.

shlok007:
@lagartoflojo I have made the required changes. Let me know if there is anything else required here before we can merge this :slightly_smiling_face:

Issue #462, opened 2014-08-18
Post-Conference Dashboard


hennevogel:
Post-Conference the dashboard should switch focus:

- [x] "Drop" progress
- [ ] "Drop" Recent Submissions/Registrations
- [ ] Add Pre-registered users vs. At-Event-Registrations
- [ ] Add Attended vs. Not-Attended
- [ ] Add statistics about the Questions 
- [ ] Sold Tickets

differentreality:
Doughnut charts for tickets are already implemented in tickets#index, they can be made visible in conferences#index, when the conference is over.

PR #1520, opened 2017-06-04
Hide recent Registrations/Submissions when conference is over

selini:
Part of #462

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -33,6 +33,50 @@ module ApplicationHelper
     result
   end
 
+  # Returns time with conference timezone
+  def time_with_timezone(time)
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -33,6 +33,50 @@ def date_string(start_date, end_date)
     result
   end
 
+  # Returns time with conference timezone
+  def time_with_timezone(time)
```
differentreality: These functions should probably not be here

differentreality:
@selini

app/helpers/application_helper.rb
selini: ok i will remove it

selini: done

differentreality:
Nice work @selini :+1:

Issue #1522, opened 2017-06-04
Replace Date.today with Date.current in the rspec specs


AEtherC0r3:
The test in https://github.com/openSUSE/osem/blob/master/spec/models/program_spec.rb#L154-L157 is passing because of a false negative. More specifically, in line 155 it uses build instead of create, so the new cfp isn't saved in the db, therefore Program#cfp_open? returns false because cfp.psesent? is false in https://github.com/openSUSE/osem/blob/master/app/models/program.rb#L140 and not because the cfp is over. If create is used instead of build then it will fail (if Date.today > Date.current). Maybe Date.current should be used instead of Date.today in general, since Date.today takes into account the server's local timezone.

Ana06:
@AEtherC0r3 good catch! the cfp should definitely be saved in thr test and `Date.current` should be used, as it is what is also used in the `cfp_open` method. 

@AEtherC0r3 will send a PR? :wink:

There is also a nice RoboCop cop to ensure that the timezone it is always used: https://rubocop.readthedocs.io/en/latest/cops_rails/#railstimezone. It would also be nice to enable it with the *flexible* option. I think there would be some offenses tricky to solve, so this should be made in a different PR. @AEtherC0r3 would you also like to do this or should I create a new issue?

AEtherC0r3:
@Ana06 I will do both of them (the fix and the cop)

Ana06:
@AEtherC0r3 great! 😉

greggawatt:
is this issued fixed?

Ana06:
No, this part is not done:

> There is also a nice RoboCop cop to ensure that the timezone it is always used: https://rubocop.readthedocs.io/en/latest/cops_rails/#railstimezone. It would also be nice to enable it with the flexible option. I think there would be some offenses tricky to solve, so this should be made in a different PR. @AEtherC0r3 would you also like to do this or should I create a new issue?

@AEtherC0r3 do you still plan to work on this? :thinking:

AEtherC0r3:
I still want to work on that but I don't think that is going to happen soon. So, @greggawatt if you want to do it, then go ahead :)

kkavita92:
@Ana06 happy to jump on this if it's still open! Let me know :)

differentreality:
@kkavita92 you can go ahead and enable the cop!

Anthematics:
[...]

PR #1526, opened 2017-06-06
Replace .today with .current in program spec

AEtherC0r3:
The cfp_open? method uses Date.current, so it should also be tested with
Date.current and not Date.today
Also, build was replaced with create to fix a false negative

This is a partial fix for #1522


differentreality:
Looks good :+1:

Ana06:
:tada:

PR #1530, opened 2017-06-07
Show conference changelog

differentreality:
This comes from #1124, it is rebased, and has some fixes.

* Use load_and_authorize_resource in versions controlller
* Add conference specific route to revision history page
* Users with role can view revision_history only for the versions they have access to
* Handle versions where conference_id is not set (records before papertrail was introduced)

### DIFF
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,17 +1,17 @@
 module Admin
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_resource :conference, find_by: :short_title
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+
+      return if @conference.blank?
+      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
+      @versions = @versions.where(conference_id: @conference.id)
     end
 
     def revert_attribute
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_attribute, @version
-
       if params[:attribute] && @version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.keys.include?(params[:attribute])
         if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
           flash[:error] = 'The item is already in the state that you are trying to revert it back to'
@@ -33,9 +33,6 @@ module Admin
     end
 
     def revert_object
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_object, @version
-
       if @version.event != 'create'
         if @version.reify.save
           flash[:notice] = 'The selected change was successfully reverted'
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -4,7 +4,7 @@ module Admin
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
 
       return if @conference.blank?
       @versions = PaperTrail::Version.where(conference_id: @conference.id).accessible_by(current_ability)
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -7,8 +7,7 @@ module Admin
       @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
 
       return if @conference.blank?
-      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
-      @versions = @versions.where(conference_id: @conference.id)
+      @versions = PaperTrail::Version.where(conference_id: @conference.id).accessible_by(current_ability)
     end
 
     def revert_attribute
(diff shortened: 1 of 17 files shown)


.rubocop.yml
```diff
@@ -127,6 +127,7 @@ Metrics/ClassLength:
 Metrics/BlockLength:
   Exclude:
     - 'spec/models/conference_spec.rb'
+    - 'spec/features/abilities_spec.rb'
```
shlok007: I think you meant `spec/features/ability_spec.rb`

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,17 @@
...
+    load_resource :conference, find_by: :short_title
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+
+      return if @conference.blank?
```
lagartoflojo: Can `@conference` really be blank? If it's not found by `load_resource`, then you get a 404.

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,17 @@
...
+      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
```
shlok007: The abilities defined for `PaperTrail::Version` in [ability.rb](https://github.com/openSUSE/osem/pull/1530/files#diff-e9c04d21b8c4d777e4231c4c6ce5f8cbR233)  means that this will return false unless `item_type` is specified.
  
```ruby
cfp_role_ability.can? PaperTrail::Version.new(conference_id: @conference.id) #=> false
cfp_role_ability.can? PaperTrail::Version.new(conference_id: @conference.id, item_type: 'Event') #=> true
```
Therefore, only organizers will be able to access the `index` action for a specific conference.

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,17 @@
 module Admin
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_resource :conference, find_by: :short_title
```
shlok007: I think we only need to load conference for the `index` action

app/controllers/admin/versions_controller.rb
differentreality: You can access /admin/revision_history outside of a conference. Code after this line does not get executed, cancancan with load_and_authorize on top has already given us a @versions variable with all the versions accessible by our user, which we show when there is no conference selected.

app/helpers/versions_helper.rb
```diff
@@ -2,35 +2,80 @@ module VersionsHelper
...
+  end
+
+  def link_to_user(user_id)
+    return 'Someone (probably via the console)' unless user_id
+
+    user = User.find_by(id: user_id)
+    if user
+      link_to user.name, admin_user_path(id: user_id)
+    else
+      name = current_or_last_object_state('User', user_id).try(:name) || PaperTrail::Version.where(item_type: 'User', item_id: user_id).last.changeset['name'].second if PaperTrail::Version.where(item_type: 'User', item_id: user_id).any?
```
lagartoflojo: I really don't like this line, but it seems PaperTrail doesn't let you `reify` the initial version of an object so there's no better way around it...

app/helpers/versions_helper.rb
```diff
@@ -2,35 +2,80 @@ module VersionsHelper
...
   # Groups functions related to change description
   ##
-  def link_if_alive(version, link_text, link_url)
-    version.item ? link_to(link_text, link_url) : link_text
+  def link_if_alive(version, link_text, link_url, conference)
+    version.item && conference ? link_to(link_text, link_url) : "#{link_text} with ID #{version.item_id}"
+  end
+
+  def link_to_conference(conference_id)
+    return 'deleted conference' if conference_id.nil?
```
differentreality: We did not have conference_id attribute in versions since the beginning, that combined with records deleted directly from the DB can make it impossible for the conference_id attribute to be calculated.
I have 9 records (with event_type Splashpage, Venue, RegistrationPeriod) that don't have a conference_id. My DB comes from events.o.o which is where I pinpointed all those special cases.

app/helpers/versions_helper.rb
```diff
@@ -2,35 +2,80 @@ module VersionsHelper
...
+
+    conference = Conference.find_by(id: conference_id)
+    if conference
+      link_to conference.short_title,
+              edit_admin_conference_path(conference.short_title)
+    else
+      short_title = current_or_last_object_state('Conference', conference_id).try(:short_title) || ''
+      " #{short_title} with ID #{conference_id}"
```
differentreality: My reasoning for introducing this is that when we look at the changelog it is very possible that we want to check who did what (possibly something they shouldn't have done). If the record is deleted, or if things got deleted in a way they shouldn't have, we won't have proper access to the information we want. An ID might prove helpful to allow for organizers to further investigate what happened manually (the conference_id will allow us to manually search the versions in our DB for the item of this particular conference. If the conference record is deleted, we cannot get that information only with its short_title).

app/helpers/versions_helper.rb
```diff
@@ -2,35 +2,80 @@ module VersionsHelper
...
+    end
+  end
+
+  def link_to_user(user_id)
```
differentreality: Oops! I will delete the one in pathshelper as we don't use it anywhere else, but versions.

app/models/ability.rb
```diff
@@ -250,9 +249,10 @@ def signed_in_with_cfp_role(user)
       (Conference.with_role(:cfp, user).pluck(:id).include? role.resource_id)
     end
 
-    can [:index, :revert_object, :revert_attribute], PaperTrail::Version, item_type: 'Event', conference_id: conf_ids_for_cfp
-    can [:index, :revert_object, :revert_attribute], PaperTrail::Version, item_type: 'Vote', conference_id: conf_ids_for_cfp
-    can [:index, :revert_object, :revert_attribute], PaperTrail::Version do |version|
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version,
+        item_type: %w(Event EventType Track DifficultyLevel EmailSettings Room Cfp Program Comment), conference_id: conf_ids_for_cfp
+    can [:index, :revert_object, :revert_attribute], PaperTrail::Version,
+        ["item_type = 'Commercial' AND conference_id IN (?) AND (object LIKE '%Event%' OR object_changes LIKE '%Event%')", conf_ids_for_cfp] do |version|
```
differentreality: This allows users with role cfp to access commercials of events (only of events, not of conferences or venues). This is the SQL equivalent for that, and we need it, in the case of a block definition of ability, for accessible_by to work
https://github.com/openSUSE/osem/pull/1124/files#r75135590
https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities-with-Blocks#fetching-records

differentreality:
@lagartoflojo do you see any more exceptions thrown? 
@shlok007 wants to jump in and contribute to the PR as well - which is more than fine with me. 

Let me know if we can squash the commits.

PR #1534, opened 2017-06-09
Add: Allow user to remove openid

### LINKED ISSUE #1468: Allow users to remove openid
Currently a user can add multiple openids from the user Account settings. However those openids cannot be removed, they should.

dPliakos:
Allow user to delete an openid connection. Inserts a _font awesome_ "times" (x) right to the openid.
Fixes #1468


app/views/devise/registrations/edit.html.haml
```diff
@@ -15,7 +15,10 @@
           %h4
             Currently the following openIDs are associated with your account
           - @openids.each do |openid|
-            %li= "#{openid.provider}:#{openid.email}"
+            %li
+              %span "#{openid.provider}:#{openid.email}"
+              = link_to user_openid_path(openid, user_id: current_user.id), method: :delete do
```
differentreality: We should probably show a confirmation message first, before proceeding to the deletion.

app/views/devise/registrations/edit.html.haml
```diff
@@ -15,7 +15,10 @@
...
+                %span.fa.fa-times
```
differentreality: I'd make this red color, and move it in front of the openid

app/views/devise/registrations/edit.html.haml
```diff
@@ -15,7 +15,10 @@
           %h4
             Currently the following openIDs are associated with your account
           - @openids.each do |openid|
-            %li= "#{openid.provider}:#{openid.email}"
+            %li
+              %span "#{openid.provider}:#{openid.email}"
```
differentreality: We don't really need the quotes on the text that we show

app/views/devise/registrations/edit.html.haml
dPliakos: Does OSEM use any defaults to confirm an action?

differentreality: You need to **manually** add the message

app/views/devise/registrations/edit.html.haml
dPliakos: The list appears a bit funny with the `X` moved to the left. 
We could remove the list bullets and use only the `X` (having both list-style role and a button role)

screenshot from 2017-06-11 15-10-12_c
screenshot from 2017-06-11 15-13-16

differentreality: Let's do that!

differentreality:
Looks good @dPliakos :D

Please squash your commits for merging!

differentreality:
Nice work :+1:

Issue #1514, opened 2017-06-02
Introduce new role 'organization admin' to manage organizations


shlok007:
Related PR: #1538
The dotted/dashed line represents the permission for specific organization/conference the user belongs to whereas solid lines represents permission to create/manage all.
 
![organization_admin_role](https://cloud.githubusercontent.com/assets/14155445/26739308/ec942bca-47ee-11e7-8db3-a06b003388b7.png)

shlok007:
Since, `organization admin` will be presiding over all the conferences and `organizers` are restricted to their conference only, it makes sense __not__ to give permission to `organizers` to create new conference anymore. 

A conference then, can __only__ be created by the site admin and the organization admin in the hosted version.

lagartoflojo:
I think, at least for now, there should only be an arrow from Site Admin -> Create organization, since that's the only thing we're allowing, right?

lagartoflojo:
As for disallowing `organizers` from creating conferences, that probably makes sense.
So does the `organization admin` choose who the `organizer`s for each conference are?

lagartoflojo:
Can you please explain what the different lines (solid, dashed, dotted) mean?

shlok007:
Since, we have made a few changes now, I'll update the diagram. The dotted/dashed line represents the permission for specific organization/conference the user belongs to whereas solid lines represents permission to create/manage all. 

> So does the organization admin choose who the organizers for each conference are?

`organization admin` behave as just as organizers do for all it's conferences. So, yes.

There can be two possibilities here:

- Give permission to `organization admins` to manage all conferences that belongs to his/her organization so when a new conference is created, all the `organization admins` will be able to manage all parts of that conference exactly as a `organizer` can. They can add an organizer( or any other role ) to its conference from the Roles section in the admin sidebar for a conference.

- As soon as an `organization admin` creates a new conference, all the `organization admins` of that organization will be assigned the  `organizer` role of that conference. In this way we wouldn't have to explicitly define permissions for `organization admin` to manage parts of the conferences since, we already have that set up for `organizers`.

No matter what we decide on, the permissions* for the rest of the roles won't be affected so they will behave as it is. So, yes `organizers` will be able to add more `organizers` just like they do now. 

*The only change will be that organizers will not be able to create conferences now!

lagartoflojo:
I think option 1. is better - organization admins have full control of everything under their organization.
[...]

PR #1538, opened 2017-06-12
Introduce organization admins

shlok007:
Reference: #1514 
This PR contains the back-end for introducing a new role `organization admins` in OSEM.

Current abilities of `organization admins` are:
- To `:manage` his/her own organization
- Can `:new` conferences.
- To create and `:manage` all conferences within his/her organization
- Abilities regarding each conference in the organization will be same as the organizers of that conference.

The changed ability of `organizers` are:
- Can no longer `:create` and  `:new` conferences


app/models/organization.rb
```diff
@@ -1,7 +1,17 @@
...
+  after_create :create_roles
+
   validates :name, presence: true
 
   mount_uploader :picture, PictureUploader, mount_on: :picture
+
+  private
+
+  def create_roles
+    Role.where(name: 'organization_admin', resource: self).first_or_create(description: "For the administrators of an organization (who shall have full access to the organization and it's conferences)")
```
lagartoflojo: it's == it is
its == possessive form of it

lagartoflojo: Maybe make the description shorter: "For the administrators of an organization and its conferences"

lagartoflojo: I think instances of this class should now have a `#roles` method, right?
Could we not say then: `roles.where(name: 'organization_admin').first_or....`

app/controllers/admin/organizations_controller.rb
```diff
@@ -45,6 +46,10 @@ def destroy
 
     private
 
+    def assign_role
+      current_user.add_role :organization_admin, @organization
```
shlok007: Yes, I'll remove this.

app/models/organization.rb
shlok007: Done! :slightly_smiling_face:

shlok007: thank you! :v:

lagartoflojo:
@shlok007 we have some conflicts 😱

shlok007:
Oops! I'll try not to rebase without messing up anything! :sweat_smile:

shlok007:
Rebased! :v:

spec/features/organization_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'spec_helper'
+
+feature Organization do
+  let!(:organization) { create(:organization) }
+  let!(:organization_admin_role) { Role.find_by(name: 'organization_admin', resource: organization) }
+  let(:organization_admin) { create(:user, role_ids: [organization_admin_role.id]) }
+  let(:admin_user) { create(:admin) }
+
+  shared_examples 'successfully updates a organization' do
```
lagartoflojo: *an* organization

spec/models/ability_spec.rb
```diff
@@ -138,9 +140,14 @@
 
     shared_examples 'user with any role' do
       before do
-        @other_conference = create(:conference)
+        @other_organization = create(:organization)
+        @other_conference = create(:conference, organization: @other_organization)
```
lagartoflojo: Instead of instance variables, why not use `let!`?

spec/features/ability_spec.rb
```diff
@@ -22,656 +26,636 @@
...
+    end
+  end
 
-    create(:schedule, program: conference3.program)
-    visit admin_conference_schedule_path(conference3.short_title, conference3.program.schedules.first)
-    expect(current_path).to eq(root_path)
+  context 'when user is cfp' do
+    before do
+      sign_in user_cfp
+      @conference = conference2
```
lagartoflojo: is `@conference` used anywhere?

spec/features/ability_spec.rb
```diff
@@ -22,656 +26,636 @@
...
+      expect(current_path).to eq(admin_conference_program_tracks_path(conference2.short_title))
+
+      visit admin_conference_roles_path(conference2.short_title)
+      expect(current_path).to eq(admin_conference_roles_path(conference2.short_title))
+
+      visit admin_conference_emails_path(conference2.short_title)
+      expect(current_path).to eq(admin_conference_emails_path(conference2.short_title))
+    end
 
-    visit admin_confer
```
lagartoflojo: Almost 700 lines!!! 😱 
How about we split this file into different files, one for each role?

lagartoflojo: Almost 700 lines!!! 😱 
How about we split this file into different files, one for each role?

spec/models/ability_spec.rb
```diff
@@ -180,8 +197,10 @@
         should be_able_to(:destroy, my_venue)
       end
 
-      it{ should be_able_to(:new, Conference) }
-      it{ should be_able_to(:create, Conference) }
+      it{ should_not be_able_to(:new, Organization)}
+      it{ should_not be_able_to(:create, Organization)}
+      it{ should_not be_able_to(:new, Conference.new) }
+      it{ should_not be_able_to(:create, Conference.new) }
```
lagartoflojo: why `Conference.new` instead of just `Conference`?

app/models/ability.rb
```diff
@@ -146,57 +148,76 @@ def signed_in_with_roles(user)
...
+  def signed_in_with_organization_admin_role(user)
+    org_ids_for_organization_admin = Organization.with_role(:organization_admin, user).pluck(:id)
+
+    can :manage, Organization, id: org_ids_for_organization_admin
+    can :new, Conference
+    can :manage, Conference, organization_id: org_ids_for_organization_admin
+    conf_ids_for_organization_admin = []
+    org_ids_for_organization_admin.each do |org_id|
+      conf_ids_for_organization_admin += Organization.find(org_id).conferences.pluck(:id)
+    end
```
lagartoflojo: I think these lines can be reduced to:

```
conf_ids_for_organization_admin = Conference.where(organization_id: org_ids_for_organization_admin).pluck(:id)
```

app/models/ability.rb
```diff
@@ -146,57 +148,76 @@ def signed_in_with_roles(user)
...
+    can [:index, :show], Role
+    can [:edit, :update], Role do |role|
+      role.resource_type == 'Organization' && (org_ids_for_organization_admin.include? role.resource_id)
+    end
+    signed_in_with_organizer_role(user, conf_ids_for_organization_admin)
+  end
+
+  def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
```
lagartoflojo: I would just rename `conf_ids_for_organization_admin_and_organizer` as `conf_ids`... such a long name makes the code hard to read, and it's obvious anyway.

shlok007: sure! will do!

spec/features/ability_spec.rb
shlok007: That sounds better. I'll do exactly that!

spec/models/ability_spec.rb
```diff
@@ -180,8 +195,10 @@
         should be_able_to(:destroy, my_venue)
       end
 
-      it{ should be_able_to(:new, Conference) }
-      it{ should be_able_to(:create, Conference) }
+      it{ should_not be_able_to(:new, Organization)}
+      it{ should_not be_able_to(:create, Organization)}
+      it{ should_not be_able_to(:new, Conference)}
+      it{ should_not be_able_to(:create, Conference) }
```
lagartoflojo: I think I figured out why these tests are failing.
From the [CanCanCan wiki](https://github.com/CanCanCommunity/cancancan/wiki/checking-abilities#checking-with-class):

> Important: If a block or hash of conditions exist they will be ignored when checking on a class, and it will return true. For example:
>
> ```ruby
> can :read, Project, :priority => 3
> can? :read, Project # returns true
> ```

Since the ability for Organizers is defined as

```ruby
can :manage, Conference, id: conf_ids
```

then

```ruby
can? :manage, Conference
```

will return `true`.
That's why the test **should** use `.new` in this case.

app/models/ability.rb
```diff
@@ -146,57 +148,73 @@ def signed_in_with_roles(user)
...
-    can :manage, Splashpage, conference_id: conf_ids_for_organizer
-    can :manage, Contact, conference_id: conf_ids_for_organizer
-    can :manage, EmailSettings, conference_id: conf_ids_for_organizer
-    can :manage, Campaign, conference_id: conf_ids_for_organizer
-    can :manage, Target, conference_id: conf_ids_for_organizer
+  def signed_in_with_organization_admin_role(user)
+    org_ids_for_organization_admin = Organization.with_role(:organization_admin, user).pluck(:id)
+
+    can :manage, Organization, id: org_ids_for_organization_admin
+    can :new, Conference
```
lagartoflojo: I think we also need a test for that.

app/models/ability.rb
```diff
@@ -146,57 +148,73 @@ def signed_in_with_roles(user)
...
+    end
+    signed_in_with_organizer_role(user, conf_ids_for_organization_admin)
+  end
+
+  def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
+    # ids of all the conferences for which the user has the 'organizer' role and
+    # conferences that belong to organizations for which user is 'organization_admin'
+    conf_ids = conf_ids_for_organization_admin.concat(Conference.with_role(:organizer, user).pluck(:id)).uniq
+    can :manage, Resource, conference_id: conf_ids
+    can :manage, Conference, id: conf_ids
```
shlok007: Following the wiki, it looks like the organizers do have permission to access `new` action of Conference:  
  
If I run this on rails console:
```ruby
Ability.new(user_organizer).can? :new, Conference
```
It returns ` true`  
 
I didn't sense anything was wrong here because if I try to access `admin/conference#new` signed in as an organizer, I get access denied error and am redirected to the root page which is exactly what we want. _strange!_

shlok007: However, making this change does pass the failing tests we have [here](https://github.com/openSUSE/osem/pull/1538#discussion-diff-125170064R201) for organizers of a conference even though there exist a hash condition. 
```ruby 
it{ should_not be_able_to(:new, Conference) }
it{ should_not be_able_to(:create, Conference) }
```
This again seem to contradict the wiki. :sweat_smile:

lagartoflojo: > if I try to access admin/conference#new signed in as an organizer, I get access denied error and am redirected to the root page which is exactly what we want. strange!

That's perfectly fine, as `load_and_authorized_resource` [checks the permissions on `Conference.new`](http://www.rubydoc.info/github/CanCanCommunity/cancancan/CanCan%2FControllerAdditions%2FClassMethods:load_resource), not on the `Conference` class.

lagartoflojo: > If I run this on rails console:
> Ability.new(user_organizer).can? :new, Conference
> It returns true

That's right. Don't confuse the `:new` permission with checking permissions on `Conference.new` :)
The documentation says:

```ruby
Ability.new(user_organizer).can? :new, Conference # => true
Ability.new(user_organizer).can? :new, Conference.new # => false, this is what load_resource does
```

So the test has to check for the `:new` permission on `Conference.new`

app/models/ability.rb
shlok007: Since, a user can be an org admin of many organizations, I thought we can allow choosing the organization in the `conference#new` page. That way, the permission for the `:new` action shall be given to all organization admins. However, in the form, they will get a drop down of only those organizations in which they belong to.

app/models/ability.rb
shlok007: >Don't confuse the :new permission with checking permissions on Conference.new :)

Thank you. I get it now. This was exactly what I was doing wrong, :sweat_smile:

spec/models/ability_spec.rb
```diff
@@ -137,17 +139,20 @@
...
-        it{ should_not be_able_to(:toggle_user, Role.find_by(name: role, resource: @other_conference)) }
-        it{ should_not be_able_to(:update, Role.find_by(name: role, resource: @other_conference)) }
-        it{ should_not be_able_to(:edit, Role.find_by(name: role, resource: @other_conference)) }
-        it{ should be_able_to(:show, Role.find_by(name: role, resource: @other_conference)) }
-        it{ should be_able_to(:index, Role.find_by(name: role, resource: @other_conference)) }
-      end
+        let!(:other_organization) { create(:organization) }
+        let!(:other_conference) { create(:conference, organization: other_organization) }
+
+        it{ should_not be_able_to(:update, Role.find_by(name: 'organization_admin', resource: other_organization)) }
```
AEtherC0r3: These lines don't seem to be indented correctly

PR #1539, opened 2017-06-13
Sample prototype for ticket pdf

siddhantbajaj:
Need some feedback for the design of ticket pdf for the conference.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -88,7 +88,7 @@ class Ability
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
     can [:new, :create], Payment, user_id: user.id
-    can [:index, :show], PhysicalTicket, user_id: user.id
+    can [:index, :show], PhysicalTicket, user: user
 
     can [:create, :destroy], Subscription, user_id: user.id
 
--- a/app/uploaders/picture_uploader.rb
+++ b/app/uploaders/picture_uploader.rb
@@ -52,6 +52,10 @@ class PictureUploader < CarrierWave::Uploader::Base
     "system/#{object_class_name}/#{mounted_as}/#{model.id}"
   end
 
+  def image
+    @image ||= MiniMagick::Image.open(file.file)
+  end
+
   # Create different versions of your uploaded files:
   version :large do
     process resize_to_fit: [300, 300]
--- a/app/views/physical_ticket/show.pdf.prawn
+++ b/app/views/physical_ticket/show.pdf.prawn
@@ -1,0 +1,62 @@
+prawn_document(filename: @file_name, page_layout: :portrait, :page_size =>'A4' ) do |pdf|
+  # Vertical Layout
+  top = pdf.bounds.top
+  bottom = pdf.bounds.bottom
+  left = pdf.bounds.left
+  right = pdf.bounds.right
+  mid_vertical = (pdf.bounds.top-pdf.bounds.bottom)/2
+  mid_horizontal = (pdf.bounds.right-pdf.bounds.left)/2
+  x = 0
+
+  pdf.move_down mid_vertical
+  pdf.dash(2, :space => 1)
+  pdf.stroke_horizontal_rule
+  pdf.stroke_vertical_line pdf.bounds.top, pdf.bounds.bottom, :at => mid_horizontal
+  pdf.move_up mid_vertical
+  pdf.draw_text "TICKET HOLDER", :at => [x,pdf.cursor-30], :size => 17
+  pdf.dash(2, :space => 0)
+  pdf.stroke_rectangle  [x, pdf.cursor-50], 230, 150
+  pdf.move_down 80
+  pdf.draw_text "NAME", :at => [x+10,pdf.cursor], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.name}", :at => [x+10,pdf.cursor-25], size: 20
+  pdf.fill_color "000000"
+  pdf.draw_text "EMAIL", :at => [x+10,pdf.cursor-50], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.email}", :at => [x+10,pdf.cursor-75], size: 20
+  pdf.fill_color "000000"
+  pdf.move_up 20
+  if @conference.picture?
+    if 7 * @conference.picture.image[:width] > 12 * @conference.picture.image[:height]
+      pdf.image "#{Rails.root}/public#{@conference.picture_url}", :at => [mid_horizontal+30, pdf.cursor], :width => 120
+    else
+      pdf.image "#{Rails.root}/public#{@conference.picture_url}", :at => [mid_horizontal+30, pdf.cursor], :height => 70
+    end
+  else
+    pdf.image "#{Rails.root}/public/img/osem-logo.png", :at => [mid_horizontal+30, pdf.cursor], :height => 70
+  end
+  pdf.move_down 70
+  pdf.draw_text "#{@conference.title}", :at => [mid_horizontal+30,pdf.cursor-30], :size => 12
+  pdf.draw_text "#{@conference.organization.name}", :at => [mid_horizontal+30,pdf.cursor-50], :size => 12
+  pdf.draw_text "#{@conference.venue.name}", :at => [mid_horizontal+30,pdf.cursor-70]
+  pdf.move_up 130
+  pdf.move_down mid_vertical
+  pdf.draw_text "EVENT", :at => [x,pdf.cursor-40], :size => 15
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@conference.title}", :at => [x,pdf.cursor-60], size: 12
+  pdf.draw_text "#{@conference.start_date.strftime('%B %d, %Y')}", :at => [x,pdf.cursor-80], size: 12
+  pdf.move_down 80
+  pdf.fill_color "000000"
+  pdf.draw_text "TICKET", :at => [x,pdf.cursor-30], :size => 15
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@physical_ticket.ticket.title}", :at => [x,pdf.cursor-50], size: 12
+  pdf.move_down 50
+  pdf.fill_color "000000"
+  pdf.draw_text "TICKET REF.", :at => [x,pdf.cursor-30], :size => 15
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@physical_ticket.ticket_purchase.id}", :at => [x,pdf.cursor-50], size: 12
+  pdf.move_down 50
+  pdf.fill_color "000000"
+  pdf.draw_text "Powered By OSEM", :at => [(mid_horizontal-left-100)/2,pdf.cursor-100], :size => 11
+  pdf.move_up 180
+end
--- a/Gemfile
+++ b/Gemfile
@@ -128,6 +128,8 @@ gem 'country_select'
 
 # as PDF generator
 gem 'prawn_rails'
+gem 'rqrcode'
+gem 'prawn-qrcode', '~> 0.2.2.1'
 
 # to render XLS spreadsheets
 gem 'axlsx', git: 'https://github.com/randym/axlsx.git'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -117,6 +117,7 @@ GEM
     chart-js-rails (0.0.6)
       railties (> 3.1)
     chronic (0.10.2)
+    chunky_png (1.3.8)
     cliver (0.3.2)
     cloudinary (1.1.6)
       aws_cf_signer
@@ -343,6 +344,9 @@ GEM
     prawn (1.0.0)
       pdf-core (~> 0.2.2)
       ttfunk (~> 1.1.1)
+    prawn-qrcode (0.2.2.1)
+      prawn (>= 0.11.1)
+      rqrcode (>= 0.4.1)
     prawn_rails (0.0.11)
       prawn (>= 0.11.1)
       railties (>= 3.0.0)
@@ -429,6 +433,8 @@ GEM
       mime-types (>= 1.16, < 3.0)
       netrc (~> 0.7)
     rolify (5.1.0)
+    rqrcode (0.10.1)
+      chunky_png (~> 1.0)
     rspec (3.0.0)
       rspec-core (~> 3.0.0)
       rspec-expectations (~> 3.0.0)
@@ -611,6 +617,7 @@ DEPENDENCIES
   phantomjs
   piwik_analytics (~> 1.0.1)
   poltergeist
+  prawn-qrcode (~> 0.2.2.1)
   prawn_rails
   rails (~> 4.2)
   rails-assets-bootstrap-markdown!
@@ -631,6 +638,7 @@ DEPENDENCIES
   redcarpet
   responders (~> 2.0)
   rolify
+  rqrcode
   rspec-activemodel-mocks
   rspec-rails
   rubocop (~> 0.48.1)
--- a/app/controllers/physical_ticket_controller.rb
+++ b/app/controllers/physical_ticket_controller.rb
@@ -8,5 +8,8 @@ class PhysicalTicketController < ApplicationController
     @physical_tickets = current_user.physical_tickets.by_conference(@conference)
   end
 
-  def show; end
+  def show
+    @file_name = "ticket_for_#{@conference.short_title}"
+    @user = @physical_ticket.user
+  end
 end


siddhantbajaj:
SAMPLE 1

screen shot 2017-06-13 at 7 11 20 am

SAMPLE 2

screen shot 2017-06-13 at 7 07 52 am

differentreality:
Personally I prefer a variation of the 2nd design.
I'd make the QR code bigger, and make the 'by the power of OSEM' less visible.

lagartoflojo:
I couldn't find the original Tito example, could you link directly to it please?
Is the idea of the second design that attendees are able to fold it?
I think "By the power of OSEM" sounds... [a bit funny](https://www.youtube.com/watch?v=V8h8snfYidg). Maybe "Powered by OSEM"?

siddhantbajaj:
screen shot 2017-06-13 at 11 56 38 pm
You can find it here https://ti.to/demo/test-event.

siddhantbajaj:
@differentreality thanks for the feedback, I'll make those changes. @lagartoflojo yes "Powered by OSEM" sounds much better. 🙂

mdeniz:
I like both (or the other way around, I don't like one over the other), so, instead of having to choose what to implement and what not.... why not to let the creator of the conference select among some templates? or even better, to upload his own background design and have a way to place the data on top of it? 😄

lagartoflojo:
@mdeniz I'd go for a single, non-configurable format for now, and make it customizable in the future, to keep things simple.

Ana06:
@mdeniz I agree with @lagartoflojo I would try to find a good one. I don't think it is needed that we have several templates, as we will complicate the app and as you say there is not that much difference. In case that organisers request this later on, we can always implement it.

Ana06:
@siddhantbajaj In case that we want to use SAMPLE 2 I would use 1 of the 4 part just your the QR, as the Power by OSEM can be combined with any of the other parts and it is not that important.

So I would give make bigger and more visible what is more important, that is the QR and the info about the conference and the ticket you bought, so basically conference name, date and ticket type.

Ana06:
@ddemaio maybe you can also provide some feedback on this :wink:

ddemaio:
Thank you @Ana06. I like the direction. This seems logical. The persons name as well as the logo of the conference should be the most predominant element on the print out. I think that is what most people want is to see the name of the person as this has been expressed to me by several people because sometimes they remember faces, but can't remember the name. I like the second design the most, but we might want to consider two layouts. A vertical and a horizontal layout for the organizer to select since they will know what type of plastic badge holder they will purchase.

siddhantbajaj:
SAMPLE 3
(Taken from https://www.attendize.com)

screen shot 2017-06-20 at 5 21 01 am

screen shot 2017-06-20 at 5 21 42 am

siddhantbajaj:
Variations of sample 2:
1. Bigger QR code, Ticket holder information more predominant.

screen shot 2017-06-22 at 4 50 08 am

Ana06:
@siddhantbajaj it seems that SAMPLE 2 is the one most people prefer. The [SAMPLE 2.1](https://github.com/openSUSE/osem/pull/1539#issuecomment-310232290) looks already better, but I would like to see some of the suggestions made on it. What about a new variation of [2.1](https://github.com/openSUSE/osem/pull/1539#issuecomment-310232290) that has:

- Move TICKET HOLDER to the first square and increase the font of the name and email, to make this info more predominant as  @ddemaio says

- Move event details to third square now, but without QR. Add in this square also the `Powered by OSEM` test.

- Move QR to the 4 square and make it bigger. So the QR alone in the square.


And after make all this changes, make also an horizontal version of this as @ddemaio suggested. I think having one design in vertical and horizontal would make sense. :+1:

siddhantbajaj:
Sample 2 (Vertical layout):

screen shot 2017-06-24 at 3 05 53 am

Ana06:
@ I would like to have the QR in the center of its square, now it is like in a side and looks strange. But apart from that I like it :+1:  Not sure if the horizontal example has the right dimensions. It should be A4 paper (29.7x21cm), and seems to be more like a square.

@mdeniz @differentreality @ChrisBr @ddemaio lagartoflojo  what do you think about this version? :smile: :

lagartoflojo:
Agreed! The latest version looks nice, just needs to center the QR.

mdeniz:
Hey, I like that new version more 👍 

My nitpicks are:
- Remove the box that surrounds the QR code, it is not really needed
- Reduce the font size for the TICKET HOLDER phrase is too big, I think that @Ana06 meant about the font size of the name and email
- Change the color of the gray text to something more dark, close to be black but not that much, now is too light and it can be more difficult to read
- And.. maybe increasing the font of the data text and decreasing the labels text, so the important thing is the name, not the label name.

Ana06:
@mdeniz 

- Remove the box that surrounds the QR code, it is not really needed
- Reduce the font size for the TICKET HOLDER phrase is too big, I think that @Ana06 meant about the font size of the name and email
- Change the color of the gray text to something more dark, close to be black but not that much, now is too light and it can be more difficult to read
- And.. maybe increasing the font of the data texts and decreasing the label texts, so the important thing is the name, not the label "NAME".

:+1:

siddhantbajaj:
Horizontal Layout:
screen shot 2017-06-26 at 8 06 42 pm

siddhantbajaj:
Vertical Layout:

screen shot 2017-06-26 at 8 09 35 pm

Ana06:
@siddhantbajaj I really like the ticket now. Good work :+1: 
@differentreality @ddemaio do you like it too? :smile: 

Although to be able to merge this, we need a way to select in the admin side if we want an horizontal or vertical design an then showing the selected one in the show view, right?

siddhantbajaj:
@Ana06 There is only one design actually. We can switch between two layouts by just changing 
`page_layout: :portrait` to `page_layout: :landscape`. So maybe we can just mention this in a comment in our file only.

shlok007:
>We can switch between two layouts by just changing
page_layout: :portrait to page_layout: :landscape. So maybe we can just mention this in a comment in our file only or an environment variable for this.

Sorry, I don't know the full context of this discussion but I just want to chime in that we are working on the [hosted version of OSEM](https://github.com/openSUSE/osem/projects/1), so can we do that via a switch in the layout instead? I think after we have a hosted version, different organizers should have their own choice instead of site administrators deciding it for them.
We can also think about having the name of the conference's organization in the ticket as well (or maybe later!)

Ana06:
@siddhantbajaj 
> @Ana06 There is only one design actually. We can switch between two layouts by just changing
> page_layout: :portrait to page_layout: :landscape. So maybe we can just mention this in a comment in our file only or an environment variable for this.

What we can do is adding the vertical one by now, as currently the ticket is completely empty. And in another PR make this configurable. But as @shlok007 says different organizers could want to use different layouts, or even the same organizer could want to have different layouts for each of his conference.

So, let's finish this only for the vertical layout and then we will open another PR for being able to select between layouts. @siddhantbajaj I created already a Trello card for this: https://trello.com/c/Y5A4kjMq/13-make-ticket-layout-configurable

app/views/conference_registrations/show.pdf.prawn
```diff
@@ -0,0 +1,148 @@
+prawn_document(filename: @pdf_filename, page_layout: :landscape, :page_size => [630.28,721.89]) do |pdf|
```
Ana06: the code should be move to `pysical_ticket#show`, right? and take into account, that admins and user should have the same view :wink:

app/views/conference_registrations/show.pdf.prawn
```diff
@@ -0,0 +1,148 @@
...
+	
+	# SAMPLE 1
```
Ana06: Please remove the code from the examples we don't want to use and leave only the final one.

app/views/conference_registrations/show.pdf.prawn
```diff
@@ -0,0 +1,148 @@
...
+  pdf.draw_text "Username: ", :at => [0, pdf.cursor]
+  pdf.draw_text "Contact: ", :at => [pdf.bounds.right - 200, pdf.cursor]
+  pdf.pad(20){pdf.stroke_horizontal_rule}
+  pdf.draw_text "Type: ", :at => [0, pdf.cursor]
+  pdf.draw_text "Amount: ", :at => [pdf.bounds.right - 200, pdf.cursor]
+  pdf.move_down 20
+  pdf.draw_text "Arrival: ", :at => [0, pdf.cursor]
+  pdf.draw_text "Departure: ", :at => [pdf.bounds.right - 200, pdf.cursor]
+  pdf.pad(20){pdf.stroke_horizontal_rule}
+  pdf.print_qr_code(@user.username, :extent=>72)
```
Ana06: the QR should be associated to the pyshical ticket :slightly_smiling_face:

Ana06:
@shlok007 
> We can also think about having the name of the conference's organization in the ticket as well (or maybe later!)

Good idea! We can add it in the third square, were the event information is. BTW, @siddhantbajaj what order REF and attendee ref exactly are?

app/views/conference_registrations/show.pdf.prawn
ChrisBr: Have you already agreed about the handling of the QR code? What happens when I scan the QR code? Get's the ticket marked as checked in and invalidated?

app/views/conference_registrations/show.pdf.prawn
siddhantbajaj: 👍

app/views/conference_registrations/show.pdf.prawn
siddhantbajaj: 👍

siddhantbajaj:
@Ana06 
> what order REF and attendee ref exactly are?

Just for the design, we need to decide what all information need to be shown in third square.

Ana06:
@siddhantbajaj 

> Just for the design, we need to decide what all information need to be shown in third square.

That's what the prototype is for :wink: 

So, I would say that in the third square we need `Ticket` (ticket type name), `Ticket ref.` (ticket id, or oder generated number to identificate the ticket), `Event` (event name). Do I miss something?

And in the second square we want the logo of the conference, the name of the conference and the conference's organization, right? So that should be fine already. :slightly_smiling_face:

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,67 @@
...
+  pdf.fill_color "000000"
+  pdf.draw_text "ORDER REF.", :at => [x,pdf.cursor-30], :size => 15
+  pdf.fill_color "808080"
+  pdf.draw_text "#{ticket_purchase.id}", :at => [x,pdf.cursor-50], size: 12
+  pdf.fill_color "000000"
+  pdf.draw_text "Powered By OSEM", :at => [(mid_horizontal-left-100)/2,pdf.cursor-100], :size => 11
+  pdf.move_up 180
+  x1 = right-mid_horizontal-180
+  y1 = top-mid_vertical-180
+  pdf.print_qr_code(@physical_ticket.id.to_s, pos: [mid_horizontal+x1/2, pdf.cursor-y1/2], extent: 180, stroke: false)
```
Ana06: I guess the QR code won't currently work. Maybe worthwhile to remove it from here and in the new PR to make it work add it again.

app/models/ability.rb
```diff
@@ -87,7 +87,7 @@ def signed_in(user)
     can :index, Ticket
     can :manage, TicketPurchase, user_id: user.id
     can [:new, :create], Payment, user_id: user.id
-    can [:index, :show], PhysicalTicket, user_id: user.id
+    can [:index, :show], PhysicalTicket, user: user
```
Ana06: That's fine but unrelated, in a different commit please! :wink:

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: 👍

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
+prawn_document(filename: @pdf_filename, page_layout: :portrait, :page_size =>'A4' ) do |pdf|
```
mdeniz: From where is coming this `@pdf_filename` from? Is this running?

siddhantbajaj: Sorry my bad, it should be `@file_name`. But yes it was working, Pdf was just opening in the browser instead of getting downloaded as there was no file name.

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
+prawn_document(filename: @file_name, page_layout: :portrait, :page_size =>'A4' ) do |pdf|
```
Ana06: you have the `pdf` view, what about the normal show? should look similar to the pdf! Could be in aother PR, but has to be done!

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
...
+  # Vertical Layout
+  user = @physical_ticket.user
```
Ana06: why don't assigning this variable in the controller?

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
...
+  ticket_purchase = @physical_ticket.ticket_purchase
```
Ana06: this is only used one, just wrote `@physical_ticket.ticket_purchase` where it is used and remove the variable!

Ana06: same with `ticket`, `organization` and `venue`. `user` is used twice so I am fine with leaving it.

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
...
+  pdf.draw_text "NAME", :at => [x+10,pdf.cursor], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{user.name}", :at => [x+10,pdf.cursor-25], size: 20
+  pdf.fill_color "000000"
+  pdf.draw_text "EMAIL", :at => [x+10,pdf.cursor-50], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{user.email}", :at => [x+10,pdf.cursor-75], size: 20
+  pdf.fill_color "000000"
+  pdf.move_up 20
+  pdf.image "#{Rails.root}/public/img/osem-logo.png", :at => [mid_horizontal+30, pdf.cursor], :height => 50
```
Ana06: this shouldn't be osem logo, but the conference image, right?

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,64 @@
...
+  pdf.fill_color "808080"
+  pdf.draw_text "#{ticket.title}", :at => [x,pdf.cursor-50], size: 12
+  pdf.move_down 50
+  pdf.fill_color "000000"
+  pdf.draw_text "TICKET REF.", :at => [x,pdf.cursor-30], :size => 15
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@physical_ticket.id}", :at => [x,pdf.cursor-50], size: 12
+  pdf.move_down 50
+  pdf.fill_color "000000"
+  pdf.draw_text "ORDER REF.", :at => [x,pdf.cursor-30], :size => 15
```
siddhantbajaj: I thought it can be ticket_purchase id. it may be useful, right?

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: Yes I'll make a card for it in trello as well.

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: Yes I think it makes sense as we already have "Powered by OSEM" thing.

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: 👍

app/views/physical_ticket/show.pdf.prawn
Ana06: I would only leave `TICKET REF.` with the `@physical_ticket.ticket_purchase.id` as the physical ticket id doesn't seems to be useful and you can get it with the purchase.

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,60 @@
...
+  pdf.draw_text "NAME", :at => [x+10,pdf.cursor], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.name}", :at => [x+10,pdf.cursor-25], size: 20
+  pdf.fill_color "000000"
+  pdf.draw_text "EMAIL", :at => [x+10,pdf.cursor-50], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.email}", :at => [x+10,pdf.cursor-75], size: 20
+  pdf.fill_color "000000"
+  pdf.move_up 20
+  if @conference.picture?
```
Ana06: if there is no conference picture, what about showing osem logo instead as it was before?

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,60 @@
...
+    pdf.image "#{Rails.root}/public#{@conference.picture_url}", :at => [mid_horizontal+30, pdf.cursor], :height => 70
```
Ana06: you set the height to 70, but do we know the dimension of the conference image? What happens if it is too width?

siddhantbajaj: According to prawn documentation:
>If only one of :width / :height are provided, the image will be scaled proportionally. When both are provided, the image will be stretched to fit the dimensions without maintaining the aspect ratio.

So I think image will scaled proportionally, but I am not too sure about it.

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: 👍

app/views/physical_ticket/show.pdf.prawn
siddhantbajaj: done. 👍

app/views/physical_ticket/show.pdf.prawn
Ana06: Yes, but I don't mean that, what happens if the image has 500 width and 70 height? 500 would be too big for the square, right? :thinking:

siddhantbajaj: Yes you are right, image will overflow from the square if width is too large. Then I guess it's best we set both height and width, right?

Ana06: no, because then you will change the dimension of the images. I think the best option would be to set only the biggest of the two. I mean, if the width is higher, then set the height. If not, set the height. This way it won't overflow.

Please when doing it, add here screenshots with a long and tall image, so we can see how it looks like :wink:

siddhantbajaj:
Image size (1000*1500)

screen shot 2017-07-04 at 7 19 13 pm

Image size (440*132)

screen shot 2017-07-04 at 7 16 05 pm

app/uploaders/picture_uploader.rb
```diff
@@ -52,6 +52,18 @@ def store_dir
     "system/#{object_class_name}/#{mounted_as}/#{model.id}"
   end
 
+  def image
+    @image ||= MiniMagick::Image.open( file.file )
```
Ana06: Rubocop doesn't like this... :sob: 
Try to remember to run Rubocop locally before pushing :wink:

app/uploaders/picture_uploader.rb
```diff
@@ -52,6 +52,18 @@ def store_dir
...
+  end
+
+  def width
+    image[:width]
```
Ana06: do we really need this method? we can just do `image[:width]` when needed

app/uploaders/picture_uploader.rb
```diff
@@ -52,6 +52,18 @@ def store_dir
...
+  end
+
+  def height
+    image[:height]
```
Ana06: same here

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -0,0 +1,62 @@
...
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.name}", :at => [x+10,pdf.cursor-25], size: 20
+  pdf.fill_color "000000"
+  pdf.draw_text "EMAIL", :at => [x+10,pdf.cursor-50], :size => 13
+  pdf.fill_color "808080"
+  pdf.draw_text "#{@user.email}", :at => [x+10,pdf.cursor-75], size: 20
+  pdf.fill_color "000000"
+  pdf.move_up 20
+  if @conference.picture?
+    if @conference.picture.width > @conference.picture.height
```
Ana06: I see a problem here. If the image is 120x119 (I am writing `widht x height`, not sure if that is correct :rofl:) you will print the image in 120x119, although it is supposed that your maximum height is 70. If it is right that our maximum height is 70 and our maximum width 120, then the condicion should look like:

``` Ruby
if 7 * @conference.picture.width > 12 * @conference.picture.height
```

With this condition:

120x119 -> 71x70
120x70 -> 120x70
1200x600 -> 120x60
600x700 -> 60x70

and so on

Ana06:
[Refine project: Creating prototypes for the ticket](https://trello.com/c/R8j0N3sy/4-refine-project-creating-prototypes-for-the-ticket)

PR #1541, opened 2017-06-13
Improved test performance.

### LINKED ISSUE #1286: Fix Travis failures in a better way

siddhantbajaj:
Closes #1286.
Both cleaning of database and loading of seeds should be done before each feature, instead of each test.

### DIFF
--- a/spec/support/database_cleaner.rb
+++ b/spec/support/database_cleaner.rb
@@ -1,6 +1,16 @@
 RSpec.configure do |config|
-  config.before(:each) do
+  config.before(:suite) do
     DatabaseCleaner.clean_with(:truncation)
     Rails.application.load_seed
+  end
+
+  config.before(:each) do |example|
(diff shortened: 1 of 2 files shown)


Ana06:
@hennevogel can you please take a look at this? It is the Travis fix we needed to revert :wink:

PR #1547, opened 2017-06-15
Add option to render html at format_helper

### LINKED ISSUE #1546: The markdown hint in text filelds is broken
In the screenshot, we can see that bellow the text field, instead of the markdown hint, it shows the html code of the hint
Example from Users#edit
markdown-hint

dPliakos:
Fixes #1546

Add optional argument at  `markdown()` function to indicate if html has to be rendered or not.
Call `markdown_hint()` with that flag as True.

### DIFF
--- a/app/helpers/format_helper.rb
+++ b/app/helpers/format_helper.rb
@@ -176,7 +176,7 @@ module FormatHelper
     (schedule == @selected_schedule) ? 'Yes' : 'No'
   end
 
-  def markdown(text)
+  def markdown(text, escape_html=true)
     return '' if text.nil?
 
     options = {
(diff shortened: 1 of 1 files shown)


app/helpers/format_helper.rb
```diff
@@ -176,20 +176,20 @@ def selected_scheduled?(schedule)
...
+  def markdown(text, allow_html=false)
     return '' if text.nil?
 
     options = {
       autolink: true,
       space_after_headers: true,
       no_intra_emphasis: true
     }
-    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(escape_html: true), options)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML.new(escape_html: !allow_html), options)
```
AEtherC0r3: We should probably call the variable escape_html to avoid confusion because of the '!'

dPliakos: I wrote it this way, because I think it's more clear when call it. 'allow' to render may be more clear than  'escape' to prevent rendering. I'm making the changes.

differentreality:
Please add screenshots of your fix

dPliakos:
Sure!
screenshot from 2017-06-20 07-09-24

AEtherC0r3:
Please squash and rebase

differentreality:
:+1: Thanks for the code @dPliakos and well done to both of you @dPliakos and @AEtherC0r3 for working on this! :tada:

PR #1549, opened 2017-06-17
Add in the database migration step to the CONTRIBUTING.md file

kormoc:
Thanks @genehack for the original diff

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,28 +35,34 @@ We are using [Vagrant](https://www.vagrantup.com/) to create our development env
     vagrant up
     ```
 
-5. Start your OSEM rails app:
+5. Deploy the initial database migration:
+
+    ```
+    vagrant exec bundle exec rake db:migrate
+    ```
+
+6. Start your OSEM rails app:
 
     ```
     vagrant exec /vagrant/bin/rails server -b 0.0.0.0
     ```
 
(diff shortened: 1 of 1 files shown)


differentreality:
Hey @kormoc, I just wanted to make a note that the `db:migrate` is **not** necessary for the initial installation. When you run `vagrant up` it loads automatically the db schema. So running `db:migrate` after that, will do nothing.

You only have to run the `db:migrate `command, if you are running vagrant, you `git pull` changes that include a new migration, and you want to apply this single migration, without re-running `vagrant up`. 

Feel free to reach out and let us know if you ran into a different situation.

genehack:
I added that line because when I pulled the *SeaGL fork* and tried to stand it up locally following the instructions, I had to manually do the migration. 

It's possible that issue had already been fixed in the automation code in the upstream.

AEtherC0r3:
I will double check this issue

hennevogel:
Here is your double check -> https://github.com/openSUSE/osem/blob/master/bootstrap.sh#L18

kormoc:
Yes, generally this isn't needed, however when switching between branches/forks, it's often a required step to re-run the migration, so I felt that having it documented is at worst is a noop and at best will prevent weird issues from lack of migrations. Sometimes it's hard to know for sure that an error is due to a missing database migration.

PR #1329, opened 2017-03-05
Fixed paper trail inconsistent results for numeric values.

siddhantbajaj:
There is a known issue in paper_trail that whenever we Query the 'versions.object' column (or 'object_changes' column), like this:

```
PaperTrail::Version.where_object(attr1: val1, attr2: val2)
PaperTrail::Version.where_object_changes(attr1: val1)
```
it evaluates inconsistent results for numeric values due to limitations of SQL wildcard matchers against the serialized objects.

So to fix this issue I have manually formed the where query instead of using where_object and where_object_changes. I have also added test for the same.

To recreate this issue you can create 2 events like this:
`Event_without_commercial{id:1}, Event_with_commercial{id:2}`
a event commercial:
`event_commercial{id:1, event_id:2}`
version for this commercial will be like:
`Version{object_change: "...id:1...commercialable_id:2"}`
**So this version will be shown in the history of both the events.**  

closes #1307

Issue #1307, opened 2017-02-20
Paper_Trail Issue


siddhantbajaj:
There is a known issue in paper_trail that whenever we Query the 'versions.object' column (or 'object_changes' column), like this:

> PaperTrail::Version.where_object(attr1: val1, attr2: val2)
> PaperTrail::Version.where_object_changes(attr1: val1)

it evaluates inconsistent results for numeric values due to limitations of SQL wildcard matchers against the serialized objects.

we are also doing  like this here:
https://github.com/openSUSE/osem/blob/master/app/controllers/admin/events_controller.rb#L53

To find out more about this issue look at the comments here:
https://github.com/openSUSE/osem/pull/1283#issuecomment-280310475

PR #1550, opened 2017-06-17
Fix broken test in Admin::EventsController

Ana06:
In the assigns versions test we were checking the ids of `event_without_commercial` and `event_commercial` were the same. That was only the case because there were the only objects in the database, but this does't need to be true and it is completely useless.

We also check that the ids of `event_commercial` and `event_commercial.commercialable` were different, but that also doesn't make any sense, as they are even different object types.

This two test worked before merging https://github.com/openSUSE/osem/pull/1541 After that, only the controller test was failing, although both of them were wrong. :bowtie: 

**We should consider improving this tests, because there are not very useful.** I'll do so or create an issue. The objective of this PR is to fix the failing tests. :wink:

### DIFF
--- a/spec/controllers/admin/events_controller_spec.rb
+++ b/spec/controllers/admin/events_controller_spec.rb
@@ -4,9 +4,13 @@ describe Admin::EventsController do
   let(:conference) { create(:conference) }
   let(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let(:organizer) { create(:user, role_ids: organizer_role.id) }
-  let!(:event_without_commercial) { create(:event, program: conference.program) }
-  let!(:event_with_commercial) { create(:event, program: conference.program) }
-  let!(:event_commercial) { create(:event_commercial, commercialable: event_with_commercial, url: 'https://www.youtube.com/watch?v=M9bq_alk-sw') }
+  # The where_object() and where_object_changes() methods of paper_trail gem are broken when having:
+  # an Event with ID 1, an Event with ID 2, and a commercial with ID 1, for event with ID 2
+  # (the numbers could be different as long as there is this matching of IDs).
+  # We implemented or own where method to solve this and those ids are for testing this case.
+  let!(:event_without_commercial) { create(:event, id: 1, program: conference.program) }
+  let!(:event_with_commercial) { create(:event, id: 2, program: conference.program) }
+  let!(:event_commercial) { create(:event_commercial, id: 1, commercialable: event_with_commercial, url: 'https://www.youtube.com/watch?v=M9bq_alk-sw') }
 
   with_versioning do
     describe 'GET #show' do
@@ -17,8 +21,6 @@ describe Admin::EventsController do
 
       it 'assigns versions' do
         versions = event_without_commercial.versions
-        expect(event_without_commercial.id).to eq event_commercial.id
-        expect(event_commercial.id).not_to eq event_commercial.commercialable_id
         expect(assigns(:versions)).to eq versions
       end
     end
--- a/spec/features/versions_spec.rb
+++ b/spec/features/versions_spec.rb
@@ -286,8 +286,6 @@ feature 'Version' do
     expect(page).to have_text('Someone (probably via the console) created new commercial')
     visit admin_conference_program_event_path(conference.short_title, event_without_commercial)
     click_link 'History'
-    expect(event_commercial.id).not_to eq event_commercial.commercialable_id
-    expect(event_without_commercial.id).to eq event_commercial.id
     expect(page).to have_no_text('Someone (probably via the console) created new commercial')
   end
 


Ana06:
@siddhantbajaj this test was broken after merging your PR #1541 . Not because of your changes, but because this was already wrong. But maybe you want to take a look. :wink:

differentreality:
No, the test is like this on purpose. It is formed like this on purpose because we had this exact point of failure (under the conditions being tested) before. And the failure even comes from paper trail itself, so with this test we can ensure our custom code will not change back to the paper trail way, as it can cause underlying discrepancies,  under certain circumstances. 

Moreover it should not be failing, as long as we are cleaning our database for every test.

siddhantbajaj:
> In the assigns versions test we were checking the ids of event_without_commercial and event_commercial were the same. That was only the case because there were the only objects in the database, but this does't need to be true and it is completely useless.

@Ana06 Actually I wrote those test here #1329 and yes you are absolutely right, I wrote those test keeping in mind that we clear the database before every test. But after the merge of #1541 it is not case now. Now coming to why we are checking these id's, actually I and @differentreality wanted to make sure that #1307 is completely resolved. So by comparing these Id's we are double checking if we have properly set up the case where this issue will arise.

Ana06:
@differentreality 

> No, the test is like this on purpose. It is formed like this on purpose because we had this exact point of failure (under the conditions being tested) before. And the failure even comes from paper trail itself, so with this test we can ensure our custom code will not change back to the paper trail way, as it can cause underlying discrepancies, under certain circumstances.

But I don't get how this test the paper trail. This two lines are not working properly. The first of them, check that `event_without_commercial` and `event_commercial` ids are different, but they are different objects, so how can it be that they have the same id? And you just create them in your test. Even more, why do we need to create `event_with_commercial` and `event_commercial` if they are not used at all for the test?

And with the second line is even clearer, as `event_commercial` and `event_commercial.commercialable`  are even different types of objects.

So, I would suggest to merge this, and then improving these tests. @differentreality are you fine with this?

Ana06:
@siddhantbajaj 

> actually I didn't get the time here #1541 to see the broken tests else I would have fix these test there only.

No problem, I also did't see it was failing until I merged it :wink: As said, we can just merge this to avoid that the tests fails and then we will focus on improving the test. :wink:

differentreality:
@Ana06 I am suggesting that we should **not** merge this PR, as long as we want to be testing a case that has previous failed us.

The answer to your questions about why we are testing the IDs like this is here:

The problem ->  https://github.com/openSUSE/osem/pull/1283#issuecomment-280310475
This describes the problem, which resulted in wrong collections return by papertrail's function, so we had to do it 'manually' #1329, #1283 
This is a particular problem, existing under very specific circumstances - it might not be possible to recreate for an existing installation of OSEM, but it is very likely to show up on a new installation.

Why I suggested we add those tests:
I think it is possible someone tries to refactor this code in the future (https://github.com/openSUSE/osem/blob/master/app/controllers/admin/events_controller.rb#L52-L56), using paper_trail's own function. 
If we remove the tests you are removing in this PR, we might have a failure, and our test suit will **not** catch it.

Why don't we empty the DB for theses tests to pass instead of removing them?

Ana06:
But @differentreality that problem would change the versions assignment, but that is not what is tested here.

- `event_without_commercial` and `event_commercial` ids are different, but they are different objects of the same model.
- 
- `event_commercial` and `event_commercial.commercialable`  are even different types of objects, so no related at all

So, this is wrong and it is impossible that it tests what you say.

To test that you can for example assign `versions` [here](https://github.com/openSUSE/osem/pull/1550/files#diff-806d0c152e59926c89ab2d66d4e7fe88R19) to the exact value you expect. This way you would avoid that this happens again. But currently this is not tested, and the lines I am removing are useless.

mdeniz:
I don't really get what the test is doing... comparing ids of 2 different objects that maybe are equal or not, and at the same time aren't changed by the action tested...

Can anyone explain why this is needed for sure?

differentreality:
I realize it might be confusing, but please read through https://github.com/openSUSE/osem/pull/1283#issuecomment-280310475 as to why I was checking those IDs. The explanation refers to event and votes, but it is the same case as with events and commercials.

Sure testing that @versions variable holds the right value is necessary. And as long as the test runs on an empty DB, and the order we create events for the test does not change, nothing else is required. 

What I am aiming to test is, whether or not @versions has the right value, when there is no other commercial, and no other event in the database, and then events and commercials are created in a specific order. That's the point of failure with paper_trail's native methods.

We are using paper_trail gem, which saves data in database table `versions`. It has a native way to search in the versions records, using `where_object()` and `where_object_changes()`.
They are broken, under certain conditions. We changed them to a manual `where()` statement, and we added a test to make sure no one changes them back. 

`where_object()` and `where_object_changes()` are only broken under specific circumstances, which are explained here https://github.com/openSUSE/osem/pull/1283#issuecomment-280310475

So for this particular test, we need an empty DB before it runs. We don't have that any more. And we specifically need to save first the event that does **not** have a commercial, and afterwards the event that has a commercial. The order, and the existence of other data matters, which is why we had those 2 expect statements there.

I agree we can refactor/ the test, or even write it in a complete different way, but just removing those 2 lines, is not enough to test what I was aiming to test here (which is the right value of @versions when we have an Event with ID 1, an Event with ID 2, and a commercial with ID 1, for event with ID 2 - obviously the numbers could be different as long as there is this matching of IDs, that caused the failure before).

Ana06:
@differentreality 

> Sure testing that @versions variable holds the right value is necessary. And as long as the test runs on an empty DB, and the order we create events for the test does not change, nothing else is required.

The DB is empty but the id could be different from 1, because we use transaction instead of truncation, that's why this fail.

> the right value of @versions when we have an Event with ID 1, an Event with ID 2, and a commercial with ID 1, for event with ID 2 - obviously the numbers could be different as long as there is this matching of IDs, that caused the failure before

But then we should enforce this in the creation of the objects, because this is the test case, not what you want to test. We could create Event and Commercial with the ID we want. I mean adding the id in the creation and removing this two lines. Would you be okey with this @differentreality?

differentreality:
Yes, absolutely. Surely it can be refactored and written better. As long as it tests the actual test case we want.

Ana06:
BTW, I wonder if we really need a feature test for this. I would say this should be only a controller test. @differentreality can I remove this from the feature test?

spec/controllers/admin/events_controller_spec.rb
```diff
@@ -17,8 +17,6 @@
 
       it 'assigns versions' do
         versions = event_without_commercial.versions
-        expect(event_without_commercial.id).to eq event_commercial.id
-        expect(event_commercial.id).not_to eq event_commercial.commercialable_id
         expect(assigns(:versions)).to eq versions
```
mdeniz: Can you replace `versions` with this `event_without_commercial.versions`

spec/controllers/admin/events_controller_spec.rb
```diff
@@ -4,9 +4,9 @@
   let(:conference) { create(:conference) }
   let(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let(:organizer) { create(:user, role_ids: organizer_role.id) }
-  let!(:event_without_commercial) { create(:event, program: conference.program) }
-  let!(:event_with_commercial) { create(:event, program: conference.program) }
-  let!(:event_commercial) { create(:event_commercial, commercialable: event_with_commercial, url: 'https://www.youtube.com/watch?v=M9bq_alk-sw') }
+  let!(:event_without_commercial) { create(:event, id: 1, program: conference.program) }
```
mdeniz: Can we have a reference of the bug, just to know why those ids are needed

spec/controllers/admin/events_controller_spec.rb
Ana06: yes, I plan to refactor this test afterwards. But by now I don't want to make this PR bigger, as we need this to get merge. It breaks the test suite. :see_no_evil:

PR #1553, opened 2017-06-22
Fix Rubocop offense & remove default configuration

Ana06:
Fix `Style/LeadingCommentSpace` Rubocop offense introduced in #1229. :bowtie: 

We had part of the default Rubocop configuration in the `.rubocop.yml` file. It is not needed as it is the default and it also don't make sense, because we only have part of it. So I removed the default configuration from `.rubocop.yml` :wink: This was caused because when enabling a cop we were moving it from `.rubocop_todo.yml` to `.rubocop.yml`, but this is only needed if we overwrite the default configuration. :confounded:


differentreality:
I am not sure if I understood correctly. 
Are we using the deleted cops? Are they being enforced? If they are enforced, I think we should include them in rubocop file, so that the file shows all cops that we are enforcing.

Ana06:
@differentreality the thing is that the cops in the [default configurations file](https://github.com/bbatsov/rubocop/tree/master/config) are always used.

So basically if the configuration files has something like [this](https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml#L288):

```
Layout/SpaceAroundOperators:
  Description: 'Use a single space around operators.'
  StyleGuide: '#spaces-operators'
  Enabled: true
```

doesn't have any efect that we have [that ](https://github.com/openSUSE/osem/pull/1553/files#diff-11a0d7906801d9dea0eccb85667ad811L88)in our .`rubocop.yml` file:

```
# Checks that operators have space around them, except for ** which should not have surrounding space.
 Style/SpaceAroundOperators:
   Enabled: true
```

And it is also not consistent as there are many other rules that are enabled by default but that are not in out `.rubocop.yml` file.

So, what I removed doesn't change the cops that are applied. Everything keeps the same, but doing things as they should be done: only adding a cop in the `.rubocop.yml` file if we want to change the default configuration. :wink:

differentreality:
:+1:

Ana06:
@differentreality can we then merge this? :slightly_smiling_face:

PR #1554, opened 2017-06-23
Fix rubocop Gemfile offense

AEtherC0r3:
Add a space to the pg gem line


Ana06:
This is fixed in https://github.com/openSUSE/osem/pull/1553

PR #1556, opened 2017-06-23
Track requests and track organizer

AEtherC0r3:
Work in progress implementation of track requests and the track organizer role

* Add the fields submitter_id, state, and cfp_active to tracks
* Add validations and the self_organized? method to the Track model
* Create a new TracksController outide of the admin namespace
* Create the relevant views for index, show, new and edit
* Modify the admin views for tracks to include extra info for self-organized tracks
* Add track organizer role and auto-create it when a self-organized track is created
* Define abilities for track organizers
* Modify the roles views and controller to handle the new role

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -94,6 +94,8 @@ Metrics/ModuleLength:
 # Offense count: 14
 Metrics/PerceivedComplexity:
   Max: 15
+  Exclude:
+    - 'app/controllers/admin/roles_controller.rb'
 
 # Offense count: 11
 # Cop supports --auto-correct.
@@ -850,6 +852,7 @@ Style/SymbolProc:
     - 'app/controllers/admin/questions_controller.rb'
     - 'app/helpers/application_helper.rb'
     - 'app/models/ability.rb'
+    - 'app/models/admin_ability.rb'
     - 'db/migrate/20140730104658_migrate_roles_for_cancancan.rb'
     - 'spec/controllers/admin/conferences_controller_spec.rb'
     - 'spec/support/flash.rb'
--- a/app/controllers/admin/roles_controller.rb
+++ b/app/controllers/admin/roles_controller.rb
@@ -15,7 +15,7 @@ module Admin
 
     def show
       @url = if @track
-               toggle_user_track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+               toggle_user_track_admin_conference_role_path(@conference.short_title, @role.name, @track)
              else
                toggle_user_admin_conference_role_path(@conference.short_title, @role.name)
              end
@@ -24,7 +24,7 @@ module Admin
 
     def edit
       @url = if @track
-               track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+               track_admin_conference_role_path(@conference.short_title, @role.name, @track)
              else
                admin_conference_role_path(@conference.short_title, @role.name)
              end
@@ -36,7 +36,7 @@ module Admin
 
       if @role.update_attributes(role_params)
         url = if @track
-                track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+                track_admin_conference_role_path(@conference.short_title, @role.name, @track)
               else
                 admin_conference_role_path(@conference.short_title, @role.name)
               end
@@ -55,7 +55,7 @@ module Admin
       state = user_params[:state]
 
       url = if @track
-              track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+              track_admin_conference_role_path(@conference.short_title, @role.name, @track)
             else
               admin_conference_role_path(@conference.short_title, @role.name)
             end
--- a/app/controllers/admin/roles_controller.rb
+++ b/app/controllers/admin/roles_controller.rb
@@ -8,14 +8,26 @@ module Admin
 
     def index
       @roles = Role.where(resource: @conference)
+      tracks = @conference.program.tracks.where.not(submitter: nil)
+      @roles += Role.where(resource: tracks)
       authorize! :index, @role
     end
 
     def show
+      @url = if @track
+               toggle_user_track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+             else
+               toggle_user_admin_conference_role_path(@conference.short_title, @role.name)
+             end
       @users = @role.users
     end
 
     def edit
+      @url = if @track
+               track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+             else
+               admin_conference_role_path(@conference.short_title, @role.name)
+             end
       @users = @role.users
     end
 
@@ -23,7 +35,13 @@ module Admin
       role_name = @role.name
 
       if @role.update_attributes(role_params)
-        redirect_to admin_conference_role_path(@conference.short_title, @role.name),
+        url = if @track
+                track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
+              else
+                admin_conference_role_path(@conference.short_title, @role.name)
+              end
+
+        redirect_to url,
                     notice: 'Successfully updated role ' + @role.name
       else
         @role.name = role_name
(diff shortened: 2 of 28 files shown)


.rubocop_todo.yml
```diff
@@ -66,7 +66,7 @@ Metrics/AbcSize:
 # Offense count: 202
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/BlockLength:
-  Max: 487
+  Max: 658
```
differentreality: Why don't you exclude the relevant file?

differentreality:
@AEtherC0r3 how is the DB table roles going to look like after your changes? Please add a table/image in comments.

app/controllers/admin/roles_controller.rb
```diff
@@ -8,22 +8,40 @@ class RolesController < Admin::BaseController
...
     def index
       @roles = Role.where(resource: @conference)
+      tracks = @conference.program.tracks.where.not(submitter: nil)
+      @roles += Role.where(resource: tracks)
       authorize! :index, @role
     end
 
     def show
+      @url = if @track
+               toggle_user_track_admin_conference_role_path(@conference.short_title, @role.name, @track.name.tr(' ', '_'))
```
differentreality: Instead of changing the track.name, it would make sense to add a short_name like conference has, and change how we navigate to tracks all together, and base it on track.short_name instead of track.id

.rubocop_todo.yml
AEtherC0r3: I tried to exclude it but for some reason it didn't work, so I ended up increasing the maximum length

AEtherC0r3:
I haven't made any changes to the roles table.
In any case, here's an image that shows track organizer roles

app/controllers/admin/roles_controller.rb
AEtherC0r3: #1565 should take care of this

differentreality:
This is dependent on #1565

AEtherC0r3:
@differentreality @sayalilunkad Sorry for the gigantic commit, properly separating the changes in different commits and fixing the conflicts was really difficult :(

AEtherC0r3:
Tracks#new


admin/Tracks#new


Tracks#index


admin/Tracks#index


Tracks#edit


admin/Tracks#edit for self organized tracks


admin/Tracks#edit for regular tracks


Tracks#show

app/controllers/admin/roles_controller.rb
lagartoflojo: Actually, Rails already has a way of turning a model into the appropriate URL parameter, via [`to_param`](http://api.rubyonrails.org/v4.2.5/classes/ActiveModel/Conversion.html#method-i-to_param).

By implementing `#to_param` like:
```ruby
def to_param
  short_title # or `name` in Role
end
```

you can then just do:

```ruby
toggle_user_track_admin_conference_role_path(@conference, @role, @track)
```

AEtherC0r3: @lagartoflojo thanks for the tip

AEtherC0r3:
The remaining rubocop offense will be fixed by #1569

sayalilunkad:
needs rebase

app/controllers/tracks_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  def new
+    @track = @program.tracks.new(color: @conference.next_color_for_collection(:tracks))
+  end
+
+  def edit; end
+
+  def create
+    @track = @program.tracks.new(track_params)
+    @track.submitter = current_user
+    @track.state = 'pending'
```
differentreality: Are you sure you need to set this here? Why don't you keep 'new' as we have in other places as well?

AEtherC0r3: 1) Can you recommend me a more suitable place?
2) It makes more sense to me for a request to be pending than to be new. If you wish I can change it back to new.

app/views/admin/tracks/index.html.haml
```diff
@@ -11,27 +11,52 @@
...
+                          off_text: 'No' }
+              - else
+                %i.fa.fa-check
+            %td
               .btn-group{role: "group"}
-                = link_to 'Edit', edit_admin_conference_program_track_path(@conference.short_title, track.short_name),
+                = link_to 'Edit', edit_admin_conference_program_track_path(@conference.short_title, track),
                 method: :get, class: 'btn btn-primary'
-                = link_to 'Delete', admin_conference_program_track_path(@conference.short_title, track.short_name),
+                = link_to 'Delete', admin_conference_program_track_path(@conference.short_title, track),
```
differentreality: For self organized tracks, destroy button should be disabled, so that the record doe snot get deleted, only moved to withdrawn/cancelled state.

differentreality:
@AEtherC0r3 please fix the errors and prepare your PR for merging! :tada:  Nice work :+1:

app/models/track.rb
```diff
@@ -14,13 +18,31 @@ class Track < ActiveRecord::Base
             uniqueness: {
               scope: :program
             }
+  validates :state, presence: true, if: :self_organized?
+  validates :cfp_active, inclusion: { in: [true, false] }, if: :self_organized?
 
   before_validation :capitalize_color
 
+  after_create :create_organizer_role, if: :self_organized?
```
differentreality: Doesn't this need to happen after we accept the track?

app/views/admin/tracks/_form.html.haml
```diff
@@ -8,9 +8,11 @@
...
 .row
   .col-md-12
-    = semantic_form_for(@track, url: (@track.new_record? ? admin_conference_program_tracks_path : admin_conference_program_track_path(@conference.short_title, @track.short_name))) do |f|
+    = semantic_form_for(@track, url: (@track.new_record? ? admin_conference_program_tracks_path : admin_conference_program_track_path(@conference.short_title, @track))) do |f|
       = f.input :name
       = f.input :short_name, hint: "A short and unique handle for the track, using only letters, numbers, underscores, and dashes. This will be used to identify the track in URLs etc. Example: 'my_awesome_track'", input_html: { required: 'required', pattern: '[a-zA-Z0-9_-]+', title: 'Only letters, numbers, underscores, and dashes.' }
       = f.input :color, input_html: {size: 6, type: 'color'}, required: true
       = f.input :description, input_html: {rows: 2, data: { provide: 'markdown-editable' } }, hint: markdown_hint
+      - if @track.self_organized?
+        = f.input :cfp_active, label: 'Allow event submitters to select this track for their proposal'
```
differentreality: Since we have that option anyway, why not make it available to all tracks, to avoid confusion? We can mark it as true  when the track is created in the admin view.

app/views/admin/tracks/index.html.haml
```diff
@@ -11,29 +11,55 @@
...
             %td
+              - if track.self_organized?
+                = link_to track.submitter.name, admin_user_path(track.submitter)
+              - else
+                N/A
+            %td
               %span.label{style: "background-color: #{track.color}; color: #{ contrast_color(track.color) }"}
                 = track.color
             %td
+              - if track.self_organized?
```
differentreality: We could always be showing the state, and mark as accepted/confirmed the tracks that are created on admin side

config/routes.rb
```diff
@@ -82,9 +86,15 @@
       resources :campaigns, except: [:show]
       resources :emails, only: [:show, :update, :index]
       resources :physical_ticket, only: [:index]
-      resources :roles, except: [ :new, :create ] do
+      resources :roles, only: [:edit]
+      resources :roles, except: [ :new, :create, :edit ] do
         member do
           post :toggle_user
+          get ':track_name' => 'roles#show', as: 'track'
```
differentreality: This still needs to be called track_id by the convention of the routes format

differentreality: What about removing all those extra lines and doing it in a more rails way? Perhaps nest the roles resources under the admin/tracks? Please think about how we could simply this route.

differentreality:
I have left some comments which can be added in your next PRs as the feature progresses. Travis failed test is not related to your PR, so I think this is good to be merged :+1:

app/models/track.rb
AEtherC0r3: This is fixed in #1597

app/views/admin/tracks/index.html.haml
AEtherC0r3: This is fixed in #1597

app/views/admin/tracks/_form.html.haml
AEtherC0r3: This is also fixed in #1597

config/routes.rb
AEtherC0r3: This is fixed in #1639

AEtherC0r3: Fixed in #1639

PR #1557, opened 2017-06-24
Move ability of admins in separate model

### LINKED ISSUE #1552: Define separate abilities for controllers in admin namespace
Since, we are using separate controllers for administrators and non- administrators, there comes a situation when we want to allow access to actions in the controller without namespace and restrict access to the same actions present in the admin namespace. 

For example: Allow access to `edit` action in the `UsersController` but restrict access to `edit` action in the `Admin::UsersController`

It makes sense to have a separate model to handle abilities for the controllers present inside the admin namespace.

The documentation in cancancan covering the above situation:
https://github.com/ryanb/cancan/wiki/Admin-Namespace

shlok007:
Closes #1552 .

When a user, signed in as admin, access any view corresponding to any controller that does not inherit [`base_controller.rb`](https://github.com/openSUSE/osem/pull/1557/files#diff-91afc22f1e63da7d7b3bdeaacc287d92), the authorization will take place from [`ability.rb`](https://github.com/openSUSE/osem/pull/1557/files#diff-e9c04d21b8c4d777e4231c4c6ce5f8cb) and not from [`admin_ability.rb`](https://github.com/openSUSE/osem/pull/1557/files#diff-8f75bcf1395a2acae3d4c762ed3e0a13) . Therefore, it becomes necessary to define permissions of a not signed in user and signed in user without any role in [`admin_ability.rb`](https://github.com/openSUSE/osem/pull/1557/files#diff-8f75bcf1395a2acae3d4c762ed3e0a13) as well.

These initial commits just moves the relevant part of model and splits the tests without any refactoring or DRYing the code.


shlok007:
More breaking changes for non-admin controller actions : 
- [ ] Admin can't access `proposals/registrations` for conference
- [ ] Admins can't access `proposals` . No `submit proposals` link for site admins in conference#index
- [ ] No administration links ( Administration, Create Conference etc) in user navbar for admins
- [ ] No notifications ( comments ) in the menu bar

app/models/admin_ability.rb
```diff
@@ -0,0 +1,241 @@
...
+    # user.is_volunteers_coordinator_of? Conference
+    # user.is_attendee_of? Conference
+    # The following is wrong because a user will only have 'cfp' role for a specific conference
+    # user.is_cfp? # This is always false
+
+    user ||= User.new
+    signed_in_with_roles(user)
+  end
+
+  def common_abilities_for_all
```
differentreality: These abilities are defined for the not_signed_in users as well, and here we are not adding anything new/more. So do we really need this whole function?

app/models/ability.rb
```diff
@@ -3,30 +3,56 @@ class Ability
...
-    # The following is wrong because a user will only have 'cfp' role for a specific conference
-    # user.is_cfp? # This is always false
-
     user ||= User.new
 
-    # This is what sets up the different abilities
     if user.new_record?
       not_signed_in
-    # Checks if the user does not have any role and is not an admin
-    elsif user.roles.any? || user.is_admin
```
shlok007: Nothing special! I just thought it better to write it in a single line rather than checking each role separately.

app/models/admin_ability.rb
shlok007: Thanks, it might be safe to remove these along with some of the abilities defined in [`common_abilities_for_roles(user)`](https://github.com/openSUSE/osem/pull/1557/files#diff-8f75bcf1395a2acae3d4c762ed3e0a13R57 ). I'll update the functions removing the abilities that are not needed.

shlok007:
@differentreality Please do have a look.

app/models/ability.rb
```diff
@@ -3,30 +3,52 @@ class Ability
...
+                           commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids_for_organizer).pluck(:id)).pluck(:id)
+    elsif conf_ids_for_cfp
+
+      can :index, Comment, commentable_type: 'Event',
+                           commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids_for_cfp).pluck(:id)).pluck(:id)
+      can :manage, Event, program: { conference_id: conf_ids_for_cfp }
+
+    elsif conf_ids_for_info_desk
+      can :manage, Registration, conference_id: conf_ids_for_info_desk
+    end
```
lagartoflojo: I don't think it's right that these are `if-elsif` conditions.
For example, a user can have `conf_ids_for_organizer == [1]` and `conf_ids_for_info_desk == [2]` (so they're an organizer of one conference, but an info-desk for another conference).
So with this code as it is, they will get all organizer abilities for Conf1 but they will NOT get the info desk permission for Conf2

app/models/ability.rb
```diff
@@ -3,30 +3,52 @@ class Ability
...
+
+    can :access, Admin
+    can :manage, :all if user.is_admin
```
lagartoflojo: better: `is_admin?`

app/models/ability.rb
```diff
@@ -3,30 +3,52 @@ class Ability
...
-    # This is what sets up the different abilities
     if user.new_record?
       not_signed_in
-    # Checks if the user does not have any role and is not an admin
     elsif user.roles.any? || user.is_admin
-      signed_in_with_roles(user)
+      common_abilities_for_admins(user)
     else
       signed_in(user)
     end
```
lagartoflojo: How about:
```ruby
if user.new_record?
  not_signed_in
else
  signed_in(user)
  common_abilities_for_admins(user) if user.roles.any? || user.is_admin?
end
```

And remove line 19: `signed_in(user)`.

This way, it is clearer to see that users with roles/admins also get the same permissions as signed in users.

PR #1565, opened 2017-06-30
Friendly urls for tracks

AEtherC0r3:
Don't use tracks ids in urls
Add short_name to tracks and use that in urls as the identifier


AEtherC0r3:
Tracks#new


Tracks#index


Tracks#edit


Url from Tracks#show

differentreality:
This looks nice @AEtherC0r3 :D :+1:

PR #1569, opened 2017-07-05
Enable Style/IndentationWidth cop

AEtherC0r3:
This was introduced in #1538

### DIFF
--- a/app/helpers/format_helper.rb
+++ b/app/helpers/format_helper.rb
@@ -102,13 +102,12 @@ module FormatHelper
     end
   end
 
-  # rubocop:disable Lint/EndAlignment
   def word_pluralize(count, singular, plural = nil)
     word = if (count == 1 || count =~ /^1(\.0+)?$/)
-      singular
-    else
-      plural || singular.pluralize
-    end
+             singular
+           else
+             plural || singular.pluralize
+           end
 
     "#{word}"
   end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -467,17 +467,6 @@ Style/IndentationConsistency:
     - 'app/models/event.rb'
     - 'spec/controllers/subscriptions_controller_spec.rb'
 
-# Offense count: 6
-# Cop supports --auto-correct.
-# Configuration parameters: Width, IgnoredPatterns.
-Style/IndentationWidth:
-  Exclude:
-    - 'app/helpers/format_helper.rb'
-    - 'app/serializers/conference_serializer.rb'
-    - 'db/migrate/20140701123203_add_events_per_week_to_conference.rb'
-    - 'lib/tasks/demo_data_for_development.rake'
-    - 'spec/models/ability_spec.rb'
-
 # Offense count: 4
 # Cop supports --auto-correct.
 Style/LeadingCommentSpace:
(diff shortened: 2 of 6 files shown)


shlok007:
There is also an indentation issue from  line [55-57](https://github.com/AEtherC0r3/osem/blob/9f0d21f6da9e18e93b68c44d457257048dadd80c/spec/models/ability_spec.rb#L55). We have excluded the indentation check for `spec/models/ability_spec.rb` in [.rubocop_todo.yml#L479](https://github.com/openSUSE/osem/blob/master/.rubocop_todo.yml#L479)  which was the reason why rubocop didn't pick these up. The tests for ability model have been refactored a bit so, it might be safe to enable it now.  
  
Can we do it in this same PR and let rubocop fix the indentation issue before you merge?

lagartoflojo:
Yeah, let's just enable the cop and let it do its thing.

AEtherC0r3:
@Ana06 @lagartoflojo I never intended to enable the cop. I only wanted to fix the indentation of lines 142-155, because they caused another cop to fail in #1556, when I added some properly indented lines after line 155

lagartoflojo:
I understand, but maybe you *can* enable the cop?

lagartoflojo:
Like, I'm questioning why we didn't enable that cop from the beginning in the first place 😝

AEtherC0r3:
@Ana06 The offenses weren't automatically corrected. The auto correct did a really bad job at correcting them, so I stepped in and did it manually.

Ana06:
@AEtherC0r3 

> @Ana06 The offenses weren't automatically corrected. The auto correct did a really bad job at correcting them, so I stepped in and did it manually.

Then mention that was manually corrected :wink: 

However, I am curious, how can it be that the autocorrect did a bad job? can you elaborate please?

AEtherC0r3:
@Ana06 
In the format_helper it only fixed the indentation of line 108. It should have fixed lines [108, 111]. And, in the demo_data_for_development rake task it tried to fix the indentation of lines 4, 6, [38, 75],  instead of  [7, 37], 76

sayalilunkad:
needs rebase

app/helpers/format_helper.rb
```diff
@@ -102,13 +102,12 @@ def class_for_todo(bool)
     end
   end
 
-  # rubocop:disable Lint/EndAlignment
```
Ana06: Unrelated to this PR, at least move this to a different commit which explain what you are doing here

app/helpers/format_helper.rb
```diff
@@ -102,13 +102,12 @@ def class_for_todo(bool)
...
   def word_pluralize(count, singular, plural = nil)
     word = if (count == 1 || count =~ /^1(\.0+)?$/)
-      singular
-    else
-      plural || singular.pluralize
-    end
+             singular
+           else
```
Ana06: You are right, Rubocop only corrected line 107 and should correct 108-110. This is a Rubocop bug and according the releases notes, it is not yet fixed in the last Rubocop version. So I thing that it is nice to report bugs to Rubocop when you find then, and not only correct this manually. I've just an open an issue: https://github.com/bbatsov/rubocop/issues/4608

Ana06:
@differentreality why did u merge this when I had requested changes? :angry:

differentreality:
The commit comes from github's button 'Update', it creates this merge commit, but once you rebase&merge the commit, this merge commits goes away.

PR #1570, opened 2017-07-06
[WIP] Make ticket layout configurable

siddhantbajaj:
Admin can set the layout type for the tickets of the conference.

screen shot 2017-07-06 at 2 15 42 pm

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -57,6 +57,7 @@ class Conference < ActiveRecord::Base
             :end_date,
             :start_hour,
             :end_hour,
+            :ticket_layout,
             :organization, presence: true
 
   validates :short_title, uniqueness: true
@@ -72,6 +73,8 @@ class Conference < ActiveRecord::Base
 
   after_create :create_free_ticket
   after_update :delete_event_schedules
+
+  enum ticket_layout: [:portrait, :landscape]
 
   ##
   # Checks if the user is registered to the conference
--- a/app/views/admin/conferences/edit.html.haml
+++ b/app/views/admin/conferences/edit.html.haml
@@ -17,6 +17,7 @@
         = image_tag @conference.picture.thumb.url
       = f.input :picture, label: false, hint: 'This will be displayed on the front page.'
       = f.hidden_field :picture_cache
+      = f.input :ticket_layout, as: :select, collection: Conference.ticket_layouts.keys, hint: "Layout type for tickets of the conference."
       = f.inputs name: 'Scheduling' do
         = f.input :timezone, as: :time_zone, hint: 'The conference time zone'
         = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', readonly: 'readonly' }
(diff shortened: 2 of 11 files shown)


Ana06:
[Make ticket layout configurable](https://trello.com/c/Y5A4kjMq/13-make-ticket-layout-configurable)

app/views/admin/conferences/edit.html.haml
```diff
@@ -11,6 +11,7 @@
       = f.input :short_title, hint: "A short title, e.g. 'oSC14', to be used in URLs"
       = f.input :description, hint: markdown_hint('A description of the conference.'), input_html: { rows: 5, data: { provide: 'markdown-editable' } }
       = f.input :color, hint: 'The color will be used eg for the dashboard.', input_html: {size: 6, type: 'color'}
+      = f.input :ticket_layout, as: :select, collection: Conference.ticket_layouts.keys, hint: "Layout type for tickets of the conference."
```
Ana06: I would exchange the ticket layout and the conference picture, I mean moving ticket layout down as the conference picture is more important :wink:

app/models/conference.rb
```diff
@@ -73,6 +74,11 @@ class Conference < ActiveRecord::Base
   after_create :create_free_ticket
   after_update :delete_event_schedules
 
+  enum ticket_layout: {
```
Ana06: Why about writing this as array: 

http://edgeguides.rubyonrails.org/4_1_release_notes.html#active-record-enums

app/views/admin/conferences/edit.html.haml
Ana06: `collection` would be written as `Conference.ticket_layouts` without the `keys` if you use and array :wink:

app/views/physical_ticket/show.pdf.prawn
```diff
@@ -1,4 +1,4 @@
-prawn_document(filename: @file_name, page_layout: :portrait, :page_size =>'A4' ) do |pdf|
+prawn_document(filename: @file_name, page_layout: @conference.ticket_layout.to_sym, :page_size =>'A4' ) do |pdf|
```
Ana06: maybe moving `@conference.ticket_layout.to_sym` to the controller

This way you can also check that this variable is correctly assigned in the tests. :wink:

Ana06:
And the tests are failing :fearful:

app/views/admin/conferences/edit.html.haml
siddhantbajaj: But `Conference.ticket_layouts` would still return a map even if we use array. So I think we would still need `Conference.ticket_layouts.keys` as `collection`, right?

siddhantbajaj:
Updated Paper Trail. This [Issue](https://github.com/airblade/paper_trail/issues/798) was causing test failure.

app/views/admin/conferences/edit.html.haml
mdeniz: Looks likes yes, it returns a Hash

Ana06: yes, sorry :confounded:

PR #1571, opened 2017-07-07
Workflow to show, assign and unassign org admins role

### LINKED ISSUE #1532: Worklow to assign organization admins to organizations
Being resolved in #1571 
Since, only site admins have the ability to create organizations for now. Site admins should be able to assign organization admins of an organization. 

I propose the following workflow:

A new button in organization#index to manage organization admin of a particular organization.
screenshot_20170608_043224

The role views for each organization to assign organization admin just like roles for conferences.
screenshot_20170608_043556

shlok007:
Closes #1532 .

#### organization/role#show for organization admins:
screenshot_20170707_071306

#### admin/organizations#index
screenshot_20170707_071334

### DIFF
--- a/app/controllers/admin/organizations_controller.rb
+++ b/app/controllers/admin/organizations_controller.rb
@@ -1,6 +1,7 @@
 module Admin
   class OrganizationsController < Admin::BaseController
     load_and_authorize_resource :organization
+    before_action :verify_user, only: [:assign_org_admins, :unassign_org_admins]
 
     def index
       @organizations = Organization.all
@@ -43,7 +44,48 @@ module Admin
       end
     end
 
+    def assign_org_admins
+      if @user.has_role? 'organization_admin', @organization
+        flash[:error] = "User #{@user.email} already has the role organization admin"
+      elsif @user.add_role 'organization_admin', @organization
+        flash[:notice] = "Successfully added role organization admin to user #{@user.email}"
+      else
+        flash[:error] = "Coud not add role organization admin to #{@user.email}"
+      end
+
+      redirect_to admins_admin_organization_path(@organization)
+    end
+
+    def unassign_org_admins
+      if @user.remove_role 'organization_admin', @organization
+        flash[:notice] = "Successfully removed role organization admin from user #{@user.email}"
+      else
+        flash[:error] = "Could not remove role organization admin from user #{@user.email}"
+      end
+
+      redirect_to admins_admin_organization_path(@organization)
+    end
+
+    def admins
+      @role = @organization.roles.first
+      @users = @role.users
+      render 'show_org_admins'
+    end
+
     private
+
+    def user_params
+      params.require(:user).permit(:email)
+    end
+
+    def verify_user
+      @user = User.find_by(email: user_params[:email])
+      unless @user
+        redirect_to admins_admin_organization_path(@organization),
+                    error: 'Could not find user. Please provide a valid email!'
+        return
+      end
+    end
 
     def organization_params
       params.require(:organization).permit(:name, :description, :picture)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -20,7 +20,13 @@ Osem::Application.routes.draw do
   end
 
   namespace :admin do
-    resources :organizations
+    resources :organizations do
+      member do
+        get :admins
+        post :assign_org_admins
+        delete :unassign_org_admins
+      end
+    end
     resources :users do
       member do
         patch :toggle_confirmation
--- a/spec/controllers/admin/organizations_controller_spec.rb
+++ b/spec/controllers/admin/organizations_controller_spec.rb
@@ -167,5 +167,32 @@ describe Admin::OrganizationsController do
         end
       end
     end
+
+    describe 'POST #assign_org_admins' do
+      let(:org_admin_role) { Role.find_by(name: 'organization_admin', resource: organization) }
+
+      before do
+        post :assign_org_admins, id: organization.id,
+                                 user: { email: user.email }
+      end
+
+      it 'assigns organization_admin role' do
+        expect(user.roles).to eq [org_admin_role]
+      end
+    end
+
+    describe 'DELETE #unassign_org_admins' do
+      let(:org_admin_role) { Role.find_by(name: 'organization_admin', resource: organization) }
+      let!(:org_admin_user) { create(:user, role_ids: [org_admin_role.id]) }
+
+      before do
+        delete :unassign_org_admins, id: organization.id,
+                                     user: { email: org_admin_user.email }
+      end
+
+      it 'unassigns organization_admin role' do
+        expect(org_admin_user.reload.roles).to eq []
+      end
+    end
   end
 end
--- a/spec/controllers/admin/roles_controller_spec.rb
+++ b/spec/controllers/admin/roles_controller_spec.rb
@@ -1,7 +1,6 @@
 require 'spec_helper'
 
 describe Admin::RolesController do
-
   let(:conference) { create(:conference) }
   let(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let(:cfp_role) { Role.find_by(name: 'cfp', resource: conference) }
--- a/spec/features/roles_spec.rb
+++ b/spec/features/roles_spec.rb
@@ -98,6 +98,51 @@ feature Role do
     end
   end
 
+  context 'organization_admin' do
+    let!(:organization) { create(:organization) }
+    let!(:org_admin_role) { Role.find_by(name: 'organization_admin', resource: organization) }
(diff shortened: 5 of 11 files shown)


spec/controllers/admin/roles_controller_spec.rb
```diff
@@ -54,6 +56,32 @@
...
+
+    it 'assigns organization_admin role' do
+      expect(user1.roles).to eq [org_admin_role]
+    end
+  end
+
+  describe 'DELETE #unassign_org_admins' do
+    before do
+      sign_in admin
+      post :unassign_org_admin, organization_id: organization.id,
```
lagartoflojo: shouldn't this be `delete`?

lagartoflojo:
Please solve the conflicts :)

app/controllers/admin/roles_controller.rb
```diff
@@ -14,12 +16,15 @@ def index
...
-             end
+      unless @conference.nil?
+        @url = if @track
+                 toggle_user_track_admin_conference_role_path(@conference.short_title, @role.name, @track)
+               else
+                 toggle_user_admin_conference_role_path(@conference.short_title, @role.name)
+               end
+      end
       @users = @role.users
+      render 'show_org_admins' if @conference.nil?
```
lagartoflojo: I feel like this controller is doing too much, with too many conditionals. Is a conference loaded? Is an organization loaded? Is there a track? etc etc...
Would it make sense to move all the Org-admin related stuff to its own controller?

Other than that, the UI works great :)

app/controllers/admin/organizations_controller.rb
```diff
@@ -43,8 +44,49 @@ def destroy
...
+    def admins
+      @role = @organization.roles.first
+      @users = @role.users
+      render 'show_org_admins'
+    end
+
     private
 
+    def user_params
+      params.require(:user).permit(:email, :state)
```
lagartoflojo: where is the param `state` being used?

app/controllers/admin/roles_controller.rb
```diff
@@ -103,6 +103,10 @@ def toggle_user
 
     protected
 
+    def user_params
+      params.require(:user).permit(:email, :state)
+    end
+
```
lagartoflojo: Why did this file change?

config/routes.rb
```diff
@@ -20,7 +20,13 @@
   end
 
   namespace :admin do
-    resources :organizations
+    resources :organizations do
+        member do
+          get :admins
+          post :assign_org_admins
+          delete :unassign_org_admins
+        end
```
lagartoflojo: too many spaces of indentation here

spec/controllers/admin/organizations_controller_spec.rb
```diff
@@ -4,6 +4,7 @@
   let!(:admin) { create(:admin) }
   let!(:organization) { create(:organization) }
   let!(:user) { create(:user) }
+  let!(:user1) { create(:user, email: 'user1@osem.io') }
```
lagartoflojo: can we not use the existing user (line 6) for these tests?

spec/features/roles_spec.rb
```diff
@@ -98,6 +98,51 @@
...
+
+    context 'for the organization it belongs to' do
+      scenario 'successfully adds role organization_admin' do
+        click_link('Admins', href: admins_admin_organization_path(organization.id))
+
+        fill_in 'user_email', with: user_with_no_role.email
+        click_button 'Add'
+        user_with_no_role.reload
+
+        expect(user_with_no_role.has_role?('organization_admin', organization)).to eq true
```
lagartoflojo: Should we not also check that the user appears in the list of admins?

spec/features/roles_spec.rb
```diff
@@ -98,6 +98,51 @@
...
+    end
+
+    context 'for the organizations it does not belong to' do
+      scenario 'does not successfully add role organization_admin' do
+        click_link('Admins', href: admins_admin_organization_path(other_organization.id))
+
+        expect(page.has_field?('user_email')).to eq false
+      end
+
+      scenario 'does not successfully removes role organization_admin' do
```
lagartoflojo: When would this ever happen?

spec/features/roles_spec.rb
```diff
@@ -98,6 +98,51 @@
...
+        user_with_no_role.reload
+
+        expect(user_with_no_role.has_role?('organization_admin', organization)).to eq true
+      end
+
+      scenario 'successfully removes role organization_admin' do
+        click_link('Admins', href: admins_admin_organization_path(organization.id))
+
+        first('tr').find('.btn-danger').click
+        expect(organization_admin.has_role?('organization_admin', organization)).to eq false
```
lagartoflojo: We should also check that the user is no longer shown on the page.

spec/models/admin_ability_spec.rb
```diff
@@ -500,6 +508,7 @@
       it{ should be_able_to(:show, my_conference.program) }
       it{ should be_able_to(:update, new_track) }
       it{ should be_able_to(:manage, my_self_organized_track) }
+      it{ should_not be_able_to(:admins, organization) }
```
lagartoflojo: these test should also check that non-admins cannot `unassign_org_admins` or `assign_org_admins`, no?

spec/controllers/admin/organizations_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe Admin::OrganizationsController do
   let!(:admin) { create(:admin) }
   let!(:organization) { create(:organization) }
-  let!(:user) { create(:user) }
+  let!(:user) { create(:user, email: 'user@osem.io') }
```
lagartoflojo: What's the reason for giving the user a static email, instead of one generated by the factory? 
In your tests, you can just use `user.email` instead of `'user@osem.io'`.

spec/controllers/admin/organizations_controller_spec.rb
lagartoflojo: @shlok007 here I'm just asking a question, so instead of updating the PR with changes, let's please discuss the question, so I can understand your problems or thought process. Then, instead of applying changes on top of changes until I have no more questions left, we'll be able to just apply the correct changes once 😃

shlok007: Oh! `user` was used in the context `logged in as user with no role` so I thought it better not to assign org admin role to user and create a new object `user1` instead. Although it won't make a difference because we are using `let!`. 
Your question made me realize that I should avoid that!

spec/models/admin_ability_spec.rb
shlok007: yeah, I think we should! :slightly_smiling_face:

spec/controllers/admin/organizations_controller_spec.rb
shlok007: This didn't cross my mind at that moment. I'll change the definition.

spec/features/roles_spec.rb
shlok007: These tests won't be needed if we only allow site admins or org admins of that particular organization to access `admins` action in organization_controller.  
  
There are two possible sane ways here:
- Allow any user with role to view organization admins of a organization
- Only allow site admins or org admins of a organization to view org admins of that organization.

If any user with role has access to view org admins then he/she will know who to contact in case of a query or problem related to conferences of that organization. So, I personally suggest the first one.
@lagartoflojo

app/controllers/admin/roles_controller.rb
shlok007: I manually undid the changes in this file that caused this. It should be fixed now!

spec/features/roles_spec.rb
lagartoflojo: Let's go for option 1 then.

spec/controllers/admin/organizations_controller_spec.rb
lagartoflojo: ok!

spec/features/roles_spec.rb
shlok007: Now, this will happen for users with any role except the site admin and the org admin of that organization!

lagartoflojo:
Although, how do non-org-admins (users with roles) get to the `/admin/organizations/:id/admins` page?

shlok007:
User with any role has access to the `index` action for organizations in admin namespace defined here https://github.com/openSUSE/osem/blob/master/app/models/admin_ability.rb#L31

lagartoflojo:
ok! let's merge then

shlok007:
There are 19 commits in this PR. Should I squash them before merge?

lagartoflojo:
Yes, you are welcome to clean up the history. It doesn't all have to be in one commit though.

shlok007:
@lagartoflojo please do have a look.

PR #1572, opened 2017-07-07
Add Call for Booths

nasia-sam:
Adds 'Booth' type for cfp

### DIFF
--- a/spec/features/organizer_ability_spec.rb
+++ b/spec/features/organizer_ability_spec.rb
@@ -108,16 +108,32 @@ feature 'Has correct abilities' do
       visit edit_admin_conference_program_path(conference.short_title)
       expect(current_path).to eq(edit_admin_conference_program_path(conference.short_title))
 
+      # Only event type exists
+      visit new_admin_conference_program_cfp_path(conference.short_title)
+      expect(current_path).to eq(new_admin_conference_program_cfp_path(conference.short_title))
+
+      # Both event and booth exists
+      cfb = create(:cfp, cfp_type: 'booths', program: conference.program)
       visit new_admin_conference_program_cfp_path(conference.short_title)
       expect(current_path).to eq root_path
+
+      visit edit_admin_conference_program_cfp_path(conference.short_title, conference.program.cfp)
+      expect(current_path).to eq(edit_admin_conference_program_cfp_path(conference.short_title, conference.program.cfp))
 
       conference.program.cfp.destroy!
       visit new_admin_conference_program_cfp_path(conference.short_title)
       expect(current_path).to eq new_admin_conference_program_cfp_path(conference.short_title)
-      create(:cfp, program: conference.program)
 
-      visit edit_admin_conference_program_cfp_path(conference.short_title, conference.program.cfp)
-      expect(current_path).to eq(edit_admin_conference_program_cfp_path(conference.short_title, conference.program.cfp))
+      # Only booth exists
+      visit new_admin_conference_program_cfp_path(conference.short_title)
+      expect(current_path).to eq(new_admin_conference_program_cfp_path(conference.short_title))
+
+      visit edit_admin_conference_program_cfp_path(conference.short_title, cfb)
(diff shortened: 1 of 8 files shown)


spec/models/program_spec.rb
```diff
@@ -256,7 +256,7 @@
     it 'returns an array with the types for which a cfp doesn\'t exist' do
       expect(program.remaining_cfp_types).to eq(Cfp::TYPES)
       create(:cfp, cfp_type: 'events', program: program, end_date: Date.current + 1)
-      expect(program.remaining_cfp_types).to eq([])
+      expect(program.remaining_cfp_types).to eq(["booths"])
```
differentreality: You need to amend the test itself, so that you are actually testing what it says it is testing.
Just adding 'booths' here, might make the test pass, but it is not actually testing everything we want to test. 

What are the test cases for the method `remaining_cfp_types`?

nasia-sam: * There are `events` so the remaining type is `booths`
* There are `booths` so the remaining type is `events`
* There are both `events` and `booth` so the remaining type is empty
* Cfp is empty so the remaining types are both of them

differentreality: Great! Now you just have to make sure you are testing all of those cases!

nasia-sam:
@differentreality  I left some comments so it will be easier for you to review the tests, I can delete them later since there are no other comments in these files

spec/features/organizer_ability_spec.rb
```diff
@@ -108,16 +108,34 @@
       visit edit_admin_conference_program_path(conference.short_title)
       expect(current_path).to eq(edit_admin_conference_program_path(conference.short_title))
 
+      # only event type exits
```
differentreality: We can keep the comment in the code anyway, please fix the typo exits -> exists and capitalize only

differentreality:
@nasia-sam nice :+1:  
let's squash those commits, so that we can merge this PR.

spec/features/organizer_ability_spec.rb
nasia-sam: should I delete the rest of them and keep just this one ?

differentreality: You should just keep the ones that are helpful to our code, ie I can tell you are calling the new action from the path, you don't have to put that in a comment on the previous line. However I found the comment about which type exists useful, as you can't tell in a glance which one exists, and which not.

spec/models/program_spec.rb
```diff
@@ -253,10 +253,28 @@
...
+
+    it 'returns an array with the types for which a cfp doesn\'t exist, when only the Booth type does' do
+      expect(program.remaining_cfp_types).to eq(Cfp::TYPES)
+      create(:cfp, cfp_type: 'booths', program: program, end_date: Date.current + 1)
+      expect(program.remaining_cfp_types).to eq(['events'])
+    end
+
+    it 'returns an empty array when all the cfp types exist' do
+      expect(program.remaining_cfp_types).to eq(Cfp::TYPES)
+      create(:cfp, cfp_type: 'events', program: program, end_date: Date.current + 1)
```
differentreality: Why do we need to set a specific end_date? Does the date matter? Isn't the default end_date from the factory enough?

nasia-sam: The conference end_date is sooner so the tests fails if we use the factory's end_date

nasia-sam: I could change the factory from `end_date { 6.days.from_now }` to `end_date { 2.days.from_now }` . Tests seem to work correctly this way

differentreality:
@nasia-sam is there anything else you want to do here? I noticed you haven't moved this PR to `In Review `

nasia-sam:
@differentreality no, it's ready I must have forgotten to move it In Review :)

differentreality:
@nasia-sam you can go ahead and merge this! I have already approved the PR, and all checks have passed. You should be able to see a button to rebase. Please choose the option: `Rebase and merge`

PR #1575, opened 2017-07-10
Ticket show page

siddhantbajaj:
Physical Ticket show page for admin and user.

screen shot 2017-07-10 at 5 44 15 pm

### DIFF
--- a/app/views/physical_ticket/show.html.haml
+++ b/app/views/physical_ticket/show.html.haml
@@ -1,0 +1,51 @@
+.container
+  .row
+    .col-md-12
+      .page-header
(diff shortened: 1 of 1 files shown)


siddhantbajaj:
Show page prototype:

screen shot 2017-07-11 at 7 05 21 pm

mdeniz:
@Ana06 sure, we need to have the same info to be printed just in case someone can't open PDFs at least

mdeniz:
I guess that @siddhantbajaj should introduce the QR in both places at the same time... to not have a non working QR here

PR #1576, opened 2017-07-10
Add token field in ticket

siddhantbajaj:
- [x] Token field in physical ticket

### DIFF
--- a/app/models/physical_ticket.rb
+++ b/app/models/physical_ticket.rb
@@ -4,4 +4,19 @@ class PhysicalTicket < ActiveRecord::Base
   has_one :conference, through: :ticket_purchase
   has_one :user, through: :ticket_purchase
   has_many :ticket_scannings
+
+  before_create :set_token
+
+  private
+
+  def set_token
+    self.token = generate_token
+  end
+
+  def generate_token
+    loop do
+      token = SecureRandom.hex(10)
+      break token unless PhysicalTicket.exists?(token: token)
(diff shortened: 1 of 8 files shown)


siddhantbajaj:
@Ana06 
Sorry I didn’t knew those task were meant to be done in order.
I though it would be better to have the Qr that need to scanned ready first and then jump to the scanner.

Ana06:
@siddhantbajaj 

> @Ana06
> Sorry I didn’t knew those task were meant to be done in order.
> I though it would be better to have the Qr that need to scanned ready first and then jump to the scanner because in that way it would easier to define the routes and all the scanning logic . I suppose we need to have to the token field for reading of QR.

Then you will render a QR that is not working. Let's do it the other way around and then we can reopen this again :wink:

app/models/physical_ticket.rb
```diff
@@ -4,4 +4,19 @@ class PhysicalTicket < ActiveRecord::Base
...
+  private
+
+  def set_token
+    self.token = generate_token
+  end
+
+  def generate_token
+    loop do
+      token = SecureRandom.hex(10)
+      break token unless PhysicalTicket.where(token: token).exists?
```
Ana06: I think you can also write this like:

`break token unless PhysicalTicket.exists?(token: token)`

siddhantbajaj: 👍

PR #1577, opened 2017-07-10
Fix new conference links

shlok007:
Only `site admins` and `organization admins` have the permission to access `conference#new` from now on. Therefore, the `New Conference` link should not be visible for others.


lagartoflojo:
Are there no tests for these links? 😢

PR #1578, opened 2017-07-10
Correct past and upcoming conferences count in admin/organizations#index

shlok007:
Display correct past and upcoming conferences count in `organizations@index` : 

screenshot_20170608_043224

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -7,6 +7,8 @@ class Conference < ActiveRecord::Base
   resourcify :roles, dependent: :delete_all
 
   default_scope { order('start_date DESC') }
+  scope :upcoming, (-> { where('end_date >= ?', Date.current) })
+  scope :past, (-> { where('end_date < ?', Date.current) })
 
   belongs_to :organization
 
--- a/spec/models/conference_spec.rb
+++ b/spec/models/conference_spec.rb
@@ -1687,4 +1687,21 @@ describe Conference do
       expect{ room.save }.to change { subject.revision }.by(1)
     end
   end
+
+  describe '.upcoming' do
+    let!(:upcoming_conference) { create(:conference) }
+    let!(:past_conference) { create(:conference, start_date: Date.current - 1.days, end_date: Date.current - 1.days) }
+    subject { Conference.upcoming }
+
+    it { is_expected.to eq [upcoming_conference] }
+  end
+
+  describe '.past' do
+    let!(:upcoming_conference) { create(:conference) }
+    let!(:past_conference1) { create(:conference, start_date: Date.current - 1.days, end_date: Date.current - 1.days) }
+    let!(:past_conference2) { create(:conference, start_date: Date.current - 2.days, end_date: Date.current - 1.days) }
+    subject { Conference.past }
+
+    it { is_expected.to eq [past_conference1, past_conference2] }
+  end
 end
--- a/app/views/admin/organizations/index.html.haml
+++ b/app/views/admin/organizations/index.html.haml
@@ -20,9 +20,9 @@
             %td
               = organization.name
             %td
-              = organization.conferences.count
+              = organization.conferences.upcoming.count
             %td
-              = organization.conferences.count
+              = organization.conferences.past.count
             %td
               .btn-group
                 = link_to 'Edit', edit_admin_organization_path(organization),


app/helpers/admin/organizations_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Admin
+  module OrganizationsHelper
+    def upcoming_conferences_count(organization)
+      organization.conferences.where('end_date >= ?', Date.current).count
```
lagartoflojo: This condition (`.where('end_date >= ?', Date.current)`) can be made into a named scope in the Conference model:

```ruby
scope :upcoming, -> { where('end_date >= ?', Date.current) }
```

Then, you can do `organization.conferences.upcoming.count`, and I would just put that directly in the HAML, no need for a helper.

app/helpers/admin/organizations_helper.rb
```diff
@@ -0,0 +1,11 @@
...
+    end
+
+    def past_conferences_count(organization)
+      organization.conferences.count - upcoming_conferences_count(organization)
```
lagartoflojo: Note that this code does TWO counts, the total and the upcoming.
Can we do this in a single query? And create a scope for it?

app/models/conference.rb
```diff
@@ -6,6 +6,8 @@ class Conference < ActiveRecord::Base
   resourcify :roles, dependent: :delete_all
 
   default_scope { order('start_date DESC') }
+  scope :upcoming_conferences, (-> { where('end_date >= ?', Date.current) })
+  scope :past_conferences, (-> { where('end_date < ?', Date.current) })
```
lagartoflojo: no need for parenthesis around `-> {}`

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,23 @@
       expect{ conference.save }.to change{ EventSchedule.count }.from(2).to(1)
     end
   end
+
+  describe '#upcoming_conferences', blah: true do
```
lagartoflojo: what's `blah`?

app/models/conference.rb
lagartoflojo: I would just call them `upcoming` and `past`, because `Conference.past_conferences` looks repetitive. `Conference.past` is pretty clear.

spec/models/conference_spec.rb
lagartoflojo: Since this is a scope, it's a class method, and it should be described as `.upcoming` (with a dot, not with a `#`)

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,23 @@
...
+    let!(:upcoming_conference) { create(:conference) }
+    let!(:past_conf) { create(:conference, start_date: Date.current - 1.days, end_date: Date.current - 1.days) }
+
+    it 'should return correct count of upcoming conferences' do
+      expect(Conference.upcoming_conferences).to eq [upcoming_conference]
+    end
```
lagartoflojo: So you can also add `subject { Conference.upcoming }` above the `let!` calls and then your test can look like:

```ruby
it { is_expected.to eq [upcoming_conference] }
```

app/models/conference.rb
shlok007: The rubocop generates a warning if the parenthesis are missing.

lagartoflojo: What rule is it? I don't like it 😶

shlok007: looks like it has been fixed here https://github.com/bbatsov/rubocop/pull/4237 . but the fix is not backported to the version we are using i.e. `0.48.1` .

lagartoflojo: I see, let's leave it like that for now then.

spec/models/conference_spec.rb
```diff
@@ -1656,4 +1656,21 @@
       expect{ conference.save }.to change{ EventSchedule.count }.from(2).to(1)
     end
   end
+
+  describe '.upcoming' do
+    let!(:upcoming_conference) { create(:conference) }
+    let!(:past_conference) { create(:conference, start_date: Date.current - 1.days, end_date: Date.current - 1.days) }
+    let!(:subject) { Conference.upcoming }
```
lagartoflojo: note that it's not `let(:subject)`, but rather just `subject`:

```ruby
subject { Conference.upcoming }
```

lagartoflojo:
Ah, there's a conflict =/

shlok007:
Squashed and rebased!

PR #1579, opened 2017-07-11
Added TicketScanning Model

siddhantbajaj:


db/migrate/20170711102511_create_ticket_scannings.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTicketScannings < ActiveRecord::Migration
+  def change
+    create_table :ticket_scannings do |t|
+      t.integer :physical_ticket_id, null: false
+
+      t.timestamps null: false
```
Ana06: The idea is to use `created_at` as registration time, right?

mdeniz: @Ana06 You mean with "registration" the moment when a user's ticket for accessing to the conference is used, right? Then the answer will be YES

PR #1585, opened 2017-07-12
Update rubocop to 0.49.1

lagartoflojo:
I ran `rubocop --auto-gen-config` to re-generate `rubocop_todo.yml`... I guess that's the way to go?


differentreality:
That's the way to go! Let's get it in after the conflicts resolution.

differentreality:
What's different when we resolve offenses prior to generating the todo file?

Ana06:
@differentreality 

> What's different when we resolve offenses prior to generating the todo file?

because when you regenerate the todo file there shouldn't be offenses. If they are something is wrong.

differentreality:
Ah you mean so that all offenses are excluded in the newly generated todo file? That makes sense. Of course we would still have to fix the offenses afterwards.

lagartoflojo:
Closed in favor of #1591

PR #1586, opened 2017-07-12
Mention organization name in navigation bar of conference views

shlok007:
- Splashpage
screenshot_20170713_024349
  
- Dashboard
screenshot_20170713_035540
  
- admin/conference#edit
screenshot_20170713_040504

I am not sure if we need all of them. I started with the intention of creating mock-ups but ended up writing the actual code. :sweat_smile:   
  
@lagartoflojo can you please take a look? 

__IMPORTANT EDIT:__
- Removed all above views and added organization name in the top left of the navigation bar.


shlok007:
There is one catch though! If someone is using OSEM as single tenant, for his own organization, he might change `OSEM_NAME` to the name of his organization. Now if we mention the name of the conference's organization in different parts of the app, it can be a useless detail.

For example: In splashpage, the name of the organization will already be present in the title of the page. and in the top left of the navigation bar.

differentreality:
I think it might be a bit over the top to mention it on the conference#show page, it would read sth  like 

`openSUSE Conference 2017 (by openSUSE)` and what it shows is not controlled by organizers.

While in the admin namespace, we could mention
`openSUSE Organization`
on the top menu bar, next to OSEM, instead of the other mentions. What do you think?

lagartoflojo:
I agree, in the splashpage it's a bit too much to show the organization name. What do you think, @shlok007?

app/views/admin/conferences/show.html.haml
```diff
@@ -1,6 +1,6 @@
 %h1
   %span.fa.fa-tachometer
-  Dashboard for #{@conference.title}
+  Dashboard for #{@conference.title} ( by #{@conference.organization.name} )
```
lagartoflojo: I'd prefer to have no spaces inside the parenthesis:
`(by #{@conference.organization.name})`

shlok007:
>While in the admin namespace, we could mention
>openSUSE Organization
>on the top menu bar, next to OSEM, instead of the other mentions. What do you think?

Sure, it will definitely be better that way!

>While in the admin namespace, we could mention
openSUSE Organization
on the top menu bar, next to OSEM, instead of the other mentions. What do you think?

Organizations tend to change `OSEM_NAME` to the name of their organization. So, when OSEM is being used by a single orgainzation, the name on the menu bar in the admin namespace will read like:   
`openSUSE Events openSUSE Organization`
`SeaGL SeaGL Organization`

lagartoflojo:
> Organizations tend to change OSEM_NAME to the name of their organization.

Maybe it doesn't make sense to show OSEM_NAME there anymore.

shlok007:
>Maybe it doesn't make sense to show OSEM_NAME there anymore.

We can drop that in the admin namespce for views that belong to a particular conference and keep it for the common views like revision history or users. Does that sound good?

lagartoflojo:
sounds good to me!

shlok007:
Also, does it make sense to show the organization name instead of OSEM_NAME in the menu bar in splashpage of the conference?

lagartoflojo:
Yup, I think so

app/views/layouts/_navigation.html.haml
```diff
@@ -7,7 +7,10 @@
         %span.icon-bar
         %span.icon-bar
         %span.icon-bar
-      = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path, class: 'navbar-brand', title: 'Open Source Event Manager'
+      - if conference.nil? || conference.new_record?
+        = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path, class: 'navbar-brand', title: 'Open Source Event Manager'
+      - else
+        = link_to (ENV['OSEM_NAME'] || "#{conference.organization.name} Organization"), organizations_path, class: 'navbar-brand', title: 'Open Source Event Manager'
```
lagartoflojo: So we are preferring to show `OSEM_NAME` before the organization name?

shlok007: Ah! no we are planning to drop that. I'll correct this line.

app/views/layouts/_navigation.html.haml
```diff
@@ -7,7 +7,10 @@
         %span.icon-bar
         %span.icon-bar
         %span.icon-bar
-      = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path, class: 'navbar-brand', title: 'Open Source Event Manager'
+      - if conference.nil? || conference.new_record?
+        = link_to (ENV['OSEM_NAME'] || 'OSEM'), root_path, class: 'navbar-brand', title: 'Open Source Event Manager'
+      - else
+        = link_to "#{conference.organization.name} Organization", organizations_path, class: 'navbar-brand', title: 'Open Source Event Manager'
```
lagartoflojo: I think this should just say `conference.organization.name`, no need to append "Organization". Other than that, LGTM.

PR #1587, opened 2017-07-13
Show conference changelog ( rebased #1530 )

shlok007:
As described in #1530 :
>This comes from #1124, it is rebased, and has some fixes.
>
>* Use load_and_authorize_resource in versions controller
>* Add conference specific route to revision history page
>* Users with role can view revision_history only for the versions they have access to
>* Handle versions where conference_id is not set (records before papertrail was introduced)

Changes covered in [`fa86267`](https://github.com/openSUSE/osem/pull/1587/commits/fa86267802eec75b1762251335517d1fcf99760d) :
* Added revision history for organizations
* Added revision history for `organization admins` role
* Added revision history for users with role organization_admin

### DIFF
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,17 +1,17 @@
 module Admin
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_resource :conference, find_by: :short_title
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+
+      return if @conference.blank?
+      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
+      @versions = @versions.where(conference_id: @conference.id)
     end
 
     def revert_attribute
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_attribute, @version
-
       if params[:attribute] && @version.changeset.reject{ |_, values| values[0].blank? && values[1].blank? }.keys.include?(params[:attribute])
         if @version.item[params[:attribute]] == @version.changeset[params[:attribute]][0]
           flash[:error] = 'The item is already in the state that you are trying to revert it back to'
@@ -33,9 +33,6 @@ module Admin
     end
 
     def revert_object
-      @version = PaperTrail::Version.find(params[:id])
-      authorize! :revert_object, @version
-
       if @version.event != 'create'
         if @version.reify.save
           flash[:notice] = 'The selected change was successfully reverted'
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -1,14 +1,18 @@
 module Admin
   class VersionsController < Admin::BaseController
-    load_resource :conference, find_by: :short_title
+    load_resource :conference, find_by: :short_title, only: :index
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      @conf_ids_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+
+      if current_user.has_role? :organization_admin, :any
+        @conferences_with_role = Organization.with_role('organization_admin', current_user).map { |org| org.conferences.pluck :short_title }.flatten
+      end
+      @conferences_with_role.uniq!
 
       return if @conference.blank?
-      authorize! :index, PaperTrail::Version.new(conference_id: @conference.id)
-      @versions = @versions.where(conference_id: @conference.id)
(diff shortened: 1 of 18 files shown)


differentreality:
The `with ID` part is missing from our test suite.

shlok007:
Yeah, will push a commit for fixing this ASAP. 

Although, I am not sure about using IDs. Keeping a hosted version in mind, they can be helpful for someone like a `site administrator` who has access to the database and can fix things the way he wants to but they aren't that helpful for users with any other role as they won't be able to access the DB and see the IDs of things. This might be a redundant detail for them.

differentreality:
Yes, that's true.

If no record is available, they can't do anything anyway. If they see the IDs they still can't do anything (and they have some extra useless info). That's the downside.

On the other hand, if you are truly looking to see what went wrong, and something did go wrong, at least the site admin, or whoever has access to the DB can investigate a bit further. In my opinion this is more important. At the same time you may able to get the ID info from the database, if you know the version ID, which is currently visible on that table. Or using the ID of the non-longer-existing record, you can search for it in the datatable and perhaps extract some more info, even without looking at the DB.

I'd like to hear what others think on this though, and ultimately it's up to @lagartoflojo who I think is going to be reviewing this.

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,23 @@
...
   class VersionsController < Admin::BaseController
-    skip_authorization_check
+    load_resource :conference, find_by: :short_title, only: :index
+    load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conferences_with_role = []
```
lagartoflojo: I don't think we need to initialize this variable

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,23 @@
...
+      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
```
lagartoflojo: Is it not possible to add `:organization_admin` to this query, instead of having the lines 10-14?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,175 +1,225 @@
...
-  = version.event == 'create' ? 'to' : 'from'
-  = 'user'
+  - role = Role.find_by(id: users_role.role_id) if users_role
+  role
+  - if role.name == 'organization_admin'
+    / organization_admin belongs to organization and not conferencess
+    - organization = Organization.find(version.conference_id) 
+    = link_if_alive version, role.name,
+            admin_organizations_path, organization
+    / edit the above path to organization admin path after #1571 PR is merged
```
lagartoflojo: We should use `-# a comment` instead of `/ a comment`, because `/`-comments get output to the HTML.
I see 4 places in this PR where it has to be changed.

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -1,175 +1,225 @@
...
   - role = current_or_last_object_state(version.item_type, version.item_id)
-  = link_if_alive version, role.name,
-          admin_conference_role_path(conference_id: Conference.find(version.conference_id).short_title, id: role.name)
+  - role_name = role.try(:name) || PaperTrail::Version.where(item_type: 'Role', item_id: version.item_id).last.changeset[:name].second
+  - if role_name == 'organization_admin'
+    / organization_admin belongs to organization and not conferencess
+    - organization = Organization.find(version.conference_id) 
+    = link_if_alive version, role_name,
+            admin_organizations_path, organization
+    / edit the above path to organization admin path after #1571 PR is merged
```
lagartoflojo: let's merge the other PR first, an then this one

app/controllers/admin/versions_controller.rb
shlok007: Organizations have multiple conferences. I can't find a way to get all conferences from a `ActiveRecord::Relation` of multiple organizations in a single line query using active record.
Do let me know if it's possible :thinking:

app/controllers/admin/versions_controller.rb
```diff
@@ -1,17 +1,22 @@
...
 
     def index
-      authorize! :index, PaperTrail::Version.new(item_type: 'User')
-      conf_ids_for_organizer = current_user.is_admin? ? Conference.pluck(:id) : Conference.with_role(:organizer, current_user).pluck(:id)
-      @versions = PaperTrail::Version.where(["conference_id IN (?) OR item_type = 'User'", conf_ids_for_organizer])
+      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+
+      if current_user.has_role? :organization_admin, :any
+        org_ids = current_user.roles.where(name: 'organization_admin').pluck(:resource_id)
+        org_ids.each { |id| @conferences_with_role.concat(Organization.find(id).conferences.pluck(:short_title)).flatten }
```
lagartoflojo: How about:

```ruby
Organization.with_role('organization_admin', current_user).map { |org| org.conferences.pluck :short_title }.flatten
```

shlok007: Oh yes, that should work :slightly_smiling_face: !

shlok007:
@lagartoflojo please do have a look..

differentreality:
About the failing test, it is searching for string
removed role organization_admin with ID #{user_role.id} **to** user #{user.name} in organization #{conference.organization.name}")

while the existing text is removed...**from** user...

shlok007:
Yeah, it's clear now. Earlier, instead of `user_role.id` I was using `role.id` which was incorrect but the test still passed for conference roles because `role.id` was equal to `user_role.id` . Now, the same test written for organization role failed and that resulted in some confusion! :stuck_out_tongue:

shlok007:
@lagartoflojo all tests seem to pass and I think we can merge it now. Please do have a look.

lagartoflojo:
🤘

PR #1591, opened 2017-07-14
Update rubocop to 0.49.1

lagartoflojo:


lagartoflojo:
@Ana06 is this OK?

lagartoflojo:
Thank you! Strange bug, because some violations DID get excluded. Anyway, it's better to fix the violations than to exclude them :)

Ana06:
@lagartoflojo 

> Thank you! Strange bug, because some violations DID get excluded. Anyway, it's better to fix the violations than to exclude them :)

Exactly, it complains about some things and not about others. :wink:

PR #1592, opened 2017-07-14
Make the input fields in sign in look bootstrapy

svbeon:
Add class form control to inputs

Before, mobile:
img_20170715_143256

After, mobile:
img_20170715_143334

Before, Desktop:
img_20170715_143316

After, desktop:
img_20170715_143439

### DIFF
--- a/app/assets/stylesheets/osem-navbar.css.scss
+++ b/app/assets/stylesheets/osem-navbar.css.scss
@@ -20,14 +20,6 @@
             color: black;
             background-color: #eeeeee;
         }
-        input[type=text],
-        input[type=password] {
-            text-align: center;
-            padding: 5px;
-            display: block;
-            margin: auto;
-            width: 255px;
-        }
         .btn-group {
             width: 100%;
             text-align: center;
--- a/app/views/layouts/_navigation.html.haml
+++ b/app/views/layouts/_navigation.html.haml
@@ -60,13 +60,13 @@
             .dropdown-menu
               - if ENV['OSEM_ICHAIN_ENABLED'] == 'true'
                 = form_tag User.ichain_login_url do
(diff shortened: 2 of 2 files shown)


differentreality:
@svbeon could you please add before/after screenshots?

svbeon:
@differentreality sure thing, I'll just have to spin up a development copy that doesn't have all my colour changes that our main copy has

svbeon:
@differentreality there we go

differentreality:
@svbeon did you just add changes to css which weren't there initially?

svbeon:
@differentreality yes I realized those had been added since the version we're running at freenode.live (1.0) and those didn't look like what I was intending to suggest with this PR

differentreality:
I don't see any difference with, or without, the CSS code any more (while form-control class is there) in XS or MD viewport. So this looks good @svbeon :+1: 

:tada: Thanks for your contribution :tada: 

Let us know what you think of OSEM, and how it is working out for your conference!
FYI we also support openid (I noticed you don't have it enabled, and wanted to make sure it was out of choice). Feel free to swing by #osem too!

svbeon:
@differentreality thanks, we might pop up with a summary of what we think later on, probably after this year's conference, maybe earlier.

We did choose to disable openid because of some bug we couldn't trace (none of us speaks Ruby) that signed in everyone that used Facebook sign in as admins, which wasn't exactly preferable 😅 (and well, beeing months behind schedule did make disabling it a better option than trying to debug)

differentreality:
@svbeon  oh, that's weird. I'd happily investigate this for you, if you send me some server logs.

svbeon:
@differentreality would be wonderful. shall see if we can fix some logs, we're in the middle of launch and don't want to mess anything up, so i'll try creating a dev copy for it

PR #1596, opened 2017-07-15
Fix user already subscribed exception

### LINKED ISSUE #1589: [production][subscriptions#create] ActiveRecord::RecordInvalid: Validation failed: User has already been taken, User already subscri...
subscriptions#create threw this 0 times
<pre>
ActiveRecord::RecordInvalid: Validation failed: User has already been taken, User already subscribed!
</pre>
[https://events.opensuse.org/conference/oS3DW/subscriptions](https://events.opensuse.org/conference/oS3DW/subscriptions)"

[See this exception on Errbit](https://errbit-opensuse.herokuapp.com/apps/5620ca99dc71fa00b1000002/problems/596487efa8d5e5000cfd37a6)

dPliakos:
Rescue exception that is produced when user tries to subscribe in a conference that has already been subscribed to it.

Fixes #1589

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -5,17 +5,19 @@ class SubscriptionsController < ApplicationController
 
   def create
     @subscription = current_user.subscriptions.build(conference_id: @conference.id)
-    if @subscription.save!
-      redirect_to root_path, notice: "You have been subscribed to receive email notifications for #{@conference.short_title}."
+    if @subscription.save
+      redirect_to root_path, notice: "You have subscribed to receive email notifications for #{@conference.title}."
     else
-      redirect_to root_path, error: subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
     end
   end
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
+
+    redirect_to(root_path, error: "You are not subscribed to #{@conference.title}.") && return unless @subscription
     if @subscription.destroy
-      redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.short_title}."
+      redirect_to root_path, notice: "You have unsubscribed and you will not be receiving email notifications for #{@conference.title}."
     else
       redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
     end
--- a/app/models/subscription.rb
+++ b/app/models/subscription.rb
@@ -1,5 +1,4 @@
 class Subscription < ActiveRecord::Base
-  validates :user_id, uniqueness: { scope: [:conference_id] }
   belongs_to :conference
   belongs_to :user
 
--- a/spec/controllers/subscriptions_controller_spec.rb
+++ b/spec/controllers/subscriptions_controller_spec.rb
@@ -24,7 +24,7 @@ describe SubscriptionsController do
 
       it 'shows success message in flash notice' do
         post :create, conference_id: conference.short_title
-        expect(flash[:notice]).to match("You have been subscribed to receive email notifications for #{conference.short_title}")
+        expect(flash[:notice]).to match("You have subscribed to receive email notifications for #{conference.title}")
       end
 
       it 'subscribes user to conference' do
@@ -47,7 +47,7 @@ describe SubscriptionsController do
 
     it 'shows success message in flash notice' do
       delete :destroy, conference_id: conference.short_title
-        expect(flash[:notice]).to match("You have been unsubscribed and now you will not be receiving email notifications for #{conference.short_title}.")
+        expect(flash[:notice]).to match("You have unsubscribed and you will not be receiving email notifications for #{conference.title}.")
     end
   end
 end


shlok007:
Hi @dPliakos !

I think we should try to reason why/where the exception is being raised and figure out a way to avoid the exception instead of rescuing it after it has been raised.

app/controllers/subscriptions_controller.rb
```diff
@@ -10,6 +10,8 @@ def create
     else
       redirect_to root_path, error: subscription.errors.full_messages.to_sentence
     end
+  rescue ActiveRecord::RecordInvalid
```
differentreality: Instead of rescuing from the exception, we can avoid it all together. 
https://stackoverflow.com/a/27421738

app/controllers/subscriptions_controller.rb
```diff
@@ -10,6 +10,8 @@ def create
...
+    redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
   end
```
differentreality: Please post a screenshot of the error message being shown

dPliakos: I forgot it, sorry.
screenshot from 2017-07-15 23-56-27

dPliakos:
@shlok007 Thank you for the suggestion!!

dPliakos:
screenshot from 2017-07-15 23-56-27
screenshot from 2017-07-16 00-15-50

app/controllers/subscriptions_controller.rb
```diff
@@ -5,21 +5,21 @@ class SubscriptionsController < ApplicationController
 
   def create
     @subscription = current_user.subscriptions.build(conference_id: @conference.id)
-    if @subscription.save!
+    save_succeed = @subscription.save
```
differentreality: Why do we need to save this in a variable?!

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
   end
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
-    if @subscription.destroy
+    if @subscription && @subscription.destroy
       redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.short_title}."
     else
-      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
+      flash[:error] = 'You are not subscribed to that conference.'
```
differentreality: Why don't you add this in the beginning of the action so that we only go into destroy method when the user is subscribed? I would add the conference title to the error message as well.

differentreality:
Your first screenshot shows 2 errors. I'd only expect to see one, sine the only error is that the user is already subscribed. Can you please investigate why we get 2?

app/controllers/subscriptions_controller.rb
dPliakos: oops.. my bad...

app/controllers/subscriptions_controller.rb
dPliakos: But what should happen if the destroy fail again even the subscription (if this is possible) ? 
if !@subscription
  # go to error
if @subscription.destoy
  # go to root_path
else 
  # does this else have to exist here?

differentreality: Yes, the second else is showing an error in case the destroy method fails. 

What you want to do is redirect to root_path with an error unless the subscription exists. 
You can do that in the beginning as you are showing, and then you can leave the destroy if-else as it is.

dPliakos:
I see that `@subscription.errors.full_messages` is an array, so replacing the `to_sentence` function with the `full_messages.second` or `full_messages[2]` would make what we want (we also could make a for loop to find the error we want, if there is the possibility that this number might change in the future).

I could not find why exactly how @subscription get it's `errors:`, but I found that they have a relationship with `validates`. So maybe there is something like "inheritance" for errors while a resource is loaded. 

Second try to subscribe:
screenshot from 2017-07-16 19-22-18

Second try to unsubscribe:
screenshot from 2017-07-16 19-21-45

differentreality:
@dPliakos can you be more specific as to which validation is related to our case? (With a file/line reference)

dPliakos:
It seems like the validation we need is here: [app/models/subscriptions.rb, line 8](https://github.com/openSUSE/osem/blob/master/app/models/subscription.rb#L8)

But it seems `conference_id` is being validated again at line 2. 

So  if we remove that line, we will have the required error and  without any other consequences. Right ?

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,24 @@ class SubscriptionsController < ApplicationController
...
-      redirect_to root_path, error: subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
     end
   end
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
-    if @subscription.destroy
+
+    if !@subscription
```
differentreality: You can extract this part into a standalone statement, so that it is not part of the if statement.
redirect_to ..., error: '...' unless @subscription

You can also redirect directly to root_path, that's the only place subscription comes from anyway.

differentreality:
`User already subscribed!` message can become 'You are already subscribed to @conference.title' so that it is similar to other messages. 
Please use conference.title instead of short_title in messages.

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
+
+    redirect_to root_path, error: "You are not subscribed to #{@conference.short_title}." and return unless @subscription
     if @subscription.destroy
       redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.short_title}."
     else
-      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
+      redirect_to  root_path, error: "You are not subscribed to #{@conference.short_title}."
```
differentreality: Remove the extra space here

app/models/subscription.rb
```diff
@@ -1,5 +1,5 @@
 class Subscription < ActiveRecord::Base
-  validates :user_id, uniqueness: { scope: [:conference_id] }
+
```
differentreality: remove empty line

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
 
   def create
     @subscription = current_user.subscriptions.build(conference_id: @conference.id)
-    if @subscription.save!
+    if @subscription.save
       redirect_to root_path, notice: "You have been subscribed to receive email notifications for #{@conference.short_title}."
```
differentreality: BTW why don't we also fix this message: You have subscribed.... for @conference.title

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
     else
-      redirect_to root_path, error: subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
     end
   end
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
+
+    redirect_to root_path, error: "You are not subscribed to #{@conference.short_title}." and return unless @subscription
```
differentreality: empty line after this one

differentreality: @conference.title instead of short_title

differentreality: If you use parenthesis for redirect, you can do `&& return`, and make rubocop happy! Let me know if you understand why this happens.

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
     if @subscription.destroy
       redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.short_title}."
     else
-      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: "You are not subscribed to #{@conference.short_title}."
```
differentreality: This error message was for our case in line 18, the error message here, in case @subscription.destroy fails should remain the same.

dPliakos: So we delete this error and leave the redirect_to as is, or delete both ?

differentreality: Take a moment to think what makes sense, what we are doing here, and why, and let me know what _you_ propose we should do :)

differentreality:
@dPliakos ping

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
     else
-      redirect_to root_path, error: subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
     end
   end
 
   def destroy
     @subscription = current_user.subscriptions.find_by(conference_id: @conference.id)
+
+    (redirect_to root_path, error: "You are not subscribed to #{@conference.title}.") && return unless @subscription
```
differentreality: Why do you need parenthesis at the beginning of the line instead of after redirect_to
`redirect(path, error: ...)`

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
     if @subscription.destroy
-      redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.short_title}."
+      redirect_to root_path, notice: "You have been unsubscribed and now you will not be receiving email notifications for #{@conference.title}."
```
differentreality: Lose the `been` here as well.

app/controllers/subscriptions_controller.rb
```diff
@@ -5,19 +5,21 @@ class SubscriptionsController < ApplicationController
...
     else
-      redirect_to root_path, error: @subscription.errors.full_messages.to_sentence
+      redirect_to root_path, error: "You are not subscribed to #{@conference.title}."
```
differentreality: But you showed that error message already, and if the user is not subscribed, will never reach this section of the code! What's confusing you here?

dPliakos: What happens if the destroy fails, even if the subscription exists. Is it possible in the first place ?

app/controllers/subscriptions_controller.rb
dPliakos: Miss-understood the syntax. it' s clear now.

app/controllers/subscriptions_controller.rb
differentreality: We have a subscription, we try to destroy it, that doesn't work for some reason.
We still have a subscription. And we show the error message. I still don't understand why you are trying to change this line in the first place.

dPliakos: The message says that there is not such a subscription. But there is (and it fails to be destroyed).

differentreality: How can I reproduce this on my machine?

dPliakos: I do not know if it possible for such an error to exist :p

differentreality: @dPliakos you just said that the error says there is no such subscription. Well, how do you know that if you can't see the error message?! I really don't understand your point. Please be clear, so that we can finish this.

dPliakos: def destroy:
  check if there is the subscription.
  if no: go back with an error
  if yes: destroy it
  //   question: Is there the possibility that this destroy() function may fail for other reason than the absence of subscription ?

differentreality: Possibly. If it does, it can show the message(s) that made it fail. What's wrong with that?

dPliakos: What message to put there? :D XD

app/controllers/subscriptions_controller.rb
dPliakos: done

app/controllers/subscriptions_controller.rb
dPliakos: produces syntax error.

dPliakos: It as my bad syntax. Fixed.

differentreality:
@dPliakos do you see the failing tests? Please fix them too!!

dPliakos:
finally :D

differentreality:
:+1:  Nice work @dPliakos :tada:

PR #1597, opened 2017-07-17
Track request acceptance and activation

AEtherC0r3:
* Add finite state machine to the track model and allow the state to be changed
* Make the state column not null and add the default value 'new' to it. Also, the regular tracks are marked as confirmed by default
* Create cfp for tracks and give to signed_in users the ability to create requests for tracks
* Turn the cfp scopes into class methods, because, if the answer is nil then a scope will return all records.
* Require a room and dates for accepted/confirmed self-organized tracks (This will be used for the scheduling later)
* Allow the track submitter to request specific dates for his tracks
* Don't allow the requests for tracks to be deleted
* Don't allow the submitter and the track organizers to edit a request after it has been accepted
* Make the cfp_active column not null and set the value to true if it is nill (the track is an old regular track)
* Restrict track selection in proposals to only the confirmed tracks with the cfp_active flag
* Make the tracks views similar to the proposals views
* Add a new field to the tracks (relevance) so the submitter of a request can explain how it relates to the conference

Important workflow notes:
* When a track request is accepted then the role of the track organizer for that track is created
* When a track is confirmed then the submitter is assigned that role
* When a track is cancelled/withdrawn then that role is revoked from all the users that have it and the track's association with events is removed

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -17,7 +17,7 @@ module Admin
 
     def index
       @events = @program.events
-      @tracks = @program.tracks
+      @tracks = @program.tracks.confirmed.cfp_active
       @difficulty_levels = @program.difficulty_levels
       @event_types = @program.event_types
       @tracks_distribution_confirmed = @conference.tracks_distribution(:confirmed)
@@ -43,7 +43,7 @@ module Admin
     end
 
     def show
-      @tracks = @program.tracks
+      @tracks = @program.tracks.confirmed.cfp_active
       @event_types = @program.event_types
       @comments = @event.root_comments
       @comment_count = @event.comment_threads.count
@@ -58,7 +58,7 @@ module Admin
 
     def edit
       @event_types = @program.event_types
-      @tracks = Track.all
+      @tracks = @program.tracks.confirmed.cfp_active
       @comments = @event.root_comments
       @comment_count = @event.comment_threads.count
       @user = @event.submitter
--- a/app/controllers/admin/tracks_controller.rb
+++ b/app/controllers/admin/tracks_controller.rb
@@ -62,7 +62,7 @@ module Admin
     private
 
     def track_params
-      params.require(:track).permit(:name, :description, :color, :short_name, :cfp_active)
+      params.require(:track).permit(:name, :description, :color, :short_name, :cfp_active, :start_date, :end_date, :room_id)
     end
   end
 end
--- a/app/controllers/admin/tracks_controller.rb
+++ b/app/controllers/admin/tracks_controller.rb
@@ -19,6 +19,7 @@ module Admin
 
     def create
       @track = @program.tracks.new(track_params)
+      @track.state = 'confirmed'
       if @track.save
         redirect_to admin_conference_program_tracks_path(conference_id: @conference.short_title),
                     notice: 'Track successfully created.'
--- a/app/controllers/admin/tracks_controller.rb
+++ b/app/controllers/admin/tracks_controller.rb
@@ -20,6 +20,7 @@ module Admin
     def create
       @track = @program.tracks.new(track_params)
       @track.state = 'confirmed'
+      @track.cfp_active = true
       if @track.save
         redirect_to admin_conference_program_tracks_path(conference_id: @conference.short_title),
                     notice: 'Track successfully created.'
@@ -60,10 +61,55 @@ module Admin
       end
     end
 
+    def restart
+      update_state(:restart, "Review for #{@track.name} started!")
+    end
+
+    def to_accept
+      update_state(:to_accept, "Track #{@track.name} marked as a possible acceptance!")
+    end
+
+    def accept
+      if @track.room && @track.start_date && @track.end_date
+        update_state(:accept, "Track #{@track.name} accepted!")
+      else
+        flash[:alert] = 'Please make sure that the track has a room and start/end dates before accepting it'
+        redirect_to edit_admin_conference_program_track_path(@conference.short_title, @track)
+      end
+    end
+
+    def confirm
+      update_state(:confirm, "Track #{@track.name} confirmed!")
+    end
+
+    def to_reject
+      update_state(:to_reject, "Track #{@track.name} marked as a possible rejection!")
+    end
+
+    def reject
+      update_state(:reject, "Track #{@track.name} rejected!")
+    end
+
+    def cancel
+      update_state(:cancel, "Track #{@track.name} canceled!")
+    end
+
     private
 
     def track_params
       params.require(:track).permit(:name, :description, :color, :short_name, :cfp_active, :start_date, :end_date, :room_id)
+    end
+
+    def update_state(transition, notice)
+      errors = @track.update_state(transition)
+
+      if errors.blank?
+        flash[:notice] = notice
+      else
+        flash[:error] = errors
+      end
+
+      redirect_back_or_to(admin_conference_program_tracks_path(conference_id: @conference.short_title))
     end
   end
 end
--- a/app/controllers/admin/tracks_controller.rb
+++ b/app/controllers/admin/tracks_controller.rb
@@ -4,6 +4,9 @@ module Admin
     load_and_authorize_resource :program, through: :conference, singleton: true
     load_and_authorize_resource through: :program, find_by: :short_name
 
+    # Show flash message with ajax calls
+    after_action :prepare_unobtrusive_flash, only: :toggle_cfp_inclusion
+
     def index; end
 
     def show
@@ -55,9 +58,13 @@ module Admin
     def toggle_cfp_inclusion
       @track.cfp_active = !@track.cfp_active
       if @track.save
-        head :ok
+        flash[:notice] = "Successfully changed cfp inclusion of #{@track.name} to #{@track.cfp_active}"
       else
-        head :unprocessable_entity
+        flash[:error] = "Failed to toggle cfp inclusion of #{@track.name} to #{@track.cfp_active}"
+      end
+
+      respond_to do |format|
+        format.js
       end
     end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,7 +56,7 @@ module ApplicationHelper
   end
 
   def tracks(conference)
-    all = conference.program.tracks.map {|t| t.name}
+    all = conference.program.tracks.map { |t| t.name if !t.self_organized? || t.confirmed? && t.cfp_active }.compact
     first = all[0...-1]
     last = all[-1]
     ts = ''
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,7 +56,7 @@ module ApplicationHelper
   end
 
   def tracks(conference)
-    all = conference.program.tracks.where(state: 'confirmed', cfp_active: true).pluck(:name)
+    all = conference.program.tracks.confirmed.cfp_active.pluck(:name)
     first = all[0...-1]
     last = all[-1]
     ts = ''
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -56,7 +56,7 @@ module ApplicationHelper
   end
 
   def tracks(conference)
-    all = conference.program.tracks.map { |t| t.name if !t.self_organized? || t.confirmed? && t.cfp_active }.compact
+    all = conference.program.tracks.where(state: 'confirmed', cfp_active: true).pluck(:name)
     first = all[0...-1]
     last = all[-1]
     ts = ''
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -39,4 +39,12 @@ module EventsHelper
       content_tag :span, 'REPLACEMENT', class: (['label', 'label-info'] + label_classes)
     end
   end
+
+  def track_selector_input(form)
+    if @program.tracks.any?
+      form.input :track_id, as: :select,
+                            collection: @program.tracks.where(state: 'confirmed', cfp_active: true).pluck(:name, :id),
+                            include_blank: true
+    end
+  end
 end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -95,6 +95,14 @@ class Ability
     can :manage, Commercial, commercialable_type: 'Event', commercialable_id: user.events.pluck(:id)
 
     can [:destroy], Openid
+
+    can [:new, :create], Track do |track|
+      track.new_record? && track.program.cfps.for_tracks.try(:open?)
+    end
+
+    can [:index, :show, :edit, :update], Track do |track|
+      user == track.submitter
+    end
   end
 
   # Abilities for users with roles wandering around in non-admin views.
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -100,8 +100,10 @@ class Ability
       track.new_record? && track.program.cfps.for_tracks.try(:open?)
     end
 
-    can [:index, :show, :edit, :update], Track do |track|
-      user == track.submitter
+    can [:index, :show, :restart, :confirm, :withdraw], Track, submitter_id: user.id
+
+    can [:edit, :update], Track do |track|
+      user == track.submitter && !(track.accepted? || track.confirmed?)
     end
   end
 
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -6,6 +6,7 @@ class Track < ActiveRecord::Base
 
   belongs_to :program
   belongs_to :submitter, class_name: 'User'
+  belongs_to :room
   has_many :events, dependent: :nullify
 
   has_paper_trail only: [:name, :description, :color], meta: { conference_id: :conference_id }
@@ -24,6 +25,10 @@ class Track < ActiveRecord::Base
             inclusion: { in: %w(new to_accept accepted confirmed to_reject rejected canceled withdrawn) },
             if: :self_organized?
   validates :cfp_active, inclusion: { in: [true, false] }, if: :self_organized?
+  validates :start_date, presence: true, if: :accepted_or_confirmed?
+  validates :end_date, presence: true, if: :accepted_or_confirmed?
+  validates :room, presence: true, if: :accepted_or_confirmed?
+  validate :valid_dates, if: :accepted_or_confirmed?
 
   before_validation :capitalize_color
 
@@ -105,6 +110,33 @@ class Track < ActiveRecord::Base
     end
   end
 
+  ##
+  # Checks if the track is accepted
+  # ====Returns
+  # * +true+ -> If the track's state is 'accepted'
+  # * +false+ -> If the track's state isn't 'accepted'
+  def accepted?
+    state == 'accepted'
+  end
+
+  ##
+  # Checks if the track is confirmed
+  # ====Returns
+  # * +true+ -> If the track's state is 'confirmed'
+  # * +false+ -> If the track's state isn't 'confirmed'
+  def confirmed?
+    state == 'confirmed'
+  end
+
+  ##
+  # Checks if the track is accepted or confirmed
+  # ====Returns
+  # * +true+ -> If the track's state is 'accepted' or 'confirmed'
+  # * +false+ -> If the track's state is neither 'accepted' nor 'confirmed'
+  def accepted_or_confirmed?
+    accepted? || confirmed?
+  end
+
   private
 
   def generate_guid
@@ -127,5 +159,29 @@ class Track < ActiveRecord::Base
   # Creates the role of the track organizer
   def create_organizer_role
     Role.where(name: 'track_organizer', resource: self).first_or_create(description: 'For the organizers of the Track')
+  end
+
+  def valid_dates
+    return unless start_date && end_date
+
+    if program && program.conference && program.conference.start_date && (start_date < program.conference.start_date)
+      errors.add(:start_date, "can't be before the conference start date (#{program.conference.end_date})")
+    end
+
+    if program && program.conference && program.conference.start_date && (end_date < program.conference.start_date)
+      errors.add(:end_date, "can't be before the conference start date (#{program.conference.end_date})")
+    end
+
+    if program && program.conference && program.conference.end_date && (start_date > program.conference.end_date)
+      errors.add(:start_date, "can't be after the conference end date (#{program.conference.end_date})")
+    end
+
+    if program && program.conference && program.conference.end_date && (end_date > program.conference.end_date)
+      errors.add(:end_date, "can't be after the conference end date (#{program.conference.end_date})")
+    end
+
+    if start_date > end_date
+      errors.add(:start_date, 'can\'t be after the end_date')
+    end
   end
 end
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -1,4 +1,5 @@
 class Track < ActiveRecord::Base
+  include ActiveRecord::Transitions
   include RevisionCount
 
   resourcify :roles, dependent: :delete_all
@@ -18,12 +19,49 @@ class Track < ActiveRecord::Base
             uniqueness: {
               scope: :program
             }
-  validates :state, presence: true, if: :self_organized?
+  validates :state,
+            presence: true,
+            inclusion: { in: %w(new to_accept accepted confirmed to_reject rejected canceled withdrawn) },
+            if: :self_organized?
   validates :cfp_active, inclusion: { in: [true, false] }, if: :self_organized?
 
   before_validation :capitalize_color
 
-  after_create :create_organizer_role, if: :self_organized?
+  state_machine initial: :pending do
+    state :new
+    state :to_accept
+    state :accepted
+    state :confirmed
+    state :to_reject
+    state :rejected
+    state :canceled
+    state :withdrawn
+
+    event :restart do
+      transitions to: :new, from: [:rejected, :withdrawn, :canceled]
+    end
+    event :readiness_to_accept do
+      transitions to: :to_accept, from: [:new]
+    end
+    event :accept do
+      transitions to: :accepted, from: [:new, :to_accept], on_transition: :create_organizer_role
+    end
+    event :confirm do
+      transitions to: :confirmed, from: [:accepted], on_transition: :assign_role_to_submitter
+    end
+    event :readiness_to_reject do
+      transitions to: :to_reject, from: [:new]
+    end
+    event :reject do
+      transitions to: :rejected, from: [:new, :to_reject]
+    end
+    event :cancel do
+      transitions to: :canceled, from: [:to_accept, :to_reject, :accepted, :confirmed], on_transition: :revoke_role_and_cleanup
+    end
+    event :withdraw do
+      transitions to: :withdrawn, from: [:new, :to_accept, :to_reject, :accepted, :confirmed], on_transition: :revoke_role_and_cleanup
+    end
+  end
 
   def conference
     program.conference
@@ -41,6 +79,30 @@ class Track < ActiveRecord::Base
 
   def to_param
     short_name
+  end
+
+  def transition_possible?(transition)
+    self.class.state_machine.events_for(current_state).include?(transition)
+  end
+
+  # Gives the role of the track_organizer to the submitter
+  def assign_role_to_submitter
+    submitter.add_role 'track_organizer', self
+  end
+
+  # Revokes the track organizer role and removes the track from events that have it set
+  def revoke_role_and_cleanup
+    role = Role.find_by(name: 'track_organizer', resource: self)
+
+    if role
+      role.users.each do |user|
+        user.remove_role 'track_organizer', self
+      end
+    end
+
+    events.each do |event|
+      event.track = nil
+    end
   end
 
   private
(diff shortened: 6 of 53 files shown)


AEtherC0r3:
When an organizer is about to accept a Track request, I need to somehow prompt him to fill in the room and dates for the track. Should I redirect him to #edit or use something like a popup?

app/controllers/admin/tracks_controller.rb
```diff
@@ -59,10 +60,50 @@ def toggle_cfp_inclusion
       end
     end
 
+    def restart
+      update_state(:restart, 'Review started!')
+    end
+
+    def ready_to_accept
+      update_state(:readiness_to_accept, 'Track request is signaling readiness for acceptance!')
```
differentreality: Can't help myself thinking of 
http://top40-charts.com/songs/lyrics.php?sid=28986&string=Catch%20Me%20When%20I%27m%20Falling

Perhaps in our case we are not signaling anything to anyone, but rather marking the track as a possible acceptance :)

sayalilunkad:
@AEtherC0r3 I would think a redirect would be more appropriate here.

app/models/track.rb
```diff
@@ -43,6 +86,70 @@ def to_param
...
+    self_organized? && (accepted? || confirmed?)
+  end
+
+  def update_state(transition)
+    error = ''
+
+    begin
+      send(transition)
+    rescue Transitions::InvalidTransition => e
+      error += "State update failed. #{e.message} "
```
differentreality: Perhaps the error message can be a bit more informative regarding which event and/or which state the function was trying to change.

app/views/admin/tracks/_change_state_dropdown.html.haml
```diff
@@ -0,0 +1,34 @@
+- if track.transition_possible? :restart
+  %li= link_to 'Start review',
+  restart_admin_conference_program_track_path(@conference.short_title, track),
+  method: :patch, id: "restart_track_#{track.id}"
+
+- if track.transition_possible? :readiness_to_accept
```
differentreality: :to_accept

app/views/admin/tracks/_change_state_dropdown.html.haml
```diff
@@ -0,0 +1,34 @@
...
+  %li= link_to 'Signal readiness to accept track request',
+  ready_to_accept_admin_conference_program_track_path(@conference.short_title, track),
+  method: :patch, id: "ready_to_accept_track_#{track.id}"
```
differentreality: ready not needed

app/views/admin/tracks/_change_state_dropdown.html.haml
```diff
@@ -0,0 +1,34 @@
+- if track.transition_possible? :restart
+  %li= link_to 'Start review',
+  restart_admin_conference_program_track_path(@conference.short_title, track),
+  method: :patch, id: "restart_track_#{track.id}"
+
+- if track.transition_possible? :readiness_to_accept
+  %li= link_to 'Signal readiness to accept track request',
```
differentreality: Possible acceptance perhaps?

app/views/admin/tracks/index.html.haml
```diff
@@ -15,6 +15,9 @@
         %th Color
```
differentreality: I'd move this column closer to the title, so that the first columns are either the same or similar to the original view, ie
name, descr, color, room, dates, submitter, state, actions
I don't think we need to show short_name at all.

app/models/track.rb
AEtherC0r3: The message of the exception already contains that info

app/views/proposals/_proposal_form.html.haml
```diff
@@ -8,7 +8,7 @@
 
     - if @program.tracks.any?
       = f.input :track_id, as: :select,
-                collection: @program.tracks.map {|track| ["#{track.name}", track.id] },
+                collection: @program.tracks.map {|track| ["#{track.name}", track.id] if !track.self_organized? || track.confirmed? && track.cfp_active }.compact,
```
differentreality: Space after curly bracket please

differentreality: Shall we extract that to a method for our program to simply our code here?

app/views/admin/tracks/index.html.haml
```diff
@@ -53,6 +61,21 @@
               - else
                 %i.fa.fa-check
             %td
+              - if track.self_organized? && track.room
```
differentreality: Room is not mandatory for regular tracks as we know them till now, but can't we just be showing this? If and when an organizer might want to assign a room at a track, the organizer should be free to do that. Same with start/end dates. What do you think?

app/controllers/admin/tracks_controller.rb
```diff
@@ -59,10 +60,55 @@ def toggle_cfp_inclusion
...
+
+    def to_accept
+      update_state(:to_accept, "Track request for #{@track.name} marked as a possible acceptance!")
+    end
+
+    def accept
+      if @track.room && @track.start_date && @track.end_date
+        update_state(:accept, "Track request for #{@track.name} accepted!")
+      else
+        flash[:notice] = 'Please make sure that the track has a room and start/end dates before accepting it'
```
differentreality: This sounds like an error, not notice

app/models/track.rb
```diff
@@ -43,6 +86,70 @@ def to_param
...
+    role = Role.find_by(name: 'track_organizer', resource: self)
+
+    if role
+      role.users.each do |user|
+        user.remove_role 'track_organizer', self
+      end
+    end
+
+    events.each do |event|
+      event.track = nil
```
AEtherC0r3: Because the track organizer can't accept and schedule events yet, I'm only removing the track from the event. When the scheduling for tracks is implemented, I will also destroy the track's schedule and revert the event back to the new state

app/models/ability.rb
```diff
@@ -89,6 +89,16 @@ def signed_in(user)
     can :manage, Commercial, commercialable_type: 'Event', commercialable_id: user.events.pluck(:id)
 
     can [:destroy], Openid
+
+    can [:new, :create], Track do |track|
+      track.new_record? && track.program.cfps.for_tracks.try(:open?)
+    end
+
+    can [:index, :show, :restart, :confirm, :withdraw], Track, submitter_id: user.id
```
differentreality: I should be able to access show action of confirmed tracks, to get their details.

app/helpers/application_helper.rb
```diff
@@ -56,7 +56,7 @@ def dynamic_association(association_name, title, form_builder, options = {})
   end
 
   def tracks(conference)
-    all = conference.program.tracks.map {|t| t.name}
+    all = conference.program.tracks.where(state: 'confirmed', cfp_active: true).pluck(:name)
```
differentreality: Let's add a scope for confirmed and cfp_active tracks.

app/views/admin/tracks/index.html.haml
```diff
@@ -23,39 +25,56 @@
...
+                = link_to track.room.name, admin_conference_venue_room_path(@conference.short_title, track.room.id)
               - else
                 N/A
             %td
-              %span.label{style: "background-color: #{track.color}; color: #{ contrast_color(track.color) }"}
-                = track.color
+              - if track.start_date
+                = track.start_date.strftime('%A, %B %-d. %Y')
+              - else
+                N/A
```
differentreality: We should remove this all together, and show nothing instead.

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,84 @@
...
+          If your track request is accepted, the conference organizers expect you to confirm that you will be able to hold it.
+          Then you will gain the Track organizer role.
+          %br
+          If your track request is rejected, you can either live with that or adapt it and resubmit it for review again.
+          %br
+          If something changes and you can't organize the track any more, you should withdraw it.
+
+        %p
+          %strong
+            Why do I need to add more information?
```
differentreality: This is not needed, is it? We don't ask people to add extra information, except for the fields that they completed during request creation.

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,84 @@
...
+        %p
+          The more information you add to your proposal, the more likely it is that the conference organizers accept your track request.
+
+        %p
+          %strong
+            Why do I need to register to the conference?
```
differentreality: Same for this one.

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,84 @@
...
+                %span.label{style: "background-color: #{track.color}; color: #{ contrast_color(track.color) }"}
+                  = track.color
+              %td
+                - if track.start_date
+                  From:
+                  = track.start_date.strftime('%A, %B %-d. %Y')
+              %td
+                - if track.end_date
+                  To:
+                  = track.end_date.strftime('%A, %B %-d. %Y')
```
differentreality: What about the room?

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,84 @@
...
+                  %span{ title: 'In review', class: 'fa fa-eye' }
+                - elsif track.state == 'accepted'
+                  %span{ title: 'Accepted', class: 'fa fa-check text-muted' }
+                - elsif track.state == 'confirmed'
+                  %spam{ title: 'Confirmed', class: 'fa fa-check text-success' }
+                - elsif %w(rejected withdrawn canceled).include? track.state
+                  %span{ title: track.state.humanize, class: 'fa fa-ban'}
+              %td
+                = link_to(conference_program_track_path(@conference.short_title, track)) do
+                  = track.name
```
differentreality: What about adding a background with the track's color, instead of having a separate column for it?

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,84 @@
...
+                  = track.color
+              %td
+                - if track.start_date
+                  From:
+                  = track.start_date.strftime('%A, %B %-d. %Y')
+              %td
+                - if track.end_date
+                  To:
+                  = track.end_date.strftime('%A, %B %-d. %Y')
+              %td.pull-right
```
differentreality: Something looks broken on the table (the horizontal line gets interrupted)

screenshot from 2017-07-23 02-35-37

app/views/admin/tracks/_form.html.haml
```diff
@@ -8,11 +8,16 @@
...
       = f.input :color, input_html: {size: 6, type: 'color'}, required: true
+      = f.input :start_date, as: :string, input_html: { id: 'registration-period-start-datepicker', start_date: @conference.start_date, end_date: @conference.end_date, readonly: 'readonly', required: @track.self_organized_and_accepted_or_confirmed? }
+      = f.input :end_date, as: :string, input_html: { id: 'registration-period-end-datepicker', readonly: 'readonly', required: @track.self_organized_and_accepted_or_confirmed? }
+      - if @conference.venue
+        = f.input :room, as: :select, collection: (@conference.venue.rooms).map {|room| ["#{room.name}", room.id]}, include_blank: true, label: 'Room', input_html: { class: 'select-help-toggle', required: @track.self_organized_and_accepted_or_confirmed? }
+      - else
+        Please add a venue with rooms, if you want to select a room for the track.
       = f.input :description, input_html: {rows: 2, data: { provide: 'markdown-editable' } }, hint: markdown_hint
```
differentreality: Let's add a hint that this will become public.

differentreality: And let's add one more field for submitter to explain how this track relates to the conference, why it should be accepted, and what's the users relation to the track's content.

AEtherC0r3: This is fixed in the non admin view

app/models/ability.rb
AEtherC0r3: You can find the details of the track in admin/Tracks#show

AEtherC0r3: The users can see the name and description of the every confirmed track included in the cfp in the conference's splashpage

app/views/admin/tracks/index.html.haml
```diff
@@ -6,59 +6,61 @@
...
-                = track.color
+              = track.start_date.strftime('%A, %B %-d. %Y') if track.start_date
             %td
-              - if track.self_organized?
-                = track.state
-              - else
-                N/A
+              = track.end_date.strftime('%A, %B %-d. %Y') if track.end_date
             %td
+              = link_to track.submitter.name, admin_user_path(track.submitter) if track.self_organized?
```
differentreality: This potentially throws an exception

```
ActiveRecord::RecordNotFound in Admin::UsersController#show
Couldn't find Conference with 'id'=152

```
app/models/user.rb:168:in `block in get_roles'
app/models/user.rb:167:in `get_roles'
app/views/admin/users/show.html.haml:25:in `block in _app_views_admin_users_show_html_haml___579614431558334695_133750840'
app/views/admin/users/show.html.haml:18:in `each'
app/views/admin/users/show.html.haml:18:in `_app_views_admin_users_show_html_haml___579614431558334695_133750840'

app/models/ability.rb
differentreality: Let's also show them then the room and the dates of the track, when available.

AEtherC0r3:
Tracks#new


Tracks#index


Tracks#show


admin/Tracks#index


admin/Tracks#show for regular tracks


admin/Tracks#show for self-organized tracks


My Tracks in the user's menu


Conference#show call for tracks in the splashpage

differentreality:
:+1: This looks good :D
I want to have another look before merging this, but in the meantime why don't you squash your commits?

@sayalilunkad what do you think? Anything to improve? Anything there might be missed?

app/models/track.rb
```diff
@@ -43,6 +86,70 @@ def to_param
...
+    self.class.state_machine.events_for(current_state).include?(transition)
+  end
+
+  # Gives the role of the track_organizer to the submitter
+  def assign_role_to_submitter
+    submitter.add_role 'track_organizer', self
+  end
+
+  # Revokes the track organizer role and removes the track from events that have it set
+  def revoke_role_and_cleanup
```
AEtherC0r3: I've updated the description

AEtherC0r3:
@sayalilunkad 
We have a drop down to select the room

If the conference doesn't have a venue or the venue doesn't have room then instead of the drop down it shows a message that informs the user to create a venue and rooms

Selecting a start date


Selecting an end date

app/views/tracks/index.html.haml
```diff
@@ -6,38 +6,74 @@
...
+        %p
+          %strong
+            What happens next with my track request?
+        %p
+          If you submit a track request, the conference organizers will review it and either accept or reject it.
+          %br
+          If your track request is accepted, the conference organizers expect you to confirm that you will be able to hold it.
+          Then you will gain the Track organizer role.
+          %br
+          If your track request is rejected, you can either live with that or adapt it and resubmit it for review again.
```
AEtherC0r3: This is the same as in [proposals](https://github.com/openSUSE/osem/blob/master/app/views/proposals/index.html.haml#L40). I thought of it as kinda playful, so I kept it like that. If you insist, I will change it

sayalilunkad:
@AEtherC0r3 the drop down seems to be broken for me.
screenshot from 2017-08-08 10-28-27
Not sure if there are changes that have not been pushed since it works for you. But you can check that, from the screenshot it looks good!

app/views/tracks/index.html.haml
sayalilunkad: Ah then let's keep it consistent!

differentreality:
@sayalilunkad it's because you have not set a venue and/or rooms. You have to add dates *and* a room to the track, in order to be able to accept the track. 

Perhaps we should come up with an idea of how to make this more understandable / visible to our user.

sayalilunkad:
@differentreality ah thanks! We could probably embed  admin/conferences/gsoc/venue to 'add a venue'  so it's easier to navigate. But it's not a blocker.

AEtherC0r3:
@sayalilunkad @differentreality Now, it is in bold and has links to admin_conference_venue and admin_conference_venue_rooms

sayalilunkad:
@AEtherC0r3 great thanks!

AEtherC0r3:
@sayalilunkad You're welcome!

spec/controllers/admin/tracks_controller_spec.rb
```diff
@@ -253,4 +258,204 @@
...
+    end
+
+    it 'changes the track\'s state to to_accept' do
+      expect(self_organized_track.state).to eq 'to_accept'
+    end
+  end
+
+  describe 'PATCH #accept' do
+    shared_examples 'fails to accept' do |start_date, end_date, room|
+      before :each do
```
differentreality: I'd re-write this part of your shared examples, and move these to the individuals examples.
The true/false values don't make a lot of sense, unless you scroll to see what they refer to. And the way you are assigning values to some of the attributes inside the shared_example is not making it easy to read/understand. 

You should be defining what you need defined and then call the shared_example, ie the common things you want to run for multiple scenarios.

You can create the self_organized_track record every time with the attributes you want, and always use that variable inside your shared example.

spec/features/tracks_spec.rb
```diff
@@ -39,7 +39,7 @@
       within('table#tracks') do
         expect(page.has_content?(track.name)).to be false
         expect(page.has_content?(track.description)).to be false
-        expect(page.assert_selector('tr', count: 1)).to be true
+        expect(page.has_content?('No data available in table')).to eq true
```
differentreality: In the feature test I would expect a test checking whether or not the Track.count has changed (in this case reduced by 1)

differentreality: Also the feature test is very important to test if this is properly working for users that submit a request (non admin), but this is only testing it for organizers.

spec/models/cfp_spec.rb
```diff
@@ -5,21 +5,35 @@
   let!(:conference) { create(:conference, end_date: Date.today) }
   let!(:cfp) { create(:cfp, start_date: Date.today - 2, end_date: Date.today - 1, program_id: conference.program.id) }
```
differentreality: I would explicitly add the type here, and rename the variable accordingly.

spec/models/cfp_spec.rb
```diff
@@ -5,21 +5,35 @@
...
   end
 
+  describe '.for_events' do
+    it 'returns the cfp for events when it exists' do
+      expect(conference.program.cfps.for_events).to be_a Cfp
+      expect(conference.program.cfps.for_events.cfp_type).to eq('events')
+    end
+
+    it 'returns nil when the cfp for events doesn\'t exist' do
+      conference.program.cfp.destroy
```
differentreality: I'd call the cfp for_events specifically here to avoid confusion.

differentreality: I would also test in a different test that program.cfp returns the cfp for events.

spec/models/track_spec.rb
```diff
@@ -23,23 +24,155 @@
...
+      context 'includes' do
+        it 'when track is confirmed' do
+          confirmed_track = create(:track, state: 'confirmed', program: @program)
+          expect(@program.tracks.confirmed.include?(confirmed_track)).to eq true
+        end
+      end
+
+      context 'excludes' do
+        %w[new to_accept accepted to_reject rejected canceled withdrawn].each do |state|
+          it "when track is #{state.humanize}" do
```
differentreality: Perhaps we can rename the tests, so that it makes more sense to read it. Right now this will read:
`Track scope #confirmed excludes when track is accepted`

spec/models/track_spec.rb
```diff
@@ -54,4 +187,226 @@
...
+      end
+    end
+  end
+
+  # accepted? and confirmed? are mutually exclusive (they can't be both true)
+  describe '#self_organized_and_accepted_or_confirmed?' do
+    context 'returns true' do
+      context 'when self_organized? returns true' do
+        before :each do
+          allow(track).to receive(:self_organized?).and_return(true)
```
differentreality: Why aren't you creating a self_organized track instead?

app/models/track.rb
```diff
@@ -1,10 +1,12 @@
...
   include RevisionCount
 
   resourcify :roles, dependent: :delete_all
 
   belongs_to :program
   belongs_to :submitter, class_name: 'User'
+  belongs_to :room
   has_many :events, dependent: :nullify
 
   has_paper_trail only: [:name, :description, :color], meta: { conference_id: :conference_id }
```
differentreality: Why only track changes for these attributes?

app/models/track.rb
```diff
@@ -18,12 +20,57 @@ class Track < ActiveRecord::Base
...
-  validates :state, presence: true, if: :self_organized?
-  validates :cfp_active, inclusion: { in: [true, false] }, if: :self_organized?
+  validates :state,
+            presence: true,
+            inclusion: { in: %w(new to_accept accepted confirmed to_reject rejected canceled withdrawn) }
+  validates :cfp_active, inclusion: { in: [true, false] }
+  validates :start_date, presence: true, if: :self_organized_and_accepted_or_confirmed?
+  validates :end_date, presence: true, if: :self_organized_and_accepted_or_confirmed?
+  validates :room, presence: true, if: :self_organized_and_accepted_or_confirmed?
+  validates :relevance, presence: true, if: :self_organized?
```
differentreality: What about the presence of description for self organized tracks?

app/views/admin/tracks/index.html.haml
```diff
@@ -6,59 +6,61 @@
         Categorize events in your conference
 .row
   .col-md-12
-    %table.table.table-hover#tracks
+    %table.table.table-hover.table-striped.table-bordered.datatable#tracks
       %thead
         %th Name
```
differentreality: We should add a field like ID, in order to be able to see which one is the oldest request, and which one is the newest.

app/views/admin/tracks/index.html.haml
```diff
@@ -6,59 +6,61 @@
...
+              = track.start_date.strftime('%A, %B %-d. %Y') if track.start_date
             %td
-              - if track.self_organized?
-                = track.state
-              - else
-                N/A
+              = track.end_date.strftime('%A, %B %-d. %Y') if track.end_date
             %td
+              = link_to track.submitter.name, admin_user_path(track.submitter) if track.self_organized?
+            %td.text-center
```
differentreality: This column is not sortable

app/views/admin/tracks/index.html.haml
```diff
@@ -6,59 +6,61 @@
...
+              = check_box_tag "#{@conference.short_title}_#{track.short_name}", track.id, track.cfp_active,
+                class: 'switch-checkbox', method: :patch,
+                url: toggle_cfp_inclusion_admin_conference_program_track_path(@conference.short_title, id: track.short_name)+"?included=",
```
differentreality: When changing this, we have no indication about it, ie there is no flash message letting the users know if this happened, successfully or not.

app/helpers/events_helper.rb
```diff
@@ -39,4 +39,12 @@ def canceled_replacement_event_label(event, event_schedule, *label_classes)
       content_tag :span, 'REPLACEMENT', class: (['label', 'label-info'] + label_classes)
     end
   end
+
+  def track_selector_input(form)
+    if @program.tracks.any?
+      form.input :track_id, as: :select,
+                            collection: @program.tracks.where(state: 'confirmed', cfp_active: true).pluck(:name, :id),
```
AEtherC0r3: I didn't have the scopes when I initially wrote that code, and then I forgot about it

differentreality:
:+1: please squash the latest commits, so that we can merge it!

app/controllers/admin/events_controller.rb
```diff
@@ -17,7 +17,7 @@ def set_timezone_for_this_request(&block)
 
     def index
       @events = @program.events
-      @tracks = @program.tracks
+      @tracks = @program.tracks.confirmed.cfp_active
```
differentreality: I am not sure we need the tracks to have cfp_active, so that we are able to move events from one track to another during evaluation period.
Let's look into that later on.

app/views/admin/tracks/index.html.haml
```diff
@@ -6,59 +7,64 @@
...
             %td
               .btn-group{role: "group"}
-                = link_to 'Edit', edit_admin_conference_program_track_path(@conference.short_title, track),
-                method: :get, class: 'btn btn-primary'
+                - if can? :edit, track
+                  = link_to 'Edit', edit_admin_conference_program_track_path(@conference.short_title, track), class: 'btn btn-primary'
                 - if can? :destroy, track
-                  = link_to 'Delete', admin_conference_program_track_path(@conference.short_title, track),
-                  method: :delete, class: 'btn btn-danger',
+                  = link_to 'Delete', admin_conference_program_track_path(@conference.short_title, track), method: :delete, class: 'btn btn-danger',
```
differentreality: When is it possible to delete a track? 
It could be when it is not self_organized, to keep things the same as before for conferences that don't have a cfp for tracks at all.

app/views/admin/tracks/toggle_cfp_inclusion.js.erb
```diff
@@ -0,0 +1,8 @@
+$('.alert').remove();
+
+track_id = <%= @track.id %>;
+track_cfp_td = $('#cfp_switch_' + track_id);
+track_cfp_value = <%= @track.cfp_active %>;
+
+track_cfp_td.attr('data-order', track_cfp_value);
+$('#tracks').DataTable().cell(track_cfp_td).invalidate();
```
differentreality: I'd add a comment, as to what this line is doing, and why we need it.

app/views/conferences/_schedule_splashpage.html.haml
```diff
@@ -10,13 +10,22 @@
         - if @conference.splashpage and @conference.program.tracks.any? and @conference.splashpage.include_tracks
           See rock-star speakers cover the topics of
       - if @conference.splashpage and @conference.splashpage.include_tracks
-        - @conference.program.tracks.each_slice(3) do |slice|
+        - @conference.program.tracks.confirmed.cfp_active.each_slice(3) do |slice|
```
differentreality: What if the cfp_active attribute is false for a self_organized track, because its cfp ended sooner?

app/views/conferences/show.html.haml
```diff
@@ -45,6 +45,10 @@
       %section#program
         = render 'schedule_splashpage'
 
+    - if @conference.program.cfps.for_tracks.try(:open?) && @conference.splashpage.include_cfp
```
differentreality: Perhaps it makes sense to show this **after** cfp for events?

differentreality:
:tada: Yay we can get this in :tada:

spec/models/cfp_spec.rb
AEtherC0r3: Program#cfp is already tested in https://github.com/openSUSE/osem/blob/master/spec/models/program_spec.rb#L243

app/views/admin/tracks/index.html.haml
AEtherC0r3: It's not possible to delete a track only when it is self-organized https://github.com/openSUSE/osem/blob/master/app/models/admin_ability.rb#L74-L76, hence the old behavior has been preserved

app/views/conferences/show.html.haml
AEtherC0r3: Why? A conference is probably going to have the call for tracks before the call for papers, so that the tracks are available for proposal submissions

app/controllers/admin/events_controller.rb
AEtherC0r3: Fixed in #1631

app/views/conferences/_conference_details.html.haml
```diff
@@ -30,6 +30,10 @@
                 = link_to "Register", new_conference_conference_registration_path(conference.short_title), class: "btn btn-default", disabled: cannot?(:new, Registration.new(conference_id: conference.id))
                 - if cannot?(:new, Registration.new(conference_id: conference.id)) && conference.registration_limit_exceeded?
                   Sorry, no places left
+            - if !current_user.nil? && current_user.tracks.where(program: conference.program).length > 0
+              = link_to "My Track Requests", conference_program_tracks_path(conference.short_title), class: 'btn btn-default'
+            - elsif can? :new, conference.program.tracks.new
+              = link_to "Submit Track Request", new_conference_program_track_path(conference.short_title), class: 'btn btn-default'
```
AEtherC0r3: haml-lint doesn't have a linter that checks for single or double quotes https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/linter/README.md, but it should have caught it through its [rubocop integration](https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/linter/README.md#rubocop). I made some tries and failed to make it catch it

spec/models/track_spec.rb
AEtherC0r3: If something changes in the way a self-organized track is defined and the track factory isn't updated accordingly, then this test might fail, because the `self_organized` trait doesn't satisfy the definition of a self-organized track. A failure of the factory shouldn't be a failure of this test. So, this way the test is independent from the factory

AEtherC0r3:
The remaining comments have been addressed in #1639

PR #1600, opened 2017-07-18
Send mail on ticket confirmation with attached ticket pdf.

siddhantbajaj:


db/migrate/20170717154257_add_ticket_confirmation_to_email_settings.rb
```diff
@@ -0,0 +1,7 @@
+class AddTicketConfirmationToEmailSettings < ActiveRecord::Migration
+  def change
+    add_column :email_settings, :send_on_ticket_confirmation, :boolean, default: false
```
Ana06: Do we really need all that?

I mean. I would say this email should always be sent. And I think we don't need to customize the message.

app/models/physical_ticket.rb
```diff
@@ -4,4 +4,12 @@ class PhysicalTicket < ActiveRecord::Base
   has_one :conference, through: :ticket_purchase
   has_one :user, through: :ticket_purchase
   has_many :ticket_scannings
+
+  after_create :send_ticket_confirmation_mail
+
+  def send_ticket_confirmation_mail
+    if conference.email_settings.send_on_ticket_confirmation?
+      Mailbot.ticket_confirmation_mail(conference, user, self).deliver_later
```
Ana06: We were talking about not sending two emails, as when paying the tickets there is an email already sent, what is the state of that?

siddhantbajaj: No unfortunately we cannot send the ticket as a attachment in stripe confirmation email. Customisation of stripe confirmation email is only limited to changing the organisation name, logo.

Ana06: so we will always send two emails in this case, right?

siddhantbajaj: Yes one for payment confirmation and the the other one for ticket confirmation.

db/migrate/20170717154257_add_ticket_confirmation_to_email_settings.rb
siddhantbajaj: But what about the case when conference does not have any tickets just registration?

siddhantbajaj: >  And I think we don't need to customize the message.

So should set the message manually or like use the same message that sent in registration confirmation email?

siddhantbajaj: > I would say this email should always be sent.

👍

Ana06: You can send a different email, but the subject can always be something like _Your tickets for ConferenceName_. So I mean that you can find a default message that can be used for all conferences, so the admin doesn't need to configure this. It is also less confusing for user, who will always receive similar email.

And in this way, you don't need to store anything in the DB. :wink:

app/views/mailbot/ticket_confirmation_template.text.erb
```diff
@@ -0,0 +1,8 @@
+Dear <%= @user.name %>,
+
+Thanks! You have successfully booked a <%= @physical_ticket.ticket.title %> ticket for the event <%= @conference.title %>.
```
Ana06: I would also add the token in the email

app/models/physical_ticket.rb
```diff
@@ -6,6 +6,7 @@ class PhysicalTicket < ActiveRecord::Base
   has_many :ticket_scannings
 
   before_create :set_token
+  after_create :send_ticket_confirmation_mail
```
Ana06: there is something I don't like with this. If you but 3 tickets, then we will send 3 different emails. I would instead send this email after creating the physical_tickets in one single email :thinking:

siddhantbajaj:
screen shot 2017-07-26 at 11 32 42 am

Issue #1601, opened 2017-07-18
The update to Ruby 2.4 broke the development environment (vagrant)


AEtherC0r3:
It looks like the package ruby2.4-devel that was added in #1588 in `bootstrap.sh` doesn't exist in openSUSE 42.2 repositories

![2017-07-18 10-43-58](https://user-images.githubusercontent.com/3638310/28314623-ff2d2568-6bc3-11e7-904d-b10d42f9ebbb.png)

differentreality:
I get the same thing on a new vagrant box, and I can't zypper in ruby2.4-devel in vagrant either. 

@rahul2240 any ideas? How did you get it working on your local instance?

rahul2240:
I am really sorry, i was continuously facing problems with setting up vagrant. so i didnt . I made these changes and osem started working on localhost. So i thought editing bootstrap.sh will work in vagrant. I apologize for my mistake.

lagartoflojo:
Do the new versions of gems from #1588 still work with the earlier Ruby version?
Maybe we ought to revert the changes to [bootstrap.sh](https://github.com/openSUSE/osem/pull/1588/files#diff-3af436e6760a0c5f1e3d16a3384853a8) for now.

AEtherC0r3:
I'm getting the following warning in the rails console
`Expected string default value for '--serializer'; got true (boolean)`
`irb: warn: can't alias context from irb_context.`
Otherwise, the current master seems to work with the ruby2.3 package

differentreality:
I am having trouble with my student because of this change which is impossible to implement in vagrant.
Is there any solution for this? Otherwise we should revert the ruby version change in bootstrap.sh ASAP.

lagartoflojo:
Unfortunately, I never got Vagrant to work on my local machine, so I use Docker and I can't really test if Vagrant works.
I would revert the changes to `bootstrap.sh` (assuming everything else works OK).

Ana06:
@lagartoflojo 

>Do the new versions of gems from #1588 still work with the earlier Ruby version?
Maybe we ought to revert the changes to bootstrap.sh for now.

>I would revert the changes to bootstrap.sh (assuming everything else works OK).

Reverting to a previous Ruby version just because we have a missing repo in vagrant makes no sense at all. It is just a small fix. :see_no_evil: 


@differentreality 
> I am having trouble with my student because of this change which is impossible to implement in vagrant.
[...]

PR #1603, opened 2017-07-20
Use ruby2.3 in vagrant

AEtherC0r3:
The package for ruby 2.4 doesn't exist in the repositories of openSUSE 42.2.
Related to #1601


Ana06:
Let's me check if I can fix Ruby 2.4 before you change it to 2.3 please!

PR #1588, opened 2017-07-13
Ruby updated to version 2.4

rahul2240:
The aim is to update the ruby version to 2.4.
The Rails version is updated to 4.2.8 which support ruby version 2.4 
Fixes issue #1563

PR #1604, opened 2017-07-20
Fix vagrant broken after updating to Ruby 2.4

Ana06:
Ruby got updated in https://github.com/openSUSE/osem/pull/1588, but there was a repo missed, so Ruby 2.4 can be installed in our Vagrant machine. :bowtie: :tada: 

Closes https://github.com/openSUSE/osem/issues/1601

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -2,6 +2,8 @@
 pushd /vagrant
 
 echo -e "\ninstalling required software packages...\n"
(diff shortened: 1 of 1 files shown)


Ana06:
@rahul2240 maybe you want to take a look! :wink:

AEtherC0r3:
Maybe we should verify that the key we are importing is the right one or at least load it over a TLS protected connection

differentreality:
What's the status of this?

PR #1605, opened 2017-07-20
Add tests for booth controller

nasia-sam:
Tests for the Admin::BoothsController. 
It tests the behavior of  #index, #new, #create, #edit, #update and #destroy actions when the user is admin.

### DIFF
--- a/spec/controllers/admin/booths_controller_spec.rb
+++ b/spec/controllers/admin/booths_controller_spec.rb
@@ -1,0 +1,148 @@
+require 'spec_helper'
+
+describe Admin::BoothsController do
+
+  let(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let(:booth) { create(:booth, title: 'Title', conference: conference) }
+  let(:admin) { create(:admin) }
+
+  context 'not logged in user' do
+
+    describe 'GET index' do
+      it 'does not render admin/booths#index' do
+        get :index, conference_id: conference.short_title
+        expect(response).to redirect_to(user_session_path)
+      end
+    end
+
+    describe 'GET show' do
+      it 'does not render admin/booths#show' do
+        get :show, id: booth.id, conference_id: conference.short_title
+        expect(response).to redirect_to(user_session_path)
+      end
+    end
+  end
+
+  context 'user is admin' do
+    before :each do
+      sign_in admin
+    end
+
+    describe 'GET index' do
+      before { get :index, conference_id: conference.short_title }
+
+      it 'assigns attributes for booths' do
+        expect(assigns(:booths)).to eq([booth])
+      end
+
+      it 'renders index template' do
+        expect(response).to render_template('index')
+      end
+    end
+
+    describe 'GET new' do
+      before { get :new, conference_id: conference.short_title }
+
+      it 'assigns attributes for booths' do
+        expect(assigns(:booth)).to be_a_new(Booth)
+      end
+
+      it 'renders new template' do
+        expect(response).to render_template('new')
+      end
+    end
+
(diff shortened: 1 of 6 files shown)


spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    describe 'GET index' do
+      it 'does not render admin/booths#index' do
+        get :index, conference_id: conference.short_title
+        expect(response).to redirect_to(user_session_path)
+      end
+    end
+
+    describe 'GET show' do
+      it 'does not render admin/booths#show' do
```
differentreality: Later on, we probably want to change this, so that a not signed in user, can access show action for confirmed booths.

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+
+        it 'shows success message' do
+          expect(flash[:notice]).to match('Booth successfully created.')
+        end
+      end
+
+      context 'create action fails' do
+        before { post :create, booth: attributes_for(:booth, title: ''), conference_id: conference.short_title }
+        it 'does not create any record' do
```
differentreality: Let's add an empty line before this one.

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
+require 'spec_helper'
+
+describe Admin::BoothsController do
+
+  let(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let(:booth) { create(:booth, conference: conference) }
```
differentreality: Let's define the title here, as we are changing it in the tests. That way we can explicitly know its initial value versus its changed one in the tests.

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+
+        it 'show success message' do
+          delete :destroy, id: booth.id, conference_id: conference.short_title
+          expect(flash[:notice]).to match('Booth successfully destroyed.')
+        end
+      end
+    end
+
+    describe 'GET #edit' do
```
differentreality: Let's move these tests closer to #update

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+      it 'renders edit template' do
+        get :edit, id: booth.id, conference_id: conference.short_title
```
differentreality: Can we move this to a before block?

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+        expect(response).to render_template('edit')
+      end
+
+      it 'assigns boot  h variable' do
```
differentreality: Typo

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+          booth.reload
+          expect(booth.title).to eq('different')
+        end
+      end
+    end
+
+    describe 'DELETE #destroy' do
+      context 'deletes successfully' do
+        it 'booth deleted' do
+          delete :destroy, id: booth.id, conference_id: conference.short_title
```
differentreality: Can we move this to a before block?

differentreality:
Nicely done! Please add a description for the PR as well. I've left some comments for refining the controller tests. As soon as they are taken care of, we can merge this!

PR #1606, opened 2017-07-20
Refactor ticket purchase

### LINKED ISSUE #1527: Refactor ticket purchases
1. The ticket_purchases controller is not authorizing TicketPurchase
https://github.com/openSUSE/osem/blob/master/app/controllers/ticket_purchases_controller.rb

2. Theoretically, a ticked_purchase can occur at any time (after the conference, after registration period is over, etc), as there is no ability limiting it

3. Ticket graphs over time, in admin/conferences#index and admin/conferences#show, show float numbers for the number of tickets sold, obviously we can't have sold 2.5 tickets. 

screenshot from 2017-06-06 18-16-10

4. At the ticket turnover distribution graph in admin/tickets#index, the amount of the turnover could be formatted into a more humanly readable way

screenshot from 2017-06-06 19-07-03

siddhantbajaj:
closes #1527

### DIFF
--- a/app/assets/javascripts/osem-dashboard.js
+++ b/app/assets/javascripts/osem-dashboard.js
@@ -55,7 +55,6 @@ $(function() {
     }
 
     function draw_line_chart(animation, $canvas){
-        var options = get_animation({}, animation);
         var chart_data = create_dataset($canvas);
         var weeks = $canvas.parent().data('weeks');
         var data = {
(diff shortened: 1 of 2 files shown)


siddhantbajaj:
image

siddhantbajaj:
@differentreality 
> Theoretically, a ticked_purchase can occur at any time (after the conference, after registration period is over, etc), as there is no ability limiting it.

True but in practical there is no way a ticket_purchase can be created after the registration period right? So should we really add a validation for this, if in case you meant to add a validation here.

PR #1566, opened 2017-07-01
Introduce Booths for admin

nasia-sam:

PR #1607, opened 2017-07-21
add booths to non admin

nasia-sam:
Create controller and views for non-admin users.
#index and #show are displayed the same way in both admin and non-admin, while both render the same form for the #new and #edit actions which is located in views/booths.
Views are due to be modified.


app/controllers/booths_controller.rb
```diff
@@ -0,0 +1,52 @@
+class BoothsController < ApplicationController
+  before_action :authenticate_user!
+  load_resource :conference, find_by: :short_title
+  load_and_authorize_resource through: :conference
+
+  def index; end
```
differentreality: Even if I am an admin/organizer, in the non-admin view I only want to see my own booths (ie booths I am the submitter/responsible of)

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,44 @@
...
+              %td
+                = image_tag(booth.picture.thumb.url, width: '20%')
+              %td
+                = link_to booth.title, conference_booth_path(@conference.short_title, booth)
+              %td
+                = link_to booth.submitter.name, user_path(booth.submitter) if booth.submitter
+              %td
+                .responsibles
+                  - booth.responsibles.each do |responsible|
+                    = link_to responsible.name, admin_user_path(responsible)
```
differentreality: This is still linking to admin.

app/controllers/booths_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      redirect_to conference_booths_path,
+                  notice: 'Booth successfully created.'
+    else
+      flash[:error] = "Creating booth failed. #{@booth.errors.full_messages.to_sentence}."
+      render :new
+    end
+  end
+
+  def edit
+    url = conference_booth_path(@conference.short_title, @booth.id)
```
differentreality: url -> @url

differentreality: @booth.id can be just @booth (rails automatically assumes the .id part)

app/models/ability.rb
```diff
@@ -75,6 +75,8 @@ def signed_in(user)
     can [:new, :create], Payment, user_id: user.id
     can [:index, :show], PhysicalTicket, user: user
 
+    can [:new, :create, :update, :index, :show], Booth
```
differentreality: A user should be able to access the actions of the booths for which the user is a submitter or a responsible. We should limit that can definition accordingly.

app/views/booths/show.html.haml
```diff
@@ -0,0 +1,58 @@
+.container
+  .row
+    .col-md-12
+      %h3
+        = image_tag(@booth.picture.thumb.url, size: '20%', alt: '')
```
differentreality: Let's only show the logo if one actually exists.

app/views/booths/show.html.haml
```diff
@@ -0,0 +1,58 @@
...
+          %td.col-md-2
+            %b Website
+          %td
+            - if @booth.website_url.present?
+              = link_to @booth.website_url,  @booth.website_url
+        %tr
+          %td.col-md-2
+            %b Submitter
+          %td
+            = link_to @booth.submitter.name, admin_user_path(@booth.submitter)
```
differentreality: This is still in admin.

differentreality:
This starts to look very nice @nasia-sam :+1:

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,44 @@
...
+              %td
+                -if can? :edit, booth
+                  = link_to 'Edit', edit_conference_booth_path(@conference.short_title, booth.id),
+                    class: 'btn btn-primary'
```
differentreality: We should also allow the users to withdraw their booth requests.

differentreality: Let's try to keep the color scheme of proposals for the buttons.

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,44 @@
+.container
+  .row
+    .col-md-12.page-header
+      %h1
+        Your booth requests for
+        = @conference.title
+      .pull-right
+        - if can? :create, Booth
+          = link_to 'Add Booth', new_conference_booth_path(@conference.short_title), class: 'button btn btn-primary'
```
differentreality: Let's put this on the same line as the title

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,44 @@
...
+  .row
+    .col-md-12
+      .margin-booth-table
+        %table.table.table-striped.table-bordered.table-hover.datatable
```
differentreality: We don't expect the user to submit more than a couple of booth requests, so making this table a datatable might be unnecessary.

nasia-sam:
booth#index for simple user

20394827_1800972756587152_313513822_o

#edit and #show are the displayed the same way as #1566 .

Users who have made a booth request are able to:
- withdraw their request
- confirm it's acceptance
- re-submit it after it's been either rejected or withdrawn.

spec/factories/booth_request.rb
```diff
@@ -2,7 +2,7 @@
   factory :booth_request do
     booth
     user
-    role 'responsible'
+    role %w[submitter responsible]
```
differentreality: So now the role attribute of a record is an array instead of a string. Why? And are you using this factory anywhere?

app/views/booths/_form.html.haml
```diff
@@ -0,0 +1,34 @@
+.container
+  .row
+    .col-md-12
+      .page-header
+        %title Request a Booth
```
differentreality: Why do we need this?

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,50 @@
...
+              %td
+                = link_to booth.title, conference_booth_path(@conference.short_title, booth)
+              %td
+                = booth.state
+              %td
+                -if can? :edit, booth
+                  = link_to 'Edit', edit_conference_booth_path(@conference.short_title, booth.id),
+                    class: 'btn btn-primary'
+                - if booth.transition_possible? :withdraw
+                  = link_to 'Withdraw',
```
differentreality: Let's ask the user for confirmation before changing the state to withdraw.

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,50 @@
...
+      %h1
+        Your booth requests for
+        = @conference.title
+  .row
+    .col-md-12
+      .margin-booth-table
+        %table.table.table-striped.table-hover
+          %thead
+            %th
+              %b ID
```
differentreality: Can you explain your reasoning why you think users should see the ID of their booth request?

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,50 @@
...
+            %tr
+              %td
+                = booth.id
+              %td
+                - if booth.logo_link
+                  = image_tag(booth.picture.thumb.url, width: '20%')
+              %td
+                = link_to booth.title, conference_booth_path(@conference.short_title, booth)
+              %td
+                = booth.state
```
differentreality: It looks like you are allowing users to see if their booth is marked as to_accept, or to_reject. However these states were supposed to be for organizers only, and users will be notified when their booth would be marked as accepted, or rejected, isn't that so?

app/views/booths/index.html.haml
nasia-sam: I thought that it might be useful if they have plenty requests. I will remove it since this is a rather rare scenario.

nasia-sam:
Update: 

new_booth_index

app/views/booths/index.html.haml
```diff
@@ -0,0 +1,51 @@
...
+                %span{ title: show_state, class: "fa #{booth_status_icon(booth)}" }
+              %td
+                - if booth.logo_link
+                  = image_tag(booth.picture.thumb.url, width: '20%')
+              %td
+                = link_to booth.title, conference_booth_path(@conference.short_title, booth)
+              %td
+                -if can? :edit, booth
+                  = link_to 'Edit', edit_conference_booth_path(@conference.short_title, booth.id),
+                    class: 'btn btn-primary'
```
differentreality: There was also a second comment about the color scheme. Let's follow the colors used in proposals#index

differentreality:
About rubocop failure, you can add the file in ignore

nasia-sam:
@differentreality  you mean exclude it ?

differentreality:
Yes, so that there is no more failures. We need to make linters go green, as always!

nasia-sam:
@differentreality  Is the review completed? Should I squash so we can merge it?

differentreality:
You are making changes in db/schema.rb file but this PR contains no migrations that change the db schema. Please remove those changes, squash your commits, and rebase.

app/models/booth.rb
```diff
@@ -35,13 +35,17 @@ class Booth < ActiveRecord::Base
     state :to_reject
     state :rejected
     state :canceled
+    state :confirmed
 
     event :restart do
-      transitions to: :new, from: [:withdrawn, :to_accept, :to_reject, :canceled]
+      transitions to: :new, from: [:withdrawn, :rejected, :canceled, :confirmed]
```
differentreality: If a booth is confirmed, this allows me to click 'Re-submit', which doesn't make a lot of sense, does it?

app/models/booth.rb
```diff
@@ -35,13 +35,17 @@ class Booth < ActiveRecord::Base
...
     end
     event :withdraw do
       transitions to: :withdrawn, from: [:new, :to_accept, :accepted, :to_reject, :rejected]
```
differentreality: I should be able to withdraw my booth, after I have confirmed. Why aren't you following the same transitions we have for proposals?

differentreality:
@nasia-sam did you check my last comment? It's not fixed yet.

differentreality:
Nice :+1: 
There a few things in the code that can be re-written in a better way, please create cards in your github project to come back to.

PR #1610, opened 2017-07-23
Route to conference#show for custom domain

shlok007:
This PR consists of the following changes:
- a new column _custom_domain_ in _conferences_ table.
- route to point to the correct _splashpage_ of a conference by finding conference by its _custom_domain_.

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -2,7 +2,6 @@ class ConferencesController < ApplicationController
   protect_from_forgery with: :null_session
   before_action :respond_to_options
   load_and_authorize_resource find_by: :short_title
-  load_resource :program, through: :conference, singleton: true, except: :index
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
@@ -10,14 +9,15 @@ class ConferencesController < ApplicationController
   end
 
   def show
-    # have to change "localhost" to ENV['OSEM_HOSTNAME'] in production 
+    # have to change "localhost" to ENV['OSEM_HOSTNAME'] in production
     check_custom_domain if request.host != 'localhost'
+    @program = @conference.program
   end
 
   private
 
   def check_custom_domain
-    @conference = @conference.custom_domain.present? ? Conference.find_by(custom_domain: request.domain) : @conference
+    @conference = @conference.nil? ? Conference.find_by(custom_domain: request.domain) : @conference
   end
 
   def respond_to_options
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -9,9 +9,16 @@ class ConferencesController < ApplicationController
     @antiquated = @conferences - @current
   end
 
-  def show; end
+  def show
+    # have to change "localhost" to ENV['OSEM_HOSTNAME'] in production 
+    check_custom_domain if request.host != 'localhost'
+  end
 
   private
+
+  def check_custom_domain
+    @conference = @conference.custom_domain.present? ? Conference.find_by(custom_domain: request.domain) : @conference
+  end
 
   def respond_to_options
     respond_to do |format|
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -1,7 +1,7 @@
 class ConferencesController < ApplicationController
   protect_from_forgery with: :null_session
   before_action :respond_to_options
-  load_and_authorize_resource find_by: :short_title
+  load_and_authorize_resource find_by: :short_title, except: :show
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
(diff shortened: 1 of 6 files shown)


app/controllers/conferences_controller.rb
```diff
@@ -2,17 +2,24 @@ class ConferencesController < ApplicationController
...
-  def show; end
+  def show
+    # have to change "localhost" to ENV['OSEM_HOSTNAME'] in production
+    check_custom_domain if request.host != 'localhost'
+    @program = @conference.program
+  end
 
   private
 
+  def check_custom_domain
```
lagartoflojo: Maybe a more intentional name for this method would be `load_conference_by_domain` or something like that. We are not checking anything here, really.

app/controllers/conferences_controller.rb
```diff
@@ -2,17 +2,24 @@ class ConferencesController < ApplicationController
...
 
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
     @antiquated = @conferences - @current
   end
 
-  def show; end
+  def show
+    # have to change "localhost" to ENV['OSEM_HOSTNAME'] in production
+    check_custom_domain if request.host != 'localhost'
```
lagartoflojo: At the top we already do `load_and_authorize_resource find_by: :short_title`, but in case I'm accessing a conference directly through its domain (ie. `www.myconf.com`), then `params[:short_title]` will be `nil`, right? What will `load_and_authorize_resource` do then? Maybe we should call `load_and_authorize_resource` only if we have `params[:short_title]`?

lagartoflojo: Also: Maybe we can call `check_custom_domain` ONLY if `params[:short_title]` is not set?
That way, we don't need to do that `request.host != 'localhost'` check (some people might use 0.0.0.0:3000 for example).

app/controllers/conferences_controller.rb
```diff
@@ -2,17 +2,24 @@ class ConferencesController < ApplicationController
...
+    @program = @conference.program
+  end
 
   private
 
+  def check_custom_domain
+    @conference = @conference.nil? ? Conference.find_by(custom_domain: request.domain) : @conference
```
lagartoflojo: Would this work?
`@conference ||= Conference.find_by(custom_domain: request.domain)`

config/initializers/domain_constraint.rb
```diff
@@ -0,0 +1,6 @@
+class DomainConstraint
+  def self.matches?(request)
+  	@domains = Conference.pluck(:custom_domain).compact
+    @domains.include?(request.domain)
```
lagartoflojo: this `domains` NEED to be an instance variable?

config/initializers/domain_constraint.rb
```diff
@@ -0,0 +1,6 @@
+class DomainConstraint
+  def self.matches?(request)
+  	@domains = Conference.pluck(:custom_domain).compact
```
lagartoflojo: Wrong indentation here

lagartoflojo: Maybe we can let the database do the filtering:

```
Conference.where.not(custom_domain: nil).pluck(:custom_domain)
```

spec/controllers/conferences_controller_spec.rb
```diff
@@ -23,6 +23,15 @@
         get :show, id: conference.short_title
         expect(response).to render_template :show
       end
+
+      it 'assigns correct conference from a custom domain' do
```
lagartoflojo: This `it` block should be within a context:

`context "accessing via a custom domain"` or something like that.

Then the conference's attribute can be set with a let:
`let(:conference) { create(:conference, custom_domain: 'thebest.conf', splashpage: create(:splashpage, public: true), venue: create(:venue)) }`

And `@request` can be set in a `before` block:

`before { @request.host = 'thebest.conf' }`

So you can reduce your test to three lines :)

lagartoflojo:
Is this still [WIP]? What's left?

shlok007:
The PR introducing views for custom domain #1624 also has these commits . We can consider closing this and work on improving the changes in #1624 itself.

lagartoflojo:
I'd prefer to have smaller PRs, so let's finish this and then you can merge master into #1624

shlok007:
Just trying to find a workaround for the Hakiri failure because of the following line introduced: https://github.com/openSUSE/osem/pull/1610/files#diff-7fbe00a3a9ebba1c66f31e2057134f9aR13

lagartoflojo:
If you check the Hakiri error, it points to this line: https://github.com/openSUSE/osem/tree/ec435b65030e6dbd9a6f86baaa418dba87ae5a33/app/views/conferences/_venue.html.haml#L35

The problem is that `link_to @conference.venue.website, @conference.venue.website`, the URL is a user-provided value, so a user could provide a URL that looks like `javascript:alert("Hacked!")`.

This is indeed an XSS vulnerability and I think it wasn't caught before because `@conference` was being loaded automatically by CanCan, so Hakiri didn't see it (wild guess here).

Anyway, unless we use something like https://github.com/jarrett/sanitize-url, it's hard to get rid of this warning right now, and it wasn't introduced by your code anyway. I would ignore it for now.

shlok007:
>Anyway, unless we use something like https://github.com/jarrett/sanitize-url, it's hard to get rid of this warning right now, and it wasn't introduced by your code anyway. I would ignore it for now.

We can create a separate issue for the same. 
That was the only thing left so I have removed the [WIP] label from the title.

shlok007:
@lagartoflojo please do review

PR #1612, opened 2017-07-25
Removed delete action form ability.rb

### LINKED ISSUE #1608: Remove delete action from ability
There is no delete action for proposals, yet there is an ability for it, which should not be there.
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L84

divyanshumehta:
There was no delete action defined for proposals, but it was there in ability.rb.So the ability was deleted. Fixes #1608


divyanshumehta:
Ok on it.

differentreality:
While you are at it, why don't you fix it here too?
https://github.com/openSUSE/osem/blob/master/app/models/admin_ability.rb#L42

divyanshumehta:
Since I had lost my previous app structure and the local branches I moved this to a new pull request.

PR #1613, opened 2017-07-25
Scan QR code.

siddhantbajaj:
Added TicketScanning controller and routes to scan qr code on PhysicalTickets.

### DIFF
--- a/app/controllers/admin/ticket_scannings_controller.rb
+++ b/app/controllers/admin/ticket_scannings_controller.rb
@@ -1,0 +1,16 @@
+module Admin
+  class TicketScanningsController < Admin::BaseController
+    before_action :authenticate_user!
+    load_resource :physical_ticket, find_by: :token
+    # We authorize manually in these actions
+    skip_authorize_resource only: [:create]
+
+    def create
+      @ticket_scanning = TicketScanning.new(physical_ticket: @physical_ticket)
+      authorize! :create, @ticket_scanning
+      @ticket_scanning.save
+      redirect_to conferences_path,
+                  notice: "Ticket with token #{@physical_ticket.token} successfully scanned."
+    end
+  end
+end
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -145,6 +145,10 @@ class AdminAbility
     can :manage, Sponsor, conference_id: conf_ids
     can :manage, SponsorshipLevel, conference_id: conf_ids
     can :manage, Ticket, conference_id: conf_ids
+    can :create, TicketScanning do |ticket_scanning|
+      conf_id = ticket_scanning.physical_ticket.ticket_purchase.conference_id
+      conf_ids.include? conf_id
+    end
     can :index, Comment, commentable_type: 'Event',
                          commentable_id: Event.where(program_id: Program.where(conference_id: conf_ids).pluck(:id)).pluck(:id)
 
@@ -219,6 +223,10 @@ class AdminAbility
     can :manage, Question, conference_id: conf_ids_for_info_desk
     can :manage, Question do |question|
       !(question.conferences.pluck(:id) & conf_ids_for_info_desk).empty?
+    end
+    can :create, TicketScanning do |ticket_scanning|
+      conf_id = ticket_scanning.physical_ticket.ticket_purchase.conference_id
+      conf_ids_for_info_desk.include? conf_id
     end
 
     # Abilities for Role (Conference resource)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -26,6 +26,7 @@ Osem::Application.routes.draw do
         patch :toggle_confirmation
       end
     end
+    resource :ticket_scanning, only: [:create]
     resources :comments, only: [:index]
     resources :conferences do
       resource :contact, except: [:index, :new, :create, :show, :destroy]
--- a/spec/controllers/admin/ticket_scannings_controller_spec.rb
+++ b/spec/controllers/admin/ticket_scannings_controller_spec.rb
@@ -1,0 +1,60 @@
+require 'spec_helper'
+
+describe Admin::TicketScanningsController do
+  let(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let(:user) { create(:user) }
+  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user) }
+  let(:physical_ticket) { create(:physical_ticket, ticket_purchase: paid_ticket_purchase) }
+
+  context 'logged in as user with no role' do
+    before :each do
+      sign_in user
+    end
+    describe 'POST #create' do
+      it 'does not create new ticket scanning' do
+        expected = expect do
+          post :create, physical_ticket_id: physical_ticket.token
+        end
+        expected.to_not change(TicketScanning, :count)
+      end
+
+      it 'redirects to root' do
+        post :create, physical_ticket_id: physical_ticket.token
+        expect(flash[:alert]).to eq('You are not authorized to access this page.')
+        expect(response).to redirect_to(root_path)
+      end
+    end
+  end
+
+  context 'logged in as admin' do
+    before :each do
+      sign_in admin
+    end
+    describe 'POST #create' do
+      context 'with valid physical_ticket' do
+        it 'creates new ticket scanning' do
+          expected = expect do
+            post :create, physical_ticket_id: physical_ticket.token
+          end
+          expected.to change { TicketScanning.count }.by(1)
+        end
+
+        it 'redirects to index' do
+          post :create, physical_ticket_id: physical_ticket.token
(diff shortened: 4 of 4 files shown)


app/controllers/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def create
+    @physical_ticket = PhysicalTicket.find_by_token(params[:token])
+
+    if @physical_ticket.present?
+      @ticket_scanning = TicketScanning.create(physical_ticket: @physical_ticket)
+      redirect_to conferences_path,
+                  notice: 'Thanks! Your ticket is scanned successfully.'
+    else
+      redirect_to conferences_path,
```
Ana06: why splitting the notice here? :thinking:

app/controllers/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  before_action :authenticate_user!
+  load_and_authorize_resource
+
+  def create
+    @physical_ticket = PhysicalTicket.find_by_token(params[:token])
+
+    if @physical_ticket.present?
+      @ticket_scanning = TicketScanning.create(physical_ticket: @physical_ticket)
+      redirect_to conferences_path,
+                  notice: 'Thanks! Your ticket is scanned successfully.'
```
Ana06: I would remove the `thanks` as QR would read by admins. Also, it is not `your` ticket` but `the` ticket. And would be nice to add the token in the message.

app/controllers/ticket_scannings_controller.rb
Ana06: and in this case it should be an error instead of a notice, right? :thinking:

app/controllers/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,17 @@
+class TicketScanningsController < ApplicationController
+  before_action :authenticate_user!
+  load_and_authorize_resource
+
+  def create
+    @physical_ticket = PhysicalTicket.find_by_token(params[:token])
```
Ana06: what about writing it in a `load_resource`??
``` Ruby
 load_resource :physical_ticket, find_by: :token
```

app/controllers/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,17 @@
+class TicketScanningsController < ApplicationController
+  before_action :authenticate_user!
+  load_and_authorize_resource
```
Ana06: why do we need the `load` here? we only have a `create`

app/controllers/ticket_scannings_controller.rb
siddhantbajaj: @Ana06 does load resource work for create action? 😕

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Admin
+  class TicketScanningsController < ApplicationController
+    before_action :authenticate_user!
+    authorize_resource
+
+    def create
+      @physical_ticket = PhysicalTicket.find_by_token(params[:id])
```
siddhantbajaj: Load resource is not working here. I am not sure if load_resource work for create action.

app/models/ability.rb
```diff
@@ -94,6 +94,7 @@ def signed_in(user)
   # Abilities for users with roles wandering around in non-admin views.
   def common_abilities_for_admins(user)
     can :access, Admin
+    can :create, TicketScanning
```
Ana06: this should be moved to like 105 and 125 (to allow only organizers and info desk people to register people) and using the id of the conference :wink:

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Admin
+  class TicketScanningsController < Admin::BaseController
+    before_action :authenticate_user!
+    before_action :find_physical_ticket_by_token
```
siddhantbajaj: I am not using `load_resource :ticket_scanning, find_by: :token` because it raises an record not found exception if the resource is not found as it is implemented by using `find_by!`.

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    load_and_authorize_resource through: :physical_ticket, singleton: true, shallow: true
```
siddhantbajaj: I am loading resource as well here though it is not needed for the create action but it is needed in defining it's abilities.

app/models/admin_ability.rb
```diff
@@ -145,6 +145,11 @@ def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
     can :manage, Sponsor, conference_id: conf_ids
     can :manage, SponsorshipLevel, conference_id: conf_ids
     can :manage, Ticket, conference_id: conf_ids
+    can :create, TicketScanning do |ticket_scanning|
+      physical_ticket = ticket_scanning.physical_ticket
```
Ana06: this have to written like that because conference id is not a field of the database in this case, right? :thinking: 
what about writing it like:

``` Ruby
conf_id = ticket_scanning.physical_ticket.ticket_purchase.conference_id
conf_ids.include? conf_id
```
I don't think that having the two extra variable are really needed :thinking:

app/controllers/admin/ticket_scannings_controller.rb
siddhantbajaj: we need a ticket scanning object with physical ticket set before the create action is actually called so that we can define the ability for create action. With just authorize resource it won't be possible.

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def create
+      if @physical_ticket.present?
+        @ticket_scanning = TicketScanning.create(physical_ticket: @physical_ticket)
+        redirect_to conferences_path,
+                    notice: "Ticket with token #{@physical_ticket.token} is scanned successfully."
```
siddhantbajaj: 👍

app/models/admin_ability.rb
siddhantbajaj: Yes you are right, I'll change this. 👍

laosb:
Using ticket ids (numeric icremental IDs) is not secure enough. They can guess an ID and make a fake QR code to attend.

Ana06:
@laosb 

> Using ticket ids (numeric icremental IDs) is not secure enough. They can guess an ID and make a fake QR code to attend.
>
> Generate a hash for each ticket and use this to check in, this will prevent those who try to guess a number.

As you can see [here](https://github.com/openSUSE/osem/blob/e2bad46d0011dccd11eb3755e6c71690fe506ea4/app/models/physical_ticket.rb#L18) @siddhantbajaj is using a token that is not the same as the `id`, just because of this reason. So what you say is already done, or did I understand you wrong?

app/controllers/admin/ticket_scannings_controller.rb
siddhantbajaj: Yes I thought about this but where should I catch the exception. After loading physical_ticket, ticket_scanning object should be loaded and authorized so I don't know where should I catch the exception. 🤔

Ana06: and I am even thinking that this shouldn't happen right? So, this would be really unexpected, so maybe even correct to raise the exception.

app/controllers/admin/ticket_scannings_controller.rb
Ana06: then may be better to authorize it inside the create action, in a similar way as this is done with events: https://github.com/openSUSE/osem/blob/master/app/controllers/proposals_controller.rb#L38

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -1,21 +1,16 @@
 module Admin
   class TicketScanningsController < Admin::BaseController
     before_action :authenticate_user!
-    before_action :find_physical_ticket_by_token
-    load_and_authorize_resource through: :physical_ticket, singleton: true, shallow: true
+    load_resource :physical_ticket, find_by: :token
+    # We authorize manually in these actions
+    skip_authorize_resource only: [:create]
```
Ana06: why is this needed? You are not authorizing the resource? I think you can remove lines 4 and 5 and everything will keep the same. :thinking:

siddhantbajaj: @Ana06 
In line 4 we are loading physical_ticket resource that is needed in our create action so I don't think we can remove that line. line 6 is necessary because of this: https://github.com/openSUSE/osem/blob/master/app/controllers/application_controller.rb#L9
We need to have authorization in every controller and if not we must skip authorization. Here we are manually authorizing therefore we are skipping authorization before method.

Ana06: I meant 5 and 6, but you are right anyway

PR #1617, opened 2017-07-28
Add booth limit 

nasia-sam:
Add column booth_limit to conference so admin won't be able to accept more booth requests than a conference can support. Method `maximum_accepted_booths?` tests whether the accepted booths have reached the booth limit or not.

### DIFF
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -47,18 +47,21 @@ module Admin
     end
 
     def accept
-      @booth.accept!
 
-      if @booth.save
-        if @conference.email_settings.send_on_booths_acceptance
-          Mailbot.conference_booths_acceptance_mail(@booth).deliver
+      if can? :accept, @booth
+        @booth.accept!
+
+        if @booth.save
+          if @conference.email_settings.send_on_booths_acceptance
+            Mailbot.conference_booths_acceptance_mail(@booth).deliver
+          end
+          redirect_to admin_conference_booths_path(conference_id: @conference.short_title),
+                      notice: 'Booth successfully accepted!'
+        else
+          redirect_to admin_conference_booths_path(conference_id: @conference.short_title)
+          flash[:error] = "Booth could not be accepted. #{@booth.errors.full_messages.to_sentence}."
         end
-        redirect_to admin_conference_booths_path(conference_id: @conference.short_title),
-                    notice: 'Booth successfully accepted!'
-      else
-        redirect_to admin_conference_booths_path(conference_id: @conference.short_title)
-        flash[:error] = "Booth could not be accepted. #{@booth.errors.full_messages.to_sentence}."
-      end
+      end  
     end
 
     def to_accept
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -47,21 +47,18 @@ module Admin
     end
 
     def accept
+      @booth.accept!
 
-      if can? :accept, @booth
-        @booth.accept!
-
-        if @booth.save
-          if @conference.email_settings.send_on_booths_acceptance
-            Mailbot.conference_booths_acceptance_mail(@booth).deliver
-          end
-          redirect_to admin_conference_booths_path(conference_id: @conference.short_title),
-                      notice: 'Booth successfully accepted!'
-        else
-          redirect_to admin_conference_booths_path(conference_id: @conference.short_title)
-          flash[:error] = "Booth could not be accepted. #{@booth.errors.full_messages.to_sentence}."
+      if @booth.save
+        if @conference.email_settings.send_on_booths_acceptance
+          Mailbot.conference_booths_acceptance_mail(@booth).deliver
         end
-      end  
+        redirect_to admin_conference_booths_path(conference_id: @conference.short_title),
+                    notice: 'Booth successfully accepted!'
+      else
+        redirect_to admin_conference_booths_path(conference_id: @conference.short_title)
+        flash[:error] = "Booth could not be accepted. #{@booth.errors.full_messages.to_sentence}."
+      end
     end
 
     def to_accept
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -74,6 +74,11 @@ class AdminAbility
     cannot :destroy, Track do |track|
       track.self_organized?
     end
+    # Can't accept a booth when booth_limit is reached
+    cannot :accept, Booth do |booth|
+      conference = booth.conference
+      conference.maximum_accepted_booths?
+    end
   end
 
   # Abilities for signed in users with roles
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -739,6 +739,15 @@ class Conference < ActiveRecord::Base
   end
 
   ##
+  #
+  # ====Returns
+  # * +True+ -> if accepted booths are equal to the booth limit
+  # * +False+ -> Accepted booths have not reached the booth limit
+  def maximum_accepted_booths?
+    booth_limit > 0 && booths.accepted.count + booths.confirmed.count >= booth_limit
+  end
+
+  ##
   # Return the current conference object to be used in RevisionCount
   #
   # ====Returns
--- a/app/views/admin/booths/_change_state_dropdown.html.haml
+++ b/app/views/admin/booths/_change_state_dropdown.html.haml
@@ -1,13 +1,17 @@
+
 - if booth.transition_possible? :accept
-  - if can? :accept, @booth
+  - if can? :accept, booth
+    - if @conference.booth_limit > 0
+      - confirm_message = ' You are able to accept '+ pluralize(@conference.booth_limit - (@conference.booths.accepted.count + @conference.booths.confirmed.count), 'more booth') + " (booth limit set to #{@conference.booth_limit}). Are you sure you want to accept this one?"
     - if @conference.email_settings.send_on_booths_acceptance
       - link = 'Accept with email'
+      - confirm_message = "By accepting this booth, an email will be sent informing the submitter for the acceptance. You may change the state to \'To accept\' until you are completely sure." + confirm_message
     - else
       - link = 'Accept booth'
     %li= link_to link,
     accept_admin_conference_booth_path(@conference.short_title, booth),
     method: :patch ,id: "accept_booth_#{booth.id}",
-    data: (@conference.booth_limit > 0 ? { confirm: 'You are able to accept '+ pluralize(@conference.booth_limit - @conference.booths.accepted.count, 'more booth') + " (booth limit set to #{@conference.booth_limit}). Are you sure you want to accept this one?" } : nil )
+    data: (confirm_message ? { confirm: confirm_message } : nil)
 
 - if booth.transition_possible? :reject
   - if @conference.email_settings.send_on_booths_rejection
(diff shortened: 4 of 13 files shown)


nasia-sam:
* admin/conference#edit
20526493_1808079169209844_587037611_n

* admin booths#index
When booth limit hasn't be set (equals to 0) (link redirects to edit_admin_conference_path)

20526643_1808079195876508_1494108204_n

When booth limit is set

20524324_1808079165876511_1791173859_o

When accepted booths are equal to the booth limit and admin tries to exceed it

20542893_1808079199209841_2117141045_o

app/views/admin/booths/index.html.haml
```diff
@@ -9,6 +9,22 @@
             = link_to 'Add Booth', new_admin_conference_booth_path(@conference.short_title), class: 'button btn btn-primary'
       %p.text-muted
         All the booth requests
+
+      - if @conference.booth_limit == 0
+        %p
+          Set the
+          = link_to 'Booth limit', edit_admin_conference_path(@conference.short_title)
+          to make sure you are not accepting more booth requests than possible.
```
differentreality: what about this wording "to make sure you are not accepting more booths than you can accommodate" ?

differentreality:
If I cannot accept the booth, why show the option as active in the drop down menu in the first place?

We should also think about the functionality a bit.  

What if I start accepting booths, and suddenly I am no longer able to do so (because of the limit)? 
What am I supposed to do then? 
- If I am supposed to just go and increase the limit, then this basically is not offering us any functionality. 
- If I need to go reject accepted booths, that's not good either.

So now that we have added the existence of booth_limit we need to make good use of it, in a way that it is helpful to the organizer and adds value to our app (If I still need to manually count the booths I want to accept, for instance, then the feature is not adding any extra value).

app/views/admin/conferences/edit.html.haml
```diff
@@ -26,4 +26,7 @@
         = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
       = f.inputs name: 'Registrations' do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events (they will still be able to register even if it has been reached). You currently have ' + pluralize(@conference.registrations.count, 'registration')
+      = f.inputs name: 'Booths' do
+        = f.input :booth_limit, as: :number, in: 0..9999,
+        hint: 'Booth limit is the maximum number of booth requets that you can accept for this conference. By setting this number (0 no limit) you can be sure that you are not going to accept more requests than the number of which your conference can support. You currently have ' + pluralize(@conference.booths.where(state: 'accepted').count, 'accepted booth') +'.'
```
differentreality: * of booths that you can accept..
* ...that you are not going to accept more booths than the conference can accommodate...

app/views/admin/booths/index.html.haml
```diff
@@ -9,6 +9,22 @@
...
+      - elsif !maximum_accepted_booths?
+        %p
+          You cannot accept more than
+          %b
+            = @conference.booth_limit
+          booth requests. (
+          = pluralize(@conference.booths.where(state: 'accepted').count, 'accepted booth')
```
differentreality: This could be moved into a scope so that we can simply do @conference.booths.accepted

app/views/admin/booths/index.html.haml
```diff
@@ -9,6 +9,22 @@
...
+          so far)
+      - else
+        %p You have reached the maximum number of accepted booths.
```
differentreality: What about this also shows the booth_limit number and has a link to the booth_limit config?

app/controllers/admin/booths_controller.rb
```diff
@@ -49,7 +51,12 @@ def destroy
     end
 
     def accept
-      update_state(:accept, 'Booth accepted!')
+      if !maximum_accepted_booths?
```
differentreality: Instead of using negation in your if statement, why not switch the contents of if and else?
So that this reads
```
if maximum_accepted_booths?
  redirect ...
else
  update_state()
end
```

app/controllers/admin/booths_controller.rb
```diff
@@ -72,6 +79,15 @@ def cancel
       update_state(:cancel, 'Booth is canceled')
     end
 
+    ##
+    #
+    # ====Returns
+    # * +True+ -> if accepted booths are equal to the booth limit
+    # * +False+ -> Accepted booths have not reached the booth limit
+    def maximum_accepted_booths?
```
differentreality: Why isn't this a method in the conference model?

nasia-sam:
@differentreality in booths#index I have added a message that informs the organizer about the accepted booths and how many more booths they can accept. However, if you think this is not enough, what if we added confirmation for the accept action where organizer is informed of the remaining booths that are possible to be accepted. I can't think of something else more straightforward. Should I go with that or do you have any other idea ?

differentreality:
Please spend some time to think it over, and document your ideas to improve this.

app/models/conference.rb
```diff
@@ -739,6 +739,15 @@ def hours_from_start_time(start_hour, end_hour)
   end
 
   ##
+  #
+  # ====Returns
+  # * +True+ -> if accepted booths are equal to the booth limit
+  # * +False+ -> Accepted booths have not reached the booth limit
+  def maximum_accepted_booths?
+    booth_limit > 0 && booths.accepted.count + booths.where(state: 'confirmed').count >= booth_limit
```
differentreality: Why aren't you using the confirmed scope here?!

nasia-sam:
* Disabled accept action when the accepted booths equal the booth limit
* Add confirmation for the accept action when booth limit is set 

20668505_1817813234903104_1279088591_n

* Make messages at booths#index that inform the organizer about the booth limit larger so they attract organizer's attention

differentreality:
@nasia-sam please rebase, so that I can review!

app/views/admin/booths/index.html.haml
```diff
@@ -9,8 +9,30 @@
             = link_to 'Add Booth', new_admin_conference_booth_path(@conference.short_title), class: 'button btn btn-primary'
       %p.text-muted
         All the booth requests
+
```
differentreality: One empty line should be enough :)

app/views/admin/booths/_change_state_dropdown.html.haml
```diff
@@ -1,7 +1,9 @@
 - if booth.transition_possible? :accept
-  %li= link_to 'Accept booth',
-  accept_admin_conference_booth_path(@conference.short_title, booth),
-  method: :patch, id: "accept_booth_#{booth.id}"
+  - unless @conference.maximum_accepted_booths?
```
differentreality: This should be added in admin_ability.rb, and then here you can simply do `if can? :accept, booth`

app/views/admin/booths/index.html.haml
```diff
@@ -9,8 +9,30 @@
...
+          (
+          = pluralize(@conference.booths.accepted.count + @conference.booths.confirmed.count, 'accepted booth')
+          so far)
+      - else
+        %p
+          You have reached the maximum number of accepted booths.
+          (
+          = link_to "#{@conference.booth_limit} booths", edit_admin_conference_path(@conference.short_title)
+          )
     .margin-booth-table
```
differentreality: What's that class doing?

app/controllers/admin/booths_controller.rb
```diff
@@ -47,7 +47,9 @@ def update
     end
 
     def accept
-      @booth.accept!
+      if can? :accept, @booth
```
differentreality: Why do you need to check this?

app/models/admin_ability.rb
```diff
@@ -120,6 +125,10 @@ def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
     can :manage, Registration, conference_id: conf_ids
     can :manage, RegistrationPeriod, conference_id: conf_ids
     can :manage, Booth, conference_id: conf_ids
+    cannot :accept, Booth do |booth|
```
differentreality: Why do we need to repeat this block in all three places?

nasia-sam: All the three of them have the `can :manage, Booth` ability

differentreality: common_abilities_for_roles (https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L20) doesn't have the can :manage, Booth definition, yet it has the cannot :accept, Booth definition (https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L78)

More importantly the can :manage, Booth definition depend on other parameters, which are unique to the individual function they appear in. For example, when you are in the organizer role (https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L109) your definition is:
`can :manage, Booth, conference_id: conf_ids`
https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L127

But when you are in the cpf role (https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L178) your definition is:
` can :manage, Booth, conference_id: conf_ids_for_cfp`
https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L186

So in one case we have the conf_ids_for_cfp and in the other we have conf_ids, each of those two variables hold different values, as such we need to define the can :manage, Booth ability in each function (once for the organizer role, one for the cfp role).

However in all places your definition reads exactly the same, and it does not take into account the role of the user.
```
cannot :accept, Booth do |booth|
      conference = booth.conference
      conference.maximum_accepted_booths?
    end
```

So my question remains:
- Why do we  need to define the exact same block 3 times?
- And based on your previous answer, why is it even defined in common_abilities where there is no can :manage, Booth definition at all?

nasia-sam: I see my mistake. I was testing it with a non-admin user and when the block was written in `signed_in_with_organizer_role`, the   `can :manage, Booth, conference_id: conf_ids_for_cfp` seemed to overwrite it (even if the block I wrote doesn't take into account of the user's role).

I defined the block in the common_abilities section because, although there is no `can :manage, Booth`, there is `can :manage, :all if user.is_admin`. (https://github.com/nasia-sam/osem/blob/cab940207e346bbf1b957abdfad7217854f25f2c/app/models/admin_ability.rb#L54)
Only by putting the block there I could prevent an admin of accepting a booth when `booth_limit ` is reached. 

Taking that into consideration, I will leave the definition of the block in `common_abilities_for_roles` as it seems to be the only section that prevents admins, organizers, _and_ cfps from accepting a booth when they shouldn't.

differentreality: That's right. And moreover it is about the order that things are defined. Even if there was no admin and no admin abilities, common_abilities call the other functions for the abilities for organizer, cfp role etc. If you wanted something (the same thing) to apply for all roles, you could put it in one place, as long as it was **after** their abilities definition, ie at the end of common_abilities.

nasia-sam: I got it now- thanks for the clarifications:)

PR #1619, opened 2017-08-02
Expose enable_starttls_auto & openssl_verify_mode.

laosb:
And align `ENV` calls to keep up with codes above.


differentreality:
We should also be defining the variables,  so let's add them in https://github.com/openSUSE/osem/blob/master/dotenv.example

differentreality:
We should also be defining the variables,  so let's add them in https://github.com/openSUSE/osem/blob/master/dotenv.example

laosb:
@differentreality Thank you for your reminder. I've added it in my last commit.

differentreality:
@laosb Nice :+1:  please squash that middle commit, so that we can merge this.

laosb:
@differentreality Thank you for your review, and I've squashed these commits!

Issue #1609, opened 2017-07-22
Only ask speakers to register, if registrations are possible for the conference


differentreality:
https://github.com/openSUSE/osem/blob/master/app/views/proposals/index.html.haml#L54 

Only show the relevant text, if there is a registration period set for the conference, otherwise it will be confusing.

Same for the tooltip link that encourages speakers to register for the conference, this only makes sense if the conference has a registration_period set.

namangupta01:
i am working on this issue

namangupta01:
Fixed in #1621 
Please close this issue!

PR #1621, opened 2017-08-03
Speaker is asked to register only when registration_period is set

namangupta01:
#1609 Speaker is only asked to register when registration_period is set.

### DIFF
--- a/app/views/proposals/_tooltip.html.haml
+++ b/app/views/proposals/_tooltip.html.haml
@@ -1,6 +1,6 @@
 - progress_status = event.progress_status
 %ul.list-unstyled
-  - if @conference.registration_open? || (@conference.user_registered? current_user)
+  - if can? :create, @conference.registrations.new
     %li{'class'=>class_for_todo(progress_status['registered'])}
       %span{'class'=>icon_for_todo(progress_status['registered'])}
         - if progress_status['registered']
--- a/app/views/proposals/_tooltip.html.haml
+++ b/app/views/proposals/_tooltip.html.haml
@@ -1,11 +1,12 @@
 - progress_status = event.progress_status
 %ul.list-unstyled
-  %li{'class'=>class_for_todo(progress_status['registered'])}
-    %span{'class'=>icon_for_todo(progress_status['registered'])}
-      - if progress_status['registered']
-        Speaker(s) registered to the conference
-      - else
-        = link_to 'Speaker(s) not registered to the conference', new_conference_conference_registration_path(event.program.conference.short_title)
+  - if @conference.registration_open? || (@conference.user_registered? current_user)
+    %li{'class'=>class_for_todo(progress_status['registered'])}
+      %span{'class'=>icon_for_todo(progress_status['registered'])}
+        - if progress_status['registered']
+          Speaker(s) registered to the conference
+        - else
+          = link_to 'Speaker(s) not registered to the conference', new_conference_conference_registration_path(event.program.conference.short_title)
   %li{'class'=>class_for_todo(progress_status['biographies'])}
     %span{'class'=>icon_for_todo(progress_status['biographies'])}
       - if progress_status['biographies']
--- a/app/views/proposals/index.html.haml
+++ b/app/views/proposals/index.html.haml
@@ -48,7 +48,7 @@
             The more information you add to your proposal, the more likely it is that the conference organizers accept your proposal.
             %br
             It will also be more likely that visitors find your proposal interesting enough to attend.
-          - if (@conference.registration_open?) && !(@conference.user_registered? current_user)
+          - if can? :create, @conference.registrations.new
             %p
               %strong
                 Why do I need to register to the conference?
@@ -74,7 +74,7 @@
 
               %td.col-md-2{style: "padding:20px 8px 20px 8px;"}
                 = link_to 'Complete your proposal', 'javascript: void(0)', "type"=>"button", "data-trigger"=>"focus", "data-toggle"=>"popover", "title"=>"Your todo list", "data-content"=>"#{render partial: 'tooltip', locals: { event: event} }"
-                - if (@conference.registration_open?) || (@conference.user_registered? current_user) 
+                - if can? :create, @conference.registrations.new
                   - progress_percentage = event.calculate_progress
                   .progress
                     %div{class: "progress-bar #{event_progress_color(progress_percentage)}", style: "width:#{progress_percentage}%;"}
--- a/app/views/proposals/index.html.haml
+++ b/app/views/proposals/index.html.haml
@@ -48,13 +48,12 @@
             The more information you add to your proposal, the more likely it is that the conference organizers accept your proposal.
             %br
             It will also be more likely that visitors find your proposal interesting enough to attend.
-
-        - if Conference.where(short_title: params[:conference_id]).first.registration_period
-          %p
-            %strong
-              Why do I need to register to the conference?
-          %p
-            Knowing the number of visitors for the conference helps the organizers plan better.
+          - if (@conference.registration_open?) && !(@conference.user_registered? current_user)
+            %p
+              %strong
+                Why do I need to register to the conference?
+            %p
+              Knowing the number of visitors for the conference helps the organizers plan better.
 
         %table.table.table-striped#events
           - @events.each do |event|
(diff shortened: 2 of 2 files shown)


differentreality:
Could you please add screenshots of how this looks with and without your change?

namangupta01:
# When the registration_period is not set:
### Before changes :
screenshot from 2017-08-04 23-07-16

### After Changes:
screenshot from 2017-08-04 23-04-36

# When registration_period is set
It is same before and after changes
screenshot from 2017-08-04 23-08-42

differentreality:
But this way, when registrations are closed, you are removing the whole pop up, which includes many other things except for registration. Why do you think this is the way to go? Can you please explain your thinking?

namangupta01:
I was thinking that registrations are necessary for the speakers to attend the conference.

namangupta01:
Registration_open? | User_registered? | Show
---------------------------|-------------------------|--------
Yes | Yes | Yes
Yes | No | Yes
No | No | No
No | Yes | Yes  (Show it through strike)

differentreality:
Yes, you are absolutely right, registration is necessary so that we know who attends the conference.

Your code, however, is removing the whole link 'Complete your proposal' which contains more information, not just the speaker registration. My question is why do we need to remove the whole thing?

namangupta01:
Yes, you are right, whole link 'Complete your proposal' should not be removed. Only speaker registration link should be removed. So for this i have updated my code.
Here are some screenshot of how does it looks like after my change.
## When Registrations are not open and Speaker have not registered
when registration are closed and speaker have not registered


## When Registrations are closed but Speaker have registered
when registration are open and user already registered


## When Registrations are open and Speaker haven't registered
when registration are open and speaker havenot register

## When Registrations are open and Speaker have registered
when registration are open and user already registered

namangupta01:
Isn't it correct? Please Explain?

differentreality:
Yeap, that's good @namangupta01 :+1:

app/views/proposals/_tooltip.html.haml
```diff
@@ -1,11 +1,12 @@
 - progress_status = event.progress_status
 %ul.list-unstyled
-  %li{'class'=>class_for_todo(progress_status['registered'])}
-    %span{'class'=>icon_for_todo(progress_status['registered'])}
-      - if progress_status['registered']
-        Speaker(s) registered to the conference
-      - else
-        = link_to 'Speaker(s) not registered to the conference', new_conference_conference_registration_path(event.program.conference.short_title)
+  - if @conference.registration_open? || (@conference.user_registered? current_user)
```
differentreality: Instead of only checking those 2 conditions here, it's best that we check if the user can create a new registration. 

**How do we know if the user is allowed create a new registration?**
We are using [CanCanCan](https://github.com/CanCanCommunity/cancancan/wiki) to define what a user can or cannot do (which action of which model the user can access). So let's have CanCanCan do its job - ability.rb file contains the necessary definitions of what a user is allowed, or not, to do.

**Why we should do that?**
Because instead of checking against specific conditions, we can check what we really care about, which is whether or not the user is allowed to perform a certain action. 

Your code is  now only checking for 2 conditions

 - if the registration is open and
 - if the user is registered

There could be more conditions in the future (in fact there already are https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L69, but let's not focus on that right now). 

Checking if the user can execute an action will always work properly, without requiring any modifications to this code here, even if there are more conditions than those two.

namangupta01: https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L69 is used to check if user can register or not, but it will also return true in some cases even when speaker have already registered which is wrong. The code according to me should be :
 `(conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user))&&!(conference.user_registered? user)`
Shouldn't it be?
And we cannot use CanCanCan in https://github.com/namangupta01/osem/blob/master/app/views/proposals/_tooltip.html.haml#L3 and https://github.com/namangupta01/osem/blob/master/app/views/proposals/index.html.haml#L77 because we are not checking if a user can register here or not here.
We can only use CanCanCan in https://github.com/namangupta01/osem/blob/master/app/views/proposals/index.html.haml#L51 because we are checking here if a user can make a registration or not.
Please, let me know if i am wrong?

differentreality: Yes, that's right about CanCanCan :+1:  
A user can only register once for every conference.

Feel free to open an issue for this, and work on it at a different PR if you want to.

differentreality: We aren't checking if a user can register indeed. My point was that we should be checking that.

https://github.com/namangupta01/osem/blob/master/app/views/proposals/_tooltip.html.haml#L9 will redirect you to create a new registration, so the user needs to have the ability to create a new registration, otherwise it doesn't make sense to redirect users to a link they cannot access. Perhaps we should take that into consideration. What do you think?

https://github.com/namangupta01/osem/blob/master/app/views/proposals/index.html.haml#L77 needs to comply with the same conditions, so that we don't show that there are 3 missing items, but only show 2 missing items in the tooltip.

namangupta01: I was thinking that there are four cases for this line https://github.com/namangupta01/osem/blob/master/app/views/proposals/_tooltip.html.haml#L9 
These are :

1. If registrations are not open and speaker haven't registered, then there should not be any link for registration in complete your proposal box. And also, we should not count the registration  **process in progress bar**. which is what i am trying to do in else block https://github.com/namangupta01/osem/blob/master/app/views/proposals/index.html.haml#L83

2.If registrations are not open and speaker have registered, then we can show them that they have already registered (which is what i am thinking of doing, which leads to not using of "user can register" statement )

3.If registrations are open and speaker haven't registered, then it should show speaker to do registration.

4.If registration are open and speaker have registered, then it should show that user have registered.

Note : When User have registered for the conference, the registration link would turn into green and also the registration link is **no more a link, its just a text showing them** that they have successfully registered.

If we follow above four statements, then we cannot use the ability of user to create new registration.
What do you think of this?

differentreality: @namangupta01 

> If we follow above four statements, then we cannot use the ability of user to create new registration.

Why not?

namangupta01: @differentreality I misunderstood something. I am correcting this. Thanks!

namangupta01: @differentreality This issue is dependent on the issue #1628 , which fixes the ability of user to create registration. Please have a look at the corresponding PR.

differentreality: Regardless of the other PR, do you understand my point about checking if user can create a new registration?

namangupta01: Yes i understood that we have to show the link and encouragement text to a user only when a user can register, if a user is not allowed to do registration or the user is already registered then we doesn't have to show them the encouragement text and tooltip link and also doesn't count the registration in progress bar as well. Is this the correct way i am thinking @differentreality ?

differentreality: Yes, exactly. 

So instead of the if condition you are using on line 3, you should check for the user's ability to register.

namangupta01: okay :+1: Should i correct my pr or wait for issue #1628 because without this my changes will not show correct output.

differentreality: You can work on the changes, as we wait for #1628 to be merged.

namangupta01: sure :+1:

differentreality: Ping

namangupta01: @differentreality respective changes are done! Have a look!

bear454:
@namangupta01 please address the linting issues https://travis-ci.org/openSUSE/osem/jobs/293980697

namangupta01:
@bear454 It's done! Please have a look!

app/views/proposals/index.html.haml
```diff
@@ -67,20 +67,27 @@
                 %small.text-muted
                   = event.event_type.title
                   = "(#{event.event_type.length} min)"
-                  = "in #{event.track.name}" if event.track
+                  = "in #{event.traistck.name}" if event.track
```
differentreality: traistck ?

namangupta01: Hey. 
Oops, It's a typos!

namangupta01: @differentreality Should i make a pr for correcting this ?

PR #1622, opened 2017-08-04
Fix the dropdown to get enough bottom padding.

laosb:
I don't know why we omitted bottom padding. This change will make the ugly dropdown a little bit neater. Sorry if you are doing this by purpose.

Before | After
-------|------
image | image

*(Screenshots may look different in many other styles since they were took in our fork, instead of this original OSEM)*


differentreality:
Thanks @laosb :+1:

Issue #1599, opened 2017-07-18
Support of Custom Domains for conferences


shlok007:
The idea is to allow an organizer of a conference to add his own domain name to the conference.

- The custom domain will point to `conference#show` or `splashpage` of a particular conference.
![screenshot_20170718_070958](https://user-images.githubusercontent.com/14155445/28296823-2164d68a-6b88-11e7-9bda-9c3c7215e0e1.png)
  
Summary of changes to be required when using a custom domain :
- Hosted version name to be changed to the name of the conference's organization.
- Find conference by the domain name and not by short title.
For example:
`osem.io/conferences/o3DW/schedule` will become `o3DW.io/schedule`
`osem.io/conferences/o3DW/register/new` will become `o3DW.io/register/new`
- Redirect user to the hosted version tenant for the routes with admin namespace
- Maintain user session in between both the tenants.
  
### Future Improvements:
- Allow sending emails from the new custom domain.

lagartoflojo:
> Find conference by the domain name and not by short title

Can the domain name be a domain with sub-domain? What about redirecting www.myconference.com to myconference.com or vice-versa? 

> Redirect user to the hosted version tenant for the routes with admin namespace

What do you mean by that?

shlok007:
>What about redirecting www.myconference.com to myconference.com or vice-versa?

Thanks, If the domain name added is `osem.com` then we can redirect `www.osem.com` to the former and vice versa. 
>Can the domain name be a domain with sub-domain?

Currently we can focus on adding `apex domains` or just `www subdomains` by adding CNAME record with their domain registrar and providing documentation for the same. Maybe work with extending support for subdomains later when this gets ready and working. Does this sound possible?

>Redirect user to the hosted version tenant for the routes with admin namespace

Since, we will be sharing the user session in both the hosted version ( `hosted-version.com` ) and the custom domain ( `myconference.com` ) I thought we can keep the custom domain for only those views that can be accessed by users with no role in the non-admin namespace for that particular conference. Any action that routes to the admin namespace `myconference.com/admin/` can be redirected to the hosted version domain `hosted-version.com/admin/conferences/myconference/admin`

@lagartoflojo Please do let me know your suggestions.

lagartoflojo:
Sure, that sounds perfectly reasonable. Let's implement the simplest version first, and then we can look into the details. 
[...]

PR #1624, opened 2017-08-04
Introduce views for Custom domain

shlok007:
Dependent on #1610 .
Refers #1599 .

- conference#custom_domain
screenshot_20170805_020149

- conference#attach_custom_domain
screenshot_20170805_020219


app/views/admin/conferences/custom_domain.html.haml
```diff
@@ -0,0 +1,82 @@
...
+          %td
+            .btn-group
+              = link_to 'Edit', attach_custom_domain_admin_conference_path(@conference.short_title),
+                method: :get, class: 'btn btn-primary'
+              = link_to 'Delete', admin_conference_path(@conference.short_title),
+                method: :delete, class: 'btn btn-danger'
+.row
+  .col-md-9.text-right
+    = link_to '#status-help', class: 'btn btn-default',  "data-toggle"=>"collapse" do
+      Help?
```
lagartoflojo: Either ask "Do you need help?" with the question mark, or just say "Help" with no question mark, but "Help?" looks funny :P

app/views/admin/conferences/custom_domain.html.haml
```diff
@@ -0,0 +1,82 @@
...
+.row
+  .col-md-12
+    .collapse#status-help
+      %h2
+        Instructions to add your own domain
+      %hr
+      %p
+        When you created the conference, the conference splash page is available in the following domain:
```
lagartoflojo: > Normally, your conference's splash page is available at

app/views/admin/conferences/custom_domain.html.haml
```diff
@@ -0,0 +1,82 @@
...
+      %p
+        To add your own domain you can follow these
+        %i
+          three
+        steps:
+      %ol
+        %li
+          Pick a domain you want to host your conference on from a domain registrar or register it with a domain registrar.
+        %li
+          Add your own domain name to OSEM ( link to conference/custom_domain#create ) and select the conference it should belong to.
```
lagartoflojo: > OSEM ( link to conference/custom_domain#create )

I guess this is supposed to be an actual link?

app/views/admin/conferences/custom_domain.html.haml
```diff
@@ -0,0 +1,82 @@
...
+        %li
+          The last step is adding a CNAME record in your registrar’s DNS Settings. Different registrars have different ways of adding a CNAME record. The following guides would help you setup a CNAME record depending upon the registrar of your domain. If it doesn’t, you probably should get in touch with your domain registrar and ask him how to register a CNAME record in their platform.
```
lagartoflojo: Whoa, long line! How about chopping it into the individual sentences?

spec/controllers/admin/conferences_controller_spec.rb
```diff
@@ -135,6 +135,15 @@
...
     end
+
+    describe 'GET #custom_domain', blah: true do
+      it 'render custom domain' do
+        get :custom_domain, id: conference.short_title
+
+        expect(response).to render_template :custom_domain
+        expect(response).to be_success
+      end
+    end
```
lagartoflojo: And tests about updating the domain?

spec/features/conference_spec.rb
```diff
@@ -68,6 +68,15 @@
...
     end
+
+    scenario 'show custom domain of a conference', feature: true, js: true do
+      conference = create(:conference, custom_domain: 'mydomain.conf')
+      sign_in user
+
+      visit custom_domain_admin_conference_path(conference.short_title)
+
+      expect(page).to have_text('mydomain.conf')
+    end
```
lagartoflojo: The point of the page is to be able to change the domain as well I guess? This test is pretty simple...

app/models/conference.rb
```diff
@@ -93,6 +93,28 @@ def user_registered? user
...
+    unless ENV['OSEM_HOSTNAME'].nil?
+      cname_record = Resolv::DNS.new.getresources(custom_domain, Resolv::DNS::Resource::IN::CNAME)
+      if cname_record.present?
+        return ENV['OSEM_HOSTNAME'] == Resolv::DNS.new.getresources(custom_domain, Resolv::DNS::Resource::IN::CNAME).first.name.to_s
+      else
+        return false
+      end
+    end
+
+    '--feature disabled--'
```
shlok007: There can be a situation in which `ENV['OSEM_HOSTNAME']` is not present. In which case it won't be possible to check if the custom domain has the correct CNAME. So, this feature will be disabled until the site admin sets the OSEM_HOSTNAME environment variable.
  
We can add the above information in the documentation about site admins about how to enable the feature.

app/models/conference.rb
```diff
@@ -93,6 +93,28 @@ def user_registered? user
...
+  # * +false+ -> If the custom domain does not have a CNAME record for the hosted version
+  def check_custom_domain
+    require 'resolv'
+
+    unless ENV['OSEM_HOSTNAME'].nil?
+      cname_record = Resolv::DNS.new.getresources(custom_domain, Resolv::DNS::Resource::IN::CNAME)
+      if cname_record.present?
+        return ENV['OSEM_HOSTNAME'] == Resolv::DNS.new.getresources(custom_domain, Resolv::DNS::Resource::IN::CNAME).first.name.to_s
+      else
+        return false
```
shlok007: That makes perfect sense. Using Services might be something we want at this stage. The models are quite lengthy. We can also consider moving the Stripe payment logic in a service too later on.

app/views/admin/conferences/custom_domain.html.haml
shlok007: Okay sure

spec/features/conference_spec.rb
shlok007: Oh! I missed the feature and controller tests for updating the custom domain. I'll add those.

app/controllers/admin/conferences_controller.rb
```diff
@@ -4,6 +4,23 @@ class ConferencesController < Admin::BaseController
...
+    def update_domain
+      @conference.assign_attributes(conference_params)
+      if @conference.save
+        redirect_to custom_domain_admin_conference_path(id: @conference.short_title),
+                    notice: 'Added new domain name to conference. This does not mean that the new domain should work. Please make sure you follow step 2 to point your domain to this hosted version'
+      else
+        redirect_to custom_domain_admin_conference_path(id: @conference.short_title),
+                    notice: 'Failed to add the new domain as custom domain to the conference'
+      end
+    end
```
shlok007: I agree :+1:

app/services/conference_domains_service.rb
```diff
@@ -0,0 +1,27 @@
+class ConferenceDomainsService
```
lagartoflojo: No tests? :'(

shlok007: Oh yes, I forgot to mention that to add tests we will need a couple of domains in which one has a CNAME record for another domain.
  
I can use `osem-demo.herokuapp.com` and `demo.osem.io` for now but the tests will fail if we ever switch the staging app to another domain.

lagartoflojo: Tests should not touch the "real" internet.
Instead, you should mock the usage of `Resolv::DNS`.

app/models/conference.rb
lagartoflojo: I understand that the feature will be disabled if OSEM_HOSTNAME is not set.
What I don't understand is why we return this funny string `'--feature disabled--'`

shlok007: Oh! Sorry about that :stuck_out_tongue: . Is a simple string `feature disabled` enough?

app/views/admin/conference_domains/show.haml
```diff
@@ -17,14 +17,15 @@
...
+            - if ENV['OSEM_HOSTNAME'].present?
+              - ck_domain = ConferenceDomainsService.new(@conference).check_custom_domain
+              - if ck_domain
+                %span.glyphicon.glyphicon-ok
+              - else
+                %span.glyphicon.glyphicon-remove
             - else
               %i
-                = ck_domain
+                feature disabled
```
lagartoflojo: would it make sense maybe to tell the user WHY the feature is disabled, and how to enable it?

shlok007: I was hoping to include it in our documentation ( Github wiki section) .

lagartoflojo: Sure, but if people see "feature disabled" here, they might not know why it's disabled, and they might not realize how to enable it. So you can link to that wiki page from here then.

shlok007:
@lagartoflojo It's getting quite difficult to rebase 14 commits with merge conflicts on each one of them.  Please do have a look at the last commits, I'll clean up the commit history and rebase.

app/views/admin/conference_domains/show.haml
shlok007: screenshot_20170823_111145

A question mark pointing to the wiki page. 
Does that sound okay?

lagartoflojo:
I think I said this before, but I don't like that the instructions in `/admin/conferences/osemdemo/conference_domains/edit` look different to those in `/admin/conferences/osemdemo/conference_domain`. I think it would be better to make them consistent.

config/routes.rb
```diff
@@ -28,6 +32,10 @@
     end
     resources :comments, only: [:index]
     resources :conferences do
+      get 'conference_domains/edit' => 'conference_domains#edit'
+      get 'conference_domain' => 'conference_domains#show'
+      patch 'conference_domain' => 'conference_domains#update'
```
lagartoflojo: Why don't you use `resource` for this?
`resource :conference_domain`
Since a conference can only have a single custom domain, it should be a singular resource.
See: http://guides.rubyonrails.org/routing.html#singular-resources

Also, the URLs generated are pretty ugly, because they repeat "conference": `conferences/osemdemo/conference_domain`
We could make it nicer, like: `conferences/osemdemo/domain`
With something like: `resource :domain, controller: 'conference_domains'`

app/views/admin/conference_domains/show.haml
```diff
@@ -0,0 +1,95 @@
+.row
```
lagartoflojo: The instructions in this file are the same as in conference_domains/_form.haml right? Maybe you can extract this into a partial?

app/views/admin/conference_domains/show.haml
```diff
@@ -0,0 +1,95 @@
...
+                method: :get, class: 'btn btn-primary'
+              = link_to 'Delete', admin_conference_domain_path(@conference.short_title),
+                method: :delete, class: 'btn btn-danger'
+.row
+  .col-md-9.text-right
+    = link_to '#status-help', class: 'btn btn-default',  "data-toggle"=>"collapse" do
+      Need Help?
+.row
+  .col-md-12
+    .collapse#status-help
```
lagartoflojo: I'm not sure if this should be collapsed.
When you set a domain for the conference, you get a flash notice saying "now follow step 3", but "step 3" is not visible anywhere, and you have to *guess* that it's under the "Need help?" link.

lagartoflojo:
Hi @shlok007, can we get this merged? :)

shlok007:
Hi! I'll work on the suggested changes. I'll make sure to update the pending PRs and get back on track by this week's end at most.

app/views/admin/conference_domains/show.haml
shlok007: There is a difference between the two. The `conference_domains/_form` partial contains a form in step 2 where the domain name can be added whereas the help text is just simple text with no such form. 

* Help text

screenshot_20170921_220653

* _form.haml 

screenshot_20170921_220715

This is the reason why I preferred not to extract it into a partial.

app/views/admin/conference_domains/show.haml
shlok007: Sounds good, I don't think there is any harm in keeping it visible at all times.

app/views/admin/conference_domains/show.haml
lagartoflojo: How about the extracting each step into its own partial and having 2 different "step 2" partials?

app/views/admin/conference_domains/show.haml
lagartoflojo: ok!

bear454:
@shlok007 : any progress?

differentreality:
@shlok007 can you please rebase?

We should figure out a way to get this merged soon, before more conflicts come up. @shlok007 will you have the time to work on the suggested changes in the next couple weeks? What do you suggest?

differentreality:
The more this sits, the more conflicts we will have to resolve, making it more difficult (and time consuming) to merge and work on the feature.

Can we merge this as is (after it is rebased obviously) and fine tune it later?
@lagartoflojo 

If @shlok007 doesn't have the time to work on it, I will happily rebase it so that it can be merged.

hennevogel:
@lagartoflojo agreed that we can merge this as is?

differentreality:
@shlok007 are you up for rebasing this? :smiley:

shlok007:
Really sorry to keep this waiting. Thank you so much for the reminder, I'll work on it tonight.

hennevogel:

PR #1629, opened 2017-08-07
Ability of user to create registration is fixed

### LINKED ISSUE #829: Allow confirmed speakers to register even after registration period is over

### LINKED ISSUE #1628: User can only  register once for every conference
https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L69 Ability of user to register should not return true when he have already register.

### LINKED ISSUE #1827: Speaker proposals allow registration outside registration period
**I'm submitting a ..**
- [X] Bug Report
- [ ] Feature Request

**Current behavior:**
After submitting a proposal (Event), a number of steps are presented to the user, to finish their proposal. One of those steps is event registration, regardless of whether the registration period is open for the conference.

**Expected behavior:**
No registration link is presented outside of registration period; authorization blocks registration outside of registration period for non-admin users.

**Steps to reproduce:**

1. Create a conference; do not set a registration period.
2. Create a Call for Papers, type: events.
3. Submit a proposal.
4. Perform the "Registration" task.

**Related code:**

**Other information:**

namangupta01:
Closes #1628 
Closes #1827 
Closes #829

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -66,7 +66,13 @@ class Ability
 
     can [:new, :create], Registration do |registration|
       conference = registration.conference
-      conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user)
+      if conference.user_registered? user
+        false
+      elsif conference.program.speakers.confirmed.include?(user) && conference.registration_period
+        true
+      else
+        conference.registration_open? && !conference.registration_limit_exceeded?
+      end
     end
 
     can :index, Organization
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -89,11 +89,131 @@ describe 'User' do
       let(:other_self_organized_track) { create(:track, :self_organized) }
 
       it{ should be_able_to(:manage, user) }
-
       it{ should be_able_to(:manage, registration_public) }
       it{ should be_able_to(:manage, registration_not_public) }
-      it{ should_not be_able_to(:new, Registration.new(conference_id: conference_with_closed_registration.id))}
-      it{ should_not be_able_to(:create, Registration.new(conference_id: conference_with_closed_registration.id))}
+
+      # Test for user can register or not
+      context 'when user is not a speaker with event confirmed' do
+        let(:conference) { create(:conference) }
+
+        context 'when the registration is closed' do
+          before :each do
+            create(:registration_period, conference: conference, start_date: Date.current - 6.days, end_date: Date.current - 6.days)
+          end
+
+          it{ should_not be_able_to(:new, Registration.new(conference: conference)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference)) }
+        end
+
+        context 'when the registration period is not set' do
+          it{ should_not be_able_to(:new, Registration.new(conference: conference)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference)) }
+        end
+
+        context 'when user has already registered' do
+          before :each do
+            create(:registration, conference: conference, user: user)
+          end
+
+          it{ should_not be_able_to(:new, Registration.new(conference: conference)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference)) }
+        end
+
+        context 'when registrations are open' do
+          before :each do
+            create(:registration_period, conference: conference)
+          end
+
+          it{ should be_able_to(:new, Registration.new(conference: conference)) }
+          it{ should be_able_to(:create, Registration.new(conference: conference)) }
+
+          context 'when user has not registered with no registration_limit_exceeded' do
+            before :each do
+              conference.registration_limit = 1
+            end
+
+            it{ should be_able_to(:new, Registration.new(conference: conference)) }
+            it{ should be_able_to(:create, Registration.new(conference: conference)) }
+          end
+
+          context 'when user has not registered with registration_limit_exceeded' do
+            before :each do
+              conference.registration_limit = 1
+              create(:registration, conference: conference, user: user2)
+            end
+
+            it{ should_not be_able_to(:new, Registration.new(conference: conference)) }
+            it{ should_not be_able_to(:create, Registration.new(conference: conference)) }
+          end
+        end
+      end
+
+      context 'when user is a speaker with event confirmed' do
+        let(:conference_with_speaker_confirmed) { create(:conference) }
+        let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+        before :each do
+          event_with_speaker_confirmed.speakers << user
+        end
+
+        context 'when registration period is not set' do
+          it{ should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+        end
+
+        context 'when speaker has already registered' do
+          before :each do
+            create(:registration, conference: conference_with_speaker_confirmed, user: user)
+          end
+
+          it{ should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+        end
+
+        context 'when registration period is set' do
+          before :each do
+            create(:registration_period, conference: conference_with_speaker_confirmed)
+          end
+
+          context 'when registrations are open' do
+            context 'when speaker has not registered' do
+              it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+              it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+            end
+
+            context 'when registration_limit_exceeded' do
+              before :each do
+                conference_with_speaker_confirmed.registration_limit = 1
+                create(:registration, conference: conference_with_speaker_confirmed, user: user2)
+              end
+
+              it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+              it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+            end
+          end
+
+          context 'when registrations are closed' do
+            before :each do
+              create(:registration_period, conference: conference_with_speaker_confirmed, start_date: Date.current - 6.days, end_date: Date.current - 6.days)
+            end
+
+            context 'when speaker has not registered' do
+              it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+              it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+            end
+
+            context 'with registration_limit_exceeded' do
+              before :each do
+                conference_with_speaker_confirmed.registration_limit = 1
+                create(:registration, conference: conference_with_speaker_confirmed, user: user2)
+              end
+
+              it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+              it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+            end
+          end
+        end
+      end
 
       it{ should be_able_to(:index, Ticket) }
       it{ should be_able_to(:manage, TicketPurchase.new(user_id: user.id)) }


db/schema.rb
```diff
@@ -14,7 +14,7 @@
 ActiveRecord::Schema.define(version: 20170721001700) do
 
   create_table "ahoy_events", force: :cascade do |t|
-    t.integer  "visit_id"
+    t.uuid     "visit_id",   limit: 16
```
Ana06: this change is unrelated, isn't it? :confused:

app/models/ability.rb
```diff
@@ -66,7 +66,7 @@ def signed_in(user)
 
     can [:new, :create], Registration do |registration|
       conference = registration.conference
-      conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user)
+      (conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user))&&!(conference.user_registered? user)
```
Ana06: This line got too complicate to read.

what about writing it like:

``` Ruby
return false if conference.user_registered? user
return true if conference.program.speakers.confirmed.include?(user)
conference.registration_open? && !conference.registration_limit_exceeded?
```

db/schema.rb
namangupta01: This get automatically changed when i run the command rake db:migrate

app/models/ability.rb
namangupta01: Thanks :) This looks good !!

db/schema.rb
Ana06: @namangupta01 it seems it comes from here: https://github.com/openSUSE/osem/pull/1229

@differentreality @eug3nix could be a migration missed?

@namangupta01 anyway it is unrelated to your changes, so please exclude it from your PR. :wink:

namangupta01:
@Ana06  Is it now good to go ?

namangupta01:
@Ana06 @differentreality can anyone one of you help me to understand why my Travis build is failing? Thanks:)
https://travis-ci.org/openSUSE/osem/builds/268890801?utm_source=github_status&utm_medium=notification

differentreality:
@namangupta01 sure! How much do you understand? And which part do you not understand?

namangupta01:
@differentreality Thank you :) I understand that the error is occurring because of the line https://github.com/namangupta01/osem/blob/e065079b6bb116780c5d06895b545299af29048e/app/models/ability.rb#L70
But i didn't understand why the error is saying **unexpected return** as my syntax is correct.
https://github.com/namangupta01/osem/blob/e065079b6bb116780c5d06895b545299af29048e/app/models/ability.rb#L70

differentreality:
@namangupta01 you said your syntax is correct, yet there is an error. 
Did you google the error to figure out what is going on?

namangupta01:
@differentreality Oh! I got why it was failing. Thanks :)

app/models/ability.rb
```diff
@@ -66,7 +66,13 @@ def signed_in(user)
 
     can [:new, :create], Registration do |registration|
       conference = registration.conference
-      conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user)
+      if conference.user_registered? user
+        false
```
differentreality: Would it make sense to add a `cannot` block for this case after this can block?

namangupta01: @differentreality I didn't get what case are you taking about? But as far as i think you are taking about  **if conference.user_registered? user**.
I donot think why we should add a cannot block for this?

namangupta01: As this can block is used to check wheather a user can register or not, so i think all the cases that depends on wheather a user can register or not should be in same block.

namangupta01:
@differentreality @Ana06  Please review this pull request

differentreality:
We should enhance our tests accordingly.

differentreality:
Great @namangupta01
Can you please show me where we test the case that the user is a speaker?

namangupta01:
@differentreality still stuck in that example :(

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,51 @@
...
+          it do
+            event_with_speaker_confirmed.speakers << user
+            should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed))
+          end
+        end
+
+        context 'when registration are closed and registration_limit_exceeded' do
+          let(:conference_with_closed_registration_and_limit_excedeed) { create(:conference, registration_limit: '1') }
+          let!(:closed_registration_period) { create(:registration_period, conference: conference_with_closed_registration_and_limit_excedeed, start_date: Date.current - 6.days, end_date: Date.current - 6.days) }
+          let(:program_with_speaker_confirmed) { create(:program, conference: conference_with_closed_registration_and_limit_excedeed) }
```
differentreality: Also not needed

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,51 @@
       it{ should_not be_able_to(:new, Registration.new(conference_id: conference_with_closed_registration.id))}
       it{ should_not be_able_to(:create, Registration.new(conference_id: conference_with_closed_registration.id))}
 
+      # Test for user can register or not
+      let(:registration_for_registering_user) { create(:registration, conference: create(:conference), user: user) }
```
differentreality: Perhaps it is easier to understand what we are testing with this, if it is called user_registration ?

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,51 @@
...
+      it{ should_not be_able_to(:new, Registration.new(conference_id: registration_for_registering_user.conference_id)) }
+      it{ should_not be_able_to(:create, Registration.new(conference_id: registration_for_registering_user.conference_id)) }
+
+      context 'when user is a speaker' do
+        context 'when speaker is confirmed' do
+          let(:conference_with_speaker_confirmed) { create(:conference) }
+          let(:program_with_speaker_confirmed) { create(:program, conference: conference_with_speaker_confirmed) }
+          let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: program_with_speaker_confirmed) }
+
+          it do
```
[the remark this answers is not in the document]
namangupta01: All the things are defined in the context what can we give more in 'it block' description?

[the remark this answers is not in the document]
namangupta01: Does this description is correct?

screen shot 2017-11-16 at 2 15 18 pm

namangupta01:
@differentreality Please have a look!

namangupta01:
@differentreality Is this fine now? Should i squash the commits now?

bear454:
@namangupta01: check out https://github.com/openSUSE/osem/pull/1834/commits/dae9cea34d7f531ecf9ae03e79a9506a7256e028

app/models/ability.rb
```diff
@@ -66,7 +66,13 @@ def signed_in(user)
...
+      elsif conference.program.speakers.confirmed.include? user
```
bear454:
```
elsif (conference.program.speakers.confirmed.include?(user) && conference.registration_period)
  # Confirmed speakers can register before or after a registration period, but there _must be_ a registration period.
```

namangupta01:
@bear454 respective changes are done and tests are implemented accordingly.

bear454:
@differentreality, @Ana06 : re-evaluate reviews, please?

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,67 @@
...
+          end
+
+          it 'checks that speaker with event confirmed with registration period set can create registration' do
+            event_with_speaker_confirmed.speakers << user
+            conference_with_speaker_confirmed.registration_period = create(:registration_period)
+            should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed))
+          end
+        end
+
+        context 'when registration are closed and registration_limit_exceeded' do
```
differentreality: This is not one single test case. I would expect that we test at least for the following:
- registration open + limit_exceeded 
- registration closed + limit **not** exceeded

namangupta01: @differentreality Will correct it 👍

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,67 @@
...
+            should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed))
+          end
+
+          it 'checks that speaker with event confirmed with registration period not set cannot create registration' do
+            event_with_speaker_confirmed.speakers << user
+            should_not be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed))
+          end
+        end
+
+        context 'when event is confirmed and the registration period is set' do
```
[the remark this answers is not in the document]
namangupta01: registration period open ? Speaker can do registration only when the registration period is set regardless of the whether registration is open or closed.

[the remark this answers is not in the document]
namangupta01: I didn't get the point of using open here and also in other comments.

spec/models/ability_spec.rb
```diff
@@ -95,6 +95,67 @@
...
+          let(:conference_with_speaker_confirmed) { create(:conference) }
+          let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+          it 'checks that speaker with event confirmed with registration period set can open new registration' do
```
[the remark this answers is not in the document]
namangupta01: Shouldn't it be `when user is a speaker with event confirmed and registration period is set the speaker can access #new action for Registration` ?

differentreality:
@namangupta01 from the comments it seems we are not on the same page on which cases we need to test for. Can you please make a table with all the necessary cases? Preferably numbered, so that we can confirm which ones we are currently covering in our tests, and which ones we aren't.

namangupta01:
@differentreality Please look at the cases! I will implement it once you verify it!

|  | User Registered? | Registration open? | Speaker's event is confirmed (if user is speaker) | Registration Set | Conference Limit Exceed?| User can register?
|---|------------------|------------------|--------------------|-----|-----|---|
|1.|no|no|no (USER OR SPEAKER)|no|no|no
|2.|no|no|no (USER OR SPEAKER)|yes|no|no|
|3.|no|no|no (USER OR SPEAKER)|yes|yes|no|
|4.|no|yes|no (USER OR SPEAKER)|yes|yes|no|
|5.|no|yes|no (USER OR SPEAKER)|yes|no|yes|
|6.|yes|-|-|-|-|-|no|
|7.|no|no|yes|yes|yes|yes|
|8.|no|no|yes|yes|no|yes|
|9.|no|no|yes|no|no|no|
|10.|no|yes|yes|yes|yes|yes|
|11.|no|yes|yes|yes|yes|yes|
|12.|no|yes|yes|yes|no|yes|

namangupta01:
Please let me know if i left some cases! And cases 1 to 3 can be implemented in one case do i have to implement it in different cases.

spec/models/ability_spec.rb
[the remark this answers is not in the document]
differentreality: So 1 case is No9 from your array (registration_period NOT set being the important part).
If the registration_period is set, what we care about is whether or not it is open.
When open, obviously the speaker of a confirmed event, and anyone else, can register.
When not open, the speaker of a confirmed event can still register. <-- That's the important case here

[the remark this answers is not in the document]
namangupta01: Yes you are right! will update it accordingly!

namangupta01:
@differentreality All tests are implemented. Have a look!

bear454:
Finite state machines FTW :)

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+
+      end
+
+      context 'when user is a speaker with event confirmed' do
+        context 'when the registration period is not set' do
+          let(:conference_with_speaker_confirmed) { create(:conference) }
+          let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+          it 'cannot access #new action for Registration' do
+            event_with_speaker_confirmed.speakers << user
```
differentreality: Move this to a before block?

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+        end
+
+        context 'when user have already registered' do
+          let(:user_registration) { create(:registration, conference: conference, user: user) }
+
+          it{ should_not be_able_to(:new, Registration.new(conference_id: user_registration.conference_id)) }
+          it{ should_not be_able_to(:create, Registration.new(conference_id: user_registration.conference_id)) }
+        end
+
+        context 'when user have not registered and registrations are open with conference_limit_not_exceeded' do
```
differentreality: The hidden not in conference_limit_not_exceeded is not very obvious. 
`registration_limit_not_exceeded` derives from the model function that checks this https://github.com/openSUSE/osem/blob/master/app/models/conference.rb#L712, changing this too much, and adding text makes it harder to read in my opinion. 
This can benefit from a small re-write.

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+          let(:registration_period) { create(:registration_period) }
+          let(:conference_with_open_registration_and_limit_not_excedeed) { create(:conference, registration_limit: '1', registration_period: registration_period) }
+
+          it{ should be_able_to(:new, Registration.new(conference: conference_with_open_registration_and_limit_not_excedeed)) }
+          it{ should be_able_to(:create, Registration.new(conference: conference_with_open_registration_and_limit_not_excedeed)) }
+        end
+
+        context 'when user have not registered and registrations are open with conference_limit_exceeded' do
```
differentreality: _has_ not

differentreality: it's registration limit, not conference. We should either use the function name as it is, or use free text.

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+          let(:registration_period) { create(:registration_period) }
+          let(:conference_with_open_registration_and_limit_excedeed) { create(:conference, registration_limit: '1', registration_period: registration_period) }
+          let(:registration_for_user_with_open_registration_and_limit_excedeed) { create(:registration, conference: conference_with_open_registration_and_limit_excedeed, user: user2) }
+
+          it{ should_not be_able_to(:new, Registration.new(conference_id: registration_for_user_with_open_registration_and_limit_excedeed.conference_id)) }
+          it{ should_not be_able_to(:create, Registration.new(conference_id: registration_for_user_with_open_registration_and_limit_excedeed.conference_id)) }
+        end
+
```
differentreality: remove empty line

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+        context 'when the registration period is not set' do
+          let(:conference_with_speaker_confirmed) { create(:conference) }
+          let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+          it 'cannot access #new action for Registration' do
+            event_with_speaker_confirmed.speakers << user
+            should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed))
+          end
+
+          it 'can access #create action for Registration' do
```
differentreality: can or cannot? You test for should_not

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,102 @@
...
+            it 'can access #new action for Registration' do
+              event_with_speaker_confirmed.speakers << user
+              should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed))
+            end
+
+            it 'can access #create action for Registration' do
+              event_with_speaker_confirmed.speakers << user
+              should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed))
+            end
+
```
differentreality: Unnecessary empty line

namangupta01:
@differentreality I didn't take the case when user is confirmed speaker, registration is open and registration limit exceeded because when the user is confirmed speaker and registration is open then it does not depend whether registration limit exceeded or not, user can always register in this case.

differentreality:
It is supposed to be able to always register. Tests are there to test whether or not our code works as expected.

namangupta01:
ok i will implement this test also

namangupta01:
@differentreality Test for the Case when user is a confirmed speaker, registration is open, and the registration limit has exceeded? has been added. Have a look!

namangupta01:
@differentreality Please review!

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,103 @@
...
+          it{ should_not be_able_to(:new, Registration.new(conference_id: registration_for_user_with_open_registration_and_limit_excedeed.conference_id)) }
+          it{ should_not be_able_to(:create, Registration.new(conference_id: registration_for_user_with_open_registration_and_limit_excedeed.conference_id)) }
+        end
+      end
+
+      context 'when user is a speaker with event confirmed' do
+        context 'when the registration period is not set' do
+          let(:conference_with_speaker_confirmed) { create(:conference) }
+          let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
```
differentreality: You are defining this in multiple places, it seems it should be defined once after line 136

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,103 @@
...
+
+          before :each do
+            event_with_speaker_confirmed.speakers << user
```
differentreality: Then you can also extract this before block after line 136

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,103 @@
...
+
+            before :each do
+              event_with_speaker_confirmed.speakers << user
+            end
+
+            it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+            it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+
+            context 'and registration_limit_exceeded' do
+              let(:conference_with_open_registration_and_limit_excedeed) { create(:conference, registration_limit: '1', registration_period: registration_period) }
```
differentreality: Based on the previous comments, instead of this let statement, perhaps you can just change the registration_limit in the existing conference?

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,103 @@
...
+          it{ should_not be_able_to(:new, Registration.new(conference: conference_with_closed_registration)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference_with_closed_registration)) }
+        end
+
+        context 'when the registration period is not set' do
+          it{ should_not be_able_to(:new, Registration.new(conference: conference)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference)) }
+        end
+
+        context 'when user has already registered' do
```
differentreality: Let's also test this for when the user is a speaker with confirmed event.

namangupta01:
@differentreality please re-review!

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
-      it{ should_not be_able_to(:new, Registration.new(conference_id: conference_with_closed_registration.id))}
-      it{ should_not be_able_to(:create, Registration.new(conference_id: conference_with_closed_registration.id))}
+
+      # Test for user can register or not
+      context 'when user is not a speaker with event confirmed' do
+        let(:conference) { create(:conference) }
+        let(:registration_period) { create(:registration_period) }
+
+        context 'when the registration is closed' do
+          let(:conference_with_closed_registration) { create(:conference) }
```
differentreality: What's the difference between conference and conference_with_closed_registration? They are both created by the same factory, yet you expect one to have a registration which is over, and the other one to not have a registration_period at all.

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+
+        context 'when user has already registered' do
+          let(:user_registration) { create(:registration, conference: conference, user: user) }
+
+          it{ should_not be_able_to(:new, Registration.new(conference_id: user_registration.conference_id)) }
+          it{ should_not be_able_to(:create, Registration.new(conference_id: user_registration.conference_id)) }
+        end
+
+        context 'when user has not registered and registrations are open with no registration_limit_exceeded' do
+          let(:conference_with_open_registration_and_limit_not_excedeed) { create(:conference, registration_limit: '1', registration_period: registration_period) }
```
differentreality: This way of associating the newly created conference with the registration_period is not very good, as it is not very explicit which conference the registration_period should belong to.

Why do you need to create a new conference here? Isn't it enough that we have a conference (or should have) with a registration period set? Then you can manipulate the dates of that registration_period and register your user in that conference.

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+      context 'when user is a speaker with event confirmed' do
+        let(:conference_with_speaker_confirmed) { create(:conference) }
+        let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+        before :each do
+          event_with_speaker_confirmed.speakers << user
+        end
+
+        context 'when speaker has already registered' do
+
```
differentreality: Please remove this empty line

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+      context 'when user is a speaker with event confirmed' do
+        let(:conference_with_speaker_confirmed) { create(:conference) }
+        let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_speaker_confirmed.program) }
+
+        before :each do
+          event_with_speaker_confirmed.speakers << user
+        end
+
+        context 'when speaker has already registered' do
```
differentreality: For this to be tested properly, you need to have a conference with an existing registration_period. If there is no registration period this test would be passing not because the already registered user cannot register, but because the registration period is not set. 

While writing your tests, imagine you have not written your code yet. Also your code can change at any time, the order of the checks can be different or totally altered. The tests need to be independent, and run successfully no matter the code. For that reason, you need to set them up properly

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+          it{ should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+        end
+
+        context 'when the registration period is not set' do
+          it{ should_not be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+          it{ should_not be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+        end
+
+        context 'And the registration period is set' do
```
differentreality: What about we keep some uniformity on the context text?

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+          let(:registration_period) { create(:registration_period) }
```
differentreality: In these tests, I would always expect the registration_period to be associated with a relevant conference immediately. It is a bit confusing how you later on associate it with a different conference.

Think about it logically, what do you do first if you are a user in OSEM?
Don't you first create a new conference, and then go setup a registration_period (for that specific conference)?
If you keep this logical order, it is way easier to read the tests later on.

This line can simply be
create(:registration_period, conference: conference_with_speaker_confirmed)

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+
+          before :each do
+            conference_with_speaker_confirmed.registration_period = registration_period
+          end
+
+          context 'with registration open' do
+
```
differentreality: Please remove empty line (there other relevant places for this to happen)

spec/models/ability_spec.rb
```diff
@@ -89,11 +89,104 @@
...
+                create(:registration, conference: conference_with_speaker_confirmed, user: user2)
+              end
+
+              it{ should be_able_to(:new, Registration.new(conference: conference_with_speaker_confirmed)) }
+              it{ should be_able_to(:create, Registration.new(conference: conference_with_speaker_confirmed)) }
+            end
+          end
+
+          context 'with registration closed' do
+            let(:event_with_speaker_confirmed) { create(:event, state: 'confirmed', program: conference_with_closed_registration.program) }
```
differentreality: Where is it defined that this conference has a closed registration?

spec/models/ability_spec.rb
namangupta01: Yes, you are right about that they both created form the same factory but there is a difference.
https://github.com/openSUSE/osem/blob/master/spec/models/ability_spec.rb#L33
That's how conference and conference_with_closed_registration are different.
The only thing is that i shouldn't have create the conference_with_closed_registration here as it is already created in https://github.com/openSUSE/osem/blob/master/spec/models/ability_spec.rb#L32

spec/models/ability_spec.rb
namangupta01: Yes we can do that! The only thing that i was trying to do is that it would make easier to understand the code!

spec/models/ability_spec.rb
namangupta01: okay 👍

spec/models/ability_spec.rb
namangupta01: check this https://github.com/openSUSE/osem/blob/master/spec/models/ability_spec.rb#L33

spec/models/ability_spec.rb
namangupta01: One of the main thing about let statement is that the value will be cached
across multiple calls in the same example but not across examples. That's why i don't think it's a problem. Please correct me if i am wrong!

spec/models/ability_spec.rb
namangupta01: Yes you are right about that tests are passing because registration period is not set.

spec/models/ability_spec.rb
namangupta01: I associate a conference with a different conference because value will be cached
across multiple calls in the same example but not across examples. So when there is another example test it won't be affect by the previous or any other test.

spec/models/ability_spec.rb
namangupta01: This test is already there.

namangupta01:
Hey @differentreality I have done the changes please have a look.

namangupta01:
@differentreality Please re-review!

namangupta01:
@bear454 Can you please see this?

bear454:
@namangupta01 I re-triggered the test suite; it's green now, but I can't change @Ana06 or @differentreality 's reviews.

differentreality:
@namangupta01 :+1:  please squash your commits

namangupta01:
sure!

namangupta01:
@differentreality Done squashing!

namangupta01:
@differentreality something needs to be change or it is good to go?

differentreality:
Sorry for the delay @namangupta01 I have been exceptionally busy the past few weeks. 

Thanks so much for working on this feature :-) Very good work :+1:

PR #1630, opened 2017-08-09
Removed delete action from abiltiy.rb and other places

### LINKED ISSUE #1608: Remove delete action from ability

divyanshumehta:
There was no delete action defined for proposals, but it was there in ability.rb So the ability was deleted. Fixes #1608


app/models/ability.rb
```diff
@@ -81,7 +81,7 @@ def signed_in(user)
       booth.users.include?(user)
     end
 
-    can [:create, :destroy], Subscription, user_id: user.id
+    can [:create], Subscription, user_id: user.id
```
differentreality: Why are you doing this?

divyanshumehta: Oh sorry edited the wrong line

PR #1631, opened 2017-08-09
Event scheduling for tracks

AEtherC0r3:
* Don't allow tracks with overlapping dates to use the same room
* Add accepted and self_organized scopes to Track
* Don't allow an event to be scheduled outside of it's track's room and dates
* Remove unnecessary abilities from AdminAbility, in order to allow CanCanCan to load @events and @event in EventsController
* Allow the track organizers to manage the events and the events' commercials of their tracks
* Create associate between Schedule and Track
* Allow the track organizers to create schedules for their tracks and schedule the events of their tracks
* Restrict events of self-organized tracks to be scheduled only in their tracks schedules
* Show scheduled events of self-organized tracks and the rooms those tracks occupy in the conference scheduled as semitransparent
* Add admin/SchedulesController#new action and related form and buttons
* Add buttons to easily access admin//Schedules#create and #show from Tracks#index and #show
* Show a unified schedule in Schedules#show
* Create tabs for conference and track schedules in admin/Schedules#index
* Add selected_schedule_id to Track
* Relax requirement for cfp_active to be enabled in Event and check it in ProposalsController
* Validate that events of self-organized tracks are scheduled only in their track's schedules

### DIFF
--- a/app/controllers/admin/tracks_controller.rb
+++ b/app/controllers/admin/tracks_controller.rb
@@ -101,6 +101,18 @@ module Admin
       update_state(:cancel, "Track #{@track.name} canceled!")
     end
 
+    def update_selected_schedule
+      if @track.update_attributes(params.require(:track).permit(:selected_schedule_id))
+        respond_to do |format|
+          format.js { render json: {} }
+        end
+      else
+        respond_to do |format|
+          format.js { render json: { errors: "The selected schedule couldn't been updated #{@track.errors.to_a.join('. ')}" }, status: 422 }
+        end
+      end
+    end
+
     private
 
     def track_params
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -49,6 +49,13 @@ class ProposalsController < ApplicationController
     # by default.
     @event.speakers = [current_user]
     @event.submitter = current_user
+
+    if Track.find_by(id: params[:event][:track_id]).try(:cfp_active) == false
+      flash.now[:error] = 'You have selected a track that doesn\'t accept proposals'
+      render action: 'new'
+      return
+    end
+
     if @event.save
       ahoy.track 'Event submission', title: 'New submission'
       redirect_to conference_program_proposals_path(@conference.short_title), notice: 'Proposal was successfully submitted.'
@@ -60,6 +67,12 @@ class ProposalsController < ApplicationController
 
   def update
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
+
+    if Track.find_by(id: params[:event][:track_id]).try(:cfp_active) == false
+      flash.now[:error] = 'You have selected a track that doesn\'t accept proposals'
+      render action: 'edit'
+      return
+    end
 
     if @event.update(event_params)
       redirect_to conference_program_proposals_path(conference_id: @conference.short_title),
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -117,7 +117,7 @@ module ApplicationHelper
 
   def concurrent_events(event)
     return nil unless event.scheduled? && event.program.selected_event_schedules
-    event_schedule = event.program.selected_event_schedules.find_by(event: event)
+    event_schedule = event.program.selected_event_schedules.find { |es| es.event == event }
     other_event_schedules = event.program.selected_event_schedules.reject { |other_event_schedule| other_event_schedule == event_schedule }
     concurrent_events = []
 
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -298,5 +298,14 @@ class AdminAbility
     can :toggle_user, Role do |role|
       role.resource_type == 'Track' && track_ids_for_track_organizer.include?(role.resource_id)
     end
+
+    # Show Events in the admin sidebar
+    can :update, Event do |event|
+      event.new_record? && conf_ids_for_track_organizer.include?(event.program.conference_id)
+    end
+
+    can :manage, Event, track_id: track_ids_for_track_organizer
+    can :manage, Commercial, commercialable_type: 'Event',
+                             commercialable_id: Event.where(track_id: track_ids_for_track_organizer).pluck(:id)
   end
 end
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -307,5 +307,18 @@ class AdminAbility
     can :manage, Event, track_id: track_ids_for_track_organizer
     can :manage, Commercial, commercialable_type: 'Event',
                              commercialable_id: Event.where(track_id: track_ids_for_track_organizer).pluck(:id)
+
+    # Show Scheduless in the admin sidebar
+    can :update, Schedule do |schedule|
+      schedule.new_record? && conf_ids_for_track_organizer.include?(schedule.program.conference_id)
+    end
+
+    # Show new track schedule button
+    can :new, Schedule do |schedule|
+      schedule.new_record? && conf_ids_for_track_organizer.include?(schedule.program.conference_id) && schedule.track.try(:new_record?)
+    end
+
+    can :manage, Schedule, track_id: track_ids_for_track_organizer
+    can :manage, EventSchedule, schedule: { track_id: track_ids_for_track_organizer }
   end
 end
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -39,10 +39,6 @@ class AdminAbility
       event.program.cfp_open? && event.new_record?
     end
 
-    can [:update, :show, :index], Event do |event|
-      event.users.include?(user)
-    end
-
     # can manage the commercials of their own events
     can :manage, Commercial, commercialable_type: 'Event', commercialable_id: user.events.pluck(:id)
 
--- a/app/models/event_schedule.rb
+++ b/app/models/event_schedule.rb
@@ -12,6 +12,8 @@ class EventSchedule < ActiveRecord::Base
   validates :event, uniqueness: { scope: :schedule }
   validate :start_after_end_hour
   validate :start_before_start_hour
+  validate :room_of_track
+  validate :during_track
 
   scope :confirmed, -> { joins(:event).where('state = ?', 'confirmed') }
   scope :canceled, -> { joins(:event).where('state = ?', 'canceled') }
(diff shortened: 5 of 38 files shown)


app/helpers/application_helper.rb
```diff
@@ -136,7 +136,7 @@ def redirect_back_or_to(options = {}, response_status = {})
 
   def concurrent_events(event)
     return nil unless event.scheduled? && event.program.selected_event_schedules
-    event_schedule = event.program.selected_event_schedules.find_by(event: event)
+    event_schedule = event.program.selected_event_schedules.find { |es| es.event == event }
```
differentreality: Why are you changing this?

app/models/event_schedule.rb
```diff
@@ -52,4 +54,26 @@ def start_before_start_hour
...
+  # Validates that the event is scheduled in the same room as it's track
+  #
+  def room_of_track
+    if event && event.track.try(:room) && event.track.room != room
+      errors.add(:room, "must be the same as the track's room (#{event.track.room.name})")
+    end
+  end
+
+  ##
+  # Validates that the event is scheduled within it's track's time slot
```
differentreality: it's -> it is
its -> it belongs to it

app/models/event_schedule.rb
```diff
@@ -52,4 +54,26 @@ def start_before_start_hour
   def conference_id
     schedule.program.conference_id
   end
+
+  ##
+  # Validates that the event is scheduled in the same room as it's track
```
differentreality: its

app/views/admin/schedules/_event.html.haml
```diff
@@ -7,12 +7,13 @@
 / subtracting the padding before calculate the number of lines
 - lines = (height - 7) / 23
 - color = event.track.try(:color).present? ? event.track.try(:color) : 'FFFFFF'
+- non_schedulable = event_schedule_id && EventSchedule.find(event_schedule_id).schedule != @schedule
```
differentreality: I'd add a parenthesis here for better readability

app/helpers/application_helper.rb
AEtherC0r3: Because of the changes I've made to Program#selected_event_schedules, it no longer returns an ActiveRecord::Relation but an Array, and arrays don't have `find_by` they just have `find`

app/models/event_schedule.rb
AEtherC0r3: Yeah, I know. My hand just adds the apostrophe by itself, when I want to express possession and I'm not really paying attention to what I'm writing :P

spec/models/track_spec.rb
```diff
@@ -246,7 +354,8 @@
       self_organized_track.cfp_active = true
       self_organized_track.save!
       @a_track_organizer.add_role 'track_organizer', self_organized_track
-      @an_event_of_the_track = create(:event, program: self_organized_track.program, track: self_organized_track)
+      @an_event_of_the_track = create(:event, program: self_organized_track.program, track: self_organized_track, state: 'confirmed')
```
differentreality: How about we change this variable name to something shorter/simpler and self-explanatory?
This is an event that has as track the self_organized_track, so some ideas:
```
event_with_self_organized_track
self_organized_track_event
event_of_self_organized_track
```

spec/models/track_spec.rb
```diff
@@ -246,7 +354,8 @@
...
+      @schedule_of_the_track = create(:schedule, program: self_organized_track.program, track: self_organized_track)
```
differentreality: I would rename this as well.
schedule_of_self_organized_track
or
self_organized_track_schedule

app/helpers/application_helper.rb
differentreality: Wouldn't we then also need to change it in https://github.com/openSUSE/osem/blob/master/app/views/proposals/show.html.haml#L117 ?

differentreality:
If my track is accepted, and confirmed, perhaps it makes sense that I am able to check in tracks#index or tracks#show and be directed in seeing the submitted events for my tracks, and scheduling my events.

app/helpers/application_helper.rb
AEtherC0r3: Yup, sorry I missed it

AEtherC0r3:
I've added buttons that send you to admin/Tracks#show

AEtherC0r3:
Buttons in Tracks#index and Tracks#show that redirect to admin/Tracks#show



Track organizer being able to access admin/Events#index and manage his/her tracks's events


admin/Schedules#show not showing the events of self-organized tracks


admin/Schedules#show showing the rooms as semitransparent in conference schedules


Showing events from selected track schedules as semitransparent in conference schedules


Track organizer being able to create and manage schedules of his/her tracks from admin/Schedules#index, admin/Tracks#index and #show




Track organizer being able to schedule events of his/her tracks

app/models/event_schedule.rb
```diff
@@ -52,4 +55,34 @@ def start_before_start_hour
   def conference_id
     schedule.program.conference_id
   end
+
+  ##
+  # Validates that the event is scheduled in the same room as its track
+  #
+  def same_room_as_track
+    if event && event.track.try(:room) && event.track.room != room
```
differentreality: You could also try this approach to make it easier to understand what your code is doing. This way when reading the line that adds the error you are focusing on the actual validation that matters (in this case the room being the same as the track room)
```
return false unless event && event.track && event.track.room
errors.add() unless event.track.room == room
```

app/views/admin/schedules/show.html.haml
```diff
@@ -11,8 +11,15 @@
...
     .col-md-2
       Selected schedule
-      = check_box_tag @conference.short_title, @schedule.id, (@schedule.id == @selected_schedule.try(:id)),
-        method: :patch, url: (admin_conference_program_path(@conference.short_title) + '?[program][selected_schedule_id]='),
+      :ruby
+        if @schedule.track
+          value = @schedule == @schedule.track.selected_schedule
+          url = update_selected_schedule_admin_conference_program_track_path(@conference.short_title, @schedule.track) + '?[track][selected_schedule_id]='
+        else
+          value = @schedule.id == @selected_schedule.try(:id)
```
differentreality: Why do we need the id here?

app/views/tracks/index.html.haml
```diff
@@ -72,6 +72,8 @@
                       method: :patch, class: 'btn btn-mini btn-success', id: "resubmit_track_request_#{track.id}"
                   - if can? :edit, track
                     = link_to 'Edit', edit_conference_program_track_path(@conference.short_title, track), class: 'btn btn-default'
+                  - if current_user.has_role? :track_organizer, track
+                    = link_to 'Administer', admin_conference_program_track_path(@conference.short_title, track), class: 'btn btn-default'
```
differentreality: Maybe 'Manage' makes more sense in this context?

differentreality:
Very good work @AEtherC0r3 :tada:  glad you managed to finish scheduling so early! This looks very good :+1: 
Please rebase & squash!

app/views/admin/schedules/show.html.haml
AEtherC0r3: I don't know, I don't think that we need it. It's there because I just copy-pasted the expression from the existing code. In the expression I wrote (https://github.com/openSUSE/osem/pull/1631#discussion-diff-134726302R16) I don't use the id, but compare the two objects directly.
@Ana06 can you help as with this?

app/controllers/proposals_controller.rb
```diff
@@ -49,6 +49,13 @@ def create
     # by default.
     @event.speakers = [current_user]
     @event.submitter = current_user
+
+    if Track.find_by(id: params[:track_id]).try(:cfp_active) == false
```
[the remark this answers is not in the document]
AEtherC0r3: We want to load the track that corresponds to the track_id in the parameters hash. I don't think that cancancan can load it, because there is no track_id in the url for it to use and we don't want it to load the track that might be associated with the event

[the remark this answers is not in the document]
differentreality: That's true our params have [:event][:track_id], but then what you are writing in this line is wrong. Did you try if this is working on your browser?

app/models/admin_ability.rb
```diff
@@ -302,5 +298,27 @@ def signed_in_with_track_organizer_role(user)
     can :toggle_user, Role do |role|
       role.resource_type == 'Track' && track_ids_for_track_organizer.include?(role.resource_id)
     end
+
+    # Show Events in the admin sidebar
+    can :update, Event do |event|
+      event.new_record? && conf_ids_for_track_organizer.include?(event.program.conference_id)
```
differentreality: We should look into rewriting this. Can update a new record does not sound very elegant.

app/controllers/admin/tracks_controller.rb
```diff
@@ -101,6 +101,18 @@ def cancel
...
     end
 
+    def update_selected_schedule
+      if @track.update_attributes(params.require(:track).permit(:selected_schedule_id))
+        respond_to do |format|
+          format.js { render json: {} }
+        end
+      else
+        respond_to do |format|
+          format.js { render json: { errors: "The selected schedule couldn't been updated #{@track.errors.to_a.join('. ')}" }, status: 422 }
```
differentreality: couldn't be updated

app/controllers/proposals_controller.rb
```diff
@@ -49,6 +49,13 @@ def create
     # by default.
     @event.speakers = [current_user]
     @event.submitter = current_user
+
+    if Track.find_by(id: params[:event][:track_id]).try(:cfp_active) == false
```
differentreality: Let's remove `== false`

app/controllers/proposals_controller.rb
```diff
@@ -61,6 +68,12 @@ def create
   def update
     @url = conference_program_proposal_path(@conference.short_title, params[:id])
 
+    if Track.find_by(id: params[:event][:track_id]).try(:cfp_active) == false
```
differentreality: Let's remove `== false`

app/models/track.rb
```diff
@@ -209,4 +223,20 @@ def valid_room
       errors.add(:room, "must be a room of #{program.conference.venue.name}")
     end
   end
+
+  ##
+  # Check that there is no other track in the same room with overlapping dates
+  def overlapping
+    return unless start_date && end_date && room && program.try(:tracks)
+    (program.tracks.accepted + program.tracks.confirmed - [self]).each do |other_track|
```
differentreality: I would name the variable `existing_track` to make it more understandable

app/models/track.rb
```diff
@@ -209,4 +223,20 @@ def valid_room
...
+      if other_track.room == room &&
+         other_track.start_date && other_track.end_date &&
+         (other_track.start_date <= start_date && other_track.end_date >= start_date ||
```
differentreality: I'd keep the same order of variables for all checks, and preferably I'd put the new track dates in the beginning, so that we can can read like this 'the start date (of the new track) is more or equal than the existing track'

differentreality:
:tada Nicely done :+1: Let's change the comments in the refactoring PR!

AEtherC0r3:
The changes have been made in #1639

PR #1632, opened 2017-08-11
Add Booth to admin sidebar and confirm from admin/booths#index

nasia-sam:
I have added 'Booths' to admin sidebar that redirects at admin/booths#index. 
All of the elements in the sidebar have a fontawesome icon, however I couldn't find any kiosk-like and I didn't know which would be more appropriate so I let it blank.

Also, I made possible to confirm a booth from admin/booths#index when organizer is also the submitter.

### DIFF
--- a/app/views/layouts/_admin_sidebar.html.haml
+++ b/app/views/layouts/_admin_sidebar.html.haml
@@ -116,7 +116,7 @@
         - if can? :update, @conference.tickets.build
           %li{class: active_nav_li(admin_conference_tickets_path(@conference.short_title)) }
             = link_to 'Tickets', admin_conference_tickets_path(@conference.short_title)
-  - if can? :manage, @conference.booths
+  - if can? :manage, @conference.booths.build
     %li
       = link_to  admin_conference_booths_path(@conference.short_title) do
         %span.fa.fa-shopping-bag
--- a/app/views/layouts/_admin_sidebar.html.haml
+++ b/app/views/layouts/_admin_sidebar.html.haml
@@ -116,7 +116,11 @@
         - if can? :update, @conference.tickets.build
           %li{class: active_nav_li(admin_conference_tickets_path(@conference.short_title)) }
             = link_to 'Tickets', admin_conference_tickets_path(@conference.short_title)
-
+  - if can? :manage, @conference.booths
+    %li
+      = link_to  admin_conference_booths_path(@conference.short_title) do
+        %span.fa.fa-shopping-bag
+        Booths
   - if (can? :manage, @conference.targets.build) || (can? :manage, @conference.campaigns.build)
     %li
       %a
--- a/app/views/layouts/_user_menu.html.haml
+++ b/app/views/layouts/_user_menu.html.haml
@@ -16,6 +16,11 @@
     = link_to(conference_program_tracks_path(@conference.short_title)) do
       %span.fa.fa-road
       My Tracks
+-if @conference && @conference.program && (@conference.program.cfps.for_booths.try(:open?) || current_user.booths.where(conference_id: @conference.id).count > 0)
+  %li
+    = link_to (conference_booths_path(@conference.short_title)) do
+      %span.fa.fa-shopping-bag
+      My Booth Requests
 %li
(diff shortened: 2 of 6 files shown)


differentreality:
@nasia-sam There are still changes to be made here before we can merge this!

differentreality:
The changes of the PR include significant visual changes, please add a screenshot in the PR description.

differentreality:
What about you also add a relevant link in the user menu for the user's booth requests?

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -116,7 +116,11 @@
         - if can? :update, @conference.tickets.build
           %li{class: active_nav_li(admin_conference_tickets_path(@conference.short_title)) }
             = link_to 'Tickets', admin_conference_tickets_path(@conference.short_title)
-
+  - if can? :manage, @conference.booths
```
differentreality: This cannot work obviously, try this with a user that is **not** an admin, and see if this ever appears. 
The `can?` statement needs either a model, or a record. You are supplying it with a collection of objects (`@conference.booths`), and this cannot be evaluated to true. 

Go back to the documentation of CanCanCan and check the correct syntax. Where does it say you can supply a collection?

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -116,11 +116,15 @@
...
-  - if can? :manage, @conference.booths
+  - if can? :manage, @conference.booths.build
     %li
       = link_to  admin_conference_booths_path(@conference.short_title) do
         %span.fa.fa-shopping-bag
         Booths
+      %ul
+        - if current_user.booths.where(conference_id: @conference.id).count > 0
+          %li
+            = link_to 'Your Requests', conference_booths_path(@conference.short_title)
```
differentreality: Why are you adding this here?

nasia-sam: My mistake- I misread your previous comment. It's fixed now:)

app/views/layouts/_user_menu.html.haml
```diff
@@ -16,6 +16,11 @@
     = link_to(conference_program_tracks_path(@conference.short_title)) do
       %span.fa.fa-road
       My Tracks
+-if @conference && current_user.booths.where(conference_id: @conference.id).count > 0
```
differentreality: For proposals, we show the link even if the user doesn't have any submissions, so that the user can go to the index page and add a new proposal from there. Did you check this (how it is done in other places of the app)?
Which way do you think we should do it?

nasia-sam: Since in the rest of the app it is shown no matter the submissions, I could also change it

app/views/layouts/_user_menu.html.haml
```diff
@@ -16,6 +16,11 @@
     = link_to(conference_program_tracks_path(@conference.short_title)) do
       %span.fa.fa-road
       My Tracks
+-if @conference && @conference.program.cfps.for_booths.try(:open?)
```
differentreality: So if the cfp for booths is closed, I cannot see the link to the submissions I made while the cfp for booths was open. Why is that?

differentreality: This comment of mine seems to have hidden. Please let me know @nasia-sam

nasia-sam: It was a thoughtless mistake. In order to make it shown when there are no any requests, I forgot that it should also be shown after the cfb is over.

app/views/layouts/_user_menu.html.haml
```diff
@@ -16,6 +16,11 @@
     = link_to(conference_program_tracks_path(@conference.short_title)) do
       %span.fa.fa-road
       My Tracks
+-if @conference && @conference.program && (@conference.program.cfps.for_booths.try(:open?) || current_user.booths.where(conference_id: @conference.id).count > 0)
```
differentreality: @nasia-sam by the way, instead of `current_user.booths.where(conference_id: @conference.id).count > 0` you can simply do `current_user.booths.any?`

PR #1634, opened 2017-08-11
Registration tickets to be set before registration period

siddhantbajaj:
screen shot 2017-08-12 at 2 03 04 am

### DIFF
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -123,7 +123,10 @@ class AdminAbility
     can :manage, Commercial, commercialable_type: 'Conference',
                              commercialable_id: conf_ids
     can :manage, Registration, conference_id: conf_ids
-    can :manage, RegistrationPeriod, conference_id: conf_ids
+    can :manage, RegistrationPeriod do |registration_period|
+      conference = registration_period.conference
+      conf_ids.include?(conference.id) && conference.tickets.for_registration.any?
+    end
     can :manage, Booth, conference_id: conf_ids
     can :manage, Question, conference_id: conf_ids
     can :manage, Question do |question|
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -26,7 +26,11 @@ class Conference < ActiveRecord::Base
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
   has_many :supporters, through: :ticket_purchases, source: :user
-  has_many :tickets, dependent: :destroy
+  has_many :tickets, dependent: :destroy do
+    def for_registration
+      where(registration_ticket: true)
+    end
+  end
   has_many :resources, dependent: :destroy
   has_many :booths, dependent: :destroy
 
--- a/app/views/admin/registration_periods/show.html.haml
+++ b/app/views/admin/registration_periods/show.html.haml
@@ -25,5 +25,12 @@
         = link_to 'Delete', admin_conference_registration_period_path,
         method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger'
     - else
-      - if can? :create, @conference.build_registration_period
-        = link_to 'New Registration Period', new_admin_conference_registration_period_path, class: 'btn btn-primary'
+      - unless @conference.tickets.for_registration.empty?
+        - if can? :create, @conference.build_registration_period
+          = link_to 'New Registration Period', new_admin_conference_registration_period_path, class: 'btn btn-primary'
+      - else
+        .h3.text-left
+          No Registration Tickets!
+          %small
+            = link_to 'Create registration tickets', new_admin_conference_ticket_path
+            before creating the registration period.
--- a/spec/features/registration_periods_spec.rb
+++ b/spec/features/registration_periods_spec.rb
@@ -6,6 +6,7 @@ feature RegistrationPeriod do
   let!(:conference) { create(:conference) }
   let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let!(:organizer) { create(:user, email: 'admin@example.com', role_ids: [organizer_role.id]) }
+  let!(:registration_ticket) { create(:registration_ticket, conference: conference) }
 
   shared_examples 'successfully' do
     scenario 'create and update registration period', js: true do
--- a/spec/controllers/admin/registration_periods_controller_spec.rb
+++ b/spec/controllers/admin/registration_periods_controller_spec.rb
@@ -5,7 +5,7 @@ describe Admin::RegistrationPeriodsController do
   # It is necessary to use bang version of let to build roles before user
   let(:conference) { create(:conference) }
   let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
-
+  let!(:registration_ticket) { create(:registration_ticket, conference: conference) }
   let(:organizer) { create(:user, role_ids: organizer_role.id) }
   let(:organizer2) { create(:user, email: 'organizer2@email.osem', role_ids: organizer_role.id) }
   let(:participant) { create(:user) }
--- a/spec/factories/tickets.rb
+++ b/spec/factories/tickets.rb
@@ -3,5 +3,8 @@ FactoryGirl.define do
     title { "#{Faker::Hipster.word} Ticket" }
     price_cents 1000
     price_currency 'USD'
+    factory :registration_ticket do
+      registration_ticket true
+    end
   end
 end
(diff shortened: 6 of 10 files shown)


app/models/conference.rb
```diff
@@ -268,6 +268,15 @@ def registration_weeks
   end
 
   ##
+  # Returns registration tickets of the conference.
+  #
+  # ====Returns
+  # * +Array+ -> registation tickets
+  def registration_tickets
```
Ana06: this could be moved inside the [`tickets` association](https://github.com/siddhantbajaj/osem/blob/25bbe207cbd782977f7181738831b4ebfc1ef748/app/models/user.rb#L54). See the _Extensions_ section in the [Rails `has_many` documentation](https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_many)

app/views/admin/registration_periods/show.html.haml
```diff
@@ -26,4 +26,11 @@
         method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger'
     - else
       - if can? :create, @conference.build_registration_period
-        = link_to 'New Registration Period', new_admin_conference_registration_period_path, class: 'btn btn-primary'
+        - unless @conference.registration_tickets.empty?
+          = link_to 'New Registration Period', new_admin_conference_registration_period_path, class: 'btn btn-primary'
```
Ana06: Although you don't show the link, the path would be still accessible and the admin could create a registration period without registration tickets.

And another concern, what is the admin do not want to make the registration compulsory?

spec/features/registration_periods_spec.rb
```diff
@@ -6,6 +6,7 @@
   let!(:conference) { create(:conference) }
   let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
   let!(:organizer) { create(:user, email: 'admin@example.com', role_ids: [organizer_role.id]) }
+  let!(:ticket) { create(:ticket, conference: conference, registration_ticket: true) }
```
Ana06: Is this line enough to test this? :thinking:

app/models/conference.rb
siddhantbajaj: 👍

app/views/admin/registration_periods/show.html.haml
siddhantbajaj: > Although you don't show the link, the path would be still accessible and the admin could create a registration period without registration tickets.

Yes we can create a validation for this in registration period.

> And another concern, what is the admin do not want to make the registration compulsory?

I am assuming if admin wants to set the registration period then it means registration is compulsory for that conference. There might be a case when user does not specifically wants to attend the conference but just want to buy some other kind of ticket like food ticket or something. In this case user can purchase the food ticket but the user won't be registered for the conference and thus cannot be marked as present. So to summarise if admin does not want to make registration compulsory then no registration period is to be set. If admin wants to set registration period then that means registration is compulsory and there must be at least one registration ticket.

spec/features/registration_periods_spec.rb
siddhantbajaj: No this line was added to make these test cases pass.

Ana06:
@siddhantbajaj 

>> Although you don't show the link, the path would be still accessible and the admin could create a registration period without registration tickets.
>
>Yes we can create a validation for this in registration period.

I think you didn't understand what I mean. You should mange this not allowing accesing the route, you can not do it with a validation. I mean that the route is not accessible im the same case that you don't show the link in the view.

>> And another concern, what is the admin do not want to make the registration compulsory?
>
> I am assuming if admin wants to set the registration period then it means registration is compulsory for that conference. There might be a case when user does not specifically wants to attend the conference but just want to buy some other kind of ticket like food ticket or something. In this case user can purchase the food ticket but the user won't be registered for the conference and thus cannot be marked as present. So to summarise if admin does not want to make registration compulsory then no registration period is to be set. If admin wants to set registration period then that means registration is compulsory and there must be at least one registration ticket.

Ok, fine with this :wink:

app/models/conference.rb
```diff
@@ -26,7 +26,11 @@ class Conference < ActiveRecord::Base
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
   has_many :supporters, through: :ticket_purchases, source: :user
-  has_many :tickets, dependent: :destroy
+  has_many :tickets, dependent: :destroy do
+    def find_registration_tickets
```
Ana06: Why not just calling this something like `for_registration?` then it will be called `tickets.for_registration?`

siddhantbajaj: 👍

app/models/admin_ability.rb
```diff
@@ -118,7 +118,10 @@ def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
     can :manage, Commercial, commercialable_type: 'Conference',
                              commercialable_id: conf_ids
     can :manage, Registration, conference_id: conf_ids
-    can :manage, RegistrationPeriod, conference_id: conf_ids
+    can :manage, RegistrationPeriod do |registration_period|
+      conference = registration_period.conference
+      conf_ids.include?(conference.id) && !conference.tickets.for_registration.empty?
```
mdeniz: I prefer to use `.any?` instead of `! + .empty?`

siddhantbajaj: Ok, I'll change this. 😉

app/models/registration_period.rb
```diff
@@ -21,4 +22,9 @@ def start_date_before_end_date
     errors
     .add(:start_date, "can't be after the end date") if start_date && end_date && start_date > end_date
   end
+
+  def presence_of_registration_ticket
+    errors
```
Ana06: instead of splitting this after the errors I would say it is more readable writing it like:

``` Ruby
if conference && conference.tickets.for_registration.empty?
  errors.add(:registration_ticket, "is not created. There should be at least one registration ticket.") 
end
```

app/models/admin_ability.rb
```diff
@@ -118,7 +118,10 @@ def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
     can :manage, Commercial, commercialable_type: 'Conference',
                              commercialable_id: conf_ids
     can :manage, Registration, conference_id: conf_ids
-    can :manage, RegistrationPeriod, conference_id: conf_ids
+    can :manage, RegistrationPeriod do |registration_period|
+      conference = registration_period.conference
+      conf_ids.include?(conference.id) && conference.tickets.for_registration.any?
```
Ana06: This will fail when creating the RegistrationPeriod, but not when accesing the route, that was more my concern.

siddhantbajaj: @Ana06 
It will also fail while accessing the `new_admin_conference_registration_period_path` because organizer of the conference will only have the `new` ability (Included in `manage`) for `registration period` when conference has at least one registration ticket. So if there are no registration tickets the  organizer cannot access `new_admin_conference_registration_period_path` as the organizer will not be authorized to access that page.

siddhantbajaj:
@Ana06 
> `new_admin_conference_registration_period_path` shouldn't be accesible when the link is not shown, that means giving an error and redirecting

I think that's exactly how it is happening right now. If there are no registration tickets then organizer does not have the `new` ability for registration period and is therefore not able to access `new_admin_conference_registration_period_path`. A authorization error is thrown and the organizer is redirected to index page.

siddhantbajaj:
@Ana06 
> And please take a look at the tests, rspec is failing and Rubocop is unhappy

Tests were failing because of our validation `presence_of_registration_ticket`. There are two fix either remove the validation as it is anyway acting as a double check because we have already define the abilities such that registration period can not be created without having at least one registration tickets. Another fix could be adding `let!(:registration_ticket) { create(:ticket, conference: conference, registration_ticket: true) }` in all the failing test files so as to make the validation pass but it would require changes in lot of test files. For now I have removed the validation but I don't have a strong opinion here.

app/models/admin_ability.rb
Ana06: you are right, I skipped the authorization in the controller :+1:

Ana06:
@siddhantbajaj 

> For now I have removed the validation but I don't have a strong opinion here. I can also validation in different PR and handle tests there just to keep this PR clean.

I thing you implemented the other alternative, right? Although the validation shouldn't be remove, you have to fix the tests, as now they have to test something different. :wink:

mdeniz:
> Another fix could be adding `let!(:registration_ticket) { create(:ticket, conference: conference, registration_ticket: true) }` in all the failing test files so as to make the validation pass but it would require changes in lot of test files

Instead of adding `registration_ticket: true` to every place you need a `registration_ticket` better to create another factory that only sets that boolean and use it where you need it.

siddhantbajaj:
> Instead of adding registration_ticket: true to every place you need a registration_ticket better to create another factory that only sets that boolean and use it where you need it.

Right, I was thinking of just that. Ok then I will make the changes. 👍

PR #1635, opened 2017-08-11
One user one registration ticket

siddhantbajaj:
A user cannot have more than one registration ticket per conference.

### DIFF
--- a/app/controllers/ticket_purchases_controller.rb
+++ b/app/controllers/ticket_purchases_controller.rb
@@ -19,7 +19,7 @@ class TicketPurchasesController < ApplicationController
                     error: 'Please get at least one ticket to continue.'
       end
     else
-      redirect_to conference_conference_registration_path(@conference.short_title),
+      redirect_to conference_tickets_path(@conference.short_title),
                   error: "Oops, something went wrong with your purchase! #{message}"
     end
   end
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -5,7 +5,8 @@ class TicketPurchase < ActiveRecord::Base
   belongs_to :payment
 
   validates :ticket_id, :user_id, :conference_id, :quantity, presence: true
-
+  validate :one_registration_ticket_per_user
+  validate :registration_ticket_already_purchased, on: :create
   validates :quantity, numericality: { greater_than: 0 }
 
   delegate :title, to: :ticket
@@ -25,18 +26,21 @@ class TicketPurchase < ActiveRecord::Base
 
   def self.purchase(conference, user, purchases)
     errors = []
-    ActiveRecord::Base.transaction do
-      conference.tickets.each do |ticket|
-        quantity = purchases[ticket.id.to_s].to_i
-        # if the user bought the ticket and is still unpaid, just update the quantity
-        purchase = if ticket.bought?(user) && ticket.unpaid?(user)
-                     update_quantity(conference, quantity, ticket, user)
-                   else
-                     purchase_ticket(conference, quantity, ticket, user)
-                   end
-
-        if purchase && !purchase.save
-          errors.push(purchase.errors.full_messages)
+    if count_purchased_registration_tickets(conference, purchases) > 1
+      errors.push('You cannot buy more than one registration tickets.')
+    else
+      ActiveRecord::Base.transaction do
+        conference.tickets.each do |ticket|
+          quantity = purchases[ticket.id.to_s].to_i
+          # if the user bought the ticket and is still unpaid, just update the quantity
+          purchase = if ticket.bought?(user) && ticket.unpaid?(user)
+                       update_quantity(conference, quantity, ticket, user)
+                     else
+                       purchase_ticket(conference, quantity, ticket, user)
+                     end
+          if purchase && !purchase.save
+            errors.push(purchase.errors.full_messages)
+          end
         end
       end
     end
@@ -71,6 +75,18 @@ class TicketPurchase < ActiveRecord::Base
     end
     Mailbot.ticket_confirmation_mail(self).deliver_later
   end
+
+  def one_registration_ticket_per_user
+    if ticket.try(:registration_ticket?) && quantity != 1
+      errors.add(:quantity, 'cannot be greater than one for registration tickets.')
+    end
+  end
+
+  def registration_ticket_already_purchased
+    if ticket.try(:registration_ticket?) && user.tickets.for_registration(conference).present?
+      errors.add(:quantity, 'cannot be greater than one for registration tickets.')
+    end
+  end
 end
 
 private
@@ -78,4 +94,10 @@ private
 def set_week
   self.week = created_at.strftime('%W')
   save!
+end
+
+def count_purchased_registration_tickets(conference, purchases)
+  conference.tickets.for_registration.inject(0) do |sum, registration_ticket|
+    sum + purchases[registration_ticket.id.to_s].to_i
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -51,7 +51,11 @@ class User < ActiveRecord::Base
   has_many :events_registrations, through: :registrations
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
-  has_many :tickets, through: :ticket_purchases, source: :ticket
+  has_many :tickets, through: :ticket_purchases, source: :ticket do
+    def for_registration conference
+      where(conference: conference, registration_ticket: true).first
+    end
+  end
   has_many :votes, dependent: :destroy
   has_many :voted_events, through: :votes, source: :events
   has_many :subscriptions, dependent: :destroy
--- a/app/views/tickets/_ticket.html.haml
+++ b/app/views/tickets/_ticket.html.haml
@@ -1,5 +1,5 @@
 %tr
-  %td.col-sm-8.col-md-6
+  %td.col-sm-8.col-md-4
     .media
       .media-body
         %h4.media-heading
@@ -7,9 +7,14 @@
         %h5.media-heading
         - unless ticket.description.blank?
           = markdown(ticket.description)
+  %td.col-sm-1.col-md-2.text-center
+    = ticket.registration_ticket? ? 'Yes' : 'No'
   %td.col-sm-1.col-md-1
-    = text_field_tag("tickets[][#{ticket.id}]", 0,
-      type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id)
+    - options = { type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id }
+    - if ticket.registration_ticket?
+      - options[:max] = 1
+      - options[:disabled] = current_user.tickets.for_registration(ticket.conference).present?
+    = text_field_tag("tickets[][#{ticket.id}]", 0, options)
   %td.col-sm-1.col-md-1.text-center
     = ticket.price.symbol
     %span{id: "price_#{ticket.id}"}
--- a/spec/models/ticket_purchase_spec.rb
+++ b/spec/models/ticket_purchase_spec.rb
@@ -34,6 +34,22 @@ describe TicketPurchase do
     it 'is valid with a quantity greater than zero' do
       should allow_value(1).for(:quantity)
     end
+
+    describe 'one_registration_ticket_per_user' do
+      let(:registration_ticket) { create(:registration_ticket) }
+      let(:ticket_purchase) { build(:ticket_purchase, ticket: registration_ticket, quantity: 1) }
+
+      it 'it is valid, if quantity for registration tickets is less than or equal to one' do
+        expect(ticket_purchase.valid?).to eq true
+      end
+
+      it 'it is not valid, if quantity for registration tickets is greater than to one' do
+        ticket_purchase.quantity = 4
+
+        expect(ticket_purchase.valid?).to eq false
+        expect(ticket_purchase.errors[:quantity]).to eq ['cannot be greater than one for registration tickets.']
+      end
+    end
   end
 
   describe 'self#purchase' do
--- a/app/views/conference_registrations/show.html.haml
+++ b/app/views/conference_registrations/show.html.haml
@@ -109,7 +109,7 @@
           You haven't bought any tickets.
           = link_to 'Please get some tickets to support us!', conference_tickets_path(@conference.short_title)
           %p
-            (Your participation won't be valid without getting a ticket)
+            (Your participation won't be valid without getting a registration ticket)
 
   .row
     .col-md-12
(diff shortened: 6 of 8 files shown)


app/models/user.rb
```diff
@@ -179,6 +179,15 @@ def get_roles
     result
   end
 
+  ##
+  # Returns registration tickets of the user for the given conference.
+  #
+  # ====Returns
+  # * +Array+ -> registation tickets
+  def registration_tickets conference
```
Ana06: this could be moved inside the [`tickets` association](https://github.com/siddhantbajaj/osem/blob/25bbe207cbd782977f7181738831b4ebfc1ef748/app/models/user.rb#L54). See the _Extensions_ section in the [Rails `has_many` documentation](https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_many)

app/views/tickets/_ticket.html.haml
```diff
@@ -8,8 +8,12 @@
         - unless ticket.description.blank?
           = markdown(ticket.description)
   %td.col-sm-1.col-md-1
-    = text_field_tag("tickets[][#{ticket.id}]", 0,
-      type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id)
+    - if ticket.registration_ticket?
+      = text_field_tag("tickets[][#{ticket.id}]", 0,
+        type: 'number', min: 0, max: 1, class: "form-control quantity", 'data-id' => ticket.id)
```
Ana06: couldn't you introduce the if directly in the max or setting the params? Most of the code here is duplicated. :thinking:

app/controllers/tickets_controller.rb
```diff
@@ -5,7 +5,11 @@ class TicketsController < ApplicationController
   authorize_resource :conference_registrations, class: Registration
   before_filter :check_load_resource, only: :index
 
-  def index; end
+  def index
```
Ana06: I am not sure if I understand what you want to do here.

This index page is shown both when buying registration tickets and when buying other kind of thicket. If you remove the registration tickets from here, where are they being rendered? :confused:

siddhantbajaj: So basically we need to avoid 2 cases here:

1. User buying more than 1 registration ticket in a single purchase. This case is avoided by setting max limit on registration tickets.

2. User buying 1 registration tickets in multiple ticket purchases so actually having more than one registration tickets. To avoid this we check if the user already has a registration ticket then we don't render registration ticket on the index page thus not giving the ability to the user to buy any more registration ticket.

siddhantbajaj:
Dependent on #1634. So closing it for now.

siddhantbajaj:
Cases Handled:

1.  **User buys more than one registration ticket of same type in a single purchase:** To handle this I have  set the max attribute to text_field_tags of registration tickets.
2.  **User buys exactly one registration ticket of different type in a single purchase:** There can be multiple registration tickets in a conference so user can easily buy exactly one of each type of registration ticket in a single purchase. To handle this before creating ticket purchase I am checking if the number of registration tickets bought in that purchase is more than one then we simply redirect the user with an error message.
3.  **User buys one registration ticket of same type in a different purchase:** To handle this if the user has already bought a registration ticket then we don't render registration tickets on the index page thus not allowing the user to buy any more registration ticket of any type.

By **type** here, I means different registration tickets available in a conference.

app/views/tickets/_ticket.html.haml
```diff
@@ -8,8 +8,10 @@
         - unless ticket.description.blank?
           = markdown(ticket.description)
   %td.col-sm-1.col-md-1
-    = text_field_tag("tickets[][#{ticket.id}]", 0,
-      type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id)
+    - options = { type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id }
+    - if ticket.registration_ticket?
```
Ana06: :art:  you could write this two lines together: 

`- options[:max] = 1 if ticket.registration_ticket?`

Ana06:
@siddhantbajaj I have a question, how does the user now what are the registration tickets? Is this indicated somehow in the view?

app/controllers/ticket_purchases_controller.rb
```diff
@@ -30,6 +31,18 @@ def index
 
   private
 
+  def count_purchased_registration_tickets
+    purchases = params[:tickets][0]
+    purchased_registration_tickets = 0
+    @conference.tickets.for_registration.each do |registration_ticket|
+      purchased_registration_tickets += purchases[registration_ticket.id.to_s].to_i
+    end
```
Ana06: I find a little bit complicated to read it. What about writing it like this:

``` Ruby
purchased_tickets = params[:tickets][0]
tickets = Ticket.find(purchased_tickets).where(registration_ticket: true)
if tickets.count > 1 || purchases[tickets.first.id] != '1'
  redirect_to conference_tickets_path(@conference.short_title),
                  error: 'You cannot buy more than one registration tickets.'
end
```

app/views/tickets/_ticket.html.haml
siddhantbajaj: 👍

app/controllers/ticket_purchases_controller.rb
siddhantbajaj: `Ticket.find(purchased_tickets).where(registration_ticket: true)`
If I am right here you are trying to find the purchased registration tickets but the problem is these tickets are not created yet. here purchased_tickets will be hash containing ticket id as key and quantity as the key (bought in that single purchase).

siddhantbajaj:
> @siddhantbajaj I have a question, how does the user now what are the registration tickets? Is this indicated somehow in the view?

Yes I missed that, registration tickets should be indicated in the views. 😉

app/controllers/ticket_purchases_controller.rb
Ana06: no, I was trying to look for the tickets using the ids from `purchased_tickets`, so `purchased_tickets.keys`, sorry:

``` Ruby
purchased_tickets = params[:tickets][0]
tickets = Ticket.find(purchased_tickets.keys).where(registration_ticket: true)
if tickets.count > 1 || purchases[tickets.first.id] != '1'
  redirect_to conference_tickets_path(@conference.short_title),
                  error: 'You cannot buy more than one registration tickets.'
end
```

siddhantbajaj: Okay so the problem is suppose there are more than 1 registration tickets in the conference and there there  id's are suppose 1 and 2 so purchase tickets would be something like this:
`{'1':'quantity', '2': 'quantity'}`
quantity can be 0 as well. So basically if there are more than 1 registration tickets then I suppose `tickets` will always be greater than 1 even if user doesn't buy any registration ticket.

siddhantbajaj:
screen shot 2017-08-23 at 6 17 52 pm

app/controllers/tickets_controller.rb
Ana06: I would show them instead, but disabling them. Because this could be a little confusing for the user. But I am fine with doing this in a different PR. :+1:

app/controllers/ticket_purchases_controller.rb
```diff
@@ -30,6 +31,18 @@ def index
 
   private
 
+  def count_purchased_registration_tickets
```
mdeniz: Rather than putting this in the controller I think it belongs to the `TicketPurchase.purchase` method, the TicketPurchase model should be the responsible of saying if a purchase is valid or not despite the reason... so, in the controller it will be just one condition more to know what happened, but all the checks should live in the model.

app/controllers/tickets_controller.rb
mdeniz: Rather than hiding the tickets I would prefer to "deactivate" the links... but showing them.

app/controllers/ticket_purchases_controller.rb
siddhantbajaj: 👍

app/controllers/tickets_controller.rb
siddhantbajaj: @mdeniz Yes absolutely a much better fix. 
@Ana06  lets do this in this PR only not many changes required. 😉

app/views/tickets/_ticket.html.haml
siddhantbajaj: Added some more conditions in this if condition therefore going back to previous implementation.

siddhantbajaj:
> @siddhantbajaj tests are missed right? and there are tests failing!!!!!

Added test. 👍

app/models/ticket_purchase.rb
```diff
@@ -71,6 +74,12 @@ def pay(payment)
     end
     Mailbot.ticket_confirmation_mail(self).deliver_later
   end
+
+  def one_registration_ticket_per_user
+    if ticket && ticket.registration_ticket? && quantity > 1
```
Ana06: this can also be written like:

```` Ruby
ticket.try(:registration_ticket?) && quantity > 1

File: app/models/ticket_purchase.rb [hunk 8]
```diff
@@ -25,18 +25,21 @@ class TicketPurchase < ActiveRecord::Base
...
-        # if the user bought the ticket and is still unpaid, just update the quantity
-        purchase = if ticket.bought?(user) && ticket.unpaid?(user)
-                     update_quantity(conference, quantity, ticket, user)
-                   else
-                     purchase_ticket(conference, quantity, ticket, user)
-                   end
-
-        if purchase && !purchase.save
-          errors.push(purchase.errors.full_messages)
+    if count_purchased_registration_tickets(conference, purchases) > 1
```
Ana06: why is not this in a validation?

app/models/user.rb
```diff
@@ -51,7 +51,11 @@ def by_conference(conference)
   has_many :events_registrations, through: :registrations
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
-  has_many :tickets, through: :ticket_purchases, source: :ticket
+  has_many :tickets, through: :ticket_purchases, source: :ticket do
+    def for_registration conference
+      where(conference: conference, registration_ticket: true)
```
Ana06: this should return only one record

app/views/tickets/_ticket.html.haml
```diff
@@ -1,15 +1,21 @@
...
           = markdown(ticket.description)
+  %td.col-sm-1.col-md-2.text-center
+    = ticket.registration_ticket? ? 'Yes' : 'No'
   %td.col-sm-1.col-md-1
-    = text_field_tag("tickets[][#{ticket.id}]", 0,
-      type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id)
+    - options = { type: 'number', min: 0, class: "form-control quantity", 'data-id' => ticket.id }
+    - if ticket.registration_ticket?
+      - options[:max] = 1
+      - if current_user.tickets.for_registration(@conference).any?
```
Ana06: this two lines can be written like

``` Ruby
options[:disabled] = current_user.tickets.for_registration(@conference).any?

File: app/models/ticket_purchase.rb (hunk 8 above)
Comment (siddhantbajaj) [reply app/models/ticket_purchase.rb]: We can not do this in validation because of the way ticket purchase feature has been implemented. Suppose a user try to buy 1 registration ticket of type A and 1 registration ticket of type B in a single purchase. Now 2 ticket purchase object will be created one for type A ticket and other for type B ticket with quantity set as one for both objects. Now looking at the at each object we cannot say it is an invalid ticket purchase object until we know the other ticket purchase objects created alongside it in a single purchase. Therefore we cannot check this in validation level because we need the whole `purchases` object.

Comment (Ana06) [reply app/models/ticket_purchase.rb]: but you can also add a validation to check that there is not another purchase_ticket with registration ticket, right?

File: app/models/ticket_purchase.rb [hunk 11]
```diff
@@ -71,6 +74,12 @@ def pay(payment)
     end
     Mailbot.ticket_confirmation_mail(self).deliver_later
   end
+
+  def one_registration_ticket_per_user
+    if ticket.try(:registration_ticket?) && quantity > 1
+      errors.add(:quantity, 'cannot be greater than one for registration tickets.')
```
Ana06: :art: maybe _can only be one_ instead of _greater than one_

spec/models/ticket_purchase_spec.rb
```diff
@@ -34,6 +34,23 @@
...
       should allow_value(1).for(:quantity)
     end
+
+    describe 'one_registration_ticket_per_user' do
+      let(:registration_ticket) { create(:registration_ticket) }
+      let(:ticket_purchase) { build(:ticket_purchase, ticket: registration_ticket, quantity: 1) }
+
+      it 'it is valid, if quantity for registration tickets is less than or equal to one' do
+        expect(ticket_purchase.valid?).to eq true
+        expect(ticket_purchase.errors.full_messages).to eq []
```
Ana06: I don't think this is needed, if it is valid there shouldn't be any errors.

siddhantbajaj: 👍

app/models/ticket_purchase.rb
siddhantbajaj: 👍

app/models/ticket_purchase.rb
siddhantbajaj: Yes. 👍

app/models/ticket_purchase.rb
```diff
@@ -25,18 +25,21 @@ class TicketPurchase < ActiveRecord::Base
...
+      ActiveRecord::Base.transaction do
+        conference.tickets.each do |ticket|
+          quantity = purchases[ticket.id.to_s].to_i
+          # if the user bought the ticket and is still unpaid, just update the quantity
+          purchase = if ticket.bought?(user) && ticket.unpaid?(user)
+                       update_quantity(conference, quantity, ticket, user)
+                     else
+                       purchase_ticket(conference, quantity, ticket, user)
+                     end
+          if purchase && !purchase.persisted? && !purchase.save
```
siddhantbajaj: `!purchase.persisted?` is added because there is a small issue with the way ticket purchase was implemented.
We expect `purchase_ticket` method to return a purchase object with attributes set but the object should not be saved. Well, this works fine for paid tickets but for unpaid tickets we call `pay` method inside `purchase_ticket` method. Inside `pay` method, we use `update_attributes` due to which an already saved purchase object is returned from `purchase_ticket` method. Now this was working fine until now as there is no issue in saving an already saved object. But now with addition of validation `one_registration_ticket_per_user` this code fail in the following case:
**suppose user buys a single registration ticket**, now the first time purchase object is saved inside `pay` method, it will work fine no validation would fail. But the next time the same purchase object is saved here `!purchase.save`, our `one_registration_ticket_per_user` will fail because this time user already has a registration ticket. Though they both represent the same registration ticket.

siddhantbajaj: Possible fix:

1. `!purchase.persisted?` so that we don't save an already saved purchase object.

2. Call `pay` method for free ticket somewhere else, outside `purchase_ticket`  method.

PR #1637, opened 2017-08-12
Mention organization name while creating a conference

### LINKED ISSUE #1524: Ability to select organizations while creating a conference
Reference #1637

shlok007:
Closes #1524 

screenshot_20170813_022752


app/controllers/admin/conferences_controller.rb
```diff
@@ -72,11 +72,15 @@ def index
 
     def new
       @conference = Conference.new
+      @organizations = {}
+      Organization.find_each do |organization|
```
lagartoflojo: I'm not very happy about this, because it will go through EVERY organization in the DB.
Can we do it somehow faster?

shlok007: How about this?
```ruby
Organization.accessible_by(current_ability, :update).map { |organization| @organizations.store(organization.name, organization.id) }
```

lagartoflojo: yeah that looks much better!
I'm just wondering, why do you use `Hash#store` instead of `Hash#[]=`?
It's not common practice, so it looks kinda funky :P

lagartoflojo: Would it be possible to say even:

```ruby
Organization.accessible_by(current_ability, :update).pluck(:name, :id)
```

shlok007: That looks a lot better. Thanks! :sweat_smile:

shlok007:
@lagartoflojo Please do have a look!

PR #1638, opened 2017-08-17
Added QR code

siddhantbajaj:
Added qr code on ticket pdf and ticket show page.

### DIFF
--- a/app/controllers/physical_ticket_controller.rb
+++ b/app/controllers/physical_ticket_controller.rb
@@ -13,6 +13,7 @@ class PhysicalTicketController < ApplicationController
     @file_name = "ticket_for_#{@conference.short_title}"
     @user = @physical_ticket.user
     @ticket_layout = @conference.ticket_layout.to_sym
+    @qrcode_image = RQRCode::QRCode.new(@physical_ticket.token).as_png(size: 180, border_modules: 0)
     respond_to do |format|
       format.html
       format.pdf do
--- a/app/pdfs/ticket_pdf.rb
+++ b/app/pdfs/ticket_pdf.rb
@@ -77,5 +77,9 @@ class TicketPdf < Prawn::Document
     move_up 180
   end
 
-  def draw_fourth_square; end
+  def draw_fourth_square
+    x = @mid_horizontal + (@right - @mid_horizontal - 180) / 2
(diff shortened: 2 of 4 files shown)


siddhantbajaj:
Screenshots:

screen shot 2017-08-17 at 12 13 04 pm

screen shot 2017-08-17 at 12 14 36 pm

app/controllers/physical_ticket_controller.rb
```diff
@@ -13,6 +13,8 @@ def show
     @file_name = "ticket_for_#{@conference.short_title}"
     @user = @physical_ticket.user
     @ticket_layout = @conference.ticket_layout.to_sym
+    @qrcode = RQRCode::QRCode.new(@physical_ticket.token)
```
mdeniz: Is this instance variable needed? Is not consumed by any of the views, isn't it?
I prefer this to be a simple local variable (or just replace this code in the next line because isn't used anymore).

app/controllers/physical_ticket_controller.rb
```diff
@@ -13,6 +13,8 @@ def show
...
+    @png = @qrcode.as_png(size: 180, border_modules: 0)
```
mdeniz: Use a better name for this variable, I mean something like `@qrcode_image` (we shouldn't care about the png format in the view, is just an image of the QRCode)

app/pdfs/ticket_pdf.rb
```diff
@@ -77,5 +77,9 @@ def draw_third_square
     move_up 180
   end
 
-  def draw_fourth_square; end
+  def draw_fourth_square
+    x1 = @right - @mid_horizontal - 180
+    y1 = bounds.top - @mid_vertical - 180
+    print_qr_code(@physical_ticket.token, pos: [@mid_horizontal + x1 / 2, cursor - y1 / 2], extent: 180, stroke: false)
```
mdeniz: If you calculate the x and the y before for clarifying the code... why not something like this?

```ruby
x = @mid_horizontal + (@right - @mid_horizontal - 180) / 2
y = cursor - (bounds.top - @mid_vertical - 180) / 2
print_qr_code(@physical_ticket.token, pos: [x, y], extent: 180, stroke: false)
```

PR #1639, opened 2017-08-17
Track related refactoring

AEtherC0r3:
* Add roles as nested routes to track (for the track organizer role)
* Allow transition from to_accept to to_reject and backwards
* Split Track#valid_dates validation to many independent ones
* Show all the confirmed tracks in the conference's splashpage
* Add comment in admin/Tracks#toggle_cfp_inclusion
* Rewrite admin/TracksController#accept spec
* Add feature spec for track requests
* Change 'In' to 'Room' in Tracks#index
* Rewrite Track#overlapping
* Refactor code in ProposalsController
* Fix typos

### DIFF
--- a/app/models/track.rb
+++ b/app/models/track.rb
@@ -31,8 +31,9 @@ class Track < ActiveRecord::Base
   validates :room, presence: true, if: :self_organized_and_accepted_or_confirmed?
   validates :relevance, presence: true, if: :self_organized?
   validates :description, presence: true, if: :self_organized?
-  validate :valid_dates
-  validate :valid_room, if: :self_organized_and_accepted_or_confirmed?
+  validate :dates_within_conference_dates
+  validate :start_date_before_end_date
+  validate :valid_room
   validate :overlapping
 
   before_validation :capitalize_color
@@ -56,7 +57,7 @@ class Track < ActiveRecord::Base
       transitions to: :new, from: [:rejected, :withdrawn, :canceled]
     end
     event :to_accept do
-      transitions to: :to_accept, from: [:new]
+      transitions to: :to_accept, from: [:new, :to_reject]
     end
     event :accept do
       transitions to: :accepted, from: [:new, :to_accept], on_transition: :create_organizer_role
@@ -65,7 +66,7 @@ class Track < ActiveRecord::Base
       transitions to: :confirmed, from: [:accepted], on_transition: :assign_role_to_submitter
     end
     event :to_reject do
-      transitions to: :to_reject, from: [:new]
+      transitions to: :to_reject, from: [:new, :to_accept]
     end
     event :reject do
       transitions to: :rejected, from: [:new, :to_reject]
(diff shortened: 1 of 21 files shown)


spec/controllers/admin/tracks_controller_spec.rb
```diff
@@ -383,12 +376,16 @@
...
+      @room = create(:room, venue: conference.venue)
+    end
+
     context 'has start_date, end_date and room' do
       before :each do
-        self_organized_track.start_date = Date.today
-        self_organized_track.end_date = Date.today
-        conference.venue = create(:venue)
-        self_organized_track.room = create(:room, venue: conference.venue)
+        self_organized_track.start_date = Date.current
```
differentreality: For this case you can simply define a variable, using let, named full_self_organized_track or something similar, that indicates that this is a self_organized track with all of its attributes.

spec/controllers/admin/tracks_controller_spec.rb
```diff
@@ -409,31 +406,80 @@
     end
 
     context 'has start_date and end_date' do
-      it_behaves_like 'fails to accept', true, true, false
+      before :each do
+        self_organized_track.start_date = Date.current
+        self_organized_track.end_date = Date.current
```
differentreality: Using the predefined track with all attributes, you can be simply removing the ones you do **not** want, which is going to reduce the lines of our tests (that's good because we have less to read and understand).

spec/controllers/admin/tracks_controller_spec.rb
```diff
@@ -409,31 +406,80 @@
...
+      end
+
+      it_behaves_like 'fails to accept'
     end
 
-    context 'has non of start_date, end_date, room' do
-      it_behaves_like 'fails to accept', false, false, false
+    context 'has none of start_date, end_date, room' do
+      before :each do
+        self_organized_track.start_date = nil
```
differentreality: We can have another variable for a track that has none of those extra attributes (dates, room) set. The same variable can be used in other cases, where you only want to set one of the attribute to non nil.

spec/features/tracks_spec.rb
```diff
@@ -60,12 +61,83 @@
       within('table#tracks') do
         expect(page.has_content?('Distribution')).to be true
         expect(page.has_content?('Events about our Linux')).to be true
-        expect(page.assert_selector('tr', count: 2)).to be true
+        expect(Track.count).to eq 1
+      end
+    end
+  end
+
+  shared_examples 'tracks' do
```
differentreality: These shared examples are all about managing / working with tracks, right?
Perhaps the name could be a little bit more specific?

app/models/track.rb
```diff
@@ -29,8 +29,12 @@ class Track < ActiveRecord::Base
   validates :room, presence: true, if: :self_organized_and_accepted_or_confirmed?
   validates :relevance, presence: true, if: :self_organized?
   validates :description, presence: true, if: :self_organized?
-  validate :valid_dates
-  validate :valid_room, if: :self_organized_and_accepted_or_confirmed?
+  validate :start_date_after_conference_start_date
```
differentreality: I'd combine the test related to the conference date into 1, dates_within_conference_dates or something similar, and I would test if start_date is within the range of conf.start_date..conf.end_date

As long as  you have the validation start_date_before_end_date, that should be sufficient. What do you think?

spec/models/track_spec.rb
```diff
@@ -84,25 +84,85 @@
...
+      end
+    end
+
+    describe '#start_date_before_end_date' do
+      before :each do
+        @conference = create(:conference, start_date: 1.day.ago, end_date: 2.days.from_now)
+      end
 
+      context 'is valid' do
+        it 'when the track\'s start date is before its end date' do
```
differentreality: This can probably be combined with https://github.com/openSUSE/osem/pull/1639/files#diff-c2212054e3536a54e738fecea403adffR117

so that we are testing track dates within conference dates (I would even have it start on conf.start_date and end on conf.end_date which is our edge case).

And that would be enough for our tests, since we will test the start_date before end_date in the other validation.

differentreality:
I'd also transform tracks#index to show table headers for 'From', 'To', 'In'

AEtherC0r3:
@differentreality I'm having trouble refactoring the refactored admin/TracksController spec. The variable I define (@this_track) isn't available in the `it_behaves_like` directive, to be passed as a variable. Also, I can't pass a track defined in a let to it. I'm including the commit for reference below. In my opinion, we should ditch the second track and, this way, get rid of the need to have a variable. This is a hybrid solution between my initially refactored code and what you proposed in the comments. What do you think?

app/views/tracks/index.html.haml
```diff
@@ -29,6 +29,13 @@
     .row
       .col-md-12
         %table.table.table-striped#tracks
+          %th
+          %th
+          %th
+          %th From
+          %th To
+          %th In
```
differentreality: Perhaps that's more meaningful if it says 'Room' or 'In room', what do you think?

AEtherC0r3: I think that 'Room' is enough

differentreality:
Nice work @AEtherC0r3 :+1:

PR #1642, opened 2017-08-22
Improving check-in process

siddhantbajaj:
Mark user as present for the conference when user's registration ticket for that conference is scanned

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -47,7 +47,11 @@ class User < ActiveRecord::Base
   has_many :event_users, dependent: :destroy
   has_many :events, -> { uniq }, through: :event_users
   has_many :presented_events, -> { joins(:event_users).where(event_users: {event_role: 'speaker'}).uniq }, through: :event_users, source: :event
-  has_many :registrations, dependent: :destroy
+  has_many :registrations, dependent: :destroy do
+    def for_conference conference
+      where(conference: conference).first
+    end
+  end
   has_many :events_registrations, through: :registrations
   has_many :ticket_purchases, dependent: :destroy
   has_many :payments, dependent: :destroy
@@ -91,6 +95,12 @@ class User < ActiveRecord::Base
 
     return false unless event_registration.present?
     event_registration.attended
+  end
+
+  def mark_attendance_for_conference conference
+    registration = registrations.for_conference(conference)
+    registration.attended = true
+    registration.save
   end
 
   def name
--- a/spec/controllers/admin/ticket_scannings_controller_spec.rb
+++ b/spec/controllers/admin/ticket_scannings_controller_spec.rb
@@ -4,7 +4,9 @@ describe Admin::TicketScanningsController do
   let(:admin) { create(:admin) }
   let(:conference) { create(:conference) }
   let(:user) { create(:user) }
-  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user) }
+  let!(:registration) { create(:registration, conference: conference, user: user) }
+  let(:registration_ticket) { create(:registration_ticket, conference: conference) }
+  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user, ticket: registration_ticket, quantity: 1) }
   let(:physical_ticket) { create(:physical_ticket, ticket_purchase: paid_ticket_purchase) }
 
   context 'logged in as user with no role' do
--- a/spec/models/ticket_scanning_spec.rb
+++ b/spec/models/ticket_scanning_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+
+describe TicketScanning do
+  let(:conference) { create(:conference) }
+  let(:user) { create(:user) }
+  let(:registration) { create(:registration, conference: conference, user: user) }
+  let(:registration_ticket) { create(:registration_ticket, conference: conference) }
+  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user, ticket: registration_ticket, quantity: 1) }
+  let(:physical_ticket) { create(:physical_ticket, ticket_purchase: paid_ticket_purchase) }
+  let(:ticket_scanning) { create(:ticket_scanning, physical_ticket: physical_ticket) }
+
+  describe 'before_create' do
+    it 'marks user as present' do
+      expect(registration.attended).to eq(false)
+      ticket_scanning
+      registration.reload
+      expect(registration.attended).to eq(true)
+    end
+  end
+end
--- a/app/models/ticket_scanning.rb
+++ b/app/models/ticket_scanning.rb
@@ -1,3 +1,13 @@
 class TicketScanning < ActiveRecord::Base
   belongs_to :physical_ticket
+
+  before_create :mark_user_present
+
+  private
+
+  def mark_user_present
+    if physical_ticket.ticket.registration_ticket?
+      physical_ticket.user.mark_attendance_for_conference(physical_ticket.conference)
+    end
+  end
 end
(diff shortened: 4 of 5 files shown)


app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -8,9 +8,14 @@ class TicketScanningsController < Admin::BaseController
     def create
       @ticket_scanning = TicketScanning.new(physical_ticket: @physical_ticket)
       authorize! :create, @ticket_scanning
-      @ticket_scanning.save
-      redirect_to conferences_path,
-                  notice: "Ticket with token #{@physical_ticket.token} successfully scanned."
+      if @physical_ticket.ticket.registration_ticket?
```
Ana06: what about doing this in a before create filter in TicketScanning?

app/models/user.rb
```diff
@@ -47,7 +47,11 @@ def by_conference(conference)
   has_many :event_users, dependent: :destroy
   has_many :events, -> { uniq }, through: :event_users
   has_many :presented_events, -> { joins(:event_users).where(event_users: {event_role: 'speaker'}).uniq }, through: :event_users, source: :event
-  has_many :registrations, dependent: :destroy
+  has_many :registrations, dependent: :destroy do
+    def for_conference conference
+      where(conference: conference)
```
Ana06: a user should have more that one registration for a conference right? So may make sense to only return one registration here

spec/controllers/admin/ticket_scannings_controller_spec.rb
```diff
@@ -4,7 +4,9 @@
   let(:admin) { create(:admin) }
   let(:conference) { create(:conference) }
   let(:user) { create(:user) }
-  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user) }
+  let!(:registration) { create(:registration, conference: conference, user: user) }
```
Ana06: is this testing this? I would say we miss tests here :thinking:

app/models/physical_ticket.rb
```diff
@@ -7,6 +7,12 @@ class PhysicalTicket < ActiveRecord::Base
 
   before_create :set_token
 
+  def mark_attendance
```
mdeniz: I think that this method should belong to the User class no to this one. I mean, is changing a registration of a user, that's user's business. You can move this code there and delegate this method here to `user.mark_attendance_for(conference)`

app/models/user.rb
mdeniz: You mean, that only can have one, right? that's why returning only one registration.

spec/controllers/admin/ticket_scannings_controller_spec.rb
mdeniz: It needs tests, sure.

app/controllers/admin/ticket_scannings_controller.rb
```diff
@@ -8,9 +8,14 @@ class TicketScanningsController < Admin::BaseController
...
+        @physical_ticket.mark_attendance
+        @ticket_scanning.save
+        redirect_to conferences_path,
+                    notice: "Ticket with token #{@physical_ticket.token} successfully scanned."
+      else
+        redirect_to conferences_path, error: 'Ticket scanned is not a registration type ticket.'
```
mdeniz: Why is this an error? I mean, you are not only going to scan registration tickets, right? It could be another kind of ticket, and scanning it could mean "consuming" it.

app/models/user.rb
Ana06: exactly

Ana06:
@siddhantbajaj ping!

app/controllers/admin/ticket_scannings_controller.rb
siddhantbajaj: True, but since we don't have any feature for consuming a ticket right now therefore I'm removing this `else` condition for now.

app/controllers/admin/ticket_scannings_controller.rb
siddhantbajaj: 👍

app/models/physical_ticket.rb
siddhantbajaj: :+1

app/models/user.rb
siddhantbajaj: 👍

spec/controllers/admin/ticket_scannings_controller_spec.rb
siddhantbajaj: done.

spec/models/ticket_scanning_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:paid_ticket_purchase) { create(:ticket_purchase, conference: conference, user: user, ticket: registration_ticket) }
+  let(:physical_ticket) { create(:physical_ticket, ticket_purchase: paid_ticket_purchase) }
+  let(:ticket_scanning) { create(:ticket_scanning, physical_ticket: physical_ticket) }
+
+  describe 'before_create' do
+    it 'marks user as present' do
+      expect(registration.attended).to eq(false)
+      ticket_scanning
+      registration.reload
+      expect(registration.attended).to eq(true)
```
Ana06: you should have two expect in the same test, split the test :wink: 

http://www.betterspecs.org/#single

spec/models/ticket_scanning_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+describe TicketScanning do
+  let(:conference) { create(:conference) }
```
Ana06: do you really need all this variables? shouldn't some of them be created by other factories? If not, we may want to modify the factories. I mean just creating what we need the factories can create the rest. :thinking:  At least we should remove the unneeded ones, if any. Improving the factories could be done in another PR.

spec/controllers/admin/ticket_scannings_controller_spec.rb
Ana06: @siddhantbajaj why are this new let exactly needed here?

spec/models/ticket_scanning_spec.rb
siddhantbajaj: Nice resource. 😉 
But here both expect are for the same behaviour so they must be present in the same test.

spec/controllers/admin/ticket_scannings_controller_spec.rb
siddhantbajaj: To test the create action of TicketScanning controller, user must registered for the conference. But yes ideally it should be handled in ticket_purchase factory because user must registered for the conference before purchasing its tickets. I'll open a different PR for modifying factories. 👍

spec/models/ticket_scanning_spec.rb
siddhantbajaj: Yes I guess right now we don need all these variables. But yes some of them can be removed by modifying some factories. 👍

spec/controllers/admin/ticket_scannings_controller_spec.rb
Ana06: here would be better, but ok! :+1:

spec/models/ticket_scanning_spec.rb
Ana06: :+1:

spec/models/ticket_scanning_spec.rb
Ana06: They are not, one is to test that by default the user attendance is false, and the other to check that is set to true afterwards.  And actually the false part is a requirement, more than something that should be tested. Anyway, let's merge this and we can improve the test afterwards. It is at least tested.

Ana06:
@siddhantbajaj the test are failing. Please when failing them, rebase and remove my two Merge branch 'master' into user-present commits :wink:

PR #1644, opened 2017-08-22
Rename "size" field to "capacity" in Room 

### LINKED ISSUE #1625: Rename "size" field to "capacity" in Room
Venue rooms have a size field which seems to be the number of attendees it can accommodate. This should be named capacity as size is an ambiguous term and probably not the first one would think of for that.

rahul2240:
Renamed "size" field to "capacity" in room
fixes #1625

### DIFF
--- a/app/views/admin/rooms/_form.html.haml
+++ b/app/views/admin/rooms/_form.html.haml
@@ -10,6 +10,6 @@
   .col-md-8
     = semantic_form_for(@room, url: (@room.new_record? ? admin_conference_venue_rooms_path : admin_conference_venue_room_path(@conference.short_title, @room))) do |f|
       = f.input :name, input_html: { autofocus: true}
-      = f.input :size, input_html: {size: 5}
+      = f.input :size, label: 'Capacity', input_html: {size: 5}
       %p.text-right
         = f.action :submit, as: :button, button_html: { class: 'btn btn-primary' }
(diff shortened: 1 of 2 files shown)


app/views/admin/rooms/_form.html.haml
```diff
@@ -10,6 +10,6 @@
   .col-md-8
     = semantic_form_for(@room, url: (@room.new_record? ? admin_conference_venue_rooms_path : admin_conference_venue_room_path(@conference.short_title, @room))) do |f|
       = f.input :name, input_html: { autofocus: true}
-      = f.input :size, input_html: {size: 5}
+      = f.input :size,label: 'Capacity', input_html: {size: 5}
```
Ana06: can you please leave a space between _size_ and _label_, so:

`= f.input :size, label: 'Capacity', input_html: {size: 5}`

just for readability :wink: I wonder why our linters are not complaining :thinking:

rahul2240:
@Ana06  Thanks! I will include screenshot next time.

app/views/admin/rooms/_form.html.haml
rahul2240: @Ana06 I did :smile:

differentreality:
@rahul2240 nice :+1: Thank you :tada:

PR #1646, opened 2017-08-22
Add #show for sponsors [wip]

nasia-sam:
* Add #show view in admin/sponsors

New fields: 
* payed, :boolean 
* swags, :boolean 
* swags_received, :boolean 
* company_address, :string
* vat_registration, :string
* has_banner, :boolean,


db/schema.rb
```diff
@@ -465,6 +465,12 @@
     t.datetime "created_at"
     t.datetime "updated_at"
     t.string   "picture"
+    t.boolean  "payed",                default: false
```
differentreality: Right spelling is paid

db/schema.rb
```diff
@@ -465,6 +465,12 @@
...
+    t.boolean  "swags",                default: false
```
differentreality: If we have `has_banner`, then it makes sense to name this `has_swag`, for naming uniformity purposes.

db/schema.rb
```diff
@@ -465,6 +465,12 @@
...
+    t.boolean  "swags_received"
+    t.string   "company_address"
```
differentreality: I think just address should be sufficient (less complicated / easier to remember)

db/schema.rb
```diff
@@ -465,6 +465,12 @@
...
+    t.string   "vat_registration"
```
differentreality: just vat is enough, doesn't help us to create very long names that are not helping us understand more about what this field is. Then we just create big names, which are potentially less easy to remember.

db/schema.rb
```diff
@@ -465,6 +465,12 @@
     t.datetime "created_at"
     t.datetime "updated_at"
     t.string   "picture"
+    t.boolean  "payed",                default: false
+    t.boolean  "swags",                default: false
+    t.boolean  "swags_received"
```
differentreality: If we go with `has_swag`, let's have that be `swag_received`, also in singular. Again this is for uniformity in names, making things simple and easy to remember and use.

config/routes.rb
```diff
@@ -100,7 +100,14 @@
 
       resources :resources
       resources :tickets
-      resources :sponsors, except: [:show]
+      resources :sponsors do
+        member do
```
differentreality: Couldn't all that be part of update action (with the right param set)?

db/schema.rb
```diff
@@ -127,8 +127,8 @@
     t.integer  "end_hour",           default: 20
     t.integer  "organization_id"
     t.integer  "ticket_layout",      default: 0
-    t.string   "custom_domain"
     t.integer  "booth_limit",        default: 0
+    t.string   "custom_domain"
```
differentreality: You have quite some unrelated changes. 

You need to use git add -p db/schema.rb

Where you can type 
y - yes for keeping (staging) the change
n - no for not staging the change
s - for splitting the change into smaller chunks (if you want to keep part of it, but not all of the presented chunk)

Please don't add/remove/change anything that is not directly related with your migrations, as it may case problems.

As a second step, you can investigate why / where from those changes originated, to make them disappear and to make sure they will not be added into your commits in the future. I believe the booths related changes in email settings is due to a migration which you changed without reversing or undoing it.

db/schema.rb
differentreality: If you have already run a migration, don't just go and change something in there. Because changes are not introduced in the schema file. 
You need to reverse the migration, make the changes, re-run it.
Of course that means that all of your values in those fields will be deleted. 
And this is obviously an option only for unmerged migrations.

app/views/admin/sponsors/show.html.haml
```diff
@@ -0,0 +1,92 @@
...
+          = @sponsor.sponsorship_level.position
+          )
+      %tr
+        %td.col-md-2
+          %b Website url
+        %td
+          = link_to @sponsor.website_url, @sponsor.website_url
+      %tr
+        %td.col-md-2
+          %b Has swags?
```
differentreality: Swag (in singular)

app/views/admin/sponsors/_form.html.haml
```diff
@@ -8,12 +8,13 @@
         = @sponsor.name
 .row
   .col-md-8
-    = semantic_form_for(@sponsor, url: (@sponsor.new_record? ? admin_conference_sponsors_path : admin_conference_sponsor_path(@conference.short_title, @sponsor))) do |f|
+    = semantic_form_for(@sponsor, url: (admin_conference_sponsors_path)) do |f|
```
differentreality: If we only need this for creating a new sponsor, why not move this to new file instead of _form? It seems that a partial is no longer needed.

app/controllers/admin/sponsors_controller.rb
```diff
@@ -4,11 +4,20 @@ class SponsorsController < Admin::BaseController
...
 
+    helper_method :add_swags
+
     def index
       authorize! :index, Sponsor.new(conference_id: @conference.id)
     end
 
-    def edit; end
+    def show
+      @sponsor.swag_index = @sponsor.swags.length
```
differentreality: Have you checked if this works for sponsors with, and sponsors without, swag?

differentreality:
The index currently looks funny, with 2 tables, one after the other (shows tables of both tabs). Please fix that.
From sponsors#show where do I go to add swag?

app/controllers/admin/sponsors_controller.rb
nasia-sam: Yes. Even if there are no swags, `@sponsor.swag` is a hash that exists in every sponsor so `@sponsor.swag_index = @sponsor.swags.length ` equals to 0.

db/schema.rb
```diff
@@ -468,6 +468,15 @@
     t.datetime "created_at"
     t.datetime "updated_at"
     t.string   "picture"
+    t.boolean  "paid",                 default: false
+    t.boolean  "has_swag",             default: false
+    t.boolean  "swag_received"
+    t.string   "address"
+    t.string   "vat"
+    t.boolean  "has_banner",           default: false
+    t.text     "swags"
```
differentreality: let's keep this singular as well!

app/views/admin/sponsors/edit.html.haml
```diff
@@ -0,0 +1,37 @@
...
+      %b Has Swags?
+      = check_box_tag @conference.short_title, @sponsor.id, @sponsor.has_swag,
+      method: :patch, url: "/admin/conferences/#{@conference.short_title}/sponsors/#{@sponsor.id}?sponsor[has_swag]=",
+      class: 'switch-checkbox', data: { size: 'small',
+                                        off_color: 'warning',
+                                        on_text: 'Yes',
+                                        off_text: 'No' }
+      - @sponsor.swags.each_with_index do |(key, value), index|
+        =render partial: 'swag_fields', locals: {f: f, index: index, type_value: value[:type].to_s, quantity_value: value[:quantity].to_i}
+      =render partial: 'swag_fields', locals: {f: f, index: @sponsor.swag_index, type_value: nil, quantity_value: nil}
```
differentreality: Why do we need this swag_index variable? Why can't this always be 0?

db/migrate/20170822173332_add_swags_to_sponsors.rb
```diff
@@ -0,0 +1,13 @@
+class AddSwagsToSponsors < ActiveRecord::Migration
+  def change
+    add_column :sponsors, :paid, :boolean, default: false
+    add_column :sponsors, :has_swag, :boolean, default: false
+    add_column :sponsors, :swag_received, :boolean
+    add_column :sponsors, :address, :string
+    add_column :sponsors, :vat, :string
+    add_column :sponsors, :has_banner, :boolean, default: false
+    add_column :sponsors, :swag, :text
+    add_column :sponsors, :courier_info, :text
```
differentreality: I don't like the variables we are introducing for this. You first need to come up with your plan, present it, and then implement it, after we discuss it and agree on it. You should have done that for the new attributes and variables you are introducing. 
The courier_info for a sponsor means very little. The swag transportation info for a sponsor means much more. Please come up with meaningful attributes, and link me to your mockups and workflows document.

db/schema.rb
differentreality: Again this is still not fixed.

db/schema.rb
```diff
@@ -577,7 +586,6 @@
     t.boolean  "is_admin",               default: false
     t.string   "username"
     t.boolean  "is_disabled",            default: false
-    t.string   "token"
```
differentreality: What's that?

db/schema.rb
```diff
@@ -48,11 +48,11 @@
     t.text     "reasoning"
     t.string   "state"
     t.string   "logo_link"
-    t.string   "website_url"
-    t.text     "submitter_relationship"
     t.integer  "conference_id"
     t.datetime "created_at",             null: false
     t.datetime "updated_at",             null: false
+    t.string   "website_url"
```
differentreality: Why are these here? Please pay some attention to your db, and this file.

app/views/admin/sponsors/show.html.haml
```diff
@@ -0,0 +1,121 @@
...
+        %td.col-md-2
+          %b Website url
+        %td
+          = link_to @sponsor.website_url, @sponsor.website_url
+
+
+      - if @sponsor.address?
+        %tr
+          %td.col-md-2
+            %b Company's address
```
differentreality: A sponsor does not have to be a company.

differentreality:
@nasia-sam is this still wip?

app/views/admin/sponsors/show.html.haml
```diff
@@ -0,0 +1,119 @@
...
+        %td
+          = check_box_tag @conference.short_title, @sponsor.id, @sponsor.has_swag,
+          method: :patch, url: "/admin/conferences/#{@conference.short_title}/sponsors/#{@sponsor.id}?sponsor[has_swag]=",
+          class: 'switch-checkbox', data: { size: 'small',
+                                            off_color: 'warning',
+                                            on_text: 'Yes',
+                                            off_text: 'No' }
+
+
+      - if @sponsor.swag.length > 0
```
differentreality: Can we do this `if @sponsor.swag.any?` ?

app/views/admin/sponsors/edit.html.haml
```diff
@@ -0,0 +1,38 @@
...
+      = f.input :vat, label: 'VAT Registration Number'
+      %b Has Swags?
+      = check_box_tag @conference.short_title, @sponsor.id, @sponsor.has_swag,
+      method: :patch, url: "/admin/conferences/#{@conference.short_title}/sponsors/#{@sponsor.id}?sponsor[has_swag]=",
+      class: 'switch-checkbox', data: { size: 'small',
+                                        off_color: 'warning',
+                                        on_text: 'Yes',
+                                        off_text: 'No' }
+      - @sponsor.swag.each_with_index do |(key, value), index|
+        = render partial: 'swag_fields', locals: {f: f, index: index, type_value: value[:type].to_s, quantity_value: value[:quantity].to_i}
```
differentreality: value here is the hash `{type: 'xyz', quantity: 10}` right?
If so, then why don't we simply pass this hash as parameter to the partial file, and get the type and quantity inside of the partial?
It will simplify our view, and it will make our code easier to read and understand.

app/controllers/admin/sponsors_controller.rb
```diff
@@ -8,7 +8,19 @@ def index
       authorize! :index, Sponsor.new(conference_id: @conference.id)
     end
 
-    def edit; end
+    def show
+      @sponsor.swag.reject! { |_key, value| value[:type].blank? || value[:quantity].blank? }
+      @sponsor.swag_transportation.reject! { |_key, value| value[:carrier_name].blank? || value[:tracking_number].blank? || value[:boxes].blank? }
+      @sponsor.swag_index = @sponsor.swag.length
```
differentreality: This variable is being used in 1 place only. Do we really have to set it here? Can't we just call the @sponsor.swag.length where we need it?

differentreality: Same goes for its declaration in edit action. And for the transportation index variable.

app/views/admin/sponsors/edit.html.haml
nasia-sam: Sure, but this isn't working when the form is rendered for new, empty fields. The partial doesn't recognize value as a hash, thus an error occurs. This is why I preferred to use `type_value` and `quantity_value`.

app/controllers/admin/sponsors_controller.rb
```diff
@@ -47,10 +59,72 @@ def destroy
...
+
+      if @sponsor.save
+        redirect_to admin_conference_sponsors_path(@conference.short_title),
+                    notice: 'Sponsor\'s state successfully updated'
+      else
+        flash[:error] = 'Sponsor\'s state  couldn\'t be updated'
+      end
+    end
+
+    def paid
```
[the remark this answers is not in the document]
nasia-sam: Organizers don't have to visit the edit form in order to change the toggle button. They can do that from the show view, so if it was part of the the update action, it wouldn't change the value of the `paid` boolean.

app/controllers/admin/sponsors_controller.rb
```diff
@@ -47,10 +59,72 @@ def destroy
...
+      @sponsor.paid = !@sponsor.paid
+      if @sponsor.save
+        flash[:notice] = 'Sponsor successfully updated.'
+      else
+        flash[:error] = 'Sponsor failed to be updated.'
+      end
+    end
+
+    def has_swag
```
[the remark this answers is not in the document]
nasia-sam: Same reason with the `paid` method.

app/controllers/admin/sponsors_controller.rb
```diff
@@ -47,10 +59,72 @@ def destroy
...
+      @sponsor.has_swag = !@sponsor.has_swag
+      if @sponsor.save
+        flash[:notice] = 'Sponsor successfully updated.'
+      else
+        flash[:error] = 'Sponsor failed to be updated.'
+      end
+    end
+
+    def swag_received
```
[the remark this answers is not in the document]
nasia-sam: Same reason with the `paid` method.

app/controllers/admin/sponsors_controller.rb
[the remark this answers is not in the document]
differentreality: I don't understand how those 2 things are related.
You have this switch checkbox in show view:
https://github.com/openSUSE/osem/pull/1646/files#diff-cd51cee616759110380e4244f398abf0R26
This is calling update action - check the server logs:
`Started PATCH "/admin/conferences/tconf0/sponsors/190?sponsor[paid]=true"`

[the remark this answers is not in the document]
nasia-sam: But still, if it's part of the update action wouldn't that mean that every time the update action is called, the booleans changes attributes?

app/views/admin/sponsors/edit.html.haml
nasia-sam: I have already tried to work on it but it turned out to be more challenging than I expected

differentreality:
Closing due to inactivity. @nasia-sam feel free to work more on this in a different PR! I'd love to see this progress :)

PR #1647, opened 2017-08-24
Move revision history of users to be accessed only by site admins

### LINKED ISSUE #1459: Organiser of a conference can remove the admin of OSEM app

shlok007:
Closes #1459 
@lagartoflojo @differentreality please do have a look.


differentreality:
So what is our workflow here?
For hosted version, who can see signups and such? Only the people who set it up? 

What about people organizing a conference, who want to help one of their speakers?

shlok007:
> For hosted version, who can see signups and such? Only the people who set it up?

Yes, that's what was initially planned!

> What about people organizing a conference, who want to help one of their speakers?

How about allowing organizers to access all users/speakers that are registered to a conference? I think that might complicate things a little but can be a solution to this problem.

differentreality:
@shlok007 sure, wanna give it a go or do you prefer that we merge this PR and open an issue for someone else to pick up?

shlok007:
I'll be happy to do it myself. Let's not merge it yet, I'll append commits here only.

differentreality:
We need the fix, so that admin rights are not taken away by organizers. What's your plan for this @shlok007

Ana06:
@shlok007 @differentreality what is the state of this?

PR #1653, opened 2017-08-25
Removed datatable if there is no data

### LINKED ISSUE #1648: Do not show datatable if there are not data
In admin/tracks#index the datatable appears even if there is no track to show. It's not useful to be there, so we should not be showing it. 

Same happens in admin/booths#index

rahul2240:
Fixes #1648
booth
tracks

### DIFF
--- a/app/views/admin/tracks/index.html.haml
+++ b/app/views/admin/tracks/index.html.haml
@@ -7,75 +7,76 @@
         Categorize events in your conference
 .row
   .col-md-12
-    %table.table.table-hover.table-striped.table-bordered.datatable#tracks
-      %thead
-        %th ID
-        %th Name
-        %th Description
-        %th Room
-        %th Start Date
-        %th End Date
-        %th Submitter
-        %th Included in Cfp
-        %th State
-        %th Actions
-      %tbody
-        - @tracks.each do |track|
-          %tr
-            %td
-              = track.id
-            %td{ style: 'padding: 15px 0px 0px 10px;' }
-              = link_to admin_conference_program_track_path(@conference.short_title, track), class: 'btn' do
-                %span.label{style: "background-color: #{track.color}; color: #{ contrast_color(track.color) }"}
-                  = track.name
-            %td
-              %p
-                = markdown(truncate(track.description))
-            %td
-              = track.room.try(:name)
-            %td
-              = track.start_date.strftime('%A, %B %-d. %Y') if track.start_date
-            %td
-              = track.end_date.strftime('%A, %B %-d. %Y') if track.end_date
-            %td
-              = link_to track.submitter.name, admin_user_path(track.submitter) if track.self_organized?
-            %td.text-center{ 'id' => "cfp_switch_#{track.id}", 'data-order' => track.cfp_active.to_s }
-              = check_box_tag "#{@conference.short_title}_#{track.short_name}", track.id, track.cfp_active,
-                class: 'switch-checkbox', method: :patch,
-                url: toggle_cfp_inclusion_admin_conference_program_track_path(@conference.short_title, id: track.short_name)+"?included=",
-                data: { size: 'small',
-                        on_color: 'success',
-                        off_color: 'warning',
-                        on_text: 'Yes',
-                        off_text: 'No' }
+    - if @tracks.any?
+      %table.table.table-hover.table-striped.table-bordered.datatable#tracks
+        %thead
+          %th ID
+          %th Name
+          %th Description
+          %th Room
+          %th Start Date
+          %th End Date
+          %th Submitter
+          %th Included in Cfp
+          %th State
+          %th Actions
+        %tbody
+          - @tracks.each do |track|
+            %tr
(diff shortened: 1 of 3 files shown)


differentreality:
The screenshot of the booth looks too grey, but seems to be OK in my local instance :+1: 
While you are at it, how would you like to also move the button in admin/booths#index to make it similar to the one in admin/tracks#index?
Please also rebase.

rahul2240:
boothbutton
After moving the button, the page is looking like this,is this ok? or you want that button below when no booth is present and at the present location when booths are present?

differentreality:
I am aiming to have uniformity in our views, so that user doesn't see one layout in one page, and a different one in another. 

So the buttons should be in the same position, same color, similar wording. Now one is up, the other down. One is blue the other one is green. One says 'New Track', the other says 'Add Booth'. That is confusing for the user. We should stick with one color schema for buttons that do the same thing, generally a common layout, and similar wording. Does that make sense to you?

I would like you to be able to understand what we need to do, and why we need it. It is not useful if I tell you to do X, and you just do X, but I'd rather you understand the purpose of what we are doing, so that you can pinpoint on your own what is required.

rahul2240:
@differentreality Thank you! :smile: 
Now i think that both pages have uniformity
booth3
track3

rahul2240:
@differentreality  Thank You! :smile: 
I think now both the pages have common layout, like the other pages
booths
tracks

rahul2240:
@differentreality Thank you so much :smile: 
I saw other pages and according to that i edit the booths and tracks view this time.Please let me know whether i did that right or not.
track1
booths

rahul2240:
@differentreality Thank you so much :smile:
I saw other pages and according to that i edit the booths and tracks view this time.Please let me know whether i did that right or not.
track1
boot

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170816203325) do
+ActiveRecord::Schema.define(version: 20170814174637) do
```
differentreality: Not sure why this shows up here. You should only introduce changes in db/schema.rb if you have a migration. Otherwise something has gone wrong. Make sure you `git pull` in your master branch, then go to your branch and rebase. In any case these changes, should **not** be here.

rahul2240: yes. Thanks i will do it. :smile:

rahul2240:
@differentreality  I removed that db schema change but I dont know why tests are failing.

rahul2240:
@differentreality I dont know why my tests are failing.

differentreality:
@rahul2240 are you on IRC? We have an irc channel #osem in freenode server.
Do you understand what the failing test is about? Have you read the test to understand what it is doing and what it is expecting?

app/views/admin/tracks/index.html.haml
```diff
@@ -7,75 +7,76 @@
...
-            %td.text-center{ 'id' => "cfp_switch_#{track.id}", 'data-order' => track.cfp_active.to_s }
-              = check_box_tag "#{@conference.short_title}_#{track.short_name}", track.id, track.cfp_active,
-                class: 'switch-checkbox', method: :patch,
-                url: toggle_cfp_inclusion_admin_conference_program_track_path(@conference.short_title, id: track.short_name)+"?included=",
-                data: { size: 'small',
-                        on_color: 'success',
-                        off_color: 'warning',
-                        on_text: 'Yes',
-                        off_text: 'No' }
+    - if !@tracks.empty?
```
differentreality: For booths you did, if booths.any?, why do you prefer this other way here for the tracks?

rahul2240:
@differentreality  i was talking about hakiri failure, sorry i should have mentioned that.I saw travis failure and i will add tests.
@Ana06 Thanks

app/views/admin/tracks/index.html.haml
rahul2240: i saw @booths.any? is used at some other place too in the booths' index.html.haml , so i used any and on tracks i used empty as i always use it because sometimes it is faster than any.
https://mensfeld.pl/2015/06/empty-vs-blank-vs-any-why-you-should-not-use-any-to-test-if-theres-anything-in-the-array/
i read it from here.

spec/features/tracks_spec.rb
```diff
@@ -42,10 +42,8 @@
 
       expected.to change { Track.count }.by(-1)
       expect(flash).to eq('Track successfully deleted.')
-      within('table#tracks') do
-        expect(page.has_content?(track.name)).to be false
-        expect(page.has_content?(track.description)).to be false
-      end
+      expect(page.has_content?(track.name)).to be false
```
Ana06: I would also like that you check that the table is not render, so that `table#tracks` is not in the page :wink:

rahul2240: @Ana06 Please let me know if i did some mistake or not :smile:

Issue #1651, opened 2017-08-25
Browsing is confusing with organization name on the top left navbar


differentreality:
When I am anywhere (eg proposals#index) I see the organization name on the top left corner of the navigation bar. That takes me to organizations#index. 

Usually on the top left we find the link to homepage. 

Also there is nothing I want to do in organizations#index (99% of the time at least).

The organization name could be the same as the app name (for the non hosted version of OSEM), as such I see the exact same thing, which is directing me to a totally different place every time. And that's confusing.

bear454:
I think what we're missing is some logic for when there's only one organization. In that case, I agree, it makes no sense to deal with the organizational level.

PR #1657, opened 2017-08-26
link to organization#new from user menu

shlok007:
To ease the usability and convey the sequence of creating an organization first before creating a conference, we can replace the `New Conference` link with `New Organization` link when no organization is present in OSEM. As soon as an organization is created we can switch it back to `New Conference` as `New Organization` can be accessed from `admin/organizations#index`


differentreality:
Bare with me here for a little bit, as workflow-wise I am not sure what our goal is (do you have a gdoc or something to link me to?):

* So if I am in the admin side

  * In admin/conferences#index, then it says OSEM (or whatever I have configured, which is nice) and takes me to the home page (which is nicer). The link 'Organizations is immediately visible for me on the left menu, so that's great, I can access that if I need to do sth there)

screenshot from 2017-08-28 11-11-47

  * In any other place of the admin, I do **not** see the configured app name (in my screenshot that's still OSEM), so my navbar is missing stuff now :( 
  * I am quite a bit missing the link to go to the home page
  * It's useful to provide quick access to organizations without cluttering the menus any further'

screenshot from 2017-08-28 11-12-45
 
On one hand it is nice to see the organization of the conference (if/when that makes sense to you - when there is more than one?), on the other hand showing both app_name and organization, which might be the same thing, will be a confusing duplication

* If I am in the non-admin side, eg. proposals#index I again see the link 'my organization'
  * It takes me to /organizations, what am I supposed to do there (as a simple user/speaker/attendee)?
  * No link to homepage. I don't like this, even less outside of admin.

screenshot from 2017-08-28 11-17-17

shlok007:
I noticed the issue #1651 and I apologise for my negligence towards it as I was not sure what the right approach should be. The link we currently have on top left in navbar are definitely not right. 
     
* As you said 99% of the time there is nothing to do is admin/organizations#index and the top left navigation bar link should obviously lead to the home page of the application to keep the browsing simple.
* I was hoping to introduce organizations along with upcoming conference in the home page. The organizations can appear as a bootstrap carousel in the home page just above 'upcoming conferences' heading. ( I remember I do have a branch with relevant code almost ready ) 
    
Considering the above two points we can keep the text of organization name with a link to home page of our application which will have a list of organizations along with conferenes too.

differentreality:
There are pages where there is no organization. Then we would need to use a different text to redirect to the same place (root_path). I think that we should:

  1. Show the root_path link always (text is the app_name, whatever that might be)
  2. Show org name if it is different than app_name
  3. In admin, make sure there is a link to take me to orgs#index (bottom of admin sidebar?)

Would that cover all our cases?

shlok007:
Yes, that sure does make perfect sense. Although, this PR wasn't really addressing #1651 , I'll append commits here resolving the issue.

differentreality:
@shlok007 great, I will wait for you to update this PR then!

differentreality:
Do you think we still need this?

We already have a link to Organizations in admin_sidebar_index

hennevogel:
I guess this is not going anywhere. Let's close this. @shlok007 please open another PR if you're up for working on this.

PR #1670, opened 2017-08-31
Showing of tracks when there are no tracks in proposals is fixed

### LINKED ISSUE #1668: Text on top of proposals#index should not show track information, when there is no track
`Proposals should fit in one of the 0 tracks: .`
That should not appear when there are no tracks.

namangupta01:
#1668

### DIFF
--- a/app/views/proposals/_encouragement_text.html.haml
+++ b/app/views/proposals/_encouragement_text.html.haml
@@ -2,7 +2,7 @@
   - if @program.event_types.any?
     You can submit proposals for
     = "#{event_types(@conference)}."
-  - if @program.tracks.any?
+  - if @program.tracks.confirmed.cfp_active.any?
     Proposals should fit in one of the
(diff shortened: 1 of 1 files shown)


app/views/proposals/_encouragement_text.html.haml
```diff
@@ -2,7 +2,7 @@
   - if @program.event_types.any?
     You can submit proposals for
     = "#{event_types(@conference)}."
-  - if @program.tracks.any?
+  - if @program.tracks.confirmed.cfp_active.first
```
differentreality: Why not `@program.tracks.confirmed.cfp_active.any?`?

namangupta01: @differentreality Yes, its better!

differentreality:
Thank you @namangupta01 :tada:

PR #1674, opened 2017-09-05
added description field to cfps

### LINKED ISSUE #1650: Add description field for cfps
So that organizers can customize the text that appears in proposals#index, tracks#index, booths#index and encourage/inform submitters about what they should focus on, and what they should know about their submissions.

ViditChitkara:
closes #1650 

screenshot from 2017-09-08 18 25 14
@differentreality Is this correct way to show description text??

Screenshots for the above feature:-
pr1
pr2
pr3
pr4
pr5

### DIFF
--- a/app/views/admin/cfps/_booths_cfp.html.haml
+++ b/app/views/admin/cfps/_booths_cfp.html.haml
@@ -11,6 +11,10 @@
 %dd
   = @cfp.end_date.strftime('%A, %B %e. %Y')
 %dt
-  Days Left:
+  Description
+%dd
+  = markdown(@cfp.description)
+%dt
+  Days Left
 %dd
   = pluralize(@cfp.remaining_days, 'day')
--- a/app/views/admin/cfps/_tracks_cfp.html.haml
+++ b/app/views/admin/cfps/_tracks_cfp.html.haml
@@ -11,6 +11,10 @@
 %dd#end_date
   = @cfp.end_date.strftime('%A, %B %-d. %Y')
 %dt
+  Description:
+%dd#description
+  = markdown(@cfp.description)
+%dt
   Days Left:
 %dd
   = pluralize(@cfp.remaining_days, 'day')
--- a/app/views/proposals/index.html.haml
+++ b/app/views/proposals/index.html.haml
@@ -6,6 +6,14 @@
         %span.notranslate
           = @conference.title
 
+
+  - if @program.cfp_open?
+    - if @program.cfp.description.present?
+      .row
+        .col-md-12
+          = markdown(@program.cfp.description)
+
+
   .row
     .col-md-12
       = render partial: 'encouragement_text'
--- a/app/views/proposals/new.html.haml
+++ b/app/views/proposals/new.html.haml
@@ -3,6 +3,11 @@
     .col-md-12
       .page-header
         %h1 New Proposal
+  - if @program.cfp_open?
+    - if @program.cfp.description.present?
+      .row
+        .col-md-12
+          = markdown(@program.cfp.description)
   .row
     .col-md-12
       = render partial: 'encouragement_text'
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -68,12 +68,13 @@ ActiveRecord::Schema.define(version: 20170924190528) do
   end
 
   create_table "cfps", force: :cascade do |t|
-    t.date     "start_date", null: false
-    t.date     "end_date",   null: false
+    t.date     "start_date",  null: false
+    t.date     "end_date",    null: false
     t.datetime "created_at"
     t.datetime "updated_at"
     t.integer  "program_id"
     t.string   "cfp_type"
+    t.text     "description"
   end
 
   create_table "comments", force: :cascade do |t|
--- a/app/controllers/admin/cfps_controller.rb
+++ b/app/controllers/admin/cfps_controller.rb
@@ -55,7 +55,7 @@ module Admin
     private
 
     def cfp_params
-      params.require(:cfp).permit(:start_date, :end_date, :cfp_type)
+      params.require(:cfp).permit(:start_date, :end_date, :description, :cfp_type)
     end
   end
 end
--- a/app/views/admin/cfps/_events_cfp.html.haml
+++ b/app/views/admin/cfps/_events_cfp.html.haml
@@ -11,6 +11,10 @@
 %dd#end_date
   = @cfp.end_date.strftime('%A, %B %-d. %Y')
 %dt
+  Description:
+%dd#description
+  = markdown(@cfp.description)
+%dt
   Days Left:
 %dd
   = pluralize(@cfp.remaining_days, 'day')
--- a/app/views/admin/cfps/_form.html.haml
+++ b/app/views/admin/cfps/_form.html.haml
@@ -8,5 +8,6 @@
       = f.input :start_date, as: :string, input_html: { id: 'registration-period-start-datepicker', start_date: @conference.start_date, end_date: @conference.end_date, readonly: 'readonly' }
       = f.input :end_date, as: :string, input_html: { id: 'registration-period-end-datepicker', readonly: 'readonly' }
       = f.input :cfp_type, as: :select, collection: (@cfp.new_record? ? @program.remaining_cfp_types : [@cfp.cfp_type] + @program.remaining_cfp_types).map {|type| ["#{type.capitalize}", type]}, include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' }
+      = f.input :description, input_html: {rows: 2, data: { provide: 'markdown-editable' } }, hint: markdown_hint
       %p.text-right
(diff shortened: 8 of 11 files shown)


db/schema.rb
```diff
@@ -577,7 +578,6 @@
     t.boolean  "is_admin",               default: false
     t.string   "username"
     t.boolean  "is_disabled",            default: false
-    t.string   "token"
```
differentreality: Not sure what happened here, but we shouldn't be making this change. Can you please take a look?

differentreality:
Very nice work @ViditChitkara  :+1: 
It would be extremely helpful if you mention the issue on the description (ie first comment of this PR) instead of the title.

app/views/admin/cfps/_tracks_cfp.html.haml
```diff
@@ -7,6 +7,10 @@
 %dd#end_date
   = @cfp.end_date.strftime('%A, %B %-d. %Y')
 %dt
+  Description:
+%dd#description
+  = @cfp.description
```
differentreality: Why isn't this in markdown as well?

db/schema.rb
ViditChitkara: Sure, I will have a look at it.

app/views/admin/cfps/_tracks_cfp.html.haml
ViditChitkara: I think it was missed out by mistake. I will make the changes right away. Thanks for your help.

differentreality: What I mean is why we don't show the description formatted (as per the markdown format the user has used).

Please add relevant screenshots of your changes so that we can see this more clearly.

app/views/proposals/_encouragement_text.html.haml
```diff
@@ -20,5 +20,10 @@
     %span.notranslate
       = @conference.title
     will only be as good as the content you present. Submit early, submit often!
+    %br
+    %b
+      Description:
+    %span
+      = @program.cfp.description
```
differentreality: Let's keep the description visible.
Do you think it makes sense to show it *before* the encouragement text? 
Encouragement text is also visible in other places (eg new form), let's show the description there as well.

differentreality:
@ViditChitkara please add new commits with your changes to help reviewers :)

app/views/admin/cfps/_tracks_cfp.html.haml
ViditChitkara: screenshot from 2017-09-08 17 20 05
This is how the description text looks as of now.

ViditChitkara: screenshot from 2017-09-08 17 19 57
This is the update form for cfp. Here decription contains markdown format link

differentreality: @ViditChitkara this is the cfps#index view, my comment was about the cfps#show view

ViditChitkara: screenshot from 2017-09-08 17 36 37
Thanks @differentreality , I have made the necessary changes.

differentreality:
I would add the screenshots on your first comment of this PR to keep them visible for future reference, the are quite useful to let us know how things changed with this PR.

If you have any questions about the rest of my comments, feel free to ping me on IRC too (freenode server, #osem).

app/views/proposals/_encouragement_text.html.haml
ViditChitkara: Yes that makes sense. So shall I make another partial and show it after the heading,(be it show or new)

differentreality: Either a partial, or you just add one line in the places where we want the description to show up.
Do you think a partial is really necessary to show just one field?

ViditChitkara: Okay I'll make the necessary changes.

app/views/proposals/index.html.haml
```diff
@@ -6,6 +6,17 @@
         %span.notranslate
           = @conference.title
 
+
+  - if @program.cfp_open?
+    - if @program.cfp.description.length>0
```
differentreality: Why do we need to check the length of the description here?

app/views/admin/cfps/_booths_cfp.html.haml
```diff
@@ -7,6 +7,10 @@
 %dd
   = @cfp.end_date.strftime('%A, %B %e. %Y')
 %dt
+  Description
+%dd
+  = markdown(truncate(@cfp.description))
```
differentreality: If we truncate this here in the show action, where are we showing the whole text on the admin side?

app/views/proposals/index.html.haml
differentreality: Why do we need to check the length of the description here?

app/views/proposals/new.html.haml
```diff
@@ -3,6 +3,14 @@
     .col-md-12
       .page-header
         %h1 New Proposal
+  - if @program.cfp_open?
+    - if @program.cfp.description.length>0
+      .row
+        .col-md-12
+          %p.lead
+            %b
+              Description:
```
differentreality: Do we actually need to show the word description? Why not just show the text?
Is it helpful if I see the following text, for example?

> Description: Note that you need to be at your booth at all times during the conference!

app/views/proposals/index.html.haml
ViditChitkara: According to me, if organizers are not giving any description we shouldn't make a div that contains "description" and description text

app/views/proposals/new.html.haml
ViditChitkara: Instead of description we could show something which catches user attention like important or so..

app/views/proposals/index.html.haml
differentreality: I completely agree, we should not create divs we don't actually need. 

But why do we need to check the length of the description and not just if @program.cfp.description

app/views/proposals/new.html.haml
differentreality: That's indeed another good idea. I was thinking to show nothing at all, just the text itself. 
The text may be a description, it may be important information, it may be something else.

app/views/proposals/index.html.haml
ViditChitkara: I guess, even if organizers leaves the description field empty, description for that particular cfp is a string of zero length and not nil.🤔🤔

app/views/proposals/new.html.haml
ViditChitkara: Yeah, so could we keep it as "important information"?
Also could we move this "important info" out of the condition checking statement, because some info like days left is shown and won't be left empty anyways?

app/views/proposals/index.html.haml
differentreality: Then cfp.description.present? which checks against nil and empty string values.

app/views/proposals/new.html.haml
differentreality: I have no idea. I think we don't have to show anything.

If organizers want to show Important there, they can do so by adding it in the beginning of their description text. Markdown gives them tons of possibilities to format this to their liking. 

So let's not show anything except for the description itself and wrap this up ;)

app/views/proposals/index.html.haml
ViditChitkara: That's a nice idea, instead of checking the length😃😃

app/views/proposals/new.html.haml
ViditChitkara: Cool! So I would fix the show page and condition statement.

app/views/admin/cfps/_booths_cfp.html.haml
ViditChitkara: Yep I should not have added truncate for show actions.😅😅

Ana06:
@ViditChitkara please rebase, the Travis run is broken, but in master is already fixed :wink:

ViditChitkara:
@Ana06 I have done rebasing. Travis is running fine now!😃😃

bear454:
@ViditChitkara please follow up on remaining issues, so we can get this merged in!

ViditChitkara:
Sure @bear454 , could you please assist me, like what more changes need to be done in this issue?

bear454:
https://github.com/openSUSE/osem/pull/1674#issuecomment-328089245

ViditChitkara:
@differentreality if you could please suggest any further changes which need to be made in this issue.

differentreality:
@ViditChitkara please update the screenshot on your first comment. 
Please also screenshots of all changed views, eg. the form is also changed, admin/cfps#index, admin#cfps#show, proposals#new.

Please also squash your commits, and rebase, so that we can merge this PR!

Thank you :+1:

ViditChitkara:
Okay @differentreality, I'll make the required changes

Issue #1666, opened 2017-08-29
config/environments/production.rb should be in .example file


differentreality:
production.rb should be renamed to production.rb.example, and production.rb should be added in gitignore.

tobru:
This would make deploying the application with Docker much more complicated as we would have to somehow generate the file or create it with a template.

differentreality:
So this would not be enough?
`cp config/environments/production.rb.example config/environments/production.rb`

tobru:
Yeah, would be a solution if the example file is ready to be used.

PR #1675, opened 2017-09-06
production file is renamed and production.rb is added in gitignore

namangupta01:
#1666


.gitignore
```diff
@@ -38,3 +38,4 @@ pickle-email-*.html
 .env.local
 docker-compose.env
 docker-compose.yml
+production.rb
```
differentreality: Please add an empty line at the end of the file to make this red symbol disappear

namangupta01: okay!

differentreality:
@hennevogel My thinking is that the users who use OSEM, change production.rb to whatever they need, then if they pull newest OSEM changes there might be conflicts.

But this really came up when I checked out https://github.com/SeaGL/osem/pull/6 to see their changes  (because I did not want to clone their whole repo, just to see a few changes) and it overwrote my production.rb

What's your thinking?

lagartoflojo:
Pulling the latest version of OSEM should not overwrite `production.rb`. If there are conflicting changes, then those conflicts will be noticed by git, and it will complain. I have never seen a Rails project with an example `production.rb` file either. Plus, this would need documentation updates before it gets merged.

@hennevogel let's try to keep our comments courteous please :)

hennevogel:
@differentreality users shouldn't change production.rb. The only place they should configure the app are the environment variables. What did it overwrite for you?

differentreality:
It overwrote several things seagl has changed, such as:
https://github.com/SeaGL/osem/blob/master/config/environments/production.rb#L36
https://github.com/SeaGL/osem/blob/master/config/environments/production.rb#L70
https://github.com/SeaGL/osem/blob/master/config/environments/production.rb#L71
https://github.com/SeaGL/osem/blob/master/config/environments/production.rb#L82

Anyways, closing this

Issue #1683, opened 2017-09-10
error in generating pdf for ticket when venue is not set


rahul2240:
physicalticketcontroller#show should contain condition on @venue.name that if it exist then only pdf will be generated otherwise venue is nil and app will show error

differentreality:
> undefined method `name' for nil:NilClass

` draw_text @conference.venue.name.to_s, at: [@mid_horizontal + 30, cursor - 70]`

app/pdfs/ticket_pdf.rb:54:in `draw_second_square'

rahul2240:
can i take this?

differentreality:
Go ahead @rahul2240, thanks for reporting and taking up on it :)

PR #1688, opened 2017-09-11
Fix error while generating pdf when venue is nil

rahul2240:
Error in generating pdf for ticket when venue is not set is fixed.Showing venue in pdf only if venue exist.
Closes https://github.com/opensuse/osem/issues/1683

### DIFF
--- a/app/pdfs/ticket_pdf.rb
+++ b/app/pdfs/ticket_pdf.rb
@@ -51,7 +51,9 @@ class TicketPdf < Prawn::Document
     move_down 70
     draw_text @conference.title.to_s, at: [@mid_horizontal + 30, cursor - 30], size: 12
     draw_text @conference.organization.name.to_s, at: [@mid_horizontal + 30, cursor - 50], size: 12
-    draw_text @conference.venue.name.to_s, at: [@mid_horizontal + 30, cursor - 70]
+    if @conference.venue
+      draw_text @conference.venue.name.to_s, at: [@mid_horizontal + 30, cursor - 70]
+    end
     move_up 130
     move_down @mid_vertical
   end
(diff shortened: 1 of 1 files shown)


rahul2240:
Can you tell me why only venue's name is mentioned in pdf?

app/pdfs/ticket_pdf.rb
```diff
@@ -51,7 +51,9 @@ def draw_second_square
     move_down 70
     draw_text @conference.title.to_s, at: [@mid_horizontal + 30, cursor - 30], size: 12
     draw_text @conference.organization.name.to_s, at: [@mid_horizontal + 30, cursor - 50], size: 12
-    draw_text @conference.venue.name.to_s, at: [@mid_horizontal + 30, cursor - 70]
+    if !@conference.venue.nil?
```
differentreality: Why not just `if conference.venue` ?

differentreality:
Hmmm perhaps @siddhantbajaj or @Ana06 can shed some light on that question. 
I assume you mean why it doesn't also show its address, right? I would agree that we should be also showing the full address of the venue on the ticket.

siddhantbajaj:
@rahul2240 Thanks for the PR. Yes I think we should indeed show the full address in the ticket. Will you take this up as well in this PR? If yes please add the screenshot of your changes. 😉

app/pdfs/ticket_pdf.rb
```diff
@@ -51,7 +51,9 @@ def draw_second_square
...
+      draw_text @conference.venue.name.to_s, at: [@mid_horizontal + 30, cursor - 70]
```
Ana06: `to_s` shouldn't be needed, right? `name` is already an string (I know this was already there, so you can add the change in a different commit) :wink:

rahul2240:
@siddhantbajaj Please review the changes.
ticket

app/pdfs/ticket_pdf.rb
rahul2240: @Ana06 Sure! :smile:

Ana06:
@rahul2240 please rebase, the Travis run is broken, but in master is already fixed :wink:

rahul2240:
@Ana06 Sorry i forgot to rebase, travis is working now :smile:

rahul2240:
@Ana06 can you close this PR if it is fixed?

differentreality:
Thank you @rahul2240  :+1: 

I don't see anything else pending for this PR, and travis is passing, so merging this :tada:

PR #1690, opened 2017-09-12
Fix travis dependencies

AEtherC0r3:
It looks like that libmariadbclient-dev is missing in the travis environment and this causes the mysql2 gem's native extensions to not be build. breaking the bundle install phase.


AEtherC0r3:
This was fixed by #1694, which got merged

Issue #1697, opened 2017-09-14
Add Currency


tobru:
How can I add a currency? We sell tickets in CHF so it would be very great to have CHF as currency available.

differentreality:
We should enable this more globally:
https://github.com/openSUSE/osem/pull/1699#issuecomment-330033169

rishabhptr:
Can I work on this?

differentreality:
Sorry I didn't update this, there is ppl working on this already. Could you please choose sth else? Thanks!

rishabhptr:
@differentreality Alright, no problem!

Ana06:
@differentreality who is working on this? is there a pr? 😕

adrianovieira:
I believe this will work if you have translation enabled for your own language.

PR #1699, opened 2017-09-15
Add currency CHF

tobru:
This PR solves #1697.

If I get it right this is the only place which needs an update to add a new currency.


differentreality:
@tobru I've been thinking, while this temporarily solves your particular problem it is not a global solution, and the more people use OSEM, the more this might come up. We should fix this for everyone.

One solution I can think of would be to include all available currencies in that drop down (this can be done with `Money::Currency.all`), and use `selectize-rails` so that it is possible to start typing the currency while it searches through the options.
(We are already using `selectize-rails` for proposals#form to select speakers.)

Thoughts?

tobru:
Of course you're right, this only solves my particular need for having CHF available. Having all currencies available would be the best case here. I'm not a Ruby developer, I'm just using OSEM for DevOpsDays Switzerland and try to solve our needs as good as possible. If someone of the OSEM team is able to add support for all currencies: go for it and close my PR.

differentreality:
Yeap makes sense! Thanks for sending your changes, we appreciate it :)
I've updated the issue as well, so it should be picked up.

By the way, can we add your conference to our wiki? 
https://github.com/openSUSE/osem/wiki/Conferences-using-OSEM
Feel free to send us the link!

tobru:
Will send you the link as soon as it's ready

Issue #1682, opened 2017-09-10
admin/physical_ticket#index should contain a amount paid option


rahul2240:
In case price for tickets change, then the admin should be able to see which User paid how much money.

siddhantbajaj:
Yes actually a good feature to have because ticket price do gets changed a lot and there is early bird discount on tickets as well. But @rahul2240 what do you propose to achieve this? 
One possible approach could be having a "amount_paid" attribute in physical_ticket model so that we can store amount paid for that ticket. @Ana06 , @mdeniz your thoughts?

rahul2240:
@siddhantbajaj I was thinking the same.

differentreality:
All that (ticket, physical_ticket, etc.) should be linked with payment in the end. Payment does hold the amount info. So we should be able to extract how much users have paid, based on the type of ticket they have purchased. Isn't that so?

rahul2240:
I was thinking to create a new column in purchase_ticket model and if the payment is done then before rendering physical ticket, the price of the ticket can be stored in the amount_paid column.right now i am doing this , if this is not the correct way then please let me know.

differentreality:
I am not sure I understand **why** we need the new column. Why are we not extracting the amount information from payments table?
Maybe I am missing something in our tickets workflow.

Ana06:
@differentreality 

> I am not sure I understand why we need the new column. Why are we not extracting the amount information from payments table?

Because in payment we have the total amount, so if you buy 3 different tickets it is impossible to know what is the price of every ticket.

However, I am not pretty sure if including the new column in `physical_ticket`, as @rahul2240 and @siddhantbajaj suggested, is the best option. Why not in `ticket_purchase`?
 Currently we are delegating this to the ticket, but as the price could change after payment, that doesn't make sense for me: https://github.com/openSUSE/osem/blob/master/app/models/ticket_purchase.rb#L14

PR #1701, opened 2017-09-16
Add paid option in physical ticket#index

rahul2240:
paid option
Paid column is added to physical_tickets#index so that admin can see which user paid how much amount
fixes https://github.com/openSUSE/osem/issues/1682

### DIFF
--- a/app/controllers/conference_registrations_controller.rb
+++ b/app/controllers/conference_registrations_controller.rb
@@ -27,8 +27,9 @@ class ConferenceRegistrationsController < ApplicationController
   end
 
   def show
-    @total_price = Ticket.total_price(@conference, current_user, paid: true)
+    @total_price = Ticket.total_price_user(@conference, current_user, paid: true)
     @tickets = current_user.ticket_purchases.by_conference(@conference).paid
+    @total_price_per_ticket = @tickets.group(:ticket_id).sum('amount_paid * quantity')
     @ticket_payments = @tickets.group_by(&:ticket_id)
     @total_quantity = @tickets.group(:ticket_id).sum(:quantity)
   end
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -55,6 +55,11 @@ class Ticket < ActiveRecord::Base
     result ? result : Money.new(0, 'USD')
   end
 
+  def self.total_price_user(conference, user, paid: false)
+    tickets = TicketPurchase.where(conference: conference, user: user, paid: paid)
+    tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
+  end
+
   def tickets_sold
     ticket_purchases.paid.sum(:quantity)
   end
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -60,12 +60,13 @@ class Ticket < ActiveRecord::Base
     tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
   end
 
-  def tickets_sold
-    ticket_purchases.paid.sum(:quantity)
+  def tickets_turnover_total(id)
+    tickets = TicketPurchase.where(ticket_id: id)
+    tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
   end
 
-  def tickets_turnover
-    tickets_sold * price
+  def tickets_sold
+    ticket_purchases.paid.sum(:quantity)
   end
 
   private
--- a/app/models/ticket_purchase.rb
+++ b/app/models/ticket_purchase.rb
@@ -52,7 +52,8 @@ class TicketPurchase < ActiveRecord::Base
       purchase = new(ticket_id: ticket.id,
                      conference_id: conference.id,
                      user_id: user.id,
-                     quantity: quantity)
+                     quantity: quantity,
+                     amount_paid: ticket.price)
       purchase.pay(nil) if ticket.price_cents.zero?
     end
     purchase
--- a/app/views/admin/physical_tickets/_physical_ticket.html.haml
+++ b/app/views/admin/physical_tickets/_physical_ticket.html.haml
@@ -1,0 +1,16 @@
+%tr
+  %td= physical_ticket.id
+  %td= physical_ticket.ticket.title
+  %td= physical_ticket.user.email
+  %td= humanized_money_with_symbol physical_ticket.ticket_purchase.amount_paid
+  %td
+    .btn-group
+      = link_to 'Show',
+      conference_physical_ticket_path(conference.short_title,
+                                      physical_ticket.token),
+      class: 'btn btn-primary'
+      = link_to 'Generate PDF',
+      conference_physical_ticket_path(conference.short_title,
+                                      physical_ticket.token,
+                                      format: :pdf),
+      class: 'button btn btn-default btn-info'
--- a/app/views/admin/tickets/index.html.haml
+++ b/app/views/admin/tickets/index.html.haml
@@ -27,7 +27,7 @@
               %td
                 = ticket.tickets_sold
               %td
-                = humanized_money_with_symbol ticket.tickets_turnover
+                = humanized_money_with_symbol ticket.tickets_turnover_total(ticket.id)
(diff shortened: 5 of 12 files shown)


rahul2240:
this PR is incomplete as the paid amount here is changing with the change in price of tickets.

rahul2240:
@Ana06 After sending the PR i realised that the issue is still not fixed, i am working on it.will send PR asap.Sorry for this incomplete PR.

rahul2240:
@Ana06 I have added a new column to ticket purchase(locally) and i am able to store paid amount in it but the problem comes when we try to show turnover in admin/tickets#index as turnover is quantity*ticket price and in conference_registrations#show because there the user will see the current price of the ticket not the price that he paid.So i am thinking of removing these for now .
I can add the actual turnover but need guidance in that.

Ana06:
@rahul2240 

>  After sending the PR i realised that the issue is still not fixed, i am working on it.will send PR asap.Sorry for this incomplete PR.

No worries, it is not that big issue :smile: 

> @Ana06 I have added a new column to ticket purchase(locally) and i am able to store paid amount in it but the problem comes when we try to show turnover in admin/tickets#index as turnover is quantity*ticket price and in conference_registrations#show because there the user will see the current price of the ticket not the price that he paid.So i am thinking of removing these for now .
I can add the actual turnover but need guidance in that.

We should always show what the user really paid, right? Do not remove it, you can fix it! :wink: What do you exactly need guidance with?

rahul2240:
admin_ticket_paid
ticket_urnover
user_register_ticket

rahul2240:
@Ana06 I just wanted to ask if these changes are fine or not so that i can work on tests.
Again sorry for incomplete PR but i wanted these changes to get confirmed first.

rahul2240:
@siddhantbajaj @Ana06 Let me know if i have done something wrong.

db/migrate/20170918160919_add_amount_paid_to_ticket_purchase.rb
```diff
@@ -0,0 +1,5 @@
+class AddAmountPaidToTicketPurchase < ActiveRecord::Migration
+  def change
+    add_column :ticket_purchases, :amount_paid, :float
```
Ana06: as said in https://github.com/openSUSE/osem/issues/1682#issuecomment-330082124 I would better add this in `ticket_purchase`

app/views/conference_registrations/show.html.haml
```diff
@@ -91,17 +91,13 @@
...
         %ul
           .col-md-12
             - @ticket_payments.each_pair do |ticket_id, tickets|
               %li
                 = @total_quantity[ticket_id]
                 = tickets.first.title
                 = word_pluralize(@total_quantity[ticket_id], 'Ticket')
-                for
-                = tickets.first.price.symbol
-                = humanized_money tickets.first.price
```
Ana06: what abour removing `tickets_turnover` method?

app/views/admin/tickets/index.html.haml
```diff
@@ -27,8 +26,6 @@
               %td
                 = ticket.tickets_sold
               %td
-                = humanized_money_with_symbol ticket.tickets_turnover
```
Ana06: why not showing the correct price instead of removing this? :confused:

app/models/ticket.rb
```diff
@@ -55,6 +55,18 @@ def self.total_price(conference, user, paid: false)
     result ? result : Money.new(0, 'USD')
   end
 
+  def self.total_price_user(conference, user, paid: false)
+    tickets = Ticket.where(conference_id: conference.id)
```
Ana06: you could write the shole method like:

``` Ruby
tickets = TicketPurchase.where(conference: conference, ticket: ticket, user: user)
tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }

File: app/models/ticket.rb [hunk 5]
```diff
@@ -75,6 +75,17 @@ def tickets_turnover
     tickets_sold * price
   end
 
+  def tickets_turnover_total(id)
+    tickets = TicketPurchase.where(ticket_id: id)
+    amount = 0
+    begin
+      tickets.each do |ticket|
```
Ana06: you can also use `inject` here.

app/models/ticket.rb
```diff
@@ -75,6 +75,17 @@ def tickets_turnover
     tickets_sold * price
   end
 
+  def tickets_turnover_total(id)
+    tickets = TicketPurchase.where(ticket_id: id)
+    amount = 0
+    begin
```
Ana06: why do you need a transaction? :thinking:

rahul2240: from ticket purchase model, i wanted the quantity and amout_paid values so i used it.

rahul2240:
@Ana06 I dont know how to break the line to pass the line's length failing test.

app/models/ticket_purchase.rb
```diff
@@ -52,7 +52,8 @@ def self.purchase_ticket(conference, quantity, ticket, user)
       purchase = new(ticket_id: ticket.id,
                      conference_id: conference.id,
                      user_id: user.id,
-                     quantity: quantity)
+                     quantity: quantity,
+                     amount_paid: ticket.price)
```
Ana06: what about calling this `price`? :thinking:  In our discussion this make sense to diferenciate this from the ticket price, but I doubt that this is easily understandable by someone else outside the discussion when reading this in some months.

app/controllers/conference_registrations_controller.rb
```diff
@@ -27,8 +27,9 @@ def new
   end
 
   def show
-    @total_price = Ticket.total_price(@conference, current_user, paid: true)
+    @total_price = Ticket.total_price_user(@conference, current_user)
```
Ana06: where are you checking now that the tickets that you are taking into account are paid? I think you removed it, in this casem why? :confused:

Ana06: and you could use `@tickets` as a parameter, as that what you want to sum, right?

app/models/ticket.rb
```diff
@@ -60,12 +60,13 @@ def self.total_price_user(conference, user)
     tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
   end
 
-  def tickets_sold
-    ticket_purchases.paid.sum(:quantity)
+  def tickets_turnover_total(id)
+    tickets = TicketPurchase.where(ticket_id: id)
+    tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
```
Ana06: I would have only one method for this line, and then calling it for different tickets, as this way you don't need to duplicate the line, and at least in one on the cases you already have the tickets.

app/views/admin/physical_tickets/_physical_ticket.html.haml
```diff
@@ -0,0 +1,16 @@
+%tr
+  %td= physical_ticket.id
+  %td= physical_ticket.ticket.title
+  %td= physical_ticket.user.email
+  %td= humanized_money_with_symbol physical_ticket.ticket_purchase.amount_paid
+  %td
+    .btn-group
+      = link_to 'Show',
+      conference_physical_ticket_path(@conference.short_title,
```
Ana06: you shouldn't use `@conference` here, that's why rubocop is complaining. You can pass the conference title as a parameter when calling the partial :wink:

app/models/ticket_purchase.rb
rahul2240: I think calling it price is better, but i used amount_paid so that others can differentiate it from ticket price.I will change it to price.

app/views/admin/physical_tickets/_physical_ticket.html.haml
rahul2240: Ok thanks. :smile:

app/controllers/conference_registrations_controller.rb
rahul2240: Removing paid: true is a mistake.
I will use @tickets :smile:  thanks

rahul2240:
@Ana06 please review.

Issue #1702, opened 2017-09-16
Total amount is not showing cents


rahul2240:
if the price of the ticket is 1.53 then in total only 1 is visible
![total](https://user-images.githubusercontent.com/25618739/30512650-d401441e-9b11-11e7-9fee-9d6b4e844c25.png)

differentreality:
Yeap, for total we are only adding the integer part of the number.

rahul2240:
@differentreality thanks.Will send the PR asap.

PR #1708, opened 2017-09-18
Fix total price for tickets

rahul2240:
While buying the ticket, user was only able to see the integer part in total.So in javascript parseInt was replaced with parseFloat to show the decimal part too.
Fixes https://github.com/openSUSE/osem/issues/1702


rahul2240:
deci

Issue #1709, opened 2017-09-18
Getting too many digits after decimal point


rahul2240:
While buying a ticket I increase the quantity of the tickets having price in decimal , for simple integers it was fine, but numbers with decimal part, i got this big total. And in most cases i am getting this when my quantity is 3 or above. Please check this.
![decim](https://user-images.githubusercontent.com/25618739/30557021-9421a59a-9cca-11e7-9097-8e0d0494079b.png)
![d](https://user-images.githubusercontent.com/25618739/30557221-31734d76-9ccb-11e7-9697-8ab8a8588c15.png)

PR #1710, opened 2017-09-20
Fix decimal part limit to 2

rahul2240:
while calculating the row total, javascript was called and on some quantity, a number with a big decimal part was appearing.In this PR it is fixed
Fixes https://github.com/opensuse/osem/issues/1709


rahul2240:
ajax

PR #1714, opened 2017-09-26
Upgrade to Rails 5

### LINKED ISSUE #1087: Update to Rails 5

AEtherC0r3:
Update config with rails app:update
Update schema.rb rails db:migrate
Add puma
Make jobs and models inherit from a ApplicationJob and ApplicationRecord
Update acts_as_list to 0.9.7 in order to fix "undefined method `sanitize_sql_hash_for_conditions'" error
Update web-console to 2.3.0 to fix a 500 internal server error
Replace before_filter with before_action
Add rails-controller-testing gem
Add prepend: :true to protect_from_forgery in ApplicationController to avoid ActionController::InvalidAuthenticityToken exceptions
Remove activeuuid
Update formtastic to 3.1.5 to fix deprecation warnings and issues with the Input class
Update ahoy_matey to 1.6.0
Update cancancan to 2.0.0 to fix issues with malformed sql queries
Fix program spec
Fix issue with the picture being nil in admin/Organizations#new and #edit and Organizations#show
Fix ActiveRecord::Base.raise_in_transactional_callbacks= deprecation warning by removing an unnecessary line in application.rb
Fix failing versions specs

Closes #1087

### DIFF
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -1,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
--- a/config/cable.yml
+++ b/config/cable.yml
@@ -6,4 +6,4 @@ test:
 
 production:
   adapter: redis
-  url: redis://localhost:6379/1
+  url: <%= ENV['OSEM_REDIS_URL'] %>
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
+# The `threads` method setting takes two numbers a minimum and maximum.
+# Any libraries that use thread pools should be configured to match
+# the maximum value specified for Puma. Default is set to 5 threads for minimum
+# and maximum, this matches the default thread size of Active Record.
+#
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
+threads threads_count, threads_count
+
+# Specifies the `port` that Puma will listen on to receive requests, default is 3000.
+#
+port        ENV.fetch("PORT") { 3000 }
+
+# Specifies the `environment` that Puma will run in.
+#
+environment ENV.fetch("RAILS_ENV") { "development" }
+
+# Specifies the number of `workers` to boot in clustered mode.
+# Workers are forked webserver processes. If using threads and workers together
+# the concurrency of the application would be max `threads` * `workers`.
+# Workers do not work on JRuby or Windows (both of which do not support
+# processes).
+#
+# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+
+# Use the `preload_app!` method when specifying a `workers` number.
+# This directive tells Puma to first boot the application and load code
+# before forking the application. This takes advantage of Copy On Write
+# process behavior so workers use less memory. If you use this option
+# you need to make sure to reconnect any threads in the `on_worker_boot`
+# block.
+#
+# preload_app!
+
+# The code in the `on_worker_boot` will be called if you are using
+# clustered mode by specifying a number of `workers`. After each worker
+# process is booted this block will be run, if you are using `preload_app!`
+# option you will want to use this block to reconnect to any threads
+# or connections that may have been created at application boot, Ruby
+# cannot share connections between processes.
+#
+# on_worker_boot do
+#   ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
+# end
+
+# Allow puma to be restarted by `rails restart` command.
+plugin :tmp_restart
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -21,7 +21,7 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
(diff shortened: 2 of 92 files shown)


differentreality:
@AEtherC0r3 please rebase

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
bear454: This should be an ENV variable.

config/puma.rb
```diff
@@ -0,0 +1,47 @@
...
+#
+environment ENV.fetch("RAILS_ENV") { "development" }
+
+# Specifies the number of `workers` to boot in clustered mode.
+# Workers are forked webserver processes. If using threads and workers together
+# the concurrency of the application would be max `threads` * `workers`.
+# Workers do not work on JRuby or Windows (both of which do not support
+# processes).
+#
+# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
```
bear454: Please un-comment this line.

config/puma.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# Use the `preload_app!` method when specifying a `workers` number.
+# This directive tells Puma to first boot the application and load code
+# before forking the application. This takes advantage of Copy On Write
+# process behavior so workers use less memory. If you use this option
+# you need to make sure to reconnect any threads in the `on_worker_boot`
+# block.
+#
+# preload_app!
```
bear454: Please uncomment this line.

config/puma.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# The code in the `on_worker_boot` will be called if you are using
+# clustered mode by specifying a number of `workers`. After each worker
+# process is booted this block will be run, if you are using `preload_app!`
+# option you will want to use this block to reconnect to any threads
+# or connections that may have been created at application boot, Ruby
+# cannot share connections between processes.
+#
+# on_worker_boot do
```
bear454: Please uncomment this block.

db/migrate/20140623100942_create_visits.rb
```diff
@@ -1,7 +1,7 @@
 class CreateVisits < ActiveRecord::Migration
   def change
     create_table :visits do |t|
-      t.uuid :visitor_id
+      t.integer :visitor_id
```
bear454: Why?

db/migrate/20140623101032_create_ahoy_events.rb
```diff
@@ -1,7 +1,7 @@
 class CreateAhoyEvents < ActiveRecord::Migration
   def change
     create_table :ahoy_events  do |t|
-      t.uuid :visit_id
+      t.integer :visit_id
```
bear454: Why?

db/migrate/20140623100942_create_visits.rb
AEtherC0r3: activeUUID is incompatible with Rails 5, so when it was removed the the uuid type no longer existed. Initially I replaced it with the integer type because that's ahoy's default and because of #1229. I've since replaced it with binary(16), since that's what it actually is, according to the activeUUID code

config/cable.yml
AEtherC0r3: I can't find documentation about how to replace the yaml config with environment variables. (My assumption is that, since this isn't a .rb file, it isn't executable, so I can't just write something like `url: ENV['REDIS_URL']`) Can you point me to the right direction?

AEtherC0r3: Thanks

AEtherC0r3:
@bear454 If we don't change those migrations then they will be broken. And Travis relies on the migrations running successfully to work https://github.com/openSUSE/osem/blob/master/.travis.yml#L25

bear454:
Why not use `db:schema:load` instead of `db:migrate` rake task for testing? I don't see any value in running through every migration every time.

differentreality:
I will extract this to a new issue and fix it within the week, so that this PR can proceed.

bear454:
#1794 is resolved by #1797. Please update the migrations @AEtherC0r3

AEtherC0r3:
@Ana06 Using environment variables is just a convention. Is there anything more to it than that, that should be included in the commit's message?

bear454:
@AEtherC0r3 _osem_ is implicitly complying with https://12factor.net/config

differentreality:
So what's the status of this? Is there something pending?

differentreality:
@bear454 you had approved this before the conflict was resolved. Is it good to be merged?

bear454:
@differentreality My last open issue was with changes to the migration history, and @AEtherC0r3 resolved that with a new migration instead, so yes, I will re-approve.

@Ana06 : please re-review as well.

differentreality:
Following the IRC conversation, @bear454 has reviewed and approved the PR. Travis has just passed. 
Since everything is working we should merge this. Merging other PRs is affecting this, making it difficult to rebase. 
I am not sure if more changes were expected, but the reviews don't mention any change to the code itself.

PR #1715, opened 2017-09-26
Added check for payments#new and changed physical_ticket to plural

### LINKED ISSUE #1703: Payments grooming
1. Do not go to payments#new, if amount is 0
Currently you can:
* go to payments#new path directly, which will show that you can pay 0 euro, 
* it will ask for your credit card information (the button will even go green, as if the transaction was successful)
* and then you will be told that 
> Invalid positive integer Please try again with correct credentials.

Relevant screenshots:

screenshot from 2017-09-17 13-25-59
screenshot from 2017-09-17 13-26-12
screenshot from 2017-09-17 13-26-40
screenshot from 2017-09-17 13-26-26

2. physical_tickets#index url is conference/abc/physical_ticket (according to our routes https://github.com/openSUSE/osem/blob/master/config/routes.rb#L125) but it would make sense to be physical_tickets (in plural)

rishabhptr:
Fixes #1703


rishabhptr:
@differentreality Done :+1: and yeah I will look into it.

differentreality:
Thank you @rishabhptr, this is now working nicely :) :+1: 

An improvement I would suggest, is that instead of adding a new function to our controller, we raise a cancancan exception. This can be added directly in the beginning of **new** action, redirect us to root_path, and we can also include our custom message letting the user know what's going on like 'Nothing to pay for!' 

What do you think? Does this make sense for you?

:tada: I will merge this and let you open an issue for this improvement and the payment after conference mentioned in a previous comment. If you want to, you can, of course, work on them as well :)

rishabhptr:
@differentreality Yes,I also think a exception with a suitable redirect message would be a good idea.

Also I cross-checked that the payments#new is accessible for conferences that have ended.I would create a issue for both :)

PR #1716, opened 2017-09-28
using postgresql as the database

stephaniewilkinson:
great project, thank you! I'm making a PR in case you are interested in the code I needed to change to get this running on postgresql. 

At the very least you can keep the `database.yml.postgresql` file I made. Happy to commit to docs if needed as well!


bear454:
There's a lot more here than enabling postgres, most of which we can't accept. Please separate out the pg sample config, and some documentation, into a separate PR.

PR #1723, opened 2017-09-30
error in revision history is removed

### LINKED ISSUE #1722: Error opening revision history
While opening revision history we are getting the error saying **Couldn't find organization**
1. When version.item_type == 'UserRoles' and role.name == 'organization_admin'
2. When version.item_type == 'Role' 
screen shot 2017-09-30 at 12 34 07 pm

namangupta01:
closes #1722


namangupta01:
When the organization is deleted roles will also deleted so it will be look like this instead of error that we were getting before
screen shot 2017-10-03 at 9 57 27 pm

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -224,7 +230,13 @@
 - unless %w(Conference Subscription Registration User Organization).include?(version.item_type)
   - if (version.item_type == 'Role' && role_name == 'organization_admin') || (version.item_type == 'UsersRole' && role.name == 'organization_admin')
     in organization
-    = link_to_organization(version.conference_id)
+    - if Organization.find_by(id: version.conference_id)
+      -# organization_admin belongs to organization and not conferences
+      - organization = Organization.find_by(id: version.conference_id)
+      = link_if_alive version, role.name,
```
bear454: You've changed the content of the link from the organization, to the role_name. This is causing the persistent test failure:

```
expected to find text 
    "added role organization_admin with ID 2 to user name75 in organization Friesen LLC"
in
    "... added role organization_admin with ID 2 to user name75 in organization organization_admin ... "

namangupta01: @bear454 Yes you are right! Correcting it! Thanks!

namangupta01: @bear454 Please have a look!

Issue #1687, opened 2017-09-11
Inform in advance about eventschedule deletion upon change of conference dates


differentreality:
From @kormoc in #1640 

I would like to request that a click though okay/cancel is given that displays the number of events this change will delete before deleting them. It was non-obvious that we had to expand, move, and then contract to change hour windows.

Thanks!

rahul2240:
can i take this?

differentreality:
@rahul2240 you can, if you want to :)

rahul2240:
@differentreality  I have to show the events that are going to delete when the conference dates are changed.Nothing else right? Please correct me if i am wrong.

differentreality:
Showing all events is an overkill (and can be a lot of data to show on a pop up). 

We should show something along those lines:
'Are you sure you want to make to change conference hours? You have already scheduled content, which will be unscheduled if you proceed. Do you wish to proceed?'

rahul2240:
Thanks 😄 I will do it.

PR #1730, opened 2017-10-02
Add popup message before eventschedule hours change

rahul2240:
Pop up window is added before editing the conference schedule hours to let the organizer know that it will unschedule the scheduled events
Closes https://github.com/openSUSE/osem/issues/1687


rahul2240:
@differentreality Please check the PR and i don't think these failures are due to my changes.

differentreality:
I have another suggestion about how to implement this:
1. Add a hint on the fields
2. Add a confirm message to save button

What do you think about this @rahul2240 ?

rahul2240:
@differentreality That will be better. 😄

rahul2240:
@differentreality  ok now?

differentreality:
@rahul2240 my suggestion was that we add a confirmation on save button, instead of the javascript popups.
Also please add new commits to the PR with your changes.

PR #1734, opened 2017-10-02
Added amount check for payments#new 

### LINKED ISSUE #1733: Use CanCan Exception to handle payments#new amount check
Instead of using a new function in the controller ,raise a CanCan exception for the same with a suitable redirect message.

rishabhptr:
Fixes #1733


differentreality:
Nice @rishabhptr :+1: Thank you, merging it :tada:

PR #1745, opened 2017-10-08
DS_Store file is added into the gitignore

### LINKED ISSUE #1744: .DS_Store file should be added in gitignore
In the Apple macOS,  **.DS_Store ( Desktop Services Store )** is a file that stores custom attributes of its containing folder such as the position of icons or the choice of a background image. It is created by **Finder** in folder that a user works with.

screen shot 2017-10-08 at 12 11 06 pm

Which should be added in **gitignore**

namangupta01:
closes #1744

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ pickle-email-*.html
 .env.local
 docker-compose.env
 docker-compose.yml
+.DS_Store


namangupta01:
screen shot 2017-10-08 at 12 24 58 pm

PR #1752, opened 2017-10-10
Added check for conference end_date in payments#new

### LINKED ISSUE #1735: Payment for tickets are still possible after conference end date
User can still make payment and edit purchase if there were any unpaid tickets even after the conference has ended.

rishabhptr:
Fixes #1735


app/controllers/payments_controller.rb
```diff
@@ -10,6 +10,9 @@ def index
 
   def new
     @total_amount_to_pay = Ticket.total_price(@conference, current_user, paid: false)
+    if @conference.end_date < Date.today
```
differentreality: I think this needs to be added in ability.rb and let cancancan handle things from there. 
This is not a link we would use directly anywhere, anyway, so we care less about the message. Cancancan will show the default message 'You are not authorized to access this page'

PR #1756, opened 2017-10-11
Add vendor picture interface; include in splash

bear454:
Vendor has a picture attribute, and was displayed in one form of the splashpage, but it wasn't exposed in the vendor editing form, nor was it included in the "map view" on the splash page.

Additionally, the map popup was rendered inline.

This commit:

* Adds a from input for editing Venue#picture, consistent with other form elements
* Displays the picture on the splash page in either style (map or static)
* Moves the splash page map popup from inline HTML to it's own partial


bear454:
Sample render:
screenshot from 2017-10-11 16-53-54

differentreality:
That looks so much better :D

differentreality:
That looks so much better :D

bear454:
Really rubocop?
```
app/views/conferences/_venue_map_marker.html.haml:2 [W] MultilinePipe: Don't use the `|` character to split up lines. Wrap on commas or extract code into helper.
```
:-1:

bear454:
@differentreality one more review, please ;-)

PR #1757, opened 2017-10-12
Fix missing link on public organizations page

bear454:
When viewing a conference, the Instance Name in the header points to /organizations. On the organizations index, there's a button to see conferences, but it does nothing. This makes it easy for public users to get lost.

This commit:
* Adds a function to the organizations controller to show child conferences
* Adds a public permission to see conferences through an organization
* Represents a subset of conferences, reusing conferences/index view, for the organization

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -4,4 +4,10 @@ class OrganizationsController < ApplicationController
   def index
     @organizations = Organization.all
   end
+
+  def conferences
+    @current = @organization.conferences.upcoming.reorder(start_date: :asc)
+    @antiquated = @organization.conferences.past
+    render '/conferences/index'
+  end
 end
--- a/app/views/conferences/index.html.haml
+++ b/app/views/conferences/index.html.haml
@@ -4,7 +4,7 @@
       .page-header
         %h2 Upcoming Conferences
   - @current.each do |conference|
-    = render partial: 'conference_details', locals: { conference: conference }
+    = render '/conferences/conference_details', conference: conference
   -if @antiquated and @antiquated.any?
     .row
       .col-md-12
@@ -17,7 +17,7 @@
             %i.fa.fa-chevron-down{ style: 'display: none' }
     #antiquated.collapse
       - @antiquated.each do |conference|
-        = render partial: 'conference_details', locals: { conference: conference}
+        = render '/conferences/conference_details', conference: conference
 
 -content_for :script_body do
   :javascript
--- a/app/views/organizations/index.html.haml
+++ b/app/views/organizations/index.html.haml
@@ -12,5 +12,7 @@
           .caption
             %h4
               = organization.name
-            %button.btn.btn-success Conferences
+            = link_to 'Conferences',
+              conferences_organization_path(organization),
+              class: 'btn btn-success'
             / = link_to 'Edit', edit_organization_path(organization), class: 'btn btn-mini btn-default'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -149,7 +149,11 @@ Osem::Application.routes.draw do
     get '/revision_history/:id/revert_object' => 'versions#revert_object', as: 'revision_history_revert_object'
     get '/revision_history/:id/revert_attribute' => 'versions#revert_attribute', as: 'revision_history_revert_attribute'
   end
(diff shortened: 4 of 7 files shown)


app/views/conferences/index.html.haml
```diff
@@ -4,7 +4,7 @@
       .page-header
         %h2 Upcoming Conferences
   - @current.each do |conference|
-    = render partial: 'conference_details', locals: { conference: conference }
+    = render partial: '/conferences/conference_details', locals: { conference: conference }
```
differentreality: This change would also have to take place in L20 of this file

.byebug_history
```diff
@@ -0,0 +1,23 @@
+c
```
differentreality: Isn't that file supposed to be in .gitignore?

spec/features/organization_spec.rb
```diff
@@ -48,4 +48,6 @@
 
     it_behaves_like 'successfully updates an organization'
   end
+
+  context ''
```
differentreality: Are you going to add sth here?

spec/controllers/organizations_controller_spec.rb
```diff
@@ -12,4 +32,27 @@
...
+  describe 'GET #conferences' do
+    before :each do
+      get :conferences, organization_id: organization.id
+    end
+
+    it 'loads the organization' do
+      expect(assigns(:organization)).to eq organization
+    end
+
+    it 'includes organizational conferences' do
```
differentreality: Perhaps easier to understand if it is simply put as `includes organization conferences`

differentreality:
This looks good :) 

I would still like to show a link to homepage instead of the organizations link on the top left corner though. https://github.com/openSUSE/osem/issues/1651

bear454:
@differentreality I'm addressing that separately, in #1771

spec/features/organization_spec.rb
bear454: Whoops.

.byebug_history
bear454: It is now; I'll update the branch.

bear454:
@differentreality please review (and merge) when you have time.

app/controllers/organizations_controller.rb
```diff
@@ -4,4 +4,11 @@ class OrganizationsController < ApplicationController
   def index
     @organizations = Organization.all
   end
+
+  def conferences
+    @organization = Organization.find(params[:organization_id])
```
differentreality: This will be loaded by cancancan on https://github.com/bear454/osem/blob/636437f277689ea1ff2d1fea87288a7829315e14/app/controllers/organizations_controller.rb#L2 if the params have id instead of organization_id, which depends on the routes.

config/routes.rb
```diff
@@ -149,7 +149,9 @@
     get '/revision_history/:id/revert_object' => 'versions#revert_object', as: 'revision_history_revert_object'
     get '/revision_history/:id/revert_attribute' => 'versions#revert_attribute', as: 'revision_history_revert_attribute'
   end
-  resources :organizations, only: [:index]
+  resources :organizations, only: [:index] do
+    get :conferences
```
differentreality: How about 
```
member do
  get :conferences
done
```
so that CanCanCan works?

bear454: @differentreality updated in https://github.com/openSUSE/osem/pull/1757/commits/09c43b1e34a471a48f561bb50e6ba474ef0942a3

app/views/organizations/index.html.haml
```diff
@@ -12,5 +12,7 @@
           .caption
             %h4
               = organization.name
-            %button.btn.btn-success Conferences
+            = link_to 'Conferences',
+              organization_conferences_path(organization),
```
differentreality: conferences_organization_path

bear454: That's not a valid route.

```
> bundle exec rake routes
...
organization_conferences GET         /organizations/:organization_id/conferences(.:format)                                              organizations#conferences
....
```

bear454: Never mind; I was on the wrong branch. I'll write a test to cover that. Nice catch!

bear454:
@differentreality please re-review

PR #1759, opened 2017-10-12
Require a version of nokogiri with known vulnerabilities resolved

bear454:


bear454:
@differentreality - please re-review

PR #1761, opened 2017-10-12
Trying travis without sudo - should run in container

bear454:
Just want to see if things still work ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 dist: trusty
 language: ruby
 cache: bundler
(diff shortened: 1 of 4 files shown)


bear454:
Container build:

screenshot from 2017-10-12 09-43-18

VM build:

screenshot from 2017-10-12 09-45-00

bear454:
So, we don't need to run in a Travis VM at this point, but it's _surprisingly_ faster to do so.

hennevogel:
I guess because there are way more people hammering on the container infrastructure :-)

bear454:
@hennevogel no, our test suite is just slow. I'm looking at approaches to get us < 5 minute runs.

PR #1762, opened 2017-10-12
Speaker emails in Events CSV/XLS export is included

### LINKED ISSUE #1743: Include speaker email in "Events" admin CSV/XLS export
This would make the export useful for emailing groups of speakers, e.g. those with Confirmed events.

Here's my workaround (no I did not write this on my own):

```sql
SELECT e.id
       ,e.title
       ,e.state
       ,u.email
  FROM events AS e
  JOIN event_users eu ON eu.event_id = e.id
  JOIN users u        ON u.id = eu.user_id
  JOIN programs p     ON p.id = e.program_id
  JOIN conferences c  ON c.id = p.conference_id
   AND c.short_title = 'seagl2017'
   AND e.state       = 'confirmed'
```

namangupta01:
Closes #1743

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -213,11 +213,7 @@ class Event < ActiveRecord::Base
 
   # Returns emails of all the speaker belongs to a particular event
   def speaker_emails
-    result = []
-    speakers.each do |speaker|
-      result << speaker.email
-    end
-    result.to_sentence
+    speakers.map(&:email).join(', ')
   end
 
   ##
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -211,6 +211,15 @@ class Event < ActiveRecord::Base
     result.to_a.to_sentence
   end
 
+  # Returns emails of all the speaker belongs to a particular event
+  def speaker_emails
+    result = Array.new
+    speakers.each do |speaker|
+      result << speaker.email
+    end
+    result.to_sentence
+  end
+
   ##
   #
   # Returns +Hash+
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -213,7 +213,7 @@ class Event < ActiveRecord::Base
 
   # Returns emails of all the speaker belongs to a particular event
   def speaker_emails
-    result = Array.new
+    result = []
     speakers.each do |speaker|
       result << speaker.email
     end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -204,11 +204,7 @@ class Event < ActiveRecord::Base
   end
 
   def speaker_names
-    result = Set.new
-    speakers.each do |speaker|
-      result.add(speaker.name)
-    end
-    result.to_a.to_sentence
+    speakers.map(&:name).join(', ')
   end
 
   # Returns emails of all the speaker belongs to a particular event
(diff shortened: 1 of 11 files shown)


namangupta01:
screen shot 2017-10-13 at 12 08 02 am

app/models/event.rb
```diff
@@ -211,6 +211,15 @@ def speaker_names
     result.to_a.to_sentence
   end
 
+  # Returns emails of all the speaker belongs to a particular event
+  def speaker_emails
+    result = []
+    speakers.each do |speaker|
+      result << speaker.email
+    end
+    result.to_sentence
```
differentreality: This can be done simply like this:
`speakers.map(&:email).join(', ')`

app/views/admin/events/_all_with_comments.xlsx.axlsx
```diff
@@ -1,8 +1,20 @@
 wb.use_shared_strings = true
 wb.add_worksheet(name: 'events with comments') do |sheet|
-  bold_style = wb.styles.add_style(b: true )
+  bold_style = wb.styles.add_style( b: true )
```
differentreality: Space is _not_ needed inside parenthesis, a minor note, but I thought I would mention it since you are trying to keep everything in the right style :)

differentreality:
Why are we adding the emails only for xls and csv but not pdf export?

namangupta01:
Yes we can add it for the pdf as well. :)

app/models/event.rb
namangupta01: I think i can do same in the speaker_names

bear454:
I'm not a fan of the `to_sentence` format for data exports. IMO there's an assumption with the CSV/XLS that it will be consumed/processed in some way, and that format will make doing so more difficult. I'd rather see a consistent separator between the data points.

app/models/event.rb
```diff
@@ -211,6 +211,11 @@ def speaker_names
     result.to_a.to_sentence
   end
 
+  # Returns emails of all the speaker belongs to a particular event
+  def speaker_emails
+    speakers.map(&:email).join(', ')
```
differentreality: Let's do this for names as well, on a separate commit.

differentreality:
Don't forget the failing tests @namangupta01

differentreality:
@namangupta01 please take care of your PR :)

namangupta01:
@differentreality yes working on it! I just have a doubt that how that we can set starting time for event?

namangupta01:
@differentreality As there are so much of columns in this export so all the column size can't be made in accordance with the real world data.
screen shot 2017-11-07 at 11 38 09 pm

differentreality:
@namangupta01 what's your doubt about start time? What are you thinking? Have you tried something? What troubles you exactly?

namangupta01:
While creating events it asks for all the details except start time so i was just wondering how to set start time for events.

differentreality:
@namangupta01 start time is only set after you schedule the event. So you need to extract that information from the event_schedule of the event. The event_schedule record needs to be for the selected schedule of the conference. Check it out, and let me know if you cannot figure it out.

namangupta01:
@differentreality Thank You! I got it!

namangupta01:
This is the best combination that i got.
screen shot 2017-11-08 at 12 06 34 am

namangupta01:
Here's hows it looks now --->
xls:
screen shot 2017-11-08 at 12 38 26 am
pdf:
screen shot 2017-11-08 at 12 06 34 am

namangupta01:
@differentreality Test is fixed please have a look!

differentreality:
OK let's squash the commits and get this in :+1:

PR #1765, opened 2017-10-13
fixed openid username

### LINKED ISSUE #1747: Set a more meaningful openid username
When creating a new account with openid, the username depends on whether or not the service is returning a username field (auth.info.username), if not the username becomes like 1234@facebook (uuid@provider), which does not mean anything to users reading it, and conveys no information whatsoever regarding who this user is. 

We should setup the username during user creation in a more meaningful way, such as: 
If the email is user_stella@gmail.com, the username should become user_stella@facebook.

ViditChitkara:
closes #1747

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -11,13 +11,13 @@ module Users
 
     def handle(provider)
       auth_hash = request.env['omniauth.auth']
-      uid = auth_hash[:uid]
+      username = auth_hash.info.email.split('@')[0]
       openid = Openid.find_for_oauth(auth_hash) # Get or create openid
       # If openid exists and is associated with a user, sign in with associated user,
       # even if the email of the associated user and the email of the provided openid are different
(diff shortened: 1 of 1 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -11,13 +11,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     def handle(provider)
       auth_hash = request.env['omniauth.auth']
-      uid = auth_hash[:uid]
+      username = auth_hash.info.email.split("@")[0]
```
differentreality: Use single quotes here, it's a styling thing, and the reason why linters have failed

ViditChitkara: Sure @differentreality

differentreality:
:+1: Nice @ViditChitkara

ViditChitkara:
ok

differentreality:
Thank you @ViditChitkara :tada:

PR #1769, opened 2017-10-17
Add hint and popup on conference#edit

### LINKED ISSUE #1687: Inform in advance about eventschedule deletion upon change of conference dates

rahul2240:
Pop up and hint are added to let the user know that changing conference hours will unschedule their scheduled changes.
Closes  #1687

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -163,6 +163,12 @@ module ApplicationHelper
     end
   end
 
+  def rescheduling_hint(affected_event_count)
+    if affected_event_count > 0
+      "You have #{affected_event_count} scheduled #{'event'.pluralize(affected_event_count)}. Changing the conference hours will unschedule those scheduled outside the conference hours."
+    end
+  end
+
   ##
   # ====Gets
   # a conference object
--- a/app/views/admin/conferences/edit.html.haml
+++ b/app/views/admin/conferences/edit.html.haml
@@ -22,8 +22,8 @@
         = f.input :timezone, as: :time_zone, hint: 'The conference time zone'
         = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', readonly: 'readonly' }
         = f.input :end_date, as: :string, input_html: { id: 'conference-end-datepicker', readonly: 'readonly' }
-        = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}
-        = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
+        = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}, hint: rescheduling_hint(@affected_event_count)
+        = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}, hint: rescheduling_hint(@affected_event_count)
       = f.inputs name: 'Registrations' do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events (they will still be able to register even if it has been reached). You currently have ' + pluralize(@conference.registrations.count, 'registration')
       = f.inputs name: 'Booths' do
--- a/app/views/admin/conferences/edit.html.haml
+++ b/app/views/admin/conferences/edit.html.haml
@@ -29,4 +29,4 @@
       = f.inputs name: 'Booths' do
         = f.input :booth_limit, as: :number, in: 0..9999,
         hint: 'Booth limit is the maximum number of booths that you can accept for this conference. By setting this number (0 no limit) you can be sure that you are not going to accept more booths than the conference can accommodate. You currently have ' + pluralize(@conference.booths.accepted.count, 'accepted booth') +'.'
-      = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}
+      = f.action :submit, as: :button, button_html: { class: 'btn btn-primary', data: { confirm: 'Are you sure you want to proceed?' } }
--- a/app/controllers/admin/conferences_controller.rb
+++ b/app/controllers/admin/conferences_controller.rb
@@ -193,6 +193,7 @@ module Admin
     def edit
       @conferences = Conference.all
       @date_string = date_string(@conference.start_date, @conference.end_date)
+      @affected_event_count = @conference.program.events.scheduled(@conference.program.selected_schedule_id).count
       respond_to do |format|
         format.html
         format.json { render json: @conference.to_json }


app/views/admin/conferences/edit.html.haml
```diff
@@ -22,11 +22,11 @@
...
       = f.inputs name: 'Registrations' do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events (they will still be able to register even if it has been reached). You currently have ' + pluralize(@conference.registrations.count, 'registration')
       = f.inputs name: 'Booths' do
         = f.input :booth_limit, as: :number, in: 0..9999,
         hint: 'Booth limit is the maximum number of booths that you can accept for this conference. By setting this number (0 no limit) you can be sure that you are not going to accept more booths than the conference can accommodate. You currently have ' + pluralize(@conference.booths.accepted.count, 'accepted booth') +'.'
-      = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}
+      = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}, onclick: 'window.confirm("Are you sure you want to proceed?")'
```
differentreality: Why are you not using `data: { confirm: 'message' }` instead?

app/views/admin/conferences/edit.html.haml
```diff
@@ -22,11 +22,11 @@
         = f.input :timezone, as: :time_zone, hint: 'The conference time zone'
         = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', readonly: 'readonly' }
         = f.input :end_date, as: :string, input_html: { id: 'conference-end-datepicker', readonly: 'readonly' }
-        = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}
-        = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
+        = f.input :start_hour, input_html: { size: 2, type: 'number', min: 0, max: 23}, hint: 'Changing the conference hours will unschedule your scheduled events.'
```
differentreality: Since you fixed styling for the first curly quote, please add a space before the last one as well. Same goes for the next line, as well as for L32

app/views/admin/conferences/edit.html.haml
```diff
@@ -22,11 +22,11 @@
...
       = f.inputs name: 'Registrations' do
         = f.input :registration_limit, as: :number, in: 0..9999, hint: 'Limit the number of registrations to the conference (0 no limit). Please note that the registration limit doesn\'t apply to speakers of confirmed events (they will still be able to register even if it has been reached). You currently have ' + pluralize(@conference.registrations.count, 'registration')
       = f.inputs name: 'Booths' do
         = f.input :booth_limit, as: :number, in: 0..9999,
         hint: 'Booth limit is the maximum number of booths that you can accept for this conference. By setting this number (0 no limit) you can be sure that you are not going to accept more booths than the conference can accommodate. You currently have ' + pluralize(@conference.booths.accepted.count, 'accepted booth') +'.'
-      = f.action :submit, as: :button, button_html: {class: 'btn btn-primary'}
+      = f.action :submit, as: :button, button_html: {class: 'btn btn-primary', data: { confirm: 'Are you sure you want to proceed?' } }
```
differentreality: Let's include on that message the useful information as well. Eg.

> 20 event are going to be unscheduled. Are you sure you want to proceed?

app/views/admin/conferences/edit.html.haml
rahul2240: ok 😄

app/views/admin/conferences/edit.html.haml
rahul2240: Yes, it will be better 😄

differentreality:
@rahul2240 waiting for you to push the updated code :)

rahul2240:
@differentreality Sorry for the delay. But i am still trying to figure out the way of counting number of events that are going to unschedule.

differentreality:
Isn't that gonna be something like program.events.scheduled(program.selected_schedule_id)?

rahul2240:
@differentreality  but this will return all the data present in EventSchedule for particular id ? But on changing start hour or end hour only those events get unschedule that doesn't fall within the range.
Please correct me if i am wrong.

rahul2240:
What if user doesn't edit start or end hour. Also on changing start  hour or end hour only those events get unscheduled that doesn't come between the range of start and end hour.

rahul2240:
@differentreality I think 1 possible solution is to fetch the value of start hour and end hour and use these values to find out the events from conference.program.event_schedules that doesnt fall in between the values of start hour and end hour and show these values to pop up.
Please let me know what you think of this solution. :)

rahul2240:
@differentreality Please review.

app/views/admin/conferences/edit.html.haml
```diff
@@ -22,8 +22,12 @@
         = f.input :timezone, as: :time_zone, hint: 'The conference time zone'
         = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', readonly: 'readonly' }
         = f.input :end_date, as: :string, input_html: { id: 'conference-end-datepicker', readonly: 'readonly' }
-        = f.input :start_hour, input_html: {size: 2, type: 'number', min: 0, max: 23}
-        = f.input :end_hour, input_html: {size: 2, type: 'number', min: 1, max: 24}
+        - if @conference.program.events.scheduled(@conference.program.selected_schedule_id).count!=0
```
bear454: Consider moving the count calculation back to the controller, e.g:
```
@affected_event_count = @conference.program.events.scheduled(@conference.program.selected_schedule_id).count
```

And adding a helper for the conditional hint:

```
# ApplicationHelper
def rescheduling_hint(affected_event_count)
  if affected_event_count > 0
    "You have #{affected_event_count} scheduled events. Changing the conference hours will unschedule those scheduled outside the conference hours."
  end
end
```

Then, all you need in the view is:

```
= f.input :start_hour, input_html: { size: 2, type: 'number', min: 0, max: 23 }, hint: rescheduling_hint(@affected_event_count)
= f.input :end_hour, input_html: { size: 2, type: 'number', min: 1, max: 24 }, hint: rescheduling_hint(@affected_event_count)
```

Less logic in the views is better.

rahul2240: :smile: thanks

app/helpers/application_helper.rb
```diff
@@ -163,6 +163,12 @@ def sign_in_path
     end
   end
 
+  def rescheduling_hint(affected_event_count)
+    if affected_event_count > 0
+      "You have #{affected_event_count} scheduled events. Changing the conference hours will unschedule those scheduled outside the conference hours."
```
differentreality: I'd use pluralize for the word events, so that: 
You have 1 scheduled events => You have 1 scheduled event

rahul2240:
@differentreality  I have added pluralize in a new commit. Please review 😄

differentreality:
That's great @rahul2240 please squash your commits and rebase so that we can merge the PR!

rahul2240:
@differentreality i have squashed the commits

differentreality:
Good work @rahul2240 :+1: 
Thank you! :tada:

PR #1771, opened 2017-10-19
Update navbar link for "one-organization" profile

### LINKED ISSUE #1651: Browsing is confusing with organization name on the top left navbar

bear454:
When osem is run with only one organization, it should function roughly the same as it did without organizations; the state in either case is that osem is functionally hosting only one organization, and therefore organizational settings can be at the top-level.

Only If multiple organizations are present, e.g. in a 'hosted' setting, do we need handle the organizational scope.

So, for the top navigational bar, only show the organizational link if there is more than one organization; otherwise show the top-level site link.

Resolves #1651 

*screenshots from specs*

Before:
capybara-201405010001114928339693

After:
capybara-201405010001136080047312

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,6 +37,7 @@ class ApplicationController < ActionController::Base
 
   def get_conferences
     @conferences = Conference.all
+    @one_organization = (Organization.count <= 1)
   end
 
   def current_ability
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -171,4 +171,18 @@ module ApplicationHelper
   def hidden_if_conference_over(conference)
     'hidden' if Date.today > conference.end_date
   end
+
+  def nav_root_link_for(conference)
+    link_text = (
+      conference.try(:organization).try(:name) ||
+      ENV['OSEM_NAME'] ||
+      'OSEM'
+    )
+    link_to(
+      link_text,
+      root_path,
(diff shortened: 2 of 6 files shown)


bear454:
@differentreality please review (and merge if approved) when you have time.

differentreality:
I think the homepage link should be there (on the top left corner) in both/all cases
More details: https://github.com/openSUSE/osem/pull/1657

bear454:
@differentreality please review again

differentreality:
@bear454 personally I love that the root_path is there at all times. I don't particularly like that the displayed name changes (sometimes it is OSEM sometimes it is the conference's organization name).

I would suggest we show
OSEM_NAME || OSEM  (-> root_path) and Organization name (linking somewhere relevant) next to it.

screenshot from 2017-11-07 17-22-32

What do you/others think?

bear454:
IMO, we don't have much control over this anyway; if we decide to impose a requirement to say "OSEM", anyone can reverse that in their fork... so we might as well create something sustainable, by using a sensible default with an easily configured option to override.

PR #1773, opened 2017-10-20
added email presence validation to openid

### LINKED ISSUE #1658: Openid needs to require email
* Openid model needs to validate for email existence, because user model requires it, and if it is not there, there is no point in creating the openid record without the user record.

That already means we should put all that into a transaction in the users/omniauth controller.

* We need to explicitly request access to user's email when authorizing with openid. If a user does **not** have a public email (I stumbled upon that with github), then things go bad. 
Also when that happens, we need to catch this and redirect to our home page with a relevant message (right now it redirects to github and the user has no idea what's going on)

ViditChitkara:
closes #1658 
fix
Here is how it looks now.


ViditChitkara:
Sure @Ana06 , I'll add a test for this.

ViditChitkara:
@Ana06 ,@differentreality  I have written tests for openid, if you could review the changes .

spec/factories/openid.rb
```diff
@@ -0,0 +1,10 @@
+# Read about factories at https://github.com/thoughtbot/factory_girl
+
+FactoryGirl.define do
+  factory :openid do
+    provider 'test'
+    email 'test@test.com'
```
Ana06: You could use Faker to generate the factory fields, so you do not always tests it with the same case.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -11,6 +11,10 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     def handle(provider)
       auth_hash = request.env['omniauth.auth']
+      if auth_hash.info.email.nil?
```
Ana06: what about testing this as well? :thinking:

ViditChitkara:
Sure @Ana06 I'm just about to complete the requested changes. So I'll squash them together.

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: @Ana06 If have written tests for this.

spec/factories/openid.rb
ViditChitkara: Done

ViditChitkara:
@Ana06 I have made the requested changes. Since I am new to writing tests, it would be really helpful of you if could see whether I have written the tests in a correct way or not.

ViditChitkara:
@differentreality shouldn't there also be some validation at schema level, since a user can create a user using a query statement ?

differentreality:
@ViditChitkara you don't necessarily need to implement db constraints for every model validation you have. I would not say it is necessary to do so. Users are supposed to use the application to do stuff, not use sql queries directly on the database.

ViditChitkara:
Ok @differentreality

spec/features/omniauth_spec.rb
```diff
@@ -49,6 +49,17 @@
       expect(flash).to eq('Could not authenticate you from Google because "Invalid credentials".')
     end
 
+    scenario 'should not signup when email is not there in auth hash' do
```
Ana06: why a feature test instead a controller test? :thinking:

ViditChitkara: @Ana06 I guess there is already a file of openid in features so I thought it would be good to write the tests there only. Also I noticed that all flash message related tests were written in feature folder. If required want I can make it in controllers too. Please correct me if I'm wrong!

app/controllers/users/omniauth_callbacks_controller.rb
differentreality: Is .nil? actually need for some reason? I would expect `if auth_hash.info.email ` to work just fine. Can you please double check?

spec/features/omniauth_spec.rb
differentreality: In the feature test I would be testing that no new Openid record is created.
I would add a controller test to test the redirection.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -11,6 +11,10 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+        flash[:error] = "Email field is missing in your #{provider} account"
+        return redirect_to new_user_registration_path
```
differentreality: Why return?

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: Yes, there is a need to check if email is nil. Instead of the highlighted line, I could use 
"unless auth_hash.info.email"

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: That was indeed by mistake.I'll correct it.

spec/features/omniauth_spec.rb
ViditChitkara: @differentreality Got it!!I'll make the changes asap

app/controllers/users/omniauth_callbacks_controller.rb
differentreality: @ViditChitkara yes, let's do that!

app/controllers/users/omniauth_callbacks_controller.rb
differentreality: Please do

spec/models/openid.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+
+describe Openid do
+  subject { create(:openid) }
+
+  describe 'validation' do
+    it { should validate_presence_of(:provider) }
```
differentreality: Please use is_expected.to instead of should

spec/features/omniauth_spec.rb
differentreality: Please do add/amend the tests, so that we can merge it! Thanks @ViditChitkara

ViditChitkara:
@differentreality apologies for the delay, will make the changes asap!!

spec/models/openid.rb
ViditChitkara: Okay

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: done

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: done

spec/models/openid.rb
```diff
@@ -0,0 +1,15 @@
...
+  subject { create(:openid) }
+
+  describe 'validation' do
+    it { is_expected.to validate_presence_of(:provider) }
+    it { is_expected.to validate_presence_of(:uid) }
+    it { is_expected.to validate_presence_of(:email) }
+  end
+
+  describe 'association' do
+    it { should belong_to(:user) }
```
differentreality: what about this? :)

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: @differentreality I realised that this "return" was intentionally written by me, since I was facing problem in redirecting when I removed "return" .
Please see if you could reproduce this error when removing return keyword

ViditChitkara:
@differentreality made the required changes, please review.

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: it was not redirecting, instead gave me an error the following line:-
https://github.com/openSUSE/osem/blob/4785293bf1348549ade1c68ef35f613eaa509a6f/app/controllers/users/omniauth_callbacks_controller.rb#L14

spec/models/openid.rb
ViditChitkara: Done

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -11,6 +11,10 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     def handle(provider)
       auth_hash = request.env['omniauth.auth']
+      unless auth_hash.info.email
```
differentreality: This needs to be 
`unless auth_hash.info.email.present?`

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -11,6 +11,10 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
+        flash[:error] = "Email field is missing in your #{provider} account"
+        return redirect_to new_user_registration_path
```
differentreality: The error you encountered tells you exactly what you need to do to fix it. 
`redirect_to new_user_registration_path and return`

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: Alright!!

app/controllers/users/omniauth_callbacks_controller.rb
ViditChitkara: Yes, this is exactly what I thought. So, I guess this is already correct?Do I need to make any changes in it?

differentreality: Read my comment :)

ViditChitkara:
@differentreality, I have made the changes. Please check

differentreality:
Nice @ViditChitkara please squash your commits!

ViditChitkara:
Sure!!

ViditChitkara:
Done squashing

PR #1774, opened 2017-10-22
Root path renders show page if only one conference exists.

### LINKED ISSUE #1259: Change homepage to next conference#show

alator21:
Fixes #1259


alator21:
The thing is that we need to access the /conferences as it is now.This issue was about changing only the root path.Can i do such change from the controller?

Ana06:
@alator21

> The thing is that we need to access the /conferences as it is now.This issue was about changing only the root path.Can i do such change from the controller?

You can access `Conference` model from the controller and redirect.

differentreality:
If you redirect inside the controller from index to show, then you will not be able to access the index action (/conferences) even if you explicitly wanted to.

Ana06:
@differentreality 

> If you redirect inside the controller from index to show, then you will not be able to access the index action (/conferences) even if you explicitly wanted to.

I don't understand what u mean :thinking:

alator21:
I think what @differentreality  means is that we should have the option to access the current index at (/conferences) current index but the root path(/) should change in case we got only one conference proposed index with only one conference. @Ana06

Ana06:
ah, ok. I agree with @differentreality then :+1:

bear454:
I did this a different way in my fork... added a redirect to Conferences#index:

```
redirect_to @current.first if @current.count == 1 and @current.first.splashpage
```

alator21:
My solution works too locally.Just doesnt pass the tests..

differentreality:
@bear454 if that happens in conferences#index we will never be able to access the index action itself, we will always be redirected to conferences#show. 

There is the idea to make this work when there is only 1 _active_ conference, so if there is osc17(past conference) and osc18 (future conference), I get redirected to conferences#show of osc18. 
But I can still go to /conferences and see everything there is to see there.

I think we need to either use a route constraint, or simply do something similar to this:
https://stackoverflow.com/questions/28784232/redirect-root-to-named-route-in-rails-4

bear454:
@differentreality that was the point of my line: redirect if and only if there is 1 current conference *and it has a splash page*. If there were two (or more) current conferences, you'd get the index.

I don't like monkeying with the routes, or rendering, because that changes the page that's represented at a URL...  / should be render Conference#Index if it renders at all.. Redirecting to /conferences/my-conf is much more RESTful than having / conditionally render either Conference#Index or Conference#show.

differentreality:
I see. current conference as in there is only 1 conference on the OSEM installation? 
If there are past conferences, we should not prevent users from seeing conferences#index (and the list of past conference), even if there is an upcoming conference.

So I think this sums it up:
If there is only 1 conference, and that conference has a splashpage, redirect to conferences#show from conferences#index

Is that what we all agree on?

Ana06:
@differentreality 

> Is that what we all agree on?

I think so! 👍 

What is the state of this then?

bear454:
@alator21 please consider closing this PR, and supporting the more flexible change in #1994 .

hennevogel:
Alright, let's continue in  #1994

PR #1777, opened 2017-11-04
reCAPTCHA on sign up (BONUS: Features)

bear454:
Optionally* use [reCAPTCHA](https://www.google.com/recaptcha) for spam/bot filtering at registration.

* Uses [Feature](https://github.com/mgsnova/feature) to switch. OSEM *needs* feature switches.

### DIFF
--- a/dotenv.example
+++ b/dotenv.example
@@ -63,3 +63,7 @@ OSEM_SMTP_OPENSSL_VERIFY_MODE=""
 
 # Enable the usage of the devise ichain plugin
 OSEM_ICHAIN_ENABLED=false
+
+# ReCAPTCHA keys
+RECAPTCHA_SITE_KEY=""
+RECAPTCHA_SECRET_KEY=""
--- a/Gemfile
+++ b/Gemfile
@@ -185,6 +185,10 @@ gem 'cloudinary'
 # for setting app configuration in the environment
 gem 'dotenv-rails'
 
+# configurable toggles for functionality
+# https://github.com/mgsnova/feature
+gem 'feature'
+
 # For countable.js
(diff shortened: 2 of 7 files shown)


bear454:
screenshot from 2017-11-03 20-57-58

differentreality:
I like this :+1:

bear454:
@hennevogel I asked for your review, because the feature control could become a core component of how osem works; I know you've been concerned with bloat, but at the same time many of us operating *osem* instances want just a bit more out of it... that means either feature controls, or a plugin architecture (or both); but for something like this, a plugin would be very challenging. Blocks of `if feature` code are only mildly annoying ;-)

dotenv.example
```diff
@@ -63,3 +63,7 @@ OSEM_SMTP_OPENSSL_VERIFY_MODE=""
 
 # Enable the usage of the devise ichain plugin
 OSEM_ICHAIN_ENABLED=false
+
+# ReCAPTCHA keys
+RECAPTCHA_SITE_KEY=""
```
hennevogel: please prefix them with `OSEM_`

hennevogel:
In the end I would like to extend this to users, like we have done in openSUSE/open-build-service#3477...

dotenv.example
bear454: The key names are defined [upstream](https://github.com/ambethia/recaptcha#rails-installation) in the rubygem; while it _can_ be configured more explictly, that means adding yet another initializer, etc., I'd rather stick with the default upstream behavior.

bear454:
I can see that some features will be driven by ENV settings, some by explicit config, and some dynamically in code. Baby Steps.

Issue #330, opened 2014-07-17
User deletion: possible or not? And how?


differentreality:
Following up on https://github.com/openSUSE/osem/issues/324 there is the question of user deletion. Here's some thoughts and concerns:
- Legally user deletion must be available (a user should be able to delete their account and an admin should be able to delete a user's account). 
- If we delete a user who has submitted proposals, we would need to delete those proposals as well (or leave them without a submitter).
  - That would cause disruptions in the application might occur:
    1. Such a proposal could be already scheduled in an upcoming or past conference, and that would leave us with a weird gap. You might recall being in a session at that time at that conference, but the schedule says there was nothing going on at that time... Weird, no?
    2. A link to that proposal could have been made public, for promotional reasons, and that link will no longer work if the proposal is deleted.
  - And what good would it do to us, to have a proposal with a submitter with name 'Unknown User' or 'User Deleted' ?

ChrisBr:
@differentreality 
Isn't that already implemented?
(http://localhost:3000/admin/users ?)

kormoc:
I think indicating that you can't delete an account when there are scheduled events makes sense. They would need to withdraw then all.

I also think wiping out the PII and making it so the account can't log would be enough and would allow them to re-register with the same email address in the future.

differentreality:
Thanks for the input @kormoc 
Withdraw does not make sense after the conference has taken place. Other than that yes, we should implement all that.

PR #1783, opened 2017-11-06
fixed comments in notifications

### LINKED ISSUE #1782: Error viewing all_comments in notifications dropdown
comments-bug
This occurs when a user is deleted and the associated comments are not destroyed.
https://github.com/openSUSE/osem/blob/be1992a03af2766aef4d39276e34fb3814f882da/app/views/admin/comments/_all_comments.html.haml#L11
The error is occuring in above line when user associated with a comment is not found.
Either we should add dependent destroy to users or show something like "Posted by: Anonymous" when user is nil.
@differentreality What do you think?

ViditChitkara:
closes #1782


differentreality:
This should be considered in regards with user deletion #330

PR #1785, opened 2017-11-08
Fix a flaky test.

bear454:
Tests were occasionally failing as html escape entities didn't match the expectation string.

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -66,12 +66,12 @@ describe ApplicationHelper, type: :helper do
       end
 
       it 'should use the environment variable' do
-        ENV['OSEM_NAME'] = Faker::Company.name
-        expect(nav_root_link_for(nil)).to match ENV['OSEM_NAME']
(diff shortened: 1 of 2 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -66,12 +66,12 @@
       end
 
       it 'should use the environment variable' do
-        ENV['OSEM_NAME'] = Faker::Company.name
-        expect(nav_root_link_for(nil)).to match ENV['OSEM_NAME']
+        ENV['OSEM_NAME'] = Faker::Company.name + "'"
```
differentreality: So we will only add a quote at the end?

bear454: I wanted to make sure there was _always_ an apostrophe in the test string, as that was the character Faker commonly spits out, which was being html encoded, but it sometimes doesn't. So, forcing an apostrophe at the end of the string makes the test behave consistently - it _always_ has to handle html encoded characters :)

PR #1792, opened 2017-11-13
Control config.force_ssl with an ENV variable

bear454:
Allow the `force_ssl` setting to be controlled with an environment variable; defaults to false just like it did before.


differentreality:
I think we should also add the variable in dotenv.example

bear454:
@differentreality ping ;-)

differentreality:
@bear454 it seems 'rebase and merge' is not possible. Can you rebase your branch manually please?

Issue #1794, opened 2017-11-14
Change Travis to use db:schema:load instead of db:migrate


differentreality:
https://github.com/openSUSE/osem/pull/1714#issuecomment-343210789

PR #1798, opened 2017-11-16
changed db:migrate to db:schema load

KyriakosZag:
#1794 = SOLVED


bear454:
I think we just need https://github.com/openSUSE/osem/pull/1798/commits/09ecc1174f782e349e9ed90facb95aff807fc325 . What are the other commits?

PR #1799, opened 2017-11-16
tests for program#any_event_for_this_date is written

### LINKED ISSUE #1388: Write test for program#any_event_for_this_date?(date)

namangupta01:
Closes #1388

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -413,6 +413,7 @@ Style/DateTime:
     - 'app/models/conference.rb'
     - 'app/models/program.rb'
     - 'spec/models/conference_spec.rb'
+    - 'spec/models/program_spec.rb'
 
 # Offense count: 488
 Style/Documentation:
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -165,7 +165,7 @@ class Program < ApplicationRecord
   # * +False+ -> If there is not any event for the given date
   def any_event_for_this_date?(date)
     parsed_date = DateTime.parse("#{date} 00:00").utc
-    EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1)).any?
+    EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1.day)).any?
   end
 
   ##
--- a/spec/models/program_spec.rb
+++ b/spec/models/program_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe Program do
   subject { create(:program) }
-  let!(:conference) { create(:conference, end_date: Date.today + 3) }
+  let!(:conference) { create(:conference, end_date: Date.current + 3) }
   let!(:program) { conference.program }
 
   describe 'association' do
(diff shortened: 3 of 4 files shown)


spec/models/program_spec.rb
```diff
@@ -240,6 +240,19 @@
     end
   end
 
+  describe '#any_event_for_this_date?' do
+    let(:event){ create(:event, program: program) }
+    let!(:event_schedule) { create(:event_schedule, event: event) }
+
+    it 'returns true if there is any event for particular date' do
+      expect(program.any_event_for_this_date?(DateTime.now.to_date)).to eq true
```
differentreality: Please use `current` instead of `now`

differentreality: You can use `Date.current`, if you only need a date

differentreality: It is not clear to me why the event is not scheduled on this particular date, because I don't know when the event is supposed to be scheduled. I would make this clear in the test, ie add a specific date in let definition.

namangupta01:
@differentreality Respective changes are done please have a look!

differentreality:
Well you still haven't listed all the available scenarios. I find it extremely helpful both for you, and the reviewer, that you do this.

spec/models/program_spec.rb
```diff
@@ -241,6 +241,48 @@
...
+      it 'returns false if date passed is empty' do
+        expect(program.any_event_for_this_date?('')).to eq false
+      end
+
+      it 'returns false if date passed is nil' do
+        expect(program.any_event_for_this_date?(nil)).to eq false
+      end
+    end
+
+    context 'when schedule is selected for the conference' do
```
differentreality: And where is the selected schedule for the conference defined?

app/models/program.rb
```diff
@@ -165,7 +165,7 @@ def languages_list
   # * +False+ -> If there is not any event for the given date
   def any_event_for_this_date?(date)
     parsed_date = DateTime.parse("#{date} 00:00").utc
-    EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1)).any?
+    EventSchedule.where(schedule: selected_schedule).where(start_time: parsed_date..(parsed_date + 1.day)).any?
```
differentreality: Why do we need this?

namangupta01: We need this because we are getting : 
screen shot 2017-12-16 at 3 24 01 am

But the behaviour should be : 
screen shot 2017-12-16 at 3 24 21 am

Adding 1 to parsed_date just increase the parsed_date by 1 sec not by one day

spec/models/program_spec.rb
namangupta01: It is defined in event_schedule factory

namangupta01: Check https://github.com/openSUSE/osem/blob/master/spec/factories/event_schedule.rb#L17

app/models/program.rb
namangupta01: Basically, this is a bug.

namangupta01:
@differentreality I think all the scenarios are there. Please tell me which test cases are left.

spec/models/program_spec.rb
differentreality: I am inclined to explicitly define a schedule in a before block, after the let definition, because this is the very thing we are testing here (what happens when there is a schedule).

.rubocop_todo.yml
```diff
@@ -422,6 +422,7 @@ Style/DateTime:
     - 'app/models/conference.rb'
     - 'app/models/program.rb'
     - 'spec/models/conference_spec.rb'
+    - 'spec/models/program_spec.rb'
```
differentreality: I would expect your commit message to say a tiny bit more as to why you added this here.

differentreality:
:+1: Nice @namangupta01 Thank you!
Once you finish the latest changes, feel free to squash your commits, and rebase.

I would keep the rubocop addition in a separate commit, but please add a bit more information on the commit message. The rest of the commits can be squashed into 1.

namangupta01:
Done!

app/models/program.rb
differentreality: Why aren't you using just Time so that the rubocop change is not needed at all?
`Time.new(2017, 12, 21, 10, 00, 00, "+00:00")`

Also did you come across other occurrences of this issue?

namangupta01: We can also use this but using DateTime is better when we need date with time instead of using hardcoded date with time.

namangupta01: And i am also getting offence in this line. Either way we have to change the rubocop.

screen shot 2017-12-21 at 12 46 12 am

differentreality:
OK I am not sure what should happen with that cop. Let's merge this for now, perhaps someone else can take a look at the cop and offer some feedback (@Ana06 ?).

Thank you @namangupta01 :+1:

PR #1805, opened 2017-11-20
Withdrawn events on conference dashboard

### LINKED ISSUE #463: Dashboard: Withdrawn Events
At the top bar information there should be the following:
- sadly the event X, Y and Z where withdrawn

chocoelho:
Resolves #463.

It adds an alert whenever events on a conference get withdrawn. 

Screenshot:
image

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -300,7 +300,7 @@ Lint/UnusedBlockArgument:
 
 # Offense count: 136
 Metrics/AbcSize:
-  Max: 86
+  Max: 98
 
 # Offense count: 258
 # Configuration parameters: CountComments, ExcludedMethods.
@@ -320,7 +320,7 @@ Metrics/LineLength:
 # Offense count: 133
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 56
+  Max: 63
 
 # Offense count: 4
 # Configuration parameters: CountComments.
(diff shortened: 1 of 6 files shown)


chocoelho:
Oh, np @Ana06. It's just that I've tried this before(last year if I'm not mistaken) and it was being applied to the admin interface. I can add it to the user side as well, without problems, but with some adjustments, such as using event titles instead of a counter because it'll make more sense on the user side.

differentreality:
I think dashboard means admin/conferences#show, otherwise it would say splashpage 

I would also like to mention that I don't particularly like the implementation with the alert. I was hoping  that you may have a better idea on how to show this @Ana06

chocoelho:
It could be another section in the dashboard view. Another way to implement this, although it would involve the alert, is to track state changing and display only withdrawn events since last user login or something like that using https://github.com/continuum/espinita.

Ana06:
@differentreality 

> I think dashboard means admin/conferences#show, otherwise it would say splashpage

ok, sorry for the confusion then 👍 

> I would also like to mention that I don't particularly like the implementation with the alert. I was hoping that you may have a better idea on how to show this @Ana06

🤔  Maybe another section on the top of the dashboard with the withdrawn events?

chocoelho:
So, section approach? @Ana06 @differentreality

chocoelho:
@Ana06 @differentreality would something like this be better?

image

chocoelho:
Same idea but with more withdrawn events:

image

Ana06:
@differentreality what do you think?

differentreality:
I like the charts :)

What about we also add a relevant icon on the top of the page? (next to Registrations, Submissions,Hours)

differentreality:
:+1: I like it :)  
How about you squash your commits?

chocoelho:
@differentreality @Ana06 I just squashed my commits into 3

differentreality:
Nice, thank you @chocoelho :+1:

PR #1809, opened 2017-11-22
Speaker is also able to see the proposal and able to modify it

### LINKED ISSUE #1482: Show proposals to all speakers, not just submitter
Users are able to see their submissions, but that only works if the user is the submitter. It should be showing the proposals where the user is just the speaker as well.

namangupta01:
Closes #1482

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -345,24 +345,34 @@ describe User do
     end
 
     describe 'proposals methods' do
-      let(:submitter) { create(:submitter, user: user) }
+      let(:submitter) { create(:user) }
+      let(:speaker) { create(:user) }
       let(:event1) { create(:event, program: conference.program) }
       let(:event2) { create(:event, program: conference.program) }
 
       before do
-        event1.event_users << create(:event_user, user: user, event_role: 'submitter')
-        event2.event_users << create(:event_user, user: user, event_role: 'submitter')
+        event1.event_users << create(:event_user, user: submitter, event_role: 'submitter')
+        event2.event_users << create(:event_user, user: submitter, event_role: 'submitter')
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -345,24 +345,35 @@
     end
 
     describe 'proposals methods' do
-      let(:submitter) { create(:submitter, user: user) }
```
Ana06: How was that working before? where is `submitter` defined? 🤔

spec/models/user_spec.rb
```diff
@@ -345,24 +345,35 @@
...
+      let(:submitter) { create(:user) }
+      let(:speaker) { create(:user) }
       let(:event1) { create(:event, program: conference.program) }
       let(:event2) { create(:event, program: conference.program) }
+      let(:event3) { create(:event, program: conference.program) }
```
Ana06: Why do you need a third event for that, use one of the others. The case when we have both a submitter and an speaker is the one we want to test. 😉

spec/models/user_spec.rb
namangupta01: It is defined in factory file 'event_users.rb'

spec/models/user_spec.rb
namangupta01: Yes you are right i will correct it!

namangupta01:
@Ana06 respective changes are done. Please have a look!

differentreality:
I think this looks good. @Ana06 ?

bear454:
Please update your review @Ana06 or it will be dismissed.

differentreality:
Thank you @namangupta01 :+1: 
Sorry this took so long!

PR #1814, opened 2017-11-24
Fixing admin user creation in docker

jesmg:
Added db:seed in order to have the deleted user, and be able to have an admin user.


docker/init.sh
```diff
@@ -42,6 +42,9 @@ fi
 echo ">>> Upgrading database..."
 bundle exec rake db:migrate
 
+echo ">>> Seed database..."
+bundle exec rake db:seed
```
AEtherC0r3: This should happen after the db:schema:load and inside the [if statement](https://github.com/openSUSE/osem/blob/master/docker/init.sh#L37-L40), so that it only runs once.

Another way to do this, is to replace db:schema:load with db:setup

differentreality:
@jesmg ping

jesmg:
I'll modify it today

jesmg:
Change done

olea:
I verified the patch and works as expected.

FYI

bear454:
@AEtherC0r3 please update your review.

PR #1818, opened 2017-11-27
Conference name repeated in revision_history on adding commercial fixed

### LINKED ISSUE #1804: Conference name repeated in revision_history on adding commercial.
When adding/updating commercial for a conference, the conf. name is repeated two times where first one links to commercial page for that conf.
I think it would be better that in case of 'commercial in conference' the link is included in commercial text as conference name is already added in the end.
screenshot from 2017-11-19 23-53-40

Harry30897:
#1804

### DIFF
--- a/app/views/admin/versions/_object_desc_and_link.html.haml
+++ b/app/views/admin/versions/_object_desc_and_link.html.haml
@@ -234,6 +234,12 @@
       = link_to_organization(version.conference_id)
     - else
       (Organization Deleted)
+  - elsif version.item_type == 'Commercial'
+    - commercial = current_or_last_object_state(version.item_type, version.item_id)
+    - commercialable = current_or_last_object_state(commercial.commercialable_type, commercial.commercialable_id)
+    - unless commercial.commercialable_type == 'Conference'
+      in conference
+      = link_to_conference(version.conference_id)
   - else
     in conference
     = link_to_conference(version.conference_id)


app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -235,5 +235,5 @@
     - else
       (Organization Deleted)
   - else
-    in conference
-    = link_to_conference(version.conference_id)
+    //in conference
```
differentreality: How is it useful to have these commented out lines on our code base?

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -235,5 +235,5 @@
     - else
       (Organization Deleted)
   - else
```
differentreality: If the else part no longer has code, why do we need to keep it?

Harry30897: I am figuring out the else part code @differentreality. Will get back to you soon with the solution. Well I am posting the screenshot for your reference.
image

app/views/admin/versions/_object_desc_and_link.html.haml
Harry30897: @differentreality  I commented out the lines because I was not sure that will it pass all the test cases or not. If it would have passes, i would certainly remove the commented out lines.

namangupta01: @Harry30897 Hey. You can locally check for the tests whether they are passing or not by running the command  ```bundle exec rspec```. And for checking the styling you can run the command ``` bundle exec rubocop -Dc .rubocop.yml```.

Harry30897: Thankyou @namangupta01  Will take care of this from now onwards.

app/views/admin/versions/_object_desc_and_link.html.haml
differentreality: In this screenshot the record for ticket (ID 23), no longer shows for which conference this was. 
This used to look like:
Stella Rouzi created new ticket abc in conference tconf0 1 day ago

differentreality:
@Harry30897 please add in the first comment of the PR a new screenshot based on the recent changes

Harry30897:
Sure @differentreality. Now the Revision History shows the tickets, events, lodgings,booths for conference correctly. Will be posting 2 screenshots for your reference.
image

image

differentreality:
Thanks @Harry30897 and sorry this took so long.

The change is also dropping 'in conference x' for event commercials. I think it is useful to have it there, sometimes we will want to filter by conference, or we want to know which conference this event is about just by looking at this list.
screenshot from 2017-12-12 18-51-36

What do you think?
Can you check if you can make it show up when it is an event (or venue) commercial, but not show twice when it is a conference commercial? Thanks!

Harry30897:
@differentreality .You are right we need to have 'in conference x' after event commercials. So as requested I changed the file according to you. Well uploading two screenshots for your reference.
image

image

differentreality:
That's nice @Harry30897 :tada: 
Can you please prepare your PR for merge? 
(Meaning squash your commits into 1, and rebase)

Thanks!

Harry30897:
Sure @differentreality .

differentreality:
Thank you @Harry30897 :-)

Issue #400, opened 2014-07-30
Splash Page: Program Section


hennevogel:
Why does the program section not show any program? For instance
- All the events which are currently on
- if no event is currently on, show some random talks
- if it's a small event (dunno how to detect that) show the full schedule

theaverageguy:
What is to be in this issue now?

hennevogel:
I still think it's a good idea to show Events which are currently on. Highlights are already done.

Shriyanshagro:
@hennevogel , Currently program section show all events 
* I don't think that it would be a good idea to display some random talks , as it would mislead the moderators and users.
* Though if anyhow we got a way to find smaller events, but still I would suggest ,not to showcase full schedule on splash page.

bear454:
I'm interested in this one ;-)

PR #1821, opened 2017-11-29
Splash Improvements

bear454:
There were a lot of niggly little issues with the splash page that annoyed me.

I'd like to think I've fixed most of them, and at least made things more maintainable for whoever wants to fix the next annoyance. 

1. If there's a link for the venue, that implies the venue is relevant, so use the venue name and link in the header. If not, use the city/country.
2. Tickets & registration at the top. Clean up the ticket design a bit.
3. Fold call for papers & call for tracks into one section. There's no good reason these two related calls for content should be stacked.
4. Renovate the highlighted sessions; typically these are mostly about the speaker, so feature them more prominently.
5. More consistent listing of tracks & a more prominent button for the full schedule.
6. Refactor booths to be less prominent; feature the description in a modal (same as sponsors)
7. Offset the map a little to get the popup closer to the center, instead of bumping the top edge.
8. List sponsor levels
9. More prominent call for sponsors.
10. Fold up some of the navigation where it makes sense, so it doesn't overlap.

osem-pr-1820

(not pictured)

* Fully linted
* Better partial separation
* Removed unnecessary markup
* More consistent markup
* Removed 18 queries!

_I assume few people run osem with a completely stock splashpage, so the improvements try to keep it easy to customize in a fork._


Ana06:
@bear454 thanks for your ideas and the PR. 😉 Next time it may be better to open an issue to this it before got to the implementation. 🤔 

> 1. If there's a link for the venue, that implies the venue is relevant, so use the venue name and link in the header. If not, use the city/country.

Why not just adding the link and keeping the other information as well? 🤔 

> 2. Tickets & registration at the top. Clean up the ticket design a bit.

I would say that the program should be first and then the registration. You want to see the program of the conference before registering, right? 🤔 

> 3. Fold call for papers & call for tracks into one section. There's no good reason these two related calls for content should be stacked.

I like the idea 👍 

> 4. Renovate the highlighted sessions; typically these are mostly about the speaker, so feature them more prominently.

I like this ideas as well 👍 I may make the picture a little bit smaller...

> 5. More consistent listing of tracks & a more prominent button for the full schedule.

I would say that what most people what to see is the whole schedule... what about giving it more important? Currently the full schedule button is at the end and it is not really visible.. 🤔  Maybe adding a title to the section also helps.

> 6. Refactor booths to be less prominent; feature the description in a modal (same as sponsors)

👍 

> 7. Offset the map a little to get the popup closer to the center, instead of bumping the top edge.

👍 

> 8. List sponsor levels

I am not sure if this will make sense for all conferences... 🤔 

> 9. More prominent call for sponsors.

👍 
> 10. Fold up some of the navigation where it makes sense, so it doesn't overlap.

👍 

BTW, I didn't take a look at the code until it is clear if we want to modify everything. 😄

Ana06:
@differentreality @ddemaio @ChrisBr what do you think? 🤔

bear454:
@Ana06:

> > 1. If there's a link for the venue, that implies the venue is relevant, so use the venue name and link in the header. If not, use the city/country.

> Why not just adding the link and keeping the other information as well? 🤔

The geography is _well represented_ in the venue section; the goal of the header is to be brief, prominent, relevant. After discussing with a number of event organizers, I found these scenarios:

* The space is a rented space, and there is no desire to promote it. In this case, the organizers would not include a URL, would not want to advertise the space. => Displays City/Country
* The space is a partner in the event. The organizer agree to promote the space, and therefore include a URL. => Displays venue name with link
* The space is an attraction/landmark. The organizers want to leverage the popularity of the space, and therefore include a URL. => Displays venue name with link

While that may need some documentation (oh man does _osem_ need documentation); I think it's a simple feature that meets the needs of organizers and improves the usefulness of the splash page.

bear454:
@Ana06:

> I would say that the program should be first and then the registration. You want to see the program of the conference before registering, right? 🤔

Good point; I'll swap those.

olea:
> > List sponsor levels
> I am not sure if this will make sense for all conferences... :thinking:

Sure it makes sense. If you are selling different categories with different prices is obvious you are selling distinction. So it's correct to me to remark it.

+1

Ana06:
@olea 

> Sure it makes sense. If you are selling different categories with different prices is obvious you are selling distinction. So it's correct to me to remark it.

and @bear454 

> Can you give me an example where it doesn't?

What I am not sure is if all the conference have different categories for sponsors. 🤔

Ana06:
@bear454 

> >> 5. More consistent listing of tracks & a more prominent button for the full schedule.
> 
> > I would say that what most people what to see is the whole schedule... what about giving it more important? 
> 
> The whole schedule is just too big to include - hence the highlighted sessions. At some point I'll get around to implementing @hennevogel 's "events right now" stuff ( https://github.com/openSUSE/osem/issues/400 ), but there's no way we can list the full schedule on the splash.

I do not want to include the whole schedule, but making this more accesible. 🤔 
> 
> > Currently the full schedule button is at the end and it is not really visible.. 🤔 
> 
> I can make the "Full Schedule" button a bit bigger; any better suggestions?

Maybe moving it up as well 🤔 

> 
> > Maybe adding a title to the section also helps.
> 
> I don't like the big "Program" line that was there; it's obvious we're listing the parts of the program ;-)

But the rest of the sections have a title, for consistency... 🤔

Ana06:
> The geography is well represented in the venue section; the goal of the header is to be brief, prominent, relevant

What I thinking about is that the city where the conference take place is normally more important than the building where it takes place. That's why I do not like this change. 🤔

olea:
> What I am not sure is if all the conference have different categories for sponsors. :thinking:

@Ana06 Yes, you are right. Some change the top one from Gold, to Platinum to Diamond. I call this «sponshorship inflation» :-D  The most interesting alternative I saw these days is to use T-Shirts size: XXL, XL, L, M... :-D

But as I checked OSEM can customize both the name and the order of sponsorship leves so I can't see what problem can be :-m

differentreality:
I am having issues with mysql in this PR, it is loading forever and mysqld is using up 100% of the CPU according to top. 
I only managed to see the changes using sqlite (and I am liking them).

I think that perhaps booths would make sense to show up on the top bar as well, as they are not exactly part of the content (ie events, tracks), but I would hold back on merging this PR just for this. 

We do have to make sure everything runs smoothly on all the DBs we are supporting though.

bear454:
@differentreality considering that this PR does not add any migrations or touch the db/schema, I think you may have unrelated db issues ;-)

bear454:
Or... does mysql just not handle complex joins as well as sqlite & postgres? 🤣

ChrisBr:
@bear454 It looks great, thank you so much :)
Also the LF NW website looks great as well 👍

Ana06:
@bear454 can you please rebase and add a new screenshot? 

My only big concern was making the whole schedule more accesible, so that people can find the link to the schedule easily... But I am also not pretty sure why...

Si it seems that are agree on the design then, right?

differentreality:
@bear454 I am still having the same problem with postgres, have you run this with data from events.opensuse.org ?

bear454:
@differentreality I don't have access to data from e.o.o; I am running this on Postgres, for LFNW, and we're not seeing any unusual performance issues (just taking 4s off of load time round-trip ;-) . If you can send me a sql dump, I'd be happy to load it up & try.

bear454:
@differentreality: c9b211a has good performance on both mariadb & postgresql - tested using AWS RDS. The problem wasn't at the SQL level really, but in allocation of ActiveRecord objects.

bear454:
Anecdotal evidence, but still:

screenshot from 2017-12-22 08-29-39

Issue #1828, opened 2017-11-30
Ticket generation fails when using Cloudinary


bear454:
**I'm submitting a ..**
- [X] Bug Report
- [ ] Feature Request

**Current behavior:**
After purchasing a ticket, confirmation email never goes out, due to pdf generation failure, due to trying to manipulate Conference#picture directly.

**Expected behavior:**
PDF should be generated; email should go out; code should never touch the 'file' part of an upload after upload; as it is probably not on the file system anymore (S3, etc.)

**Steps to reproduce:**

1. Setup an _osem_ instance using Cloudinary (as suggested in the 'Deploy to Heroku' config)
2. Register a user with a ticket.

**Related code:**

app/pdfs/ticket_pdf.rb:L46

```
if 7 * @conference.picture.image[:width] > 12 * @conference.picture.image[:height]
```

app/uploaders/picture_uploader.rb:L55

```
def image
  @image ||= MiniMagick::Image.open(file.file)
end
```

**Other information:**

```
NoMethodError: undefined method `file' for #<Cloudinary::CarrierWave::CloudinaryFile:0x007f33423549a0>
1 File "/app/app/uploaders/picture_uploader.rb", line 56 in image
2 File "/app/app/pdfs/ticket_pdf.rb", line 46 in draw_second_square
3 File "/app/app/pdfs/ticket_pdf.rb", line 15 in initialize
4 File "/app/app/mailers/mailbot.rb", line 17 in new
5 File "/app/app/mailers/mailbot.rb", line 17 in block in ticket_confirmation_mail
6 File "/app/app/mailers/mailbot.rb", line 16 in each
7 File "/app/app/mailers/mailbot.rb", line 16 in ticket_confirmation_mail
```

PR #1834, opened 2017-11-30
Fix speaker registration issues

bear454:
#### Checklist

- [x] My branch is up-to-date with the Upstream `master` branch.
- [X] I have added necessary documentation (if appropriate).

#### Changes proposed in this pull request:

* A confirmed speaker cannot register if registration is not open; registration must be open to register. Period.
* Ticket generation doesn't depend on local file system, but uses carrierwave formats instead.

#### Fixes Issue

* #1827 
* #1828

### DIFF
--- a/app/pdfs/ticket_pdf.rb
+++ b/app/pdfs/ticket_pdf.rb
@@ -1,3 +1,5 @@
+require 'open-uri'
+
 class TicketPdf < Prawn::Document
   def initialize(conference, user, physical_ticket, ticket_layout, file_name)
     super(page_layout: ticket_layout, page_size: 'A4', filename: file_name)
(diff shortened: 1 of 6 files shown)


app/models/ability.rb
```diff
@@ -66,7 +66,7 @@ def signed_in(user)
 
     can [:new, :create], Registration do |registration|
       conference = registration.conference
-      conference.registration_open? && !conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user)
+      conference.registration_open? && (!conference.registration_limit_exceeded? || conference.program.speakers.confirmed.include?(user))
```
Ana06: As I just wrote in https://github.com/openSUSE/osem/issues/1827#issuecomment-348426404, this is made in purpose. The expected condition is the one we already have:

``` Ruby
(conference.registration_open? && !conference.registration_limit_exceeded?) || conference.program.speakers.confirmed.include?(user)
```

Please remove the first commit 😄

differentreality:
Registration thingies look OK, I also agree with the approach of 

Allow confirmed speakers to register for the conference:
  - if the conference accepts registrations (otherwise don't show the link in proposals tooltip)
  - even if the dates for registration have passed
  - even if the registration limit has exceeded (because well, the speakers will be there anyway)

@namangupta01 has also been working on this in https://github.com/openSUSE/osem/pull/1629
@bear454 would you like to take a look at that PR and perhaps suggest the changes there so that his PR can be merged?
Then this PR can be about tickets and perhaps the addition of some extra tests for registration.

bear454:
okay, I'll withdraw that commit from this PR, so I can at least get the other part fixed. :/

bear454:
Based on changes to #1629 , I removed the permissions commit from this PR. @Ana06 please update your review.

PR #1841, opened 2017-12-02
Autofocus tickets

Astradeni:
To improve user experience


differentreality:
:+1: I like the focus on the first (most relevant) field of the form. We should do that.
Did you check if there are more forms that need this?

PR #1845, opened 2017-12-02
autofill conference end date

### LINKED ISSUE #1844: autofill conference end date
## Description

For every date you choose for as the begin for a conference, when you go to choose end date it always show the current month.

It could be good that the behaviour for the ending date show at least the same month where the conference begin. 

- **old-behaviour**
old-behaviour

### Steps to Reproduce

1. register a new conference
2. select start date from the calendar (any month ahead which isn't the current one)
3. select the end date - it will be showed the current month

**Expected behavior:** after the beginig date was defined, calendar for end date starts showing at least the same one as showed in the start date

**Reproduces how often:** I reproduced it every time I've tried.

**Version:** I've tried against master branch

- **new Expected behaviour**
new-behaviour

adrianovieira:
based on start date autofill end date (fix #1844)

- **new behaviour**

new-behaviour

### DIFF
--- a/app/assets/javascripts/osem-datepickers.js
+++ b/app/assets/javascripts/osem-datepickers.js
@@ -79,6 +79,9 @@ $(function () {
 
   $("#conference-start-datepicker").on("dp.change",function (e) {
       $('#conference-end-datepicker').data("DateTimePicker").setMinDate(e.date);
+      if (!$('#conference-end-datepicker').val()) {
+         $('#conference-end-datepicker').data("DateTimePicker").setDate(e.date);
+      }
   });
   $("#conference-end-datepicker").on("dp.change",function (e) {
       $('#conference-start-datepicker').data("DateTimePicker").setMaxDate(e.date);


app/assets/javascripts/osem-datepickers.js
```diff
@@ -79,6 +79,7 @@ $(function () {
 
   $("#conference-start-datepicker").on("dp.change",function (e) {
       $('#conference-end-datepicker').data("DateTimePicker").setMinDate(e.date);
+      $('#conference-end-datepicker').val()?$('#conference-end-datepicker').val():$('#conference-end-datepicker').data("DateTimePicker").setDate(e.date);
```
Ana06: Although this work, I find this syntax quite difficult to read. And I don't like that you are assigning again the datepicker in case that if was already assigned. What about this:

``` Javascript
unless $('#conference-end-datepicker').val(){
  $('#conference-end-datepicker').data("DateTimePicker").setDate(e.date);
}
```

And maybe we should also assign `$('#conference-end-datepicker')` do a variable as it used several times. 🤔

adrianovieira: I understand your point of view.

Sorry, but I don't know `unless` (I've tried to find some documentation about, but I didn't find one).

So, an alternative could be:

```javascript
if (!$('#conference-end-datepicker').val()) {
   $('#conference-end-datepicker').data("DateTimePicker").setDate(e.date); 
}
```

Ana06: yes, sorry, `unless` doesn't exist in javascript 🙈 

> So, a better alternative could be:
>
> ```javascript
> if (!$('#conference-end-datepicker').val()) {
>    $('#conference-end-datepicker').data("DateTimePicker").setDate(e.date); 
> }
> ```

This looks great 👍

adrianovieira:
My belief is that this was finished.

Issue #1667, opened 2017-08-29
Persistent Storage directories unclear


tobru:
Looking at the provided `docker-compose.yml.example` it looks like all persistent data is stored under `/data`. But that doesn't seem to be the case, at least the uploaded conference logo is stored under `/osem/public/system/conferences/logos`. Would it be possible to make sure everything which must be persistent (uploads, generated files, ...) will be saved under `/data`?

PR #1851, opened 2017-12-04
Fix the docker persistence for uploaded pictures: #1833 #1667

### LINKED ISSUE #1833: Rebuilding containers lost loaded images
- [x]  bug report
- [ ] enhancement

1. setup docker containers with docker-compose as expected
2. containers up, you start to add details and content for a conference
3. particularly you add pictures as: conf logo, venue logo and lodging pictures
4. everything is fine, sun shines, live is beautiful
5. you rebuild containers: docker-compose up --build -d
6. with the new container up all the data is still there as expected but... not any of the pictures previously uploaded.

Every time we rebuild persistence is broken for the pictures.

olea:
Well: this seems to work and fixes #1833 #1667


olea:
@TheAssassin would you like to check if this PR too, please?

bear454:
JFYI: this only applies when using Carrierwave's local storage, not when using any cloud-based asset host (e.g. direct S3, Cloudinary, etc.)

TheAssassin:
@olea the permissions of the directories you create are a bit too generous. I used `0770` instead of `0755`. Is there a special purpose for which this is required?
Also, can't we put that data into the `$DATA_DIR`? Putting the data into a separate directory but other data near the source tree seems like a bad idea to me.

The `VOLUME`s could be written in a single line by the way.

Is it really necessary to create the entire chain like `/osem/ /osem/tmp /osem/tmp/uploads/`? I am pretty confident it'll work like `mkdir -p` and create the intermediate directories if they don't exist.

olea:
@bear454 

> JFYI: this only applies when using Carrierwave's local storage, not when using any cloud-based asset host (e.g. direct S3, Cloudinary, etc.)

Sorry but I don't understand the implications of that. My  production platform is a VM Centos 7 with local volumes :-m

olea:
> @olea the permissions of the directories you create are a bit too generous. I used 0770 instead of 0755. Is there a special purpose for which this is required?

I used exactly the same permissions OSEM generates :-m 

> Also, can't we put that data into the $DATA_DIR? Putting the data into a separate directory but other data near the source tree seems like a bad idea to me.

I understand. But using just one volume AFAIK requires modifications of OSEM code, which I tried to avoid for simplicity and maintainability

> 
> The VOLUMEs could be written in a single line by the way.

I did for readibility, but don't have strong opinions here.
>
> Is it really necessary to create the entire chain like /osem/ /osem/tmp /osem/tmp/uploads/? I am pretty confident it'll work like mkdir -p and create the intermediate directories if they don't exist.

I'll fix it.

Thanks for your time.

bear454:
@olea 

> @bear454

>> JFYI: this only applies when using Carrierwave's local storage, not when using any cloud-based asset host (e.g. direct S3, Cloudinary, etc.)

> Sorry but I don't understand the implications of that. My production platform is a VM Centos 7 with local volumes :-m

"Just For Your Information"

Dockerfile
```diff
@@ -38,7 +38,10 @@ RUN chown -R osem.root /osem/ && \
 # data directory is used to cache the secret key in a file
 ENV DATA_DIR /data
 RUN install -d -m 0770 -o osem -g root $DATA_DIR
-VOLUME ["$DATA_DIR"]
+# data persistence for uploaded files (logos, other pictures, etc)
+RUN mkdir 0775 -p /osem/tmp/cache /osem/tmp/uploads/ && \
+       chown -R  osem.osem /osem/tmp/
+VOLUME ["$DATA_DIR"] ["/osem/tmp/uploads/"] ["/osem/public/system/"]
```
TheAssassin: That line should be `VOLUME ["$DATA_DIR", "/osem/tmp/uploads/", "/osem/public/system/"]`.

TheAssassin:
I didn't even know that this syntax works or exists at all. The syntax is meant to be a JSON-like array. I'd recommend to change it as suggested in my comment.

Other than that, it looks good to me.

bear454:
Please update as @TheAssassin suggested, then I'll be happy to merge in.

TheAssassin:
@bear454 you can do it, too, I guess, if you want to speed up things. I'm pretty sure @olea won't mind.

bear454:
Fixed as described & merged in manually.

bear454:
Scratch that; I can't push it to `master`. I'll wait for @olea to update...

Dockerfile
olea: Done.

PR #1853, opened 2017-12-05
Fix email text overflow in proposal#show

### LINKED ISSUE #1838: Text overlap on public proposal page
Speaker details either gets overlapped or gets hidden in case event has its own commercial.

screenshot from 2017-12-01 17-19-43
screenshot from 2017-12-01 17-20-58

rishabhptr:
Fixes #1838
Before:
screenshot from 2017-12-01 17-20-58

screenshot from 2017-12-01 17-19-43
After:
screenshot from 2017-12-05 17-08-41

### DIFF
--- a/app/views/proposals/show.html.haml
+++ b/app/views/proposals/show.html.haml
@@ -36,8 +36,10 @@
             .col-md-8
               %h4
                 = link_to speaker.name, user_path(speaker.id)
+                %br
                 - if speaker.email_public?
-                  = "(#{speaker.email})"
+                  = mail_to "#{ speaker.email }" do
+                    %i.fa.fa-envelope-o.fa-2x
               - if speaker.affiliation?
                 .text-muted
                   from


rishabhptr:
@Ana06  Yes, I just made the text smaller so that it can accommodate average email length which it was not doing so before.
Also I thought it wont look nice if we wrap/break the email.

One styling can be to let it be in one line and full email shows up on hover in case its too long.What do you think?
screenshot from 2017-12-05 18-28-55

bear454:
What's the value in printing the email address there at all?

Consider putting up an [email icon](http://fontawesome.io/icon/envelope/) with a mailto link, instead of providing such easy fodder for spam harvesters ;-)

differentreality:
Email icon is implemented in https://github.com/openSUSE/osem/issues/1870

differentreality:
@rishabhptr please show the current behavior in your screenshots (first comment of this PR should show what this PR is about)

rishabhptr:
@differentreality As you can see in my screenshot in first comment the email is overlapped by the commercial or enters event data area.

As proposed I have made changes and now it looks like this : 
screenshot from 2017-12-10 19-34-09

Ana06:
@bear454 

> What's the value in printing the email address there at all?
> 
> Consider putting up an email icon with a mailto link, instead of providing such easy fodder for spam harvesters ;-)

I would say we could just directly remove it. In this page I would just show the speaker name, nickname company and description and I would move the email to the user show page

adrianovieira:
Hi!
From my perspective showing some social media icons on that page could be good.

On the other hand, if I were you what I really did care about would be a revision which doesn't show speaker biography on this page (Or maybe it be a -very- short description). My belief is that it could be a long page if they have some speakers (2+) for a presentation/workshop.

adrianovieira:
Yes! I agree with you @bear454

differentreality:
@rishabhptr please make sure that the description of the PR (first comment) depicts the actual situation (right now the screenshot is an old one, and it doesn't show what the PR is actually doing, so we should fix that. That way people seeing this PR can have a clear idea of what it is doing by the first commit, without having to scroll down forever)

differentreality:
I agree on the improvements suggested (adding social media icons, relevant to https://github.com/openSUSE/osem/issues/1081), however for a new PR.

PR #1867, opened 2017-12-08
Redirect to conference#show

### LINKED ISSUE #1259: Change homepage to next conference#show

davidsf:
In conferences#index, if there are only one current conferences and has a public splashpage, redirect to it

I need to change some tests from root_path to conference_path(conference). I don't kwown if is better to change the conference object (to set splashpage not public for example) in the test and remain root_path. 

fix #1259

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -6,6 +6,10 @@ class ConferencesController < ApplicationController
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
     @antiquated = @conferences - @current
+
+    redirect_to @current.first if @current.count == 1 &&
+                                  @current.first.splashpage &&
+                                  @current.first.splashpage.public
   end
 
   def show
--- a/spec/features/cfp_ability_spec.rb
+++ b/spec/features/cfp_ability_spec.rb
@@ -75,7 +75,7 @@ feature 'Has correct abilities' do
       # Event, booth, track cfps exist
       call_for_tracks = create(:cfp, cfp_type: 'tracks', program: conference.program)
       visit new_admin_conference_program_cfp_path(conference.short_title)
-      expect(current_path).to eq root_path
+      expect(current_path).to eq conference_path(conference)
 
       # Booth and track cfps exist
       conference.program.cfp.destroy!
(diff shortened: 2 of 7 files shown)


bear454:
Please close https://github.com/openSUSE/osem/pull/1774 as well.

differentreality:
My concern is that while there is a current conference people are not able to see a past conference, even if they wanted to (because every time they go to conferences#index they are redirected to conferences#show of the current conference, so if they want to access last year's conference page, they either need to know and type the direct link, or they cannot do it).

Doesn't anyone else share that concern? That we are practically disabling index action/view as long as there is 1 upcoming conference? What if this is in the hosted version? There is 1 upcoming conference, and you cannot see the list of 10 past conferences (different communities/conferences).
(Hosted version issue might be solved with the domain names option that is being developed)

I would agree to the change if it was Conference.all.count == 1, but then of course if this is an annual conference, the second year it happens, you still get the #index, not the #show action.

I'd like more feedback on this.

spec/features/cfp_ability_spec.rb
```diff
@@ -75,7 +75,7 @@
       # Event, booth, track cfps exist
       call_for_tracks = create(:cfp, cfp_type: 'tracks', program: conference.program)
       visit new_admin_conference_program_cfp_path(conference.short_title)
-      expect(current_path).to eq root_path
+      expect(current_path).to eq conference_path(conference)
```
differentreality: We should not change all the tests. The root_path is the default path when CanCanCan's authorization fails. I would create another conference, so that we don't fall into the exception.

When reading these tests, which are about abilities, thus about CanCanCan, it makes more sense to read root_path

bear454:
Your opinion is valid @differentreality , that's why I added you for review here ;-)

I see both cases as valid... maybe this is an ideal place to use a feature (https://github.com/openSUSE/osem/blob/master/config/initializers/feature.rb) switch?

Ana06:
This is a duplicated of https://github.com/openSUSE/osem/pull/1867 Isn't it?

app/controllers/conferences_controller.rb
```diff
@@ -6,6 +6,10 @@ class ConferencesController < ApplicationController
   def index
     @current = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
     @antiquated = @conferences - @current
+
+    redirect_to @current.first if @current.count == 1 &&
```
differentreality: Based on what we agreed on #1774 

> If there is only 1 conference, and that conference has a splashpage, redirect to conferences#show from conferences#index

this should read `Conference.count == 1`

differentreality:
@bear454 perhaps it is :)

I can see why we want to see conferences#index (for a list of ongoing, and past conferences), as well as why people might want to always redirect to the conferences#show of their upcoming conference.
Let people choose.

Unless anyone objects, I will close #1867 and #1774 and update the issue (#1259) accordingly.

hennevogel:
@differentreality please do so

bear454:
See #1994 .

Issue #1863, opened 2017-12-07
How to connect to fresh OSEM docker deployment


vmisson:
Hello,

I have deployed OSEM instance with the docker compose file.
I'm able to open the default web page but I was not able to find the admin user / password info to log on the app... 

Sorry if it's a stupid question...
Thanks in advance.

differentreality:
@vmisson have you created a user? You can use the sign up link to do so.

vmisson:
Yes I already try this but I got a "Error 500: Internal Server Error". It seems to be related to SMTP misconfiguration.
Anything special to do with mailhog?

vmisson:
Even if SMTP is not working my account is created on the app and I'm able to login. 
But it's a user account... how can I be admin ?

differentreality:
@vmisson your first user should be an admin, it might be that docker is not running the seeds. We should have a look at that, thanks:)

To add the attribute is_admin to a user manually execute the following command:
rake user:admin ['email_of_existing_user']

vmisson:
Thanks for your feedback @differentreality 

I just try but I get the following error:
osem@2309942c09b3:/osem$  rake user:admin ['vincent.misson@gmail.com']
NOTE: Gem.datadir is deprecated; use spec.datadir instead. It will be removed on or after 2016-10-01.
Gem.datadir called from /usr/local/bundle/gems/rdoc-generator-fivefish-0.3.0/lib/rdoc/generator/fivefish.rb:29.
rake aborted!
NoMethodError: undefined method `web_console' for #<Rails::Application::Configuration:0x00000000052cf600>
Did you mean?  console
/usr/local/bundle/gems/railties-4.2.9/lib/rails/railtie/configuration.rb:95:in `method_missing'
/osem/config/environments/development.rb:5:in `block in <top (required)>'
/usr/local/bundle/gems/railties-4.2.9/lib/rails/railtie.rb:210:in `instance_eval'
/usr/local/bundle/gems/railties-4.2.9/lib/rails/railtie.rb:210:in `configure'
/usr/local/bundle/gems/railties-4.2.9/lib/rails/railtie.rb:182:in `configure'
/osem/config/environments/development.rb:1:in `<top (required)>'
/usr/local/bundle/gems/activesupport-4.2.9/lib/active_support/dependencies.rb:274:in `require'
/usr/local/bundle/gems/activesupport-4.2.9/lib/active_support/dependencies.rb:274:in `block in require'
[...]

PR #1868, opened 2017-12-09
Fix example file to work with docker mailhog

vmisson:
Issue seems to be link with this commit : https://github.com/openSUSE/osem/commit/da735fe1c9b9e685cb9b87987f3048cfebbe7a3e
More information here : https://github.com/openSUSE/osem/issues/1863


docker-compose.env.example
```diff
@@ -32,9 +32,9 @@ SECRET_KEY_BASE=changemechangemechangeme"
 # you should comment out or remove the mailhog service from docker-compose.yml,
 # too
 OSEM_EMAIL_ADDRESS="osem@mailhog"
-OSEM_SMTP_AUTHENTICATION="login"
```
bear454: String quoting is perfectly valid in YAML ( http://www.yaml.org/spec/1.2/spec.html#style/flow/double-quoted )... I don't see where this change (and the next line) are necessary.

docker-compose.env.example
```diff
@@ -32,9 +32,9 @@ SECRET_KEY_BASE=changemechangemechangeme"
...
-OSEM_SMTP_ADDRESS="mailhog"
-OSEM_SMTP_PORT="1025"
```
bear454: I assume this is the real problem, as this is parsed as a `String`, whereas without the quotes, it's an `Integer`.

Ana06: @bear454  if I understand correctly, we need to only remove the quotes for `OSEM_SMTP_PORT`, right?

@vmisson can you leave the quotes in the rest so we can get this merge? 😉

PR #1871, opened 2017-12-09
Change email text by an icon in proposal page

adrianovieira:
The newest behaviour

new-behaviour


adrianovieira:
Hi! We have two PR (and two issues) to solve the same behaviour. I believe it was my fault opening the that issue and that PR to a work that was only in stand by for a few days. 

Then one of them need to be closed without been merged. Maybe this one

PR #1878, opened 2017-12-11
optional abstract

### LINKED ISSUE #1092: proposal#new with empty abstract is not possible

namangupta01:
Closes #1092

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -38,7 +38,7 @@ class Event < ActiveRecord::Base
   validate :abstract_limit
   validate :before_end_of_conference, on: :create
   validates :title, presence: true
-  validates :abstract, presence: true
+  validates :abstract, presence: true, if: 'event_type.minimum_abstract_length > 0'
   validates :event_type, presence: true
   validates :program, presence: true
   validates :speakers, presence: true


app/models/event.rb
```diff
@@ -38,7 +38,7 @@ class Event < ApplicationRecord
   validate :abstract_limit
   validate :before_end_of_conference, on: :create
   validates :title, presence: true
-  validates :abstract, presence: true
+  validates :abstract, presence: true, if: 'event_type.minimum_abstract_length > 0'
```
Ana06: what about validating that abstract lenght is at least the `minimum_abstract_length` instead of doing this only for the empty case?

Ana06:
@namangupta01 do you need help with this? Please let us know if you need help 😉

app/models/event.rb
namangupta01: @Ana06 As far as i understand you are saying that i should validate for abstract_length to at least `minimum_abstract_length` which is already done in https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L38
Problem here is that when `minimum_abstract_length` is set to 0 words and `abstract_length` is also 0, code is validating for the presence of the abstract but our abstract is `nil` so we should not validate the presence of abstract when `minimum_abstract_length` is 0 and `abstract_length` is also 0 which is what i am trying to do here.
Please let me know if i misunderstood something!

bear454:
@Ana06 the record would only become invalid if edited again, at which time the editor would be able to add an abstract, so I think that's a non-issue.

bear454:
@namangupta01 : can you please update this PR and resolve any remaining test issues, so we can merge this?

namangupta01:
yes sure!

app/models/event.rb
Ana06: @namangupta01 I'd remove this validation, as we don't really want that abstract is present. We want that abstract is longer than the minimum length and longer than the maximum and that's all. Instead I'll ensure that this case is covered in [`abstract_limit`](https://github.com/openSUSE/osem/blob/master/app/models/event.rb#L284)

Ana06: BTW, passing string to be evaluated in `:if` and `:unless` conditional in validations has been removed in Rails 5.2, so we shouldn't use to make updating easier. :wink:

Ana06: BTW, passing string to be evaluated in `:if` and `:unless` conditional in validations has been removed in Rails 5.2, so we shouldn't use to make updating easier. :wink:

hennevogel:
@namangupta01 this seems stale, closing. feel free to re-open this PR when you're ready to address the feedback! Thank you :bouquet:

PR #1764, opened 2017-10-13
Adding issue and pull request templates

akshitahluwalia:
#### Checklist

- [x] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
- [x] My branch is up-to-date with the Upstream `master` branch.
- [x] The tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works(if appropriate).
- [ ] I have added necessary documentation (if appropriate).

#### Short description of what this resolves:

This Pull Request create a uniform template for the users to create an issue and a pull request. It encourages the uniformity and decreases the time to scan through the description of the issues and pull requests.

#### Changes proposed in this pull request:

- Creating PULL_REQUEST_TEMPLATE.md 
- Creating ISSUE_TEMPLATE.md 

#### Fixes Issue

#1763 

@bear454 @differentreality @Ana06
Please look into this PR.

Issue #1763, opened 2017-10-13
Pull Request and Issue Templates


akshitahluwalia:
The repository lacks uniformity in pull requests and issues templates. There should be a uniform way in which the issues and pull requests are described. This enables complete documentation of the issue and pull request.

These templates would be picked up default by the github when a person chooses to create a pull request or issue.

@bear454 @differentreality @Ana06
Please suggest any further changes needed to enhance the flow.

akshitahluwalia:
@differentreality @bear454 @Ana06 Please look into this.

PR #1880, opened 2017-12-12
Add templates for github issues & pull requests

bear454:
re: #1763
based on: #1764

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,15 @@
+**I'm submitting a ..**
+- [ ] Bug Report
+- [ ] Feature Request
+
+**Current behavior:**
+<!-- Describe the bug -->
+
+**Expected correct behavior:**
+<!-- Describe what should happen, without the bug. -->
+
+**Steps to reproduce:**
+<!-- If you are able to illustrate the bug or feature request with an example, please provide steps to reproduce it. -->
+
+**Other information:**
+<!-- List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to fix, Stack Overflow links, forum links, etc. -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,19 @@
+**Checklist**
+
+- [ ] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
+- [ ] My branch is up-to-date with the upstream `master` branch.
+- [ ] The tests pass locally with my changes.
+- [ ] I have added tests that prove my fix is effective or that my feature works(if appropriate).
+- [ ] I have added necessary documentation (if appropriate).
+
+**Short description of what this resolves/which [issues](https://github.com/openSUSE/osem/issues) does this fix?:**
+
+<!-- List the issue number resolved with this change; if there is no open issue, describe the problem this request solves -->
+
+-
+
+**Changes proposed in this pull request:**
+
+<!-- Summarize the changes, using declarative language. -->
+
+-


differentreality:
I would **not** make the templates mandatory, but if you think they are helpful to new contributors, I don't object in having them.

@hennevogel @lagartoflojo @ChrisBr @Ana06 what do you think?

lagartoflojo:
I think that these templates are not supposed to be useful for new contributors, but rather for existing contributors, who want to receive quality issue reports, feature requests and pull requests. In that regard, I think this is a good starting point, and if we see that people don't use the templates correctly, we can very easy adjust them in the future. 👍

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
++**I'm submitting a ..**
```
lagartoflojo: Are these `++` at the beginning of each line necessary?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
++- [ ] Bug Report
++- [ ] Feature Request
++
++**Current behavior:**
++<!-- Describe the bug and how it creeped in. -->
```
lagartoflojo: If you asked me "how it creeped in," I would try to look for the specific commit that changed the line the triggered the bug, and I don't think that's what we expect here. I'd just leave it at "Describe the bug."

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
++
++**Expected behavior:**
```
lagartoflojo: Just to be more explicit, I'd say "Expected correct behavior"

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
++<!-- Describe what the behavior would be without the bug. -->
```
lagartoflojo: How about:

> Describe what should happen without the bug

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
++
++**Steps to reproduce:**
++<!-- If you are able to illustrate the bug or feature request with an example, please provide steps to reproduce -->
```
lagartoflojo: reproduce **it**

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
++
++**Related code:**
```
lagartoflojo: Not sure if we need this section. Bugs will not necessarily come from developers.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
++#### Checklist
++
++- [ ] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
++- [ ] My branch is up-to-date with the Upstream `master` branch.
++- [ ] The tests pass locally with my changes.
++- [ ] I have added tests that prove my fix is effective or that my feature works(if appropriate).
```
lagartoflojo: missing space between `works(if`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
++#### Short description of what this resolves:
++
++
++#### Changes proposed in this pull request:
++
++-
++-
++-
++
++#### Fixes Issue
```
lagartoflojo: The ISSUE_TEMPLATE.md uses `**asterisks**` for each section, but in this file we use `#### four hashes`. I would normalize and use the same one in both files (I don't have any preference *which*).

bear454:
Thanks for the corrections @lagartoflojo !

bear454:
@lagartoflojo your review got dismissed when I rebased onto master. Would you mind checking the box one more time?

PR #1881, opened 2017-12-12
Link to an existing video with embeddable content

hennevogel:
Otherwise the Commercial validation hits and this breaks automatic deployment of demo.osem.io


differentreality:
Hmmm thought we had merged that already!

PR #1883, opened 2017-12-12
link_to_organization method in version is fixed

### LINKED ISSUE #1724: not returning link to organization
link_to_organization method in versions_helper.rb is not returning link to organization
[https://github.com/openSUSE/osem/blob/master/app/helpers/versions_helper.rb](url)

namangupta01:
Closes #1724 
As link_to_conference is returning link to edit_admin_conference_path same is done for link_to_organization which is returning the link to edit_admin_organization_path

### DIFF
--- a/app/helpers/versions_helper.rb
+++ b/app/helpers/versions_helper.rb
@@ -9,9 +9,14 @@ module VersionsHelper
   def link_to_organization(organization_id)
     return 'deleted organization' unless organization_id
 
-    org = Organization.find_by(id: organization_id)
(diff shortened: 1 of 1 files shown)


differentreality:
Well conference doesn't really have a show page that shows the information found in the form :) 
That's not the case with organization though. Moreover that link is visible to non organization admins, and for them the link will not work. Please try this locally with users with proper rights (not an admin): a conference organizer (not organization admin) should be able to click that link.

Ana06:
@namangupta01 @differentreality what is the state of this?

hennevogel:
according to @differentreality this is moot. closing

PR #1886, opened 2017-12-12
Use distinct instead of uniq in user.rb because deprectation

davidsf:
Fix deprecation warning in user.rb:

```
DEPRECATION WARNING: uniq is deprecated and will be removed from Rails 5.1 (use distinct instead) (called from block in <class:User> at /vagrant/app/models/user.rb:48)
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,8 +45,8 @@ class User < ApplicationRecord
   attr_accessor :login
 
   has_many :event_users, dependent: :destroy
-  has_many :events, -> { uniq }, through: :event_users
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -45,7 +45,7 @@ def by_conference(conference)
   attr_accessor :login
 
   has_many :event_users, dependent: :destroy
-  has_many :events, -> { uniq }, through: :event_users
+  has_many :events, -> { distinct }, through: :event_users
   has_many :presented_events, -> { joins(:event_users).where(event_users: {event_role: 'speaker'}).uniq }, through: :event_users, source: :event
```
AEtherC0r3: Here too

differentreality:
@AEtherC0r3 is this good to go?

PR #1896, opened 2017-12-16
Add unregistered speakers scope to program

ChristinaAN:
And use it in Conference#registration_limit_exceeded?

### DIFF
--- a/app/models/program.rb
+++ b/app/models/program.rb
@@ -51,6 +51,10 @@ class Program < ApplicationRecord
     def registered(conference)
       joins(:registrations).where('registrations.conference_id = ?', conference.id)
     end
+
+    def unregistered(conference)
+      self - registered(conference)
+    end
   end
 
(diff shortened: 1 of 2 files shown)


differentreality:
@ChristinaAN nice work :+1: 
Please check the failure in linters, and fix the problem so that all checks pass :-)
Thanks!

app/models/program.rb
```diff
@@ -51,6 +51,10 @@ def confirmed
     def registered(conference)
       joins(:registrations).where('registrations.conference_id = ?', conference.id)
     end
+
+    def unregistered(conference)
+      self - self.registered(conference)
```
Ana06: this `self` is not needed, that's why Rubocop is complaining. You should write like:

``` Ruby
self - registered(conference)
```

differentreality:
This looks good :+1: Thank you @ChristinaAN, merging :tada:

PR #1900, opened 2017-12-16
Added autofocus in a number of forms

Astradeni:
Added autofocus in a number of forms, plus correction of a few spelling mistakes

### DIFF
--- a/app/views/admin/campaigns/_form.html.haml
+++ b/app/views/admin/campaigns/_form.html.haml
@@ -11,7 +11,7 @@
   .col-md-8
     = semantic_form_for(@campaign, url: (@campaign.new_record? ? admin_conference_campaigns_path : admin_conference_campaign_path(@conference.short_title, @campaign))) do |f|
       = f.inputs do
-        = f.input :name
+        = f.input :name, input_html: { autofocus: true }
         = f.inputs name: 'UTM Parameters' do
           = f.input :utm_campaign, label: 'Campaign', hint: 'Groups all of the content from one campaign. E.g. 20percentpromocode'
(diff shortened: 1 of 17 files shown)


app/views/admin/campaigns/_form.html.haml
```diff
@@ -11,7 +11,7 @@
   .col-md-8
     = semantic_form_for(@campaign, url: (@campaign.new_record? ? admin_conference_campaigns_path : admin_conference_campaign_path(@conference.short_title, @campaign))) do |f|
       = f.inputs do
-        = f.input :name
+        = f.input :name, input_html: { autofocus: true }
```
Ana06: Is the `input_html` needed?  I mean, this should work as well:

``` Ruby
= f.input :name, autofocus: true
```

Doesn't it? 🤔

differentreality: It does not.

app/views/admin/commercials/index.html.haml
```diff
@@ -14,7 +14,7 @@
   .row
     .col-md-6
       = semantic_form_for(@commercial, url: admin_conference_commercials_path(conference_id: @conference.short_title)) do |f|
-        = f.input :url, label: 'URL', as: :string, input_html: { required: 'required' },
+        = f.input :url, label: 'URL', as: :string, input_html: { required: 'required' }, input_html: { autofocus: true },
```
bear454: Please fold this into the existing hash. For example:

```
..., input_html: { required: 'required', autofocus: true }, ...
```

Astradeni:
Done... Thanks for the suggestion!

screenshot_20180104_011304

differentreality:
Thank you @Astradeni :+1:

PR #1906, opened 2017-12-21
Fixes the leak of names of all the users

### LINKED ISSUE #1884: Personal information leak when editing proposals
### Problem:

Proposal submitters has access to personal data of (probably?) all the OSEM registered users. 

This is not acceptable and probably could be violating data privacy laws.

Verified in master branch at b9712209

### Expected behaviour:

3rd persons personal information should not be exposed to non-admin users.

### Steps to reproduce:

1. register at OSEM
2. submits a proposal
3. go to edit your submited proposal at, say, http://localhost:9292/conferences/MyConf/program/proposals/2/edit
4. edit the Speakers field
5. field displays a list of another users to select your co-speakers
6. this list exposes information for any other registered users: name username and email.

### Mitigation

You can do a quick fix with this patch:

```
diff --git a/app/views/proposals/_proposal_form.html.haml b/app/views/proposals/_proposal_form.html.haml
index 4b66240c..71bea2eb 100644
--- a/app/views/proposals/_proposal_form.html.haml
+++ b/app/views/proposals/_proposal_form.html.haml
@@ -4,7 +4,7 @@
 
     = f.input :subtitle, as: :string
 
-    = speaker_selector_input f
+    -# = speaker_selector_input f
 
     = track_selector_input f
```

At the cost of losing the feature of adding other speakers by the submitter.

anmoliiitd823:
Closes #1884 
Before 
screenshot from 2017-12-21 15-23-07

After
screenshot from 2017-12-21 16-03-02

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -136,17 +136,17 @@ module ApplicationHelper
     safe_join(event.speakers.map{ |speaker| link_to speaker.name, admin_user_path(speaker) }, ',')
   end
 
-  def speaker_selector_input(form)
+def speaker_selector_input(form)
     users = User.active.pluck(:id, :name, :username, :email).map { |user| [user[0], user[1].blank? ? user[2] : user[1], user[2], user[3]] }.sort_by { |user| user[1].downcase }
     form.input :speakers, as: :select,
-                          collection: options_for_select(users.map {|user| ["#{user[1]} (#{user[2]}) #{user[3]}", user[0]]}, @event.speakers.map(&:id)),
+                          collection: options_for_select(users.map {|user| ["#{user[2]} #{user[3]}", user[0]]}, @event.speakers.map(&:id)),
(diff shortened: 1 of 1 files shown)


differentreality:
Hi @anmoliiitd823 
aren't those 2 screenshots the same? Or am I missing something?

Can you please explain how you are solving the issue by removing id field from the collection?

In the future, please prefer to work on issues that are labeled, which means the maintainers have confirmed them. Thank you :)

anmoliiitd823:
As during the sign up form it is written that other users can only view the usernames.So I only removed the name of the user.
screenshot from 2017-12-18 00-05-45
In the Before screenshot any user can view the real names of other users also which violates the statement written during the sign-up form.

differentreality:
@anmoliiitd823 thank you for trying this out and Happy New Year!
I am pretty sure the point of the issue was the whole idea of the search box itself.

I would strongly suggest you focus on issues that are labeled, starting with a few labeled Junior or good first issue.

I will close this PR - don't let that discourage you though :)

Issue #1789, opened 2017-11-10
Remove data table from program/cfps


rahul2240:
we can create only 3 types of cfps, so i don't think data tables is required here.

Ana06:
@rahul2240 what do you exactly mean? 🤔

PR #1907, opened 2017-12-21
remove datatable in program/cfps

evris99:
In response to issue #1789 .
Before:
screenshot from 2017-12-27 14-53-27
After:
screenshot from 2017-12-27 14-55-17


differentreality:
Thanks @evris99 :+1: 

And Happy New Year :tada:

PR #1914, opened 2017-12-22
currency of a single ticket could be updated now

### LINKED ISSUE #1913: Not able to update the free ticket's currency, created when a new conference is created
**I'm submitting a ..**
- [X] Bug Report
- [ ] Feature Request

**Current behavior:**

The free ticket created at time of conference creation has default currency as USD. 
This ticket's currency is not updating, even when this is the only conference ticket available.

ticket-bug-1

The ticket's currency should update whenever it is the only ticket of the conference
**Steps to reproduce:**

1) Make a new conference (or there should be a single ticket)
2) Try to update the free ticket's currency

ViditChitkara:
closes #1913

**Checklist**


app/models/ticket.rb
```diff
@@ -72,8 +72,11 @@ def tickets_sold
   private
 
   def tickets_of_conference_have_same_currency
-    unless Ticket.where(conference_id: conference_id).all?{|t| t.price_currency == price_currency }
-      errors.add(:price_currency, 'is different from the existing tickets of this conference.')
+    tickets = Ticket.where(conference_id: conference_id)
+    unless tickets.count.zero? || (tickets.count == 1 && self == tickets.first)
```
differentreality: How about instead of this unless statement you use
`return true if condition`

ViditChitkara: That's a nice idea, but I guess return if condition should be fine (for line 76). There is no need to return 'true'. What do you think?

differentreality:
@ViditChitkara are you going to take care of this?

ViditChitkara:
Yes, I'll take care of this!! Just pushing the changes

ViditChitkara:
@differentreality I have made the changes!! Could you please have a look at this?

differentreality:
This looks good @ViditChitkara 
There is an extra empty file probably by mistake

I would like to add tests that would have caught the bug, how are your rspec skills? :)

ViditChitkara:
Still learning about rspec. Will give it a go!!😃😃

ViditChitkara:
@differentreality I have made the changes! Please have a look!

ViditChitkara:
@differentreality Any changes required for this??

ViditChitkara:
@differentreality rebased!!

spec/models/ticket_spec.rb
```diff
@@ -177,4 +178,25 @@
       end
     end
   end
+
+  describe 'currency updation' do
```
differentreality: This should say `currency update`

hennevogel:
I guess we can live with the wrong work in the `spec`, merging.

Issue #1937, opened 2018-01-06
NoMethodError in revision history


rahul2240:
![abc](https://user-images.githubusercontent.com/25618739/34639727-65e6b39a-f30b-11e7-84eb-48e25271fca7.png)

Bug: NoMethodError in revision history

Steps to reproduce:
1) Create organization
2) Create conference
3) Delete organization
4) Visit revision history page

PR #1943, opened 2018-01-07
Fix NoMethodError in revision history

rahul2240:
Bug: Roles are deleted as the organization is deleted, so finding roles from usersrole's role id will return nil because role doesn't exist for that role id.

Fix: Previous version of role model is required if organization is deleted.

closes https://github.com/openSUSE/osem/issues/1937

**Checklist**


app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -9,7 +9,8 @@
 
 - when 'UsersRole'
   - users_role = current_or_last_object_state(version.item_type, version.item_id)
-  - role = Role.find_by(id: users_role.role_id) if users_role
+  - if users_role
+    - role = Role.find_by(id: users_role.role_id) || PaperTrail::Version.where(item_type: 'Role', item_id: version.item_id).where(id: users_role.role_id)
```
Ana06: could this still be `nil`? should we maybe add a `.try` in `role.name`? 🤔

rahul2240: yes right using .try will be better.

rahul2240:
@Ana06 I have added information to commit message. :smile:

app/views/admin/versions/_object_desc_and_link.html.haml
```diff
@@ -9,13 +9,13 @@
 
 - when 'UsersRole'
   - users_role = current_or_last_object_state(version.item_type, version.item_id)
-  - role = Role.find_by(id: users_role.role_id) if users_role
+  - role = Role.find_by(id: users_role.role_id) || PaperTrail::Version.where(item_type: 'Role', item_id: version.item_id).where(id: users_role.role_id)
```
differentreality: version.item_id is the ID of the UsersRole record, not the id of a Role record. 

Why not find role like this:
`- role = current_or_last_object_state('Role', users_role.role_id)`

differentreality: And for name
- role_name = role.try(:name) || PaperTrail::Version.where(item_type: 'Role', item_id: users_role.role_id).last.changeset[:name].second

So that we can ensure showing something.

rahul2240:
@differentreality Please review.

app/views/admin/versions/_object_desc_and_link.html.haml
differentreality: While you are at it, you may want to remove the (Role Deleted) and (Organization Deleted) and substitute them with something more meaningful, like we do in the rest of the cases.

Issue #1942, opened 2018-01-07
User with accepted track request can see unregistered speakers of all events.


rishabhptr:
**I'm submitting a ..**
- [x] Bug Report

**Current behavior:**
When a user with a accepted track request tries to manage his track he can see the unregistered speakers of all events not those restricted to his track under reports option.
![screenshot from 2018-01-07 20-07-15](https://user-images.githubusercontent.com/26951824/34650382-635999fe-f3e6-11e7-98d8-c6d8cad841fd.png)

**Expected correct behavior:**
For other information about events in the reports page we are initialising instance variable according to the user's ability to access those events so here also user should be shown missing speakers in events under his track.

**Steps to reproduce:**
After track request is accepted, click on manage option, you could see users who are speakers and haven't registered yet.

differentreality:
What is the manage option?

rishabhptr:
@differentreality It is available in track#index when your track request is selected and you confirm it.
![screenshot from 2018-01-07 22-34-12](https://user-images.githubusercontent.com/26951824/34651786-19c86a80-f3fb-11e7-9985-9542ef067a7a.png)

membrq:
Can I claim this?

membrq:
I'm having trouble understanding this issue. What determines whether a speaker is "checked in" vs missing? Is it an attribute that gets set somewhere?

Where's the best place to put this code? `reports/_missing_speakers.html.haml`? My first thought was to add an `if` condition around ln 21-22 and if that's able to work, I'd have to find a way to update ln 6 dynamically. Would appreciate feedback!

differentreality:
People (speakers included) can register to a conference. When they arrive there is the attribute **attended** which can be enabled. That's what "checked in" means.

It is very helpful if you provide links to the files/lines you are referring to so that we are all at the same page. Assuming you refer to https://github.com/openSUSE/osem/blob/master/app/views/admin/reports/_missing_speakers.html.haml#L21 I think it makes sense to change the value of the variable @missing_event_speakers altogether.

membrq:
Ah ok, that helps to know. I thought the `missing` attribute had to do with their registration status.

Also, thanks for your input. That was the file I was referring to and I'll make sure to include links next time!

PR #1965, opened 2018-01-13
Added condition to user scope for missing speakers report.

membrq:
**Checklist**

- #1942 User with accepted track request can see unregistered speakers of all events. 

- Added condition that checks whether the signed-in user and the missing speaker have an event in common. If they do, the missing speaker should be displayed in the `missing speakers report`.

This is a WIP. Let me know if any of my assumptions about the issue are incorrect.

### DIFF
--- a/app/views/admin/reports/_missing_speakers.html.haml
+++ b/app/views/admin/reports/_missing_speakers.html.haml
@@ -19,14 +19,15 @@
       - @missing_event_speakers.each do |event_user|
         - speaker = event_user.user
         - event = event_user.event
-        %tr
-          %td= link_to speaker.name, admin_user_path(speaker)
-          %td
-            - if @conference.user_registered?(speaker)
-              = link_to 'Yes', admin_conference_registrations_path(@conference.short_title)
-            - else
-              No
-          %td= link_to event.title, edit_admin_conference_program_event_path(@conference.short_title, event)
-          %td= event.room.name if event.room
-          %td= event.time.to_date if event.time
-          %td= event.time.strftime('%H:%M') if event.time
+        - if @events.include?(event)
+          %tr
+            %td= link_to speaker.name, admin_user_path(speaker)
+            %td
+              - if @conference.user_registered?(speaker)
+                = link_to 'Yes', admin_conference_registrations_path(@conference.short_title)
+              - else
+                No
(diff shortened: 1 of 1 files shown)


differentreality:
@membrq there is a recommendation on how to implement this in the issue: change the value calculated in the controller. Did you have trouble implementing it?

membrq:
I went with a different approach because I hit a roadblock with making changes in the controller. I'm revisiting it now.

I tried modifying the `@missing_event_speakers` query in https://github.com/openSUSE/osem/blob/master/app/controllers/admin/reports_controller.rb so it only returns missing speakers who share events with the current user. Here are a few of the things I've tried.

Option 1: (doesn't raise error; doesn't give expected results)
```
current_user_events = @events.pluck(:id) #from what I understand @events = current user's events
@missing_event_speakers = EventUser.joins(:event)
                                .where('event_role = ? and program_id = ?', 'speaker', @program.id)
                                .where(event_id: current_user_events)
                                .where.not(user_id: attended_registrants_ids)
                                .includes(:user, :event)
```

Option 2: (raises error)
```
@missing_event_speakers = EventUser.joins(:event)
                                .where('event_role = ? and event_id = ?', 'speaker', @events.pluck(:id))
                                .where.not(user_id: attended_registrants_ids)
                                .includes(:user, :event)
```

Option 3: (raises error)
```
@missing_event_speakers = EventUser.joins(:event)
                                .where('event_role = ? and program_id = ? and event_id = ?', 'speaker', @program.id, @events.pluck(:id))
                                .where.not(user_id: attended_registrants_ids)
                                .includes(:user, :event)
```

Both options 2 and 3 raise errors with line 6 here: https://github.com/openSUSE/osem/blob/master/app/views/admin/reports/_missing_speakers.html.haml. 

I haven't worked much work ActiveRecord Query. Am I off base with this approach?

differentreality:
@membrq your 1st option seems to be correct. And you don't even need to create the extra variable
`.where(event_id: @events.pluck(:id))`

Is it possible that your user is an organizer of the conference, and not just track organizer? Organizers have access everywhere, and as a result @events would not include only the track events.

membrq:
Ok, thanks for the feedback. I'll work on it this week.

hennevogel:
@membrq any news on this?

differentreality:
@membrq shall we work together to finish this?

hennevogel:

PR #1892, opened 2017-12-13
hoptoad gem is replaced with airbrake gem

namangupta01:
Closes #1891

PR #1993, opened 2018-03-02
Reduce deprecation warnings

bear454:
Reduce deprecations as much as possible; the only remaining deprecations are from Hoptoad, which would be addressed by PR #1892 .

* Update _coveralls_ gem (and mandatory depedencies)
* Update _money-rails_ gem (and mandatory dependencies)
* Update `linear-gradient` CSS to match CSS standards ( `to DIR` argument, and drop deprecated browser-specific variants)


bear454:
@Ana06: description updated.

PR #1994, opened 2018-03-04
Allow root URL to redirect to a conference via ENV var.

bear454:
**Checklist**

- #1259

- Unlike #1259 or #1774, this change:
- allows for root to show `conferences#index` by default, 
- avoids the overhead of running the `conferences#index` action and all it's loading work by working in routing, 
- allows `/` to redirect to any specific conference (so you can have one conference current, but start on the next conference ;-)
- peforms a redirect, maintaining the RESTful url (e.g. `/ => 301 /conferences/FOO` ) instead of directly rendering `conferences#show` on the root URL.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -263,6 +263,8 @@ group :test do
   gem 'json-schema'
   # For using 'assigns' in tests
   gem 'rails-controller-testing'
+  # For managing the environment
+  gem 'climate_control'
 end
 
 group :development, :test do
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -263,6 +263,8 @@ group :test do
   gem 'json-schema'
   # For using 'assigns' in tests
   gem 'rails-controller-testing'
+  # For managing the environment
```
hennevogel: What do we need this for if it does not work with rspec? :nerd_face:

bear454: *hope* for the future :grin:

hennevogel: I guess hope dies last then...

PR #2006, opened 2018-03-18
[wip] Fix ahoy tracking and visits

differentreality:
Ahoy hasn't been working for a while. 

A recent migration attempting to make it work by changing ahoy_events.visit_id to integer type cannot run in existing installations due to a content type mismatch. (Problem still remains!)
https://github.com/openSUSE/osem/blob/master/db/migrate/20171118113113_change_visit_id_type_of_ahoy_events_to_integer.rb

This PR:

* updates ahoy gem to the latest version
* changes the ahoy initializer as per https://github.com/ankane/ahoy/blob/master/docs/Ahoy-2-Upgrade.md#activerecordtokenstore

Since things weren't working before, I haven't gone with the uuid solution https://github.com/ankane/ahoy/blob/master/docs/Ahoy-2-Upgrade.md#activerecordstore (which when tried didn't immediately work)

Old fields are **not** deleted - if we want to do something with them, and visit.visitor_id is no longer being used.

Campaigns' counts now show up properly!


bear454:
@differentreality can you please look into the Gemfile issues found on Travis CI ?

PR #2012, opened 2018-03-23
Specify url input type for client-side validation

michaelkl:
**Checklist**

- [x] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
- [x] My branch is up-to-date with the upstream `master` branch.
- [x] The tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works(if appropriate).
- [ ] I have added necessary documentation (if appropriate).

**Short description of what this resolves/which [issues](https://github.com/openSUSE/osem/issues) does this fix?:**

- Add Validation to lodgings website. #1948

**Changes proposed in this pull request:**

- Change `<input type="text">` to `<input type="url">` for `website_link` control of Lodging edit form.

PR #2017, opened 2018-03-27
Fix tests failures

michaelkl:
**Checklist**

- Fix some tests failures

- Fix test failures in some timezones.
- Fix test failures caused by logging changes of `updated_at` attribute.
- Fix test failures caused by hardcoded roles order.

### DIFF
--- a/app/models/event_type.rb
+++ b/app/models/event_type.rb
@@ -2,7 +2,8 @@ class EventType < ApplicationRecord
   belongs_to :program, touch: true
   has_many :events, dependent: :restrict_with_error
 
-  has_paper_trail meta: { conference_id: :conference_id }
+  has_paper_trail meta: { conference_id: :conference_id },
+                  ignore: %i[updated_at]
 
   validates :title, presence: true
   validates :length, numericality: {greater_than: 0}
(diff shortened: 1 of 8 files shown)


michaelkl:
@bear454 Tests fixes extracted from #2012

bear454:
Hmm... While I don't have any objections to the changes... I wonder how you're testing, as the suite passes on TravisCI (most of the time ;-).

michaelkl:
My timezone is UTC+5. I'm running the project with PostgreSQL as a database (because I have it locally for other projects and I was too lazy to setup MariaDB for a single project :-) ). I've also tried sqlite with exactly the same result.
I guess something breaks when we save timezone-unaware values to database and extract them later as `ActiveSupport::TimeWithZone`.

bear454:
I'm running `osem` on Postgres as well, in UTC+~~8~~7 (P**DS**T - I _hate_ daylight savings, but that's another rant), on Ruby 2.4.1 - https://lfnw.org

michaelkl:
I'm on Ruby 2.3.5.
[Example of failures I get](https://gist.github.com/michaelkl/a5181e00b21b2261d5b119948774bfcd)

bear454:
Okay, so it must be a ruby-2.3 thing. Since the tests still pass on 2.4, I don't see any reason not to approve.

PR #2032, opened 2018-04-07
Update paper_trail and show ticket price properly in versions

differentreality:
* Update paper_trail gem to the latest version
* Small refactor in the file with the changed values table to make it more obvious where each **td** is created
* Fix how ticket price shows up in the revision_history 
  * show price (not price_cents) 
  * show currency 
It showed price_cents 4500, now it shows $45.


bear454:
Hi @differentreality , please correct the test failure: https://travis-ci.org/openSUSE/osem/jobs/363495488#L4724

PR #2064, opened 2018-06-03
 make sure that the registed link is there but only accessible to admins

NeimadTL:
bugfix on index proposals/index so the registed link is only accessible to admins

### DIFF
--- a/app/views/proposals/index.html.haml
+++ b/app/views/proposals/index.html.haml
@@ -76,7 +76,9 @@
                   = event.event_type.title
                   = "(#{event.event_type.length} min)"
                   = "in #{event.track.name}" if event.track
-                  
+                  - if event.require_registration && current_user.is_admin
+                    %br
+                    = link_to registered_text(event), registrations_conference_program_proposal_path(@conference.short_title, event), class: 'btn btn-xs btn-danger'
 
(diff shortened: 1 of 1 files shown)


differentreality:
Why should the link only be available for admins? Not even conference organizers? What about proposal submitters?

NeimadTL:
I thought the conference organizers were the admins of the conferences they've created.
So the rule should be : link available for admins `or` conference organizers `or` proposal submitters. Is that correct @differentreality ?

NeimadTL:
`if event.require_registration && (current_user.is_admin || event.event_users.include?(current_user))`
@differentreality, can you tell me if this is better please ?

differentreality:
The check should be based on the ability of the user, and not the user's role. 
So we should check if the user can access the controller#action the button takes us to.

To check the abilities of each user we use [CanCanCan](https://github.com/CanCanCommunity/cancancan) gem and in order to check a certain ability you have to use a syntax like this:
`can? :index, @proposal`
Please read the gem's manual for more details: https://github.com/CanCanCommunity/cancancan/wiki/Checking-Abilities

If you haven't already found it, there is information on the roles in our wiki: https://github.com/openSUSE/osem/wiki/Roles-and-Permissions

hennevogel:

PR #2075, opened 2018-06-15
change unless else to if else

bear454:
(cherry picked from commit 7fa38e2697a8bb5da288ec5b264b7ac95612203b)

**Checklist**

- the condition was unnecessarily complex by using unless with an else clause.

- Simplify logic by reversing behavior to an if/else.
- *Bonus*: fix a typo


spec/models/ticket_spec.rb
```diff
@@ -192,7 +192,7 @@
     end
   end
 
-  describe 'currency updation' do
+  describe 'currency update' do
```
Ana06: how is this related to the change? it was also not in the cherry picked commit. :confused:

bear454: It was from a second commit in the parent PR; I thought it trivial; but I can submit as a separate PR.

PR #2077, opened 2018-06-15
Update schema

Ana06:
It seems like `EventUser` was missed in the schema. :confused:


bear454:
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L240-L247 ?

differentreality:
@Ana06 it's there as event_users
https://github.com/openSUSE/osem/blob/master/db/schema.rb#L240

Ana06:
It was created when starting the project from scratch... maybe I did something wrong... :confused:

PR #2078, opened 2018-06-19
Boostrap switch cleanup

bear454:
**Checklist**

- Repetitive arguments on boostrapSwitch instances
- bootstrapSwitch can only be set up on page load

- DRY up the configuration using default values
-  Set up an external function for setting up a bootstrapSwitch


differentreality:
We should probably reflect the changes in the wiki page as well
https://github.com/openSUSE/osem/wiki/Bootstrap-Switch-Checkbox

I think the switch use is not 100% straight forward, and I have found myself checking the wiki section more than a few times, so I believe it is useful to have the information there, and makes it super easy to just copy-paste the code from the wiki and adapt in the code.

PR #2094, opened 2018-08-22
Add email for proposal submission

differentreality:
Adds an email template for when a proposal is submitted.

This is related to https://github.com/SeaGL/osem/issues/18 which I think is a good idea to have.


bear454:
:+1:

PR #2171, opened 2018-09-05
Feautre/setup task

hennevogel:


lib/tasks/setup.rake
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+
+# Everything that needs to be done before you can use this app
```
Ana06: Why is this needed? In my opinion the `setup` name implies this already

lib/tasks/setup.rake
```diff
@@ -0,0 +1,49 @@
...
+    copy_example_file(".env.#{Rails.env}", 'dotenv.example')
+  end
+
+  desc 'Bootstrap the application'
+  task bootstrap: [:configure, :environment] do
+    puts 'Creating the database...'
+    begin
+      # Only do this if the database does not exist...
+      Rake::Task['db:version'].invoke
+      # rubocop:disable Style/RescueStandardError
```
Ana06: Why is that needed? Using `rescue` should be the default in Rubocop :confused:

hennevogel: This is disabling the reporting the `rescue` in the middle :)

Ana06: @hennevogel that is what it looks like, but I think it shouldn't be needed, because the cop is enforcing `rescue` and not complaining about it (at least is what the documentation says). :confused:

hennevogel: No it enforces rescuing `StandardError` instead of all exceptions (like we do in this case)

PR #2182, opened 2018-09-18
Put database.yml into the code base

hennevogel:
It works by default, we have a mechanism to configure it (dotenv.example)
and we want to be a 12factor app, right? Let's do this...


hennevogel:
> for established forks that are already maintaining their own database.yml...

@bear454 you mean *development* forks right?

bear454:
Forks that actually merge back from master. LFNW, for example, is still ~40 commits ahead of master.

hennevogel:
@bear454 what would stop you from rebasing?

PR #2184, opened 2018-09-19
Drop knapsack

hennevogel:
It constantly fails with "Mysql2::Error: This connection is in use by...".
Might be because of database cleaner? But until we can get this stable,
rip it out again.

Replaced by a matrix based on test type. Sorry @bear454


bear454:
It looks like there's some API problem - Travis isn't finishing tests, and Codecov is getting triggered a lot... I recommend waiting until we get a clean test out of Travis before merging.

hennevogel:
@bear454 I'm pushing every 5 minutes ;-)

hennevogel:
Alright, I'm happy. here we go...

bear454:
Whoah...when did the ability specs cut down to 3mins ? They used to be 17!?

PR #2185, opened 2018-09-19
User schedule

differentreality:
User can select favorite events, and show only their individual schedule
Continuing #1142 by @Ana06 

* Introduce an association between User and Event
* Show individual User schedule


Ana06:
@differentreality thanks for taking care of this! :cupid:

bear454:
Just a little lint to remove:

```
app/models/event.rb:35:3: C: Rails/HasAndBelongsToMany: Prefer has_many :through to has_and_belongs_to_many.
  has_and_belongs_to_many :favourite_users, class_name: 'User'
  ^^^^^^^^^^^^^^^^^^^^^^^
app/models/user.rb:79:3: C: Rails/HasAndBelongsToMany: Prefer has_many :through to has_and_belongs_to_many.
  has_and_belongs_to_many :favourite_events, class_name: 'Event'
  ^^^^^^^^^^^^^^^^^^^^^^^
app/controllers/proposals_controller.rb:11:1: C: Layout/EmptyLines: Extra blank line detected.
```

hennevogel:
I know it feels weird but try [out code-review-by-commit](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/). It makes things so much faster :cowboy_hat_face:

hennevogel:
@differentreality once more! You can make it! ;-)

bear454:
@differentreality: do you think you'll get this wrapped up soon? I'd *love* to have this in by mid-February, for LFNW2019!

hennevogel:
go stella, go stella

Im rooting for you

differentreality:
Rebased :grinning:

differentreality:
What else would we need for this?

bennettscience:
Any chance of having this merged in soon-ish? We're hosting an event on Feb 6th and this would really help out!

differentreality:
Yeah I'd love to see this merged as well! Still pending review though..
I should probably take a look at it again too.

bennettscience:
@differentreality If I manually pull this commit into my local copy, I should just be able to migrate the database with the new relationship, right?

app/controllers/proposals_controller.rb
```diff
@@ -83,11 +83,22 @@ def update
...
                   notice: 'Proposal was successfully updated.'
     else
-      flash.now[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
+      flash[:error] = "Could not update proposal: #{@event.errors.full_messages.join(', ')}"
       render action: 'edit'
     end
   end
 
+  def toggle_favorite
+    user = User.find(params[:favourite_user_id])
```
cycomachead, line 92: I've finally started working on merging this into my own fork for an event this summer. It seems to me that this allows anyone to toggle anyone else's favorite events. Should it not just use `current_user`? Since you should be logged in and have a cookie?

I can't really see a use case as well for someone else (such as an admin) toggling another user's favorites so I think the `favourite_user_id` parameter could probably just be removed, but I might be missing something.

app/controllers/schedules_controller.rb
```diff
@@ -52,9 +53,17 @@ def events
     @events_schedules = @program.selected_event_schedules(
       includes: [:room, { event: %i[track event_type speakers submitter] }]
     )
+
+    @events_schedules = @events_schedules.select{ |e| e.event.favourite_users.exists?(current_user.id) } if @events_schedules && current_user && @favourites
```
cycomachead, line 57: This may be a slightly different use case, but I think related enough...
How would you feel about showing a schedule which is favorite events + events where you are speaking? 

I am thinking about creating an `event_schedule_for_user(user)` method, which is essentially your personal conference agenda. My app has assigned volunteers which could also how up there.

app/views/schedules/events.html.haml
```diff
@@ -5,6 +5,13 @@
   %h1.text-center
```
cycomachead, line 5: The above line `  -if @events_schedules.any?` probably should be removed if this is merged in. It leads to a weird looking case when you have no events in your favorites schedule and the tabs disappear.

PR #2187, opened 2018-09-20
Consistently use sqlite dev/test dbs

bear454:

- The default database.yml caused dev & test sqlite dbs to move from the prior
location, `osem/db/[ENV].sqlite3` to `osem/osem_[ENV]`;

- gitignore Add osem_development & osem_test
- Use the default locations for dev & test sqlite dbs

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -3,6 +3,13 @@
   if ENV['OSEM_DB_ADAPTER'] == 'mysql2'
     encoding = 'utf8'
   end
+
+  development_database = ENV['OSEM_DB_NAME'] || 'osem_development'
+  test_database = ENV['OSEM_DB_NAME'] || 'osem_test'
+  if ENV['OSEM_DB_ADAPTER'] == 'sqlite3'
+    development_database = "db/#{development_database}.sqlite3"
+    test_database = "db/#{test_database}.sqlite3"
+  end
 %>
 
 
@@ -19,14 +26,14 @@ default: &default
 
 development:
   <<: *default
-  database: osem_development
+  database: <%= development_database %>
 
 # Warning: The database defined as "test" will be erased and
 # re-generated when you run "rake".
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: osem_test
+  database: <%= test_database %>
 
 production:
   <<: *default


config/database.yml
```diff
@@ -13,20 +13,20 @@ default: &default
...
   username: <%= ENV['OSEM_DB_USER'] || 'postgres' %>
   password: <%= ENV['OSEM_DB_PASSWORD'] || 'mysecretpassword' %>
-  database: <%= ENV['OSEM_DB_NAME'] || 'postgres' %>
   pool: 5
   timeout: 5000
 
 development:
   <<: *default
-  database: osem_development
+  database: <%= ENV['OSEM_DB_ADAPTER'] == 'sqlite3' ? "db/development.sqlite3" : "osem_development" %>
```
hennevogel: Hm can you do this above where we set the encoding too? Something like

```
encoding = 'unicode'
db_prefix = ''
if ENV['OSEM_DB_ADAPTER'] == 'mysql2'
     encoding = 'utf8'
elsif  ENV['OSEM_DB_ADAPTER'] == 'sqlite3'
   db_prefix = 'db/'
end
...
development:
  <<: *default
  database: <%= db_prefix + 'osem_development' %>
```

config/database.yml
```diff
@@ -13,20 +13,20 @@ default: &default
...
 # re-generated when you run "rake".
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: osem_test
+  database: <%= ENV['OSEM_DB_ADAPTER'] == 'sqlite3' ? "db/test.sqlite3" : "osem_test" %>
 
 production:
   <<: *default
+  database: <%= ENV['OSEM_DB_NAME'] || 'osem_production' %>
```
hennevogel: This would break docker...

.gitignore
```diff
@@ -38,3 +38,5 @@ docker-compose.override.yml
 .DS_Store
 .byebug_history
 .buildconfig
+osem_development
+osem_test
```
hennevogel: If we have the prefix then this is not needed right?

bear454: Does it hurt to exclude them?

config/database.yml
bear454: Can you elaborate on that?

config/database.yml
bear454: We need the suffix too... although there's no hard requirement, it's conventional to have `.sqlite3` files.

bear454:
Never mind; I resolved this locally with `.env.test`:

```
OSEM_DB_ADAPTER=sqlite3
OSEM_DB_NAME="db/test.sqlite3"
```

PR #2088, opened 2018-07-13
AJAX Datatable ✨

bear454:
**Checklist**

- [x] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
- [x] My branch is up-to-date with the upstream `master` branch.
- [x] The tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works(if appropriate).

**Short description of what this resolves/which [issues](https://github.com/openSUSE/osem/issues) does this fix?:**

- Every page in _osem_ that uses databales has the same problem: it _actually_ loads the **entire** data set, and then shows a paginated view locally. Whether it's 7 Tickets, or 1400+ Users, it loads _all the data_.

**Changes proposed in this pull request:**

- Set up a framework for using datatables built-in AJAX functionality to load _only the data that will be shown_.
- Add a single reference implementation on one of the worst offenders: the administrative User's list.

TODO: Reimplement the rest of the datatables, once the design/framework is accepted.

PR #2189, opened 2018-09-22
Ajax datatable admin registrations

bear454:

- Registrations is a beastly monster, loading thousands of records in hundreds of queries.

- Stop the madness, using AJAX.
- Follows on #2088

### DIFF
--- a/app/views/admin/registrations/index.html.haml
+++ b/app/views/admin/registrations/index.html.haml
@@ -8,9 +8,9 @@
         = "(#{@registrations.length})" if @registrations
         .btn-group.pull-right
           - if can? :read, Registration
-            = link_to 'Export CSV', admin_conference_registrations_path(@conference.short_title, format: :csv), class: 'btn btn-success'
-            = link_to 'Export PDF', admin_conference_registrations_path(@conference.short_title, format: :pdf), class: 'btn btn-success'
-            = link_to 'Export XLS', {format: :xlsx}, class: 'btn btn-success'
+            = link_to 'Export CSV', { format: :csv }, class: 'btn btn-success'
+            = link_to 'Export PDF', { format: :pdf }, class: 'btn btn-success'
+            = link_to 'Export XLS', { format: :xlsx }, class: 'btn btn-success'
       %p.text-muted
         All the people who registered to your event
   .col-md-4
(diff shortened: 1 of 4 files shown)


bear454:
@hennevogel - review this already :P

app/views/admin/registrations/index.html.haml
```diff
@@ -21,70 +21,96 @@
...
-      = render partial: 'questions', locals: { registration: registration }
 
-.modal.fade{ id: 'questions', 'role' => 'dialog', 'aria-hidden' => 'true' }
-  .modal-dialog
-    .modal-content
-      .modal-header
-        %h3{id: 'question-modal-header'}
-          Questions for Foobar
-      .modal-body#question-modal-body
+:javascript
```
hennevogel: I really think you should put this into a .js file in the asset pipeline...

bear454: Yeah, I go back and forth on that... and here's what I've come up with:

* If it's used on one page, it belongs in that page.
* If it's used more than once, it should be in the asset pipeline.

hennevogel: Yeah the only overhead you would generate people downloading those ~80 lines once :-D But okay

app/views/admin/registrations/index.html.haml
```diff
@@ -21,70 +21,96 @@
...
-                  = registration.departure.strftime('%d %b %H:%M')
-                - else
-                  n/a
-              -if @conference.questions.any?
-                %td
-                  = link_to 'Questions','#', class: 'btn btn-success question-btn', 'data-id' => index, 'data-name' => registration.name
-              %td{ 'data-order' => registration.attended.to_s }
-                = check_box_tag "#{@conference.short_title}_#{registration.id}", registration.id, registration.attended,
-                  class: 'switch-checkbox',
-                  url: toggle_attendance_admin_conference_registration_path(@conference.short_title, id: registration.id)+"?attended="
```
differentreality: @bear454 you have completed removed attendance from the registrations. Why?

bear454: Because it's useless. It's not an effective sort, or filter, you can't get an overall count of it, and it's only set if you use one of the check-in methods (which are _terrible_ upstream).

differentreality: I understand the concerns, however I don't agree with the removal. I am pretty sure there are some conferences using this feature. 

I didn't expect a PR that aims to improve performance, to actually remove a feature we had :/ 
If this was discussed, I've clearly missed the conversation.

PR #2194, opened 2018-09-26
Start making some helpers for redundant visual elements (widgets)

bear454:


hennevogel:
> just from a readablity perspective, big_statistic(... says a lot more to me than render :partial...

Au contraire! Mon ami. I know what `render :partial` does. I have **no** idea what `big_statistic` does unless I open the helper code and study it...

PR #2199, opened 2018-09-27
Conservative update of haml-rails 0.5.3 -> 0.8.2

bear454:
Preceeds #2155


bear454:
Moving to a newer version of haml-rails will require moving to haml 5, and there seem to be some issues with haml-lint there, but we can go this far without major version changes of dependencies.

PR #2200, opened 2018-09-29
Update gem active_model_serializers

bear454:
Supercedes #2136

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -35,8 +35,11 @@ GEM
       erubis (~> 2.7.0)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.0, >= 1.0.3)
-    active_model_serializers (0.9.4)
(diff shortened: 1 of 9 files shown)


bear454:
@hennevogel : can you help debug the test failures?

_Every_ failure reported starts with a MySQL error, either:

```
Mysql2::Error: This connection is in use by: #<Thread:0x0000000001aecd28 sleep>: SELECT  `users`.* FROM `users` WHERE `users`.`id` = 2 ORDER BY `users`.`id` ASC LIMIT 1
```

... or ...

```
ActiveRecord::StatementInvalid:
              Mysql2::Error: MySQL client is not connected: SELECT @@FOREIGN_KEY_CHECKS
```

I'm seeing _no errors_ locally on sqlite.

hennevogel:
@hennevogel : can you help debug the test failures? 

Not really. Same errors we saw with the other test parallelization :-/

PR #2203, opened 2018-10-08
Add a registration link to user menu

bear454:
screenshot from 2018-10-08 07-48-50


bear454:
To make it easier to get back to the registration page, which also serves as a 'homepage' for ticketing and surveys. For conferences that run with `OSEM_ROOT_CONFERENCE` set, the links on `/conferences` aren't obvious.

app/views/layouts/_user_menu.html.haml
```diff
@@ -19,6 +19,10 @@
   = link_to(edit_user_path(current_user.id)) do
     %span.fa.fa-user
     Edit Profile
+- if @conference && ((current_user && @conference.user_registered?(current_user)) || @conference.registration_open?)
+  %li
+    = link_to(conference_conference_registration_path(@conference)) do
+      = fa_icon('id-badge', text: "#{@conference.short_title} registration")
```
differentreality: I'd rather we have
```
%span.fa.fa-id-badge
      = @conference.short_title
      registration

```
like in other places

bear454: Updated in 9b2734d8.

PR #2204, opened 2018-10-08
Fix tickets

bear454:
Resolve ~two~three small issues on ticketing:

* Prevent math errors in calculating ticket 'turnover' _(yes this can still be a problem)_
* Gracefully fail out of image handling in ticket PDF ( I see this issue, in particular, when using a database copy to test, that doesn't have access to the same Cloudinary account - e.g. the same asset host).
* ticket_purchases_total no longer requires a redunant ID. `Ticket.find(X).ticket_puchases_total(X)` ?!

### DIFF
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -63,11 +63,10 @@ class Ticket < ApplicationRecord
     tickets.inject(0){ |sum, ticket| sum + (ticket.amount_paid * ticket.quantity) }
   end
 
-  def tickets_turnover_total(id)
-    ticket = Ticket.find(id)
-    return Money.new(0, 'USD') unless ticket
-    sum = ticket.ticket_purchases.paid.total
-    Money.new(sum, ticket.price_currency)
+  def tickets_turnover_total
+    purchases = ticket_purchases.paid
+    total = purchases.sum { |purchase| (purchase.price_cents * purchase.quantity) }
+    Money.new(total, price_currency)
   end
 
   def tickets_sold
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -517,7 +517,7 @@ class Conference < ApplicationRecord
     if tickets && ticket_purchases
       tickets.each do |ticket|
         result[ticket.title] = {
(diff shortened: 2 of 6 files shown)


app/models/ticket.rb
```diff
@@ -64,10 +64,13 @@ def self.total_price_user(conference, user, paid: false)
   end
 
   def tickets_turnover_total(id)
-    ticket = Ticket.find(id)
-    return Money.new(0, 'USD') unless ticket
-    sum = ticket.ticket_purchases.paid.total
-    Money.new(sum, ticket.price_currency)
+    tickets = TicketPurchase.where(ticket_id: id).paid
```
Ana06: what is the difference between this code and the one before? :thinking:

app/models/ticket.rb
```diff
@@ -64,10 +64,13 @@ def self.total_price_user(conference, user, paid: false)
...
+    if tickets.blank?
+      Money.new(0, Ticket.find(id).price_currency)
+    else
+      total = tickets.inject(0) { |sum, ticket| sum + (ticket.price_cents * ticket.quantity) }
```
Ana06: `sum` instead of `inject` :see_no_evil:

app/models/ticket.rb
```diff
@@ -64,10 +64,13 @@ def self.total_price_user(conference, user, paid: false)
   end
 
   def tickets_turnover_total(id)
-    ticket = Ticket.find(id)
-    return Money.new(0, 'USD') unless ticket
-    sum = ticket.ticket_purchases.paid.total
-    Money.new(sum, ticket.price_currency)
+    tickets = TicketPurchase.where(ticket_id: id).paid
+    if tickets.blank?
```
Ana06:
```ruby
return Money.new(0, Ticket.find(id).price_currency) if tickets.blank?
```
instead of `if-else`

bear454:
@Ana06 can you please review again?

bear454:
Fair enough, @Ana06 ; I'll analyze the data, and see if I can find some other underlying problem. In the meantime, I split out the PDF fix  into PR#2221.

PR #2206, opened 2018-10-09
Add splashpage option for shuffling highlight tiles

bear454:
Highlights show up, by default in order of ID, and this may imply somepreference of one highlight over another (especially if they render in multiple rows). As a solution, each user is presented with a randomly shuffled set of highlights. Since this is done in JS, the cache of highlighted sessions is preserved, maintaining server-side performance.

### DIFF
--- a/app/views/conferences/_highlights.haml
+++ b/app/views/conferences/_highlights.haml
@@ -4,7 +4,7 @@
       Program Highlights
 .row
   .col-md-12
-    .row.row-centered
+    .row.row-centered.shuffle-deck
       - highlights.each do |event|
         - speaker = event.speakers_ordered.first
         .col-md-3.col-sm-3.col-centered.col-top.highlights
--- a/app/assets/javascripts/osem.js
+++ b/app/assets/javascripts/osem.js
@@ -31,6 +31,19 @@ $(function () {
         });
     });
 
+    /**
+     * Randomize order of parallel elements by shuffling the decks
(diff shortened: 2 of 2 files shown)


hennevogel:
Hm, why make this an option?

bear454:
Didn't want to impose my opinions on everyone, and didn't want to necessarily change legacy data/views.

hennevogel:
> Didn't want to impose my opinions on everyone

But you should! It's a good one :-)

bear454:
Alright; if you insist.

app/views/conferences/_highlights.haml
```diff
@@ -4,7 +4,7 @@
       Program Highlights
 .row
   .col-md-12
-    .row.row-centered
+    .row.row-centered.shuffle-deck
       - highlights.each do |event|
```
hennevogel: Thinking about it. Would it be the same as 

```ruby
@highlights = @conference.highlighted_events.eager_load(:speakers).shuffle
```
in the controller?

Sorry I did not catch this earlier.

bear454: It's not the same! The highlights are part of the cache key for the partial caching, and changing the order changes the key... it would invalidate the cache on every view!

> Since this is done in JS, the cache of highlighted sessions is preserved, maintaining server-side performance.

hennevogel: aight

PR #2219, opened 2018-10-19
Event reservations switch

bear454:
So... for LFNW we don't want people submitting proposals to be able to set event registration. It just doesn't work for _our_ event... so I added a switch.

In backporting this to upstream OSEM, I couldn't help fixing a few other nitpicks... like having the "Call for Booths" under the heading "Call for Papers". *sigh*.


bear454:
screenshot from 2018-10-19 11-39-05

screenshot from 2018-10-19 13-10-23

PR #2221, opened 2018-10-22
Gracefully fail out of image handling in ticket PDF

bear454:
(cherry picked from commit 730d53c8b642e3d04434301b4c9d4b38ef775d89)

### DIFF
--- a/app/pdfs/ticket_pdf.rb
+++ b/app/pdfs/ticket_pdf.rb
@@ -49,7 +49,11 @@ class TicketPdf < Prawn::Document
     if @conference.picture?
       conference_image = case @conference.picture.ticket.url[0, 4]
                          when 'http', 'ftp:' # CDNs
-                           open(@conference.picture.ticket.url)
+                           begin
+                             open(@conference.picture.ticket.url)
+                           rescue OpenURI::HTTPError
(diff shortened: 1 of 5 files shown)


app/pdfs/ticket_pdf.rb
```diff
@@ -49,18 +49,22 @@ def draw_second_square
     if @conference.picture?
       conference_image = case @conference.picture.ticket.url[0, 4]
                          when 'http', 'ftp:' # CDNs
-                           open(@conference.picture.ticket.url)
+                           begin
+                             open(@conference.picture.ticket.url)
+                           rescue OpenURI::HTTPError
+                             nil
+                           end
```
Ana06: Why do you explicitly need to return `nil`?

bear454: Well, the `nil` return isn't explicitly necessary, but the rescue block has to have _some content_, and doing an explicit return is more elegant than, say `# awkward blank rescue block`( https://stackoverflow.com/questions/3672445/having-a-ruby-block-command-silently-fail-without-a-blank-rescue-block )

bear454:
It's been 2 months... merging.

PR #2223, opened 2018-10-24
Provide an opportunity for including conference-specific CSS ￼

bear454:
There is currently no top-level selector for views that is conference-specific, so all conferences on one OSEM instance end up sharing all the same CSS.

In order to allow individual conferences to have unique CSS, a class for the current conference is applied to the <body> tag, and an scss template provided with an example of conference-specific CSS.

_Bonus: unlinting, to make the project a better place._


Ana06:
@bear454 I am trying to find out why 82b79bb was added but in the commit description there is no info, do you remember why you moved this out? (that's  why commit descriptions are so important....)

bear454:
82b79bb has a commit description: `Unlint Rspec/DescibeClass #BetterPlace`. The description is a brief summary both of _what_ was changed - the linter TODOs for `Rspec/DescibeClass` were removed, and the described lints corrected. The summary also explains _why_  - to leave the world a better place than I found it.

Ana06:
@bear454 thanks for your explanation, but I still don't understand why this was needed. The todo file (`.rubocop_todo.yml`) is generated automatically and it includes the offenses that we want to fix but we didn't fix yet. This exists done because when you add Rubocop to an already existing project normally there are too many offenses and you don't want to have to fix everything at one, but at least to not allow to introduce new offenses. On the other hand, the `.rubocop.yml` file is where you add the offenses you don't plan to fix because of any reason. So moving this imply we don't want to fix those offenses, not following the default Ruby style guide. My question is then, why don't we want to fix the offenses?

bear454:
IIRC, this was in conjunction with a Rubocop update which changed the rule definition. I fixed every instance of that rule that is fixable (  'spec/models/booth_spec.rb', 'spec/models/registration_spec.rb' ) and removed one entry that had already been fixed ( 'spec/models/admin_ability_spec.rb' ). The view specs, and the ability spec do not conform to Rubocop's naming expections, and therefore cannot be fixed. For example, 'https://github.com/openSUSE/osem/blob/master/spec/views/layouts/application.haml_spec.rb#L3' cannot be described using a class instead of a string, as there is no class representation of the view.

I wish you had asked these questions when the PR was fresh, instead of waiting 5 months.

PR #2225, opened 2018-10-26
Goodbye PhantomJS

### LINKED ISSUE #2150: Switch to headless chromium
We currently use the phantomjs headless browser in capybara through poltergeist. Development of phantomjs has stoppped so we should replace it with something maintained. People move to headless chromium or firefox. We should too...

bear454:
* Drop phantomjs; use selenium to run Chrome or Firefox, visible or headless. (Headless Chrome by default - including features testing in Travis.) Resolves #2150 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
   - "echo `phantomjs -v`"
 addons:
   mariadb: '10.2'
+  chrome: stable
 notifications:
   email:
     on_success: change
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ before_install:
   - "echo 'gem: --no-ri --no-rdoc' > ~/.gemrc"
   - "echo `phantomjs -v`"
 addons:
-  mariadb: '10.2'
   chrome: stable
 notifications:
   email:
@@ -23,15 +22,12 @@ notifications:
     on_success: change
     on_failure: change
 before_script:
-  - mysql -u root -e 'create database osem_test;'
   - RAILS_ENV=test bundle exec rake db:bootstrap --trace
 script:
   - "./travis_script.sh $TEST_SUITE"
 env:
   global:
-    - OSEM_DB_ADAPTER=mysql2
-    - OSEM_DB_USER=root
-    - OSEM_DB_PASSWORD=''
+    - OSEM_DB_ADAPTER=sqlite3
     - OSEM_DB_HOST=''
     - OSEM_DB_PORT=''
     - OSEM_DB_NAME='osem_test'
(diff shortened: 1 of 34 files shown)


bear454:
@hennevogel please review...

dotenv.example
```diff
@@ -115,3 +115,12 @@
 # Speeds up turn around times of tests
 # OSEM_FACTORY_LINT=false
 
+# How should browser tests be performed?
```
hennevogel: Why does this need to be configurable?

bear454: _I_ want to test on Chrome _and_ Firefox, but I don't see the need to run the whole suite on both browsers all the time, so, it's configurable. In this way, I can either

```
rspec spec/features
````

to just run the suite... or 

```
OSEM_TEST_DRIVER=firefox rspec spec/features
```

to, for example, see it in Firefox. Makes me happy.

Comment (hennevogel) [reply dotenv.example]: >  Makes me happy.

If that's the case, ship it! :heart_eyes:

**Comment by bear454** [comment]
@hennevogel : so the database issue is mysql only... are you okay with testing on sqlite instead?

File: .travis.yml [hunk 2]
```diff
@@ -23,17 +22,12 @@ notifications:
...
 script:
   - "./travis_script.sh $TEST_SUITE"
 env:
   global:
-    - OSEM_DB_ADAPTER=mysql2
-    - OSEM_DB_USER=root
-    - OSEM_DB_PASSWORD=''
-    - OSEM_DB_HOST=''
-    - OSEM_DB_PORT=''
+    - OSEM_DB_ADAPTER=sqlite3
```
hennevogel: Phew this is very far away from any production instance I know. Might be a temporary fix, but not a solution.

bear454: Well, that's supposed to be the point of ActiveRecord, that the database behind the ORM doesn't matter. If we're worried about full-stack production testing, we'd need to run parallel on MySQL & Postgres. IMO the only thing interim about this as a test db, is that we're not running in-memory. Yet.

bear454: Not to be argumentative, but just to be clear: the `osem` code didn't change, but now the tests pass. This is a pretty clear indicator that the interaction of capybara+database_cleaner+mysql is the culprit for our flaky tests, not the test suite itself.

hennevogel: > capybara+database_cleaner+mysql is the culprit for our flaky tests,

Nobody debates this.

hennevogel: Define `good` :wink: I fear we are going to stay with this a long time. Because it's so convenient...

bear454:
I'm going with sqlite for now. I'll file an issue for moving travis to a "production-grade" database, when the test suite can be run reliably.

hennevogel:
:+1:

PR #2276, opened 2018-11-28
Fix up a flappy test

bear454:
The link `#description_link` reveals a field, and the time for revealing sometimes caused the `#event_description` field not be found.


Ana06:
@bear454 I know that we haven't been really fast with the PR reviews in the last time, but I think we shouldn't merge PR without a review. If something is important asking for a review directly to people may help. :thinking:

PR #2297, opened 2018-12-28
Use chartkick globally

bear454:
Charting has been flaky, but with the introduction of chartkick it doesn't need to be so convoluted.

* Clean up legacy charting imports
* Update existing donut and line charts
* Create helpers for parsing out existing chart data
* Move chart partials to a more common path


bear454:
screenshot from 2018-12-27 19-48-52

screenshot from 2018-12-27 19-50-55

differentreality:
I fully support the change :+1: 
With a few more tickets available things don't look that well though. 
What about we put the legend on the bottom? (it will make it easier to click the rectangular below the download icon)

osem
screenshot from 2018-12-28 13-28-36
screenshot from 2018-12-28 13-31-00

bear454:
Thanks for the reviews @differentreality !

PR #2299, opened 2018-12-28
Move splashpage logdgings descriptions to a modal

bear454:
screenshot from 2018-12-27 21-03-53


differentreality:
Why do we want to change this?

bear454:
Right now, the entire lodging description is included in the splashpage, while sponsor descriptions are not; some sponsors have suggested this makes sponsorship feel 'less important' to the event than lodgings, as they get less 'real estate' on the splashpage. I agree, and using a common pattern for both provided a nice balance.

bear454:
I'm going to reimplement this as a splashpage switch instead.

PR #2301, opened 2018-12-28
Add markdown formatting for sponsor descriptions

bear454:


differentreality:
Perhaps we should amend the admin/sponsors#index view as well to use the markdown format, but I am fine with merging just this as well.

PR #2306, opened 2018-12-29
Admin ticketing

bear454:
Allow admins to 'give' tickets to users, and create a class of tickets that can only be given by admins. This is a great way to handle providing material to sponsor staff.

screenshot from 2018-12-28 17-16-44

### DIFF
--- a/app/controllers/admin/tickets_controller.rb
+++ b/app/controllers/admin/tickets_controller.rb
@@ -52,7 +52,12 @@ module Admin
     private
 
     def ticket_params
-      params.require(:ticket).permit(:conference, :title, :url, :description, :conference_id, :price_cents, :price_currency, :price, :registration_ticket)
+      params.require(:ticket).permit(
+        :conference, :conference_id,
+        :title, :url, :description,
+        :price_cents, :price_currency, :price,
+        :registration_ticket, :visible
+      )
     end
   end
 end
--- a/app/controllers/admin/tickets_controller.rb
+++ b/app/controllers/admin/tickets_controller.rb
@@ -38,13 +38,30 @@ module Admin
       end
     end
 
+    def give
+      ticket_purchase = @ticket.ticket_purchases.new(gift_ticket_params)
+      recipient = ticket_purchase.user
+      if ticket_purchase.save
+        redirect_to(
+          admin_conference_ticket_path(@conference.short_title, @ticket),
+          notice: "#{recipient.name} was given a #{@ticket.title} ticket."
+        )
+      else
+        redirect_back(
+          fallback_location: admin_conference_ticket_path(@conference.short_title, @ticket),
+          error:             "Unable to give #{recipient.name} a #{@ticket.title} ticket: " +
+                             ticket_purchase.errors.full_messages.to_sentence
+        )
+      end
+    end
+
     def destroy
       if @ticket.destroy
         redirect_to admin_conference_tickets_path(conference_id: @conference.short_title),
-                    notice: 'Ticket successfully destroyed.'
+                    notice: 'Ticket successfully deleted.'
       else
         redirect_to admin_conference_tickets_path(conference_id: @conference.short_title),
-                    error: 'Ticket was successfully destroyed.' \
+                    error: 'Deleting ticket failed! ' \
                     "#{@ticket.errors.full_messages.join('. ')}."
       end
     end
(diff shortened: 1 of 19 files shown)


Ana06:
@bear454 can you please provide more details about this? :pray:

bear454:
I don't know how to explain it much better, other than to rearrange the words a bit.

As an administrator, I should be able to give a ticket to a user.
As an administrator, I should be able to create tickets that general users cannot view or select.

hennevogel:
>  As an administrator, I should be able to create tickets that general users cannot view or select.

Why? Why can't you just give them "normal" tickets?

bear454:
At LFNW, we give tickets to sponsored staff, booth staff, and speakers, which allows for provide them different material at checkin; we also use the tickets in our badge design.

I'm sure you can understand that you wouldn't want the general public to be able to choose "speaker gift" ticket, for example.

hennevogel:
So the ticket is different to distinguish people at checkin, okay. Do you use 'free' tickets too? I would rather do this with `User` roles so once you give people a staff role, their `TicketPurchase` takes this into account. Same for speakers probably.

bear454:
> Do you use 'free' tickets too?

No, we don't use 'free tickets' or 'registration tickets'.

bear454:
> I would rather do this with `User` roles so once you give people a staff role, their `TicketPurchase` takes this into account. Same for speakers probably.

I could see creating a dynamic role for speakers, but I don't have any way to tie users to sponsored staff; we don't use the 'booth' feature.

bear454:
* Is there an objection to having an admin interface for 'giving' a ticket? We currently have admin interfaces for creating users, sponsors, registrations, but not tickets.
* Is there an objection to having tickets that aren't public? We have a non-public state for schedules, registration, all the splashpage components.

sbernhard:
I really like the way this feature works. Sponsers, companies having a booth often gets free tickets.

cycomachead:
I second this feature. This includes a solution for #2567. Hiding and showing tickets in general is useful, especially since we'll have early registrations and some temporary tickets. 

The conference we are running this summer will likely also gift tickets as a result of sponsorships or for a few community members. The fact that this workflow is explicit is a bonus, to me. The clarity around tickets that are normally X being counted as 0 is helpful, too, from an audit and budget perspective.

spec/controllers/admin/tickets_controller_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+  let(:admin) { create(:admin) }
+  let(:conference) { create(:conference) }
+  let!(:ticket) { create(:ticket, conference: conference) }
+  let(:new_title) { Faker::Hipster.sentence }
+
+  context 'admin is signed in' do
+    before { sign_in admin }
+
+    describe 'GET #index' do
+      before { get :index, conference_id: conference }
```
cycomachead: This now needs to be updated, but that's pretty quick.

https://github.com/snap-cloud/snapcon/commit/231b2a6d7215e9b330572cc1ef3fa2746a217c2d

Merged this in my fork and it's working well!

app/controllers/admin/tickets_controller.rb
```diff
@@ -38,21 +38,50 @@ def update
       end
     end
 
+    def give
+      ticket_purchase = @ticket.ticket_purchases.new(gift_ticket_params)
```
cycomachead, line 42: After using this for a conference, the workflow works pretty well! But, this leads to a somewhat weird scenario when a user has an _unpaid_ ticket. Sometimes we want to give a sponsored ticket to someone who started to go through the purchase process and when this happens it requires console access to first delete their ticket. I wonder if it makes sense to have a "replace" option, or if an unpaid ticket exists to just update it to paid with a value of $0.

cycomachead:
I'm now a couple days out from running the 4th event where we've successfully used this. One thing I just realized is that the #give method in this PR doesn't trigger the follow to create a physical ticket, and send a confirmation email. Both those seemed like useful things as we had a couple people that were confused. 

Simple fix: https://github.com/snap-cloud/snapcon/commit/9ae89b56cc16b84f4d201c736784aa2e3f2ebf4b (ignore accidentally removing the `.visible` filter in ticket_purchases)

I will say, aside from the few people who attempt to buy a ticket then we need to give them one, this has worked quite well for my uses.  Here's all of them:

* I give all the conference admins registration tickets so they don't need to worry. (This is because in my fork I just check for registered? to a conference, though I support I could do registered? or admin role...) Still, in other events I've worked with the team still have tickets.
* I give volunteers tickets (same deal as above)
* We have people pay on behalf of others. So someone will be 10 non-reg tickets and give us the list of emails of people to buy tickets for.
* This makes it easy to give scholarship tickets. These are people who are often just attending, but requested a reduced ticket price.
* Hiding/Showing tickets is useful for "early bird" pricing and admin-only tickets.

I all these cases, I'm using registration tickets, since the common case is for someone to buy a reg ticket. It's quite nice that I and a couple others on our team can just drop in an email address to make sure someone is ready go.

PR #2307, opened 2018-12-29
Enable registration to proceed without requiring a ticket.

bear454:
Requiring a ticket after registration is convoluted; while it may be necessary for some workflows, it definitely isn't for others, and the core of _osem_ doesn't need it... so let's make it optional.

### DIFF
--- a/app/views/admin/registration_periods/show.html.haml
+++ b/app/views/admin/registration_periods/show.html.haml
@@ -8,29 +8,38 @@
   .row
     .col-md-8
       %dl.dl-horizontal
-        %dt
-          Start Date
-        %dd
-          = @registration_period.start_date
-        %dt
-          End Date
-        %dd
-          = @registration_period.end_date
+        %dt Start Date
+        %dd= @registration_period.start_date
+        %dt End Date
+        %dd= @registration_period.end_date
+        %dt Ticket required?
+        %dd= @conference.tickets.for_registration.any? ? 'Yes' : 'No'
+
+.row
+  .col-md-12
+    %p
+      If you would like to require a ticket purchase for registration, at least
+      one
+      = link_to 'ticket', admin_conference_tickets_path(@conference)
+      must be set as a registration ticket, and the registration process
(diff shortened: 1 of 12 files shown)


app/views/admin/registration_periods/show.html.haml
```diff
@@ -8,29 +8,38 @@
...
+        %dd= @conference.tickets.for_registration.any? ? 'Yes' : 'No'
+
+.row
+  .col-md-12
+    %p
+      If you would like to require a ticket purchase for registration, at least
+      one
+      = link_to 'ticket', admin_conference_tickets_path(@conference)
+      must be set as a registration ticket, and then the registration process
+      will not process without the purchase (or selection of a free) single
```
differentreality: What will the process not process? 
I assume we mean that a ticket must be set as a registration ticket, so that the registration process proceeds with the purchase of a registration ticket, or the selection of a free one?

app/views/conference_registrations/show.html.haml
```diff
@@ -97,33 +97,44 @@
   - if @conference.tickets.any?
     .row
       .col-md-12
+        %h4
+          %span.fa-stack
+            %i.fa.fa-square-o.fa-stack-2x
+            %i.fa.fa-ticket.fa-stack-1x
+          Tickets
         -if @tickets.any?
```
differentreality: Since we are doing all the rest, please add a space between the dash and the if

app/views/admin/registration_periods/show.html.haml
bear454: Have you ever read something so many times that you can't actually read it anymore? :/

bear454: I reworded in bedb6fad . Thanks for catching that... I would have read past it 100 times ;-)

app/views/conference_registrations/show.html.haml
bear454: Fixed in bedb6fad .

PR #2318, opened 2019-01-29
Simplify app.json

hennevogel:
Use system libraries for building nokogiri and do not require to configure so many environment variables.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,67 +1,21 @@
 {
   "name": "osem",
   "description": "Open Source Event Manager",
+  "repository": "https://github.com/openSUSE/osem",
   "env": {
     "CLOUDINARY_URL": {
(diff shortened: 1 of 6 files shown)


app.json
```diff
@@ -1,67 +1,21 @@
...
-    "RAILS_ENV": {
-      "required": false
-    }
+    "NOKOGIRI_USE_SYSTEM_LIBRARIES": "1",
+    "RAILS_ENV": "production",
+    "RACK_ENV": "production"
   },
   "scripts": {
-    "postdeploy": "bundle exec rake db:reset data:demo"
+    "postdeploy": "bundle exec rake db:schema:load db:seed"
```
differentreality: I am sad we removed the demo data :cry: It was so helpful having a few things setup for the review apps

PR #2346, opened 2019-02-20
Update binding_of_caller

Ana06:
There is a bug for the old version which affectes some Ruby versions.
Generate Gemfile with our current Ruby version 2.5.0.


Ana06:
how did the coverage change? :unamused:

Issue #2349, opened 2019-02-20
Calls dates are broken


Ana06:
In the Call for Events/Booths/Tracks, start and end dates are broken, which makes that creating then is not anymore possible.

![image](https://user-images.githubusercontent.com/16052290/53104561-dd0f1900-352f-11e9-8416-d496dd03293e.png)

Ana06:
Those are not the only datapickers which are broken

PR #2354, opened 2019-02-21
Fix Datepickers in Call for Content

Ana06:
Datapickers in Call for Events/Booths/Tracks were broken which made impossible to create them (as the date is required).

Those are not the only Datepickers broken and a similar solution should work for the rest as well.

Fixes https://github.com/openSUSE/osem/issues/2349

image


Ana06:
I haven't managed to change the date in the datepickers by using Javascript. See issue at:

https://github.com/TrevorS/bootstrap3-datetimepicker-rails/issues/66

I also tried to use capybara-bootstrap-datepicker gem, but also didn't work. See my branch [datepickers-gem](https://github.com/openSUSE/osem/compare/master...Ana06:datepickers-gem?expand=1) and the issue https://github.com/akarzim/capybara-bootstrap-datepicker/issues/14

Because of that I have disabled the test. I even doubt if we need a feature test for this. The functionality was completely wrong and the test didn't fail anyway.

Ana06:
I haven't managed to change the date in the datepickers by using Javascript. See issue at:

https://github.com/TrevorS/bootstrap3-datetimepicker-rails/issues/66

I also tried to use capybara-bootstrap-datepicker gem, but also didn't work. See my branch [datepickers-gem](https://github.com/openSUSE/osem/compare/master...Ana06:datepickers-gem?expand=1) and the issue https://github.com/akarzim/capybara-bootstrap-datepicker/issues/14

Anyway, it doesn't make sense to test this with a feature test as it is just testing that the controller actions work properly. Also, he functionality was completely wrong and the test didn't fail. So, I've replaced those feature test by controller tests.

ehrenfeu:
Hi @Ana06,

for me disabling (removing) the `pickTime` from `osem-datepicker.js` worked well, i.e. something like this:

```diff
diff --git a/app/assets/javascripts/osem-datepickers.js b/app/assets/javascripts/osem-datepickers.js
index 598204ae..a682f3dc 100644
--- a/app/assets/javascripts/osem-datepickers.js
+++ b/app/assets/javascripts/osem-datepickers.js
@@ -53,7 +53,6 @@ $(function () {
    });
 
    $("#conference-start-datepicker").datetimepicker({
-       pickTime: false,
        useCurrent: false,
        format: "YYYY-MM-DD",
        // conference-start-day >= Current_date
```

Unfortunately I don't have the time right now to finish it up and do a PR, but maybe this still helps...?

Cheers,
Niko

Ana06:
@ehrenfeu

> for me disabling (removing) the pickTime from osem-datepicker.js worked well, i.e. something like this:

 this is exactly the change I am introducing and there is a PR to do this in more places https://github.com/openSUSE/osem/pull/2391

The only "problem" with this PR was adapting the tests after the change.

Issue #1213, opened 2016-10-10
Replace vagrant with docker 


alexdor:
Docker is overall lighter than vagrant. Its benefits include, but are not limited to :
- Better resource management 
- Smaller images 
- Fewer and lighter dependencies 
- Less bandwidth required per image download

differentreality:
@alexdor @lagartoflojo can help you with docker

lagartoflojo:
Hi @alexdor, have you had a look at https://github.com/openSUSE/osem/pull/933? Is something like that what you're proposing?

lagartoflojo:
While that PR was closed, we could re-open the discussion about either keeping Vagrant or moving development to Docker/docker-compose.

alexdor:
Hi @lagartoflojo, my proposal is similar but not the same. For example, I would move the DB to separate container cause it's closer to docker's philosophy that way. Yeah we should reopen the discussion, should we do it before creating a pull request?

lagartoflojo:
I agree with respect to the DB. Ideally, we would run mysql/mariadb even in dev machines, to be as close to a production env as possible.

But we should definitely agree on the direction we want to take before writing code. There's already a prototype in #933, and I think it's enough to have a fruitful discussion.

Personally, I prefer the Docker / docker-compose solution. I tried using Vagrant locally but it did not work for me, so I also used Docker for osem in the end.

Docker can be used as a [provider for Vagrant](https://www.vagrantup.com/docs/docker/), but I don't think we need it. Using is just more complexity for very little gain. It would mean maintaining both a Vagrantfile and a Dockerfile, as well as requiring contributors to have both tools installed.

Right now, Docker is very easy to install, even on [Windows](https://docs.docker.com/docker-for-windows/) and on [macOS](https://docs.docker.com/docker-for-mac/), so it should be quite easy for any future contributor to get started. Also, building osem on Docker is quite easy: `docker-compose build`. All dependencies are automatically downloaded, no need for additional plug-ins like in Vagrant.

@hennevogel @differentreality @ChrisBr @kalabiyau thoughts?

hennevogel:
Introducing docker as vagrant backend is pretty easy. I have a branch in the [hackweek tool](https://github.com/SUSE/hackweek/compare/vagrant-docker-backend?expand=1) that does this. But anyway, using docker for development brings two major PITAs:

[...]

PR #2355, opened 2019-02-21
Get rid of vagrant

Ana06:
We are using Docker now instead.

Closes https://github.com/openSUSE/osem/issues/1213


bear454:
I haven't been involved in the decision making here, so I'll defer to @hennevogel .

hennevogel:
@bear454 involve yourself in #1213 :wink:

config/routes.rb
```diff
@@ -14,11 +14,6 @@
                path:        'accounts'
   end
 
-  # Use letter_opener_web to open mails in browser (e.g. necessary for Vagrant)
-  if Rails.env.development?
-    mount LetterOpenerWeb::Engine, at: "/letter_opener"
```
hennevogel: You can also drop the gem from the bundle then.

bear454:
> @bear454 involve yourself in #1213

We're happy with native Heroku at this point, so it's all moot for me. I'd say though, out of courtesy, that someone should check in with the other contributors to the vagrant code and make sure they're not still dependent on it.

Ana06:
@bear454 

> We're happy with native Heroku at this point, so it's all moot for me. I'd say though, out of courtesy, that someone should check in with the other contributors to the vagrant code and make sure they're not still dependent on it.

We can wait a bit before getting this merge, in case someone has a problem with it. But I think changing from vagrant to docker is not a big issue and having to maintain two development environments is extra work.

Issue #2330, opened 2019-02-19
The links to talks are broken


Ana06:
For example:

https://events.opensuse.org/conferences/oSC17/program/proposal/1388
https://events.opensuse.org/conferences/oSC18/program/proposal/1690

I am linking this pages and also openSUSE from blog post and other communication channels, so we should fix this fast.

Ana06:
This is also broken from the links in the app. This is a really visual bug

Ana06:
I am working on it.

Ana06:
The url was changed to https://events.opensuse.org/conferences/oSC18/program/proposals/1690 (proposals with s)

PR #2357, opened 2019-02-21
Add /conferences/:conference_id/program/proposal/:id

Ana06:
We changed this route (because of renaming proposal to proposals in):
https://github.com/openSUSE/osem/commit/daf1f805bd16fbf22fbe25f5224dd022b5e082cc

Keep the old route for compatibility. We may need to add more routes for the same reason.

Fixes https://github.com/openSUSE/osem/issues/2330

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -170,6 +170,7 @@ Osem::Application.routes.draw do
       end
     end
     resource :program, only: [] do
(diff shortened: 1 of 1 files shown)


Ana06:
@hennevogel we need to deploy this ASAP

PR #2362, opened 2019-02-22
Fix scheduling

### LINKED ISSUE #2353: It is not possible to create an schedule
According to @ddemaio, it is not possible to create an schedule. This is important for the conferences coming on.

hennevogel:
Can't cache the event as the data is changed the scheduling js.
We also can't verify the token in the EventSchedules controller
as we are making susequent ajax calls and have no means to refresh
the token for the second request.

Partly fixes #2353


hennevogel:
The scroll thing from jauery-ui draggable doesn't work. can't figure out why...

Issue #2363, opened 2019-02-22
Conference should be a link in the proposal page


Ana06:
![screenshot_2019-02-22-07-02-43-877_com android chrome](https://user-images.githubusercontent.com/16052290/53222992-18186600-3670-11e9-8764-7d015380244e.png)

PR #2366, opened 2019-02-22
Add link to conference in proposal page

Ana06:
Closes https://github.com/openSUSE/osem/issues/2363

### DIFF
--- a/app/views/proposals/show.html.haml
+++ b/app/views/proposals/show.html.haml
@@ -83,7 +83,7 @@
                 = @event.room.name
           .col-md-12
             %dt Conference:
-            %dd= @event.program.conference.title
+            %dd= link_to @event.program.conference.title, conference_path(@conference)
           .col-md-12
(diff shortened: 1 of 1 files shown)


Ana06:
@hennevogel @bear454 @differentreality anyone up to review this small change? :smile:

app/views/proposals/show.html.haml
```diff
@@ -83,7 +83,7 @@
                 = @event.room.name
           .col-md-12
             %dt Conference:
-            %dd= @event.program.conference.title
+            %dd= link_to @event.program.conference.title, conference_path(@conference.short_title)
```
bear454: We shouldn't need to put `@conference.short_title`, just `@conference`, because of [`Conference#to_param`](https://github.com/openSUSE/osem/blob/master/app/models/conference.rb#L725-L727)... but that's just a nitpick.

Issue #2334, opened 2019-02-19
Revision history is broken


Ana06:
The [revision history link](https://events.opensuse.org/admin/revision_history) in the profile dropdown is broken:

![image](https://user-images.githubusercontent.com/16052290/53010850-30556e80-343f-11e9-8de7-d77487e4fcaf.png)

rahul2240:
@Ana06 Can I work on it?

Ana06:
@rahul2240 sure! :wink:

AndrewKvalheim:
I’ve been unable to reproduce this. Is it specific to the openSUSE instance?

hennevogel:
This should be fixed by https://github.com/openSUSE/osem/pull/3094

PR #2385, opened 2019-03-02
fix revision history broken bug

rahul2240:
Fixes https://github.com/openSUSE/osem/issues/2334

### DIFF
--- a/app/controllers/admin/versions_controller.rb
+++ b/app/controllers/admin/versions_controller.rb
@@ -6,12 +6,15 @@ module Admin
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
-
+      if Conference.any?
+        @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+        @conferences_with_role.uniq!
+      end
       if current_user.has_cached_role? :organization_admin, :any
         @conferences_with_role = Organization.with_role('organization_admin', current_user).map { |org| org.conferences.pluck :short_title }.flatten
+        @conferences_with_role.uniq!
       end
-      @conferences_with_role.uniq!
+
 
       return if @conference.blank?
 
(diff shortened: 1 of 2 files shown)


app/controllers/admin/versions_controller.rb
```diff
@@ -6,12 +6,15 @@ class VersionsController < Admin::BaseController
     load_and_authorize_resource class: PaperTrail::Version
 
     def index
-      @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
-
+      if Conference.any?
```
Ana06: why is this needed? how is this fixing the bug?

app/controllers/admin/versions_controller.rb
```diff
@@ -6,12 +6,15 @@ class VersionsController < Admin::BaseController
...
+        @conferences_with_role = current_user.is_admin? ? Conference.pluck(:short_title) : Conference.with_role([:organizer, :cfp, :info_desk], current_user).pluck(:short_title)
+        @conferences_with_role.uniq!
+      end
       if current_user.has_cached_role? :organization_admin, :any
         @conferences_with_role = Organization.with_role('organization_admin', current_user).map { |org| org.conferences.pluck :short_title }.flatten
+        @conferences_with_role.uniq!
       end
-      @conferences_with_role.uniq!
```
Ana06: why did you need to move this line?

app/views/admin/versions/index.html.haml
```diff
@@ -8,8 +8,9 @@
           %span.caret
         %ul.dropdown-menu
           %li= link_to 'All Conferences & Users',  admin_revision_history_path
-          - @conferences_with_role.each do |conference_short_title|
-            %li= link_to conference_short_title,  admin_conference_revision_history_path(conference_id: conference_short_title)
+          - if !@conferences_with_role.nil?
```
Ana06: this can only be `nil` because `@conferences_with_role.uniq!` was moved.

rahul2240:
@Ana06 I misunderstood the issue, my bad. When I was working on it, I got undefined method for nil:nilclass error in revision history so i thought, it is due to Conference is nil and checking for the conditions as shown in this commit I fixed that but I am no longer able to recreate that, also revision history link in dropdown is working for me.

Issue #2365, opened 2019-02-22
Schedule not viewable


ddemaio:
The schedule is not viewable when logged in and also when logged out. 

When signed out, it is says "You are not authorized to access this page. Maybe you need to sign in?"

When signed in, it says "You are not authorized to access this page." 
![schedule](https://user-images.githubusercontent.com/8996626/53236602-fc738680-3694-11e9-9c6e-5cb4b760bcd2.png)

Ana06:
Related to https://github.com/openSUSE/osem/issues/1343

differentreality:
@Ana06 can you reproduce this locally?

Zennoe:
Thank you for the PR. Can confirm that this issue was fixed.

PR #2389, opened 2019-03-05
Fix several problems with permissions

Ana06:
This fixes an important issue affecting events.opensuse.org:
- Allow to see the schedule without login #2365

There are many things weird in the permissions. I will take a closer look, but I wanted to fix the priority bug first.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -56,10 +56,6 @@ class Ability
         conference = registration.conference
         conference.registration_open? && registration.new_record? && !conference.registration_limit_exceeded?
       end
-
-      can [:new, :create], Event do |event|
-        event.program.cfp_open? && event.new_record?
-      end
     end
   end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -57,10 +57,6 @@ class Ability
         conference.registration_open? && registration.new_record? && !conference.registration_limit_exceeded?
       end
 
-      can :show, Event do |event|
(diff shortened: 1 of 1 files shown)


app/models/ability.rb
```diff
@@ -23,8 +23,8 @@ def not_signed_in
       conference.splashpage&.public == true
     end
     # Can view the schedule
-    can [:schedule, :events], Conference do |conference|
-      conference.program.cfp && conference.program.schedule_public
+    can [:show, :events], Schedule do |schedule|
```
hennevogel: :schedule and :show are two different [abilities](https://github.com/CanCanCommunity/cancancan/wiki/Custom-Actions). both are needed.

hennevogel:
See #2390

app/models/ability.rb
Ana06: why? there is not `schedule` method in the (conference controller)[https://github.com/openSUSE/osem/blob/master/app/controllers/conferences_controller.rb] :confused:

Ana06:
@hennevogel even if the fist commit is not correct (which I think it is), I think we shouldn't close this PR, because bc721fda0b9e56b3669af3fc321c02a8e76eebee and 50418e0  are not included in #2390

hennevogel:
@Ana06 rebase

Ana06:
I won't have time for this in the close future. Closing so that someone else can take care of it.

PR #2390, opened 2019-03-05
Various update fixes

hennevogel:


lib/tasks/several_schedules.rake
```diff
@@ -6,8 +6,8 @@ namespace :data do
   task move_events_attributes: :environment do
     Program.all.each do |program|
       schedule = Schedule.create(program: program)
-      program.selected_schedule = schedule
-      program.save
+      program.selected_schedule = schedule unless program.selected_schedule
```
Ana06: did `selected_schedule` exist in openSUSE? Before there was only one schedule, so this attribute was not there. Was it?

app/models/ability.rb
```diff
@@ -31,6 +31,10 @@ def not_signed_in
       event.state == 'confirmed'
     end
 
+    can [:show, :events], Schedule do |schedule|
```
Ana06: I still thing that https://github.com/openSUSE/osem/pull/2389/commits/ea312eb728a20770ac9da0132bbc9214840b2e93 was correct... there is not `schedule` method in the [conference controller](https://github.com/openSUSE/osem/blob/master/app/controllers/conferences_controller.rb) :confused:

lib/tasks/version.rake
```diff
@@ -10,8 +10,12 @@ namespace :data do
         version.update_attributes(conference_id: version.item_id)
 
       elsif version.item_type == 'Event'
-        event = (version.item || version.reify || version.next.reify)
-
+        event = version.item
```
Ana06: how can happen that `version.item` is `null` but `version.item_id` references a valid event?

Ana06:
@hennevogel is c8f0895fd76776f65ac1771c85185816cf27f056 related or solving https://github.com/openSUSE/osem/issues/2334?

app/models/ability.rb
hennevogel: you can also freely invent your own "method" names and check for them.

lib/tasks/several_schedules.rake
Ana06: or is this in case you run it after another schedule has been created? That makes sense...

lib/tasks/version.rake
differentreality: If the item is let's say an Event, then the version record keeps version.item_id, let's say number 9 (as in 9 being the id of the Event record). Now that's a field on the database and the record won't get delete once Event.find(9) is destroyed.

`version.item` tries to give you the actual object Event.find(9), which may or may not be in our database. 

`version.reify` returns the item as it was one version before (if the version is the one where the event was destroyed, then it builds it as it was before its destruction)

lib/tasks/several_schedules.rake
hennevogel: the latter yes

CHANGES.md
```diff
@@ -0,0 +1,49 @@
+# Changes in OSEM 1.1
```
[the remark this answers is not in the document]
hennevogel: Ok, you're right.

hennevogel:
@Ana06 please review again

Issue #2356, opened 2019-02-21
Flickering test: Event as a participant signed_in user submits a valid proposal


Ana06:
```
  1) Event as a participant signed_in user submits a valid proposal
     Failure/Error: page.find('#event_description')
     
     Capybara::ElementNotFound:
       Unable to find visible css "#event_description"
     # ./spec/features/proposals_spec.rb:124:in `block (3 levels) in <top (required)>'
```

hennevogel:
This is flickering hard.

PR #2408, opened 2019-03-08
Fix flickering submit valid proposal test

Ana06:
The `find` was added to try to fix the flickering test: https://github.com/openSUSE/osem/commit/ac3337bd1ed93462ee6d8cae5651fe8d633305d8

The reason why this fail is that the Javascript is not loaded yet.
`find` and `fill_in` have the same `default_max_wait_time`. Because of
that adding `find` doesn't solve the problem. Increase wait time
instead.

Fixes https://github.com/openSUSE/osem/issues/2356

### DIFF
--- a/spec/features/proposals_spec.rb
+++ b/spec/features/proposals_spec.rb
@@ -121,8 +121,7 @@ feature Event do
       fill_in 'event_abstract', with: 'Lorem ipsum abstract'
       click_link 'Do you require something special?'
 
-      page.find('#event_description')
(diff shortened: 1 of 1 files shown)


Ana06:
I am not sure if the time limit would be enough and also not sure why a simple collapse needs that much time. I also think that we have too many feature tests in OSEM... some of them should be replace by controller and model tests.

Ana06:
This doesn't seem to fix the test....

PR #2413, opened 2019-03-10
track organizers should only add users for their tracck

differentreality:
**Checklist**

- [ ] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
- [ ] My branch is up-to-date with the upstream `master` branch.

- It is possible for a track organizer to toggle users for other roles (conference organizer, other track organizers).

- Track organizers should only add/remove users from the role track_organizer which refers to the track(s) they are organizing.

This clearly needs better testing :/

### DIFF
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -292,6 +292,8 @@ class AdminAbility
       role.resource_type == 'Conference' || role.resource_type == 'Track'
     end
 
+    cannot :toggle_user, Role
+
     can :toggle_user, Role do |role|
       role.resource_type == 'Track' && track_ids_for_track_organizer.include?(role.resource_id)
     end


app/models/admin_ability.rb
```diff
@@ -292,6 +292,8 @@ def signed_in_with_track_organizer_role(user)
       role.resource_type == 'Conference' || role.resource_type == 'Track'
     end
 
+    cannot :toggle_user, Role
+
     can :toggle_user, Role do |role|
       role.resource_type == 'Track' && track_ids_for_track_organizer.include?(role.resource_id)
```
Ana06: The permissions you wan is exactly that here, right? Why do you need the `cannot`?

differentreality: You are right, I just realized what I stumbled upon was an organization admin being able to make all these fancy changes! Thanks!

PR #2431, opened 2019-03-14
Bring back some of the contribution guidelines

hennevogel:


Ana06:
I guess no answer is also an answer.... :wink:

lagartoflojo:
@Ana06 As I haven't been involved in OSEM for a long time, I don't feel entitled to an opinion tbh; so yes, my non-answer is also an answer.

PR #2437, opened 2019-03-15
Disable capybara animation

Ana06:
Turn off jquery animation to speed up the Capybara JS specs.

Co-authored-by: Stephan Kulow stephan@kulow.org

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -60,6 +60,8 @@ RSpec.configure do |config|
   #     --seed 1234
   config.order = 'random'
 
+  Capybara.disable_animation = true
+
   Capybara.register_driver :firefox do |app|
     Capybara::Selenium::Driver.new(app, browser: :firefox)
   end


hennevogel:
how much did it speed things up? :-)

Ana06:
@hennevogel 

> how much did it speed things up? :-)

It is more about the animation happening faster to avoid (or easy debug) tests involving animations than the speed we gain with this.

spec/spec_helper.rb
```diff
@@ -60,6 +60,8 @@
   #     --seed 1234
   config.order = 'random'
 
+  Capybara.disable_animation = true
```
hennevogel: Can you mention why we do this?

Ana06: do you think it is needed? It is a capybara option, which can be check directly in capybara. I am not sure we should duplicate information.

hennevogel: well its not for what the commit message says. please either a comment or the right message.

PR #2442, opened 2019-03-17
Add condition for missing_speakers in a track.

### LINKED ISSUE #1942: User with accepted track request can see unregistered speakers of all events.

rishabhptr:
Fixes #1942

**Checklist**

A user could see missing speakers of events which not accessible by him ( not under his track) in the reports option.

### DIFF
--- a/spec/controllers/admin/reports_controller_spec.rb
+++ b/spec/controllers/admin/reports_controller_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Admin::ReportsController do
+
+  let(:conference) { create(:conference, start_date: Date.current - 1.day) }
+  let!(:admin) { create(:admin) }
+  let!(:user1) { create(:user) }
+  let!(:user2) { create(:user) }
+  let!(:venue) { create(:venue, conference: conference) }
+  let!(:room) { create(:room, venue: venue) }
+  let!(:track_submitter) { create(:user) }
+  let!(:self_organized_track) { create(:track, :self_organized, submitter_id: track_submitter.id, program: conference.program, name: 'My awesome track', start_date: Date.current, end_date: Date.current, room: room, state: 'confirmed') }
+  let!(:track) { create(:track, program: conference.program, color: '#800080') }
+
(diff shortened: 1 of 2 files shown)


rishabhptr:
@Ana06 Sure, but I am not sure where the test for this would go, as I couldn't find a file for this, we are just testing the track organizer's ability to access events in track_organizer_ability_spec.rb . Thanks!

differentreality:
So any progress with the tests? :) 

A test that checks that we actually return a collection of users that are speakers of this particular track (and not everything) would have caught the bug!

rishabhptr:
@Ana06 @differentreality Sorry for the delay and thanks for the suggestions, I did make a new file for this and followed a similar approach in which only user of event with that particular track is returned.

Also while writing the test I noticed one more thing, we are checking for role submitter for missing speakers, but that might not always be the case as even admin can create any event under a particular track and add users as speakers, it will show only the admin as missing speakers and not the actual speaker.   :)

spec/controllers/admin/reports_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  context 'track organizer is signed in' do
+    before :each do
+      sign_in(track_submitter)
+      self_organized_track.assign_role_to_submitter
+    end
+
+    describe 'GET #index' do
+      it 'initialises missing speakers' do
+        get :index, conference_id: conference.short_title
+        expect(response).to render_template('index')
```
Ana06: Please split this to have single expectation specs, check: http://www.betterspecs.org/

rishabhptr:
@Ana06 It's best to resolve in single PR. As when a user creates a proposal he is both submitter and speaker but organizer is only the submitter when creating an event. So I've just changed it to speaker in EventUser query. :smile:

Issue #2455, opened 2019-03-22
Form dates are always in read only mode


lbenore:
After creating a new conference you are unable to setup the registration period because the start and end dates are always set to read only. The track dates have the same issue the start and end dates are always set to read only.

![image](https://user-images.githubusercontent.com/28565816/54834027-eac9d280-4c84-11e9-8713-b3df7ec1ce2b.png)

cyrillefr:
Hello, 
I came across the same issue these days too, regarding the "Registration period" form.
The component does not show up because of the readonly attributes defined in the haml form.
The problem lies in the app/assets/javascripts/osem-datepickers.js file.

There are typos while instantiating the component:
`$('#registration-period-start-datapicker').datetimepicker({ ...`
It should be datepicker instead of datapicker.
I hint there is a similar problem with call for paper since there is a similar typo.(But in a haml form this time).

But there is the readonly attribute that prevents the input to be updated. Getting rid of the readonly attribute will allow that, however it will possible to put a non valid entry in that input.
My suggestion would be to keep the readonly but, to pass an ignoreReadonly at the inception of the Bootstrap datepicker component in JavaScript, that allows change of date but prevents to enter manually a non valid date.

After some changes in the code, I have also noticed JS errors (calls on  mispelled methods on datepicker).

I volunteer for this issue since I have already done some work on it.
I can check for cfp, tracks and conference since they are using the same datepicker component.

Ana06:
@cyrillefr go ahead!

@rahul2240 you wanted to do some related work as well, although I though that where improvements and not broken stuff. Please be careful not to work on the same as @cyrillefr (you can also review his PR when he send one and work together to fix all the problems) :wink:

rahul2240:
@Ana06 Yes, I was talking about the broken stuff :sweat_smile: I thought I was not allowed to make those changes so I dropped those commits, but I can help in reviewing the code. Thank you :smile:

lbenore:
@cyrillefr I also noticed the datepickers on the program page for the voting dates has the same issue.

![image](https://user-images.githubusercontent.com/28565816/55006753-ddbe2380-4fa3-11e9-9c59-d650d6d10b26.png)

cyrillefr:
Thank you @lbenore, you are right. I have looked after this part too.

PR #2466, opened 2019-03-26
Fixes datetime pickers

cyrillefr:
**Checklist**

- read only attributes in haml to prevent non valid manual date
- typos in haml and in JS
- undefined methods in JS

- Fixes https://github.com/openSUSE/osem/issues/2455

- Add readonly on datetime pickers (haml)
- Fix some typos (haml)
- Update method names in JavaScrip that do not exit for bootstrap component
- Do some name matching between in JS and Haml

### DIFF
--- a/app/assets/javascripts/osem-datepickers.js
+++ b/app/assets/javascripts/osem-datepickers.js
@@ -4,7 +4,6 @@ $(function () {
   $("input[id^='datetimepicker']").datetimepicker({
     useCurrent: false,
     sideBySide: true,
-    autoclose: true,
     format: 'YYYY-MM-DD HH:mm'
   });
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/osem-datepickers.js
```diff
@@ -4,8 +4,8 @@ $(function () {
   $("input[id^='datetimepicker']").datetimepicker({
     useCurrent: false,
     sideBySide: true,
-    autoclose: true,
-    format: 'YYYY-MM-DD HH:mm'
+    format: 'YYYY-MM-DD HH:mm',
+    ignoreReadonly: true
```
Ana06: why is this needed? where is readonly set? why not removing it there?

app/views/admin/conferences/new.html.haml
```diff
@@ -9,7 +9,7 @@
                                 input_html: { required: 'required', pattern: '[a-zA-Z0-9_-]+', title: 'Only letters, numbers, underscores, and dashes.' }, prepend: conferences_url + '/'
       = f.inputs 'Scheduling' do
         = f.input :timezone, as: :time_zone, default: Time.zone.name, hint: 'Please select in what time zone your conference will take place.'
-        = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', required: 'required' }
-        = f.input :end_date, as: :string, input_html: { id: 'conference-end-datepicker', required: 'required' }
+        = f.input :start_date, as: :string, input_html: { id: 'conference-start-datepicker', required: 'required', readonly: 'readonly' }
```
Ana06: why do you set readonly?

app/assets/javascripts/osem-datepickers.js
cyrillefr: Hi, this is is a 2 ways strategy.
By setting readonly in the haml, I prevent the user from entering non valid dates. But, by doing so, updating from the datetimepicker will not work. So, I added ignoreReadonly, so that user still cannot have access manually to the input, but the datetime picker has access to it.
cf. http://eonasdan.github.io/bootstrap-datetimepicker/Options/#ignorereadonly

app/views/admin/conferences/new.html.haml
cyrillefr: To prevent user to input manually a date. And to be coherent with the first form I encountered (Registration Period) that was already read only.
To insure however that the datetime picker can update with a readonly attribute in html, I had to add an ignoreReadonly in the datepicker js file. (See my comment above).

Ana06: I wouldn't prevent the user to manually input a date. I would make it coherent the other way around.

cyrillefr: Ok, I will update the changes in that way and remove the readonly attributes.

cyrillefr: I have pushed the changes. I hope this is ok.

differentreality:
Are pickers working for you now?

Issue #1664, opened 2017-08-29
Expected string default value for '--serializer'; got true (boolean)


tobru:
When starting the application in Docker, there is this error message displayed:

```
>>> Upgrading database...
Expected string default value for '--serializer'; got true (boolean)
>>> Precompiling assets...
Expected string default value for '--serializer'; got true (boolean)
```

differentreality:
This also happens for me without docker

```
=> Booting WEBrick
=> Rails 4.2.9 application starting in development on http://localhost:3001
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
Expected string default value for '--serializer'; got true (boolean)
[2017-08-29 11:22:56] INFO  WEBrick 1.3.1
[2017-08-29 11:22:56] INFO  ruby 2.3.3 (2016-11-21) [x86_64-linux]
[2017-08-29 11:22:56] INFO  WEBrick::HTTPServer#start: pid=27554 port=3001

```

namangupta01:
@differentreality This error is because of the gem **Thor 0.19.4**. Which can be resolved by downgrading the **Thor** version to **0.19.1**. I added the **gem 'thor', '0.19.1'**. And then run the command **bundle update**. It works for me. Here is the screenshot.
![screenshot from 2017-08-31 22-29-40](https://user-images.githubusercontent.com/20709985/29935410-4881e5b4-8e9c-11e7-93d8-fa1716966615.png)

namangupta01:
Starting of server occurs without error but by doing this when we are requesting home page it is showing error.
![screenshot from 2017-08-31 23-17-47](https://user-images.githubusercontent.com/20709985/29937283-ac21a5ea-8ea2-11e7-8a69-91970e8e199f.png)

differentreality:
That's probably because in your closed PR you updated tons of gems instead of just updating thor. You should **not** do that. For this issue, you should only update thor gem, if that's where the problem is, and any gem that is a dependency to thor's new version (if there is one).

namangupta01:
@differentreality Thanks! Now i have only added ** gem 'thor', '0.19.1'** in gem file and after running sudo bundle update thor, it's working fine without showing any error. :)

PR #2481, opened 2019-04-01
Do not lock gems

Ana06:
Do not lock gems to an specific version if there is not a reason for it. This allow depfu to update them. The gems that were not in the last version where updated. :bowtie:


Gemfile
```diff
@@ -87,8 +87,6 @@ gem 'jquery-ui-rails', '~> 4.2.1'
 # for languages validation
 gem 'iso-639'
 
-gem 'thor', '~> 0.19'
```
bear454: We don't need _thor_ anymore?

Ana06: no, as I say in the commit message, thor was set to 1.19.1 because of a bug, but it is not longer needed. Check #1664

PR #2483, opened 2019-04-03
Add uniqueness validation for votes

### LINKED ISSUE #2020: Add validation for user votes
**I'm submitting a ..**
- [ ] Bug Report

**Current behavior:**

A user is able to vote on an event more than once. 

Each user should only vote 1 time for each event.

**Steps to reproduce:**

I saw the bug happening in events.opensuse.org (where I deleted the second vote). I couldn't reproduce. 
The voting process is an ajax call, I am thinking that perhaps it was somehow triggered twice, not necessarily by user's actions.

We should validate against this, so that our database does not have the same person voting on the same proposal.

kaustubh-nair:
**Checklist**

 Fixes #2020 

Prevents duplicate addition of votes for events

### DIFF
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -6,6 +6,8 @@ class Vote < ApplicationRecord
 
   has_paper_trail ignore: [:updated_at], meta: { conference_id: :conference_id }
 
+  validates :event, uniqueness: { scope: :user }
+
   delegate :name, to: :user
 
   private
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -6,7 +6,7 @@ class Vote < ApplicationRecord
 
   has_paper_trail ignore: [:updated_at], meta: { conference_id: :conference_id }
 
-  validates :event, uniqueness: { scope: :user }
+  validates :user_id, uniqueness: { scope: :event_id }
 
   delegate :name, to: :user
 
--- a/app/models/vote.rb
+++ b/app/models/vote.rb
@@ -6,7 +6,7 @@ class Vote < ApplicationRecord
 
   has_paper_trail ignore: [:updated_at], meta: { conference_id: :conference_id }
 
-  validates :user_id, uniqueness: { scope: :event_id }
+  validates :user, uniqueness: { scope: :event }
 
   delegate :name, to: :user
 
--- a/spec/models/vote_spec.rb
+++ b/spec/models/vote_spec.rb
@@ -9,9 +9,5 @@ describe Vote do
 
   describe 'validation' do
     it { is_expected.to validate_uniqueness_of(:user_id).scoped_to(:event_id) }
-
-    it 'has a valid factory' do
-      expect(build(:vote)).to be_valid
-    end
   end
 end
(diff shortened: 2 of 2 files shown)


spec/models/vote_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Vote do
```
Ana06: you may want to check [`validate_uniqueness_of`](https://www.rubydoc.info/github/thoughtbot/shoulda-matchers/master/Shoulda%2FMatchers%2FActiveRecord:validate_uniqueness_of) instead :wink:

app/models/vote.rb
```diff
@@ -6,6 +6,8 @@ class Vote < ApplicationRecord
 
   has_paper_trail ignore: [:updated_at], meta: { conference_id: :conference_id }
 
+  validates :event, uniqueness: { scope: :user }
```
Ana06: I would write this the other way arround:

```ruby
validates :user, uniqueness: { scope: :event }
```

The error message is slightly different. :smile:

spec/models/vote_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:user) { create(:user) }
+  let(:event) { create(:event) }
+  let(:vote) { create(:vote, event: event, user: user) }
+
+  describe 'validation' do
+    it { is.expected_to validate_uniqueness_of(:user).scoped_to(:event) }
+
+    it 'has a valid factory' do
```
kaustubh-nair: Probably unnecessary right now but I let it be because it might be useful later

Ana06: In what way could it be useful later? i don't think leaving unneeded code is a good idea. We also have to maintain this code. If we need it in the future, we will add it then :wink:

app/models/vote.rb
```diff
@@ -6,6 +6,8 @@ class Vote < ApplicationRecord
 
   has_paper_trail ignore: [:updated_at], meta: { conference_id: :conference_id }
 
+  validates :user_id, uniqueness: { scope: :event_id }
```
Ana06: Why are you using the id instead of the relationship?

spec/models/vote_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Vote do
+  let(:user) { create(:user) }
```
Ana06: `user` and `event` are not needed. `let(:vote) { create(:vote) }` will create them. (That's why we use factories, to save all this work :wink: )

differentreality: @kaustubh-nair would you please take care of the review comments, so that we can merge this?

kaustubh-nair: Oh sorry @differentreality 
I'll make the changes and push again asap

Ana06:
@kaustubh-nair do you need help to address the requested changes?

Ana06:
Closing due to inactivity. @kaustubh-nair in case you address the requested changes reopen it again so that we review it.

PR #2515, opened 2019-05-04
Some prerequisites to update to Rails 5.1

Ana06:
The PR to update to Rails 5.1 is getting too big. Let's split out the things that can be already fixed without the need to update. :smile:

### DIFF
--- a/spec/models/role_spec.rb
+++ b/spec/models/role_spec.rb
@@ -4,8 +4,8 @@ require 'spec_helper'
 
 describe Role do
   let(:conference) { create(:conference) }
-  let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
-  let!(:cfp_role) { Role.find_by(name: 'cfp', resource: conference) }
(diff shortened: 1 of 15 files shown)


spec/models/role_spec.rb
```diff
@@ -4,8 +4,8 @@
 
 describe Role do
   let(:conference) { create(:conference) }
-  let!(:organizer_role) { Role.find_by(name: 'organizer', resource: conference) }
-  let!(:cfp_role) { Role.find_by(name: 'cfp', resource: conference) }
+  let!(:organizer_role) { Role.find_or_create_by(name: 'organizer', resource: conference) }
+  let!(:cfp_role) { Role.find_or_create_by(name: 'cfp', resource: conference) }
```
hennevogel: Why isn't this a factory?

Ana06: Because there should be there, but it seems that after the update they are not anymore. There are that many things in those tests that need to be changed, that I would focus on finishing the Rails update (there is still a long way to achieve this and there is a security issue) and refactor/rewrite the tests&code afterwards

Issue #2511, opened 2019-05-02
OSEM Installation using Docker doesn't work


medwinz:
**I'm submitting a ..**
- [x] Bug Report
- [ ] Feature Request

**Current behavior:**

When following INSTALL.md to install osem using docker in openSUSE Leap 15.0, it doesn't work as expected. When I run 
```
docker-compose -f docker-compose.yml.production-example build
```
it produce some errors. The errors are:
```
ost-install message from rails-assets-markdown:
This component doesn't define main assets in bower.json.
Please open new pull request in component's repository:
https://github.com/evilstreak/markdown-js
Removing intermediate container b64c91e84b75
 ---> 457949717fa5
Step 6/8 : RUN export RAILS_ENV=production; bundle exec rake assets:clobber assets:precompile
 ---> Running in d17ce16eb04a
rake aborted!
LoadError: cannot load such file -- rubocop/rake_task
/osem/lib/tasks/rubocop.rake:3:in `<top (required)>'
/osem/Rakefile:9:in `<top (required)>'
(See full trace by running task with --trace)
ERROR: Service 'production_web' failed to build: The command '/bin/sh -c export RAILS_ENV=production; bundle exec rake assets:clobber assets:precompile' returned a non-zero code: 1
```

**Expected correct behavior:**

Images are build.

**Steps to reproduce:**

In openSUSE Leap 15.0 follow the step in INSTALL.md and run 
```
docker-compose -f docker-compose.yml.production-example build
```

**Other information:**

I can install it by adding tmp directory in osem directory and give it sticky bit.
```
cd osem
mkdir tmp
chmod 777 tmp/
chmod o+t tmp/
```
Then running 
```
docker-compose up -d
```
The container is up and running and I can access it through browser

![Screenshot_20190502_190524](https://user-images.githubusercontent.com/1616257/57074204-6cddfa00-6d0d-11e9-8492-f7b71460cb21.png)

Without adding tmp directory with sticky bit in osem directory there will be a problem with the permission when trying to acces it using browser. Below is the example
![Screenshot_20190426_183850](https://user-images.githubusercontent.com/1616257/57074022-09ec6300-6d0d-11e9-87dc-9a81bf6c7908.png)

OndraX:
I'm having the same issue, though with the current version of the Dockerfile.production and the tmp file commands in osem directory don't seem to fix it.

```
[...]

PR #2516, opened 2019-05-07
Only define rubocop tasks in the dev env

hennevogel:
This broke loading the rake tasks in the production environemt as we only haverubocop in development.


hennevogel:
Also fixes #2511

Issue #2527, opened 2019-06-03
Invite users to become booth responsible


rahul2240:
User can invite other users to become booth responsible while creating a new or editing an existing booth

**Changes in the PR will be:** 
- [ ] Add option to invite user for booth responsibles
- [ ] Install ‘devise_invitable’ gem and configure
- [ ] Edit mail template that will be used for invitations
- [ ] Install 'mailkick' gem
- [ ] Add unsubscribe link in the email

Ana06:
@rahul2440

> I have a local PR here.

Why local?

rahul2240:
@Ana06 We have to create a PR for GSOC branch, right? Since it doesn't exist right now so I created it locally.

PR #2535, opened 2019-06-09
Invitation for booth responsibles

rahul2240:
Related to #2527 

**Checklist**

**Need**

Whenever a conference is organized, people submit booth request for that conference. It is possible to add responsibles to booth but only registered users can be added as responsibles. 

With this PR, booth submitter will be able to invite unregistered users to join the app and handle the booth. Making it easier for the submitter to add users as responsibles only one time rather than waiting for users to first join and then adding them as responsibles. 

**Workflow**
  1. User submits a booth request.
  2. This invitation field will be available in new booth and edit booth form for submitters.
  3. User can enter any number of emails that are needed to be invited.
  4. If user with that email already exists in user db then he/she will become booth responsible. 
  5. If user with that email doesn't exist in user db then that user will be created and an invitation mail will be sent to the user with an invitation token to join the app. 
  6. Once the user joins the app, they will be able to access the booth.

**new booth view**
ready

**mail**
accept1

**Registartion page after accepting link**
accept2

**After registration my booth option available**
accept3

**columns added to User**
db

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'rails-i18n'
 # as authentification framework
 gem 'devise'
 gem 'devise_ichain_authenticatable'
+gem 'devise_invitable'
 
 # for openID authentication
 gem 'omniauth'
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -129,7 +129,8 @@ module Admin
 
     def booth_params
       params.require(:booth).permit(:title, :description, :reasoning, :state, :picture, :conference_id,
-                                    :created_at, :updated_at, :submitter_relationship, :website_url, responsible_ids: [])
+                                    :created_at, :updated_at, :submitter_relationship, :website_url, :invite_responsible,
+                                    responsible_ids: [])
     end
   end
 end
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -4,6 +4,7 @@ module Admin
   class BoothsController < Admin::BaseController
     load_and_authorize_resource :conference, find_by: :short_title
     load_and_authorize_resource through: :conference
+    include Invitation
 
     def index
       @file_name = "booths_for_#{@conference.short_title}"
@@ -38,6 +39,7 @@ module Admin
       @booth.submitter = current_user
 
       if @booth.save
+        booth_responsible_invite
         redirect_to admin_conference_booths_path,
                     notice: 'Booth successfully created.'
       else
@@ -54,7 +56,7 @@ module Admin
       @url = admin_conference_booth_path(@conference.short_title, @booth.id)
 
       @booth.update_attributes(booth_params)
-
+      booth_responsible_invite
       if @booth.save
         redirect_to admin_conference_booths_path,
                     notice: "Successfully updated booth for #{@booth.title}."
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -9,6 +9,8 @@ class ApplicationController < ActionController::Base
   # Ensure every controller authorizes resource or skips authorization (skip_authorization_check)
   check_authorization unless: :devise_controller?
 
+  before_action :devise_configure_permitted_parameters, if: :devise_controller?
+
   def store_location
     # store last url - this is needed for post-login redirect to whatever the user last visited.
     return unless request.get?
@@ -64,5 +66,9 @@ class ApplicationController < ActionController::Base
 
   def not_found
     raise ActionController::RoutingError.new('Not Found')
+  end
+
+  def devise_configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:accept_invitation, keys: [:username])
   end
 end
--- a/app/controllers/booths_controller.rb
+++ b/app/controllers/booths_controller.rb
@@ -5,6 +5,7 @@ class BoothsController < ApplicationController
   load_resource :conference, find_by: :short_title
   load_and_authorize_resource through: :conference
   skip_authorize_resource only: [:withdraw, :confirm, :restart]
+  include Invitation
 
   def index
     @booths = current_user.booths.where(conference_id: @conference.id).uniq
@@ -22,6 +23,7 @@ class BoothsController < ApplicationController
     @booth.submitter = current_user
 
     if @booth.save
+      booth_responsible_invite
       redirect_to conference_booths_path,
                   notice: 'Booth successfully created.'
     else
@@ -38,6 +40,7 @@ class BoothsController < ApplicationController
     @url = conference_booth_path(@conference.short_title, @booth.id)
     @booth.update_attributes(booth_params)
 
+    booth_responsible_invite
     if @booth.save
       redirect_to conference_booths_path,
                   notice: 'Booth successfully updated!'
--- a/app/controllers/booths_controller.rb
+++ b/app/controllers/booths_controller.rb
@@ -94,6 +94,7 @@ class BoothsController < ApplicationController
 
   def booth_params
     params.require(:booth).permit(:title, :description, :reasoning, :state, :picture, :conference_id,
-                                  :created_at, :updated_at, :submitter_relationship, :website_url, responsible_ids: [])
+                                  :created_at, :updated_at, :submitter_relationship, :website_url, :invite_responsible,
+                                  responsible_ids: [])
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,8 +49,8 @@ class User < ApplicationRecord
                       [:ichain_authenticatable, :ichain_registerable, :omniauthable, omniauth_providers: []]
                     else
                       [:database_authenticatable, :registerable,
-                       :recoverable, :rememberable, :trackable, :validatable, :confirmable,
-                       :omniauthable, omniauth_providers: [:suse, :google, :facebook, :github]]
+                       :recoverable, :rememberable, :trackable, :validatable, :invitable,
+                       :confirmable, :omniauthable, omniauth_providers: [:suse, :google, :facebook, :github]]
                     end
 
   devise(*devise_modules)
--- a/app/views/booths/_form.html.haml
+++ b/app/views/booths/_form.html.haml
@@ -29,4 +29,18 @@
         plugins: ['remove_button'],
         minItems: 2
       } )
+      $('#booth_invite_responsible').selectize({
+        plugins: ['remove_button'],
+        delimiter: ',',
+        persist: false,
+        create: function(input) {
+          if (!/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(input)){
+            return {}
+          }
+          return {
+              value: input,
+              text: input
+          }
+        }
+      } )
     });
--- a/app/views/booths/_form.html.haml
+++ b/app/views/booths/_form.html.haml
@@ -12,6 +12,8 @@
           hint: 'e.g. employee, comunity manager, etc'
         = f.input :website_url
         = responsibles_selector_input f
+        = f.input :invite_responsible,
+          hint: 'This field is to invite unregistered users using their email to become booth responsibles.'
         = image_tag f.object.picture.thumb.url if f.object.picture?
         = f.input :picture
 
--- a/app/views/devise/invitations/edit.html.haml
+++ b/app/views/devise/invitations/edit.html.haml
@@ -1,0 +1,20 @@
+.container
+  .row
+    .col-md-6.col-md-offset-3
+      .panel.panel-default
+        .panel-heading
+          %h3.panel-title
+            %h2= t 'devise.invitations.edit.header'
+        .panel-body
+          = form_for(resource, as: resource_name, url: invitation_path(resource_name), html: { method: :put }) do |f|
+            = render 'devise/shared/error_messages', resource: resource
+            = f.hidden_field :invitation_token, readonly: true
+            - if f.object.class.require_password_on_accepting
+              = f.label :username
+              = f.text_field :username, class: 'form-control'
+              = f.label :password
+              = f.password_field :password, class: 'form-control'
+              = f.label :password_confirmation
+              = f.password_field :password_confirmation, class: 'form-control'
+            %br
+            = f.submit t('devise.invitations.edit.submit_button'), class: 'btn btn-default btn-primary'
--- a/spec/controllers/admin/booths_controller_spec.rb
+++ b/spec/controllers/admin/booths_controller_spec.rb
@@ -77,6 +77,18 @@ describe Admin::BoothsController do
         it 'shows success message' do
           expect(flash[:notice]).to match('Booth successfully created.')
         end
+
+        it 'creates a new user on inviting booth responsible' do
+          expect(User.where(email: 'user@example.com')).to exist
+        end
+
+        it 'does not create a user with invalid email on inviting booth responsible' do
+          expect(User.where(email: 'example')).not_to exist
+        end
+
+        it 'invited user should be a part of booth responsibles' do
+          expect(Booth.last.responsibles.ids).to include(User.find_by(email: 'user@example.com').id)
+        end
       end
 
       context 'create action fails' do
@@ -113,7 +125,7 @@ describe Admin::BoothsController do
 
     describe 'PATCH #update' do
       context 'updates suchessfully' do
-        before { patch :update, params: { id: booth.id, booth: attributes_for(:booth, title: 'different'), conference_id: conference.short_title } }
+        before { patch :update, params: { id: booth.id, booth: attributes_for(:booth, title: 'different', invited_users: 'user@example.com, example'), conference_id: conference.short_title } }
         it 'redirects to admin booth index path' do
           expect(response).to redirect_to admin_conference_booths_path
         end
@@ -125,6 +137,19 @@ describe Admin::BoothsController do
         it 'updates booth' do
           booth.reload
           expect(booth.title).to eq('different')
+        end
+
+        it 'creates a new user on inviting booth responsible' do
+          expect(User.where(email: 'user@example.com')).to exist
+        end
+
+        it 'does not create a user with invalid email on inviting booth responsible' do
+          expect(User.where(email: 'example')).not_to exist
+        end
+
+        it 'invited user should be a part of booth responsibles' do
+          booth.reload
+          expect(booth.responsibles.ids).to include(User.find_by(email: 'user@example.com').id)
         end
       end
     end
--- a/spec/factories/booths.rb
+++ b/spec/factories/booths.rb
@@ -12,5 +12,6 @@ FactoryBot.define do
 
     submitter { create(:user) }
     responsible_ids { [create(:user).id] }
+    invite_responsible { 'user@example.com, example' }
   end
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -161,6 +161,9 @@ GEM
       warden (~> 1.2.3)
     devise_ichain_authenticatable (0.3.2)
       devise (>= 2.2)
+    devise_invitable (2.0.1)
+      actionmailer (>= 5.0)
+      devise (>= 4.6)
     diff-lcs (1.3)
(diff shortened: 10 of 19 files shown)


app/controllers/admin/booths_controller.rb
```diff
@@ -38,6 +38,24 @@ def create
       @booth.submitter = current_user
 
       if @booth.save
+
+        if booth_params[:invited_users]
+          emails_array = booth_params[:invited_users].split(',')
+
+          emails_array.each do |email|
+            if User.find_by(email: email).nil? || ((!User.find_by(email: email).confirmed? &&
```
Ana06: why need the email to be confirmed?

Ana06: why are you looking by email if you also have the id in the view?

Ana06: how can it happen that the user doesn't exist? shouldn't the user get an error in this case?

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -77,6 +77,14 @@
         it 'shows success message' do
           expect(flash[:notice]).to match('Booth successfully created.')
         end
+
+        it 'creates a new user' do
```
Ana06: how is this related? what are you testing here?

Ana06:
@rahul2240 you also need to fix the lint offenses and tests (in general it is a good practice to run the test locally before sending the PR :wink:)

Gemfile
```diff
@@ -224,6 +224,9 @@ gem 'nokogiri', '>= 1.8.1'
 # memcached binary connector
 gem 'dalli'
 
+# unsubscribe from emails
+gem 'mailkick'
```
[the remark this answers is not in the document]
rahul2240: @Ana06 Sure I will create small PR from now.

app/controllers/admin/booths_controller.rb
rahul2240: > why need the email to be confirmed?

I want that only unconfirmed users can receive a new invitation so I used negation before that.
 

> why are you looking by email if you also have the id in the view?

@Ana06 Which id? I am confused.


> how can it happen that the user doesn't exist? shouldn't the user get an error in this case?

@Ana06  What if email is not valid (say: abc) so in that case we must add a condition to check it and if an email id is not correct then i have added a javascript check for email id (this condition will prevent error if user has disabled javascript). So I found no reason for sending an error to user since he cannot add any string other than email.

spec/controllers/admin/booths_controller_spec.rb
rahul2240: @Ana06 yes sorry i will explain it more.

rahul2240:
> @rahul2240 you also need to fix the lint offenses and tests (in general it is a good practice to run the test locally before sending the PR )

@Ana06 I have 1 lint only for 105 lines used whose limit is 100, so can you tell me what i should do and i dont know what error is breaking the other tests as it is something related to Selenium::WebDriver::Error::UnknownError:
            unknown error: Chrome failed to start: exited abnormally

so i send a PR to ask about it.

app/models/user.rb
```diff
@@ -59,7 +59,7 @@ def by_conference(conference)
 
   has_many :openids
 
-  attr_accessor :login
+  attr_accessor :login, :invitation_message
```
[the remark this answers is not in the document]
rahul2240: @Ana06 because we need custom messages for users(example: to become booth responsible or to become organizer etc). So I thought having an attribute will be beneficial for any custom message invitation.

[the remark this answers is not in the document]
Ana06: why don't you pass it as a parameter in `invite`? The invitation doesn't belong to the user model.

app/controllers/admin/booths_controller.rb
Ana06: @rahul2240 

> I want that only unconfirmed users can receive a new invitation so I used negation before that.

I am not sure I understand you. Can you elaborate?

> @Ana06 Which id? I am confused.

User id

> What if email is not valid (say: abc) so in that case we must add a condition to check it and if an email id is not correct then i have added a javascript check for email id (this condition will prevent error if user has disabled javascript). So I found no reason for sending an error to user since he cannot add any string other than email.

you can always modify the Javascript from the user side. But I would say before entering on those details, we should focus on the workflow. Please check https://github.com/openSUSE/osem/pull/2535#discussion_r292991717

app/models/user.rb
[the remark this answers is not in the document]
rahul2240: I will try it :)

Ana06:
@rahul2240 

>  so i send a PR to ask about it.

but you didn't ask :rofl: It is fine. Let's check if I can help you. :wink: 

> @Ana06 I have 1 lint only for 105 lines used whose limit is 100, so can you tell me what i should do

When a class gets to big, it is normally because there are things on it that doesn't belong there. So, you need to split things out (moving model related logic to the models, using concerns, introduce new controllers, etc.). invite is currently in the User model, but it is not necessarily related to the user. It is maybe time to put everything related with the invitation (in both user model and booth controller) out and together.

> and i dont know what error is breaking the other tests as it is something related to Selenium::WebDriver::Error::UnknownError:
unknown error: Chrome failed to start: exited abnormally

This is not related to your PR, it is failing in other PRs as well. We need to fix it, but not as part of this PR.

app/controllers/admin/booths_controller.rb
rahul2240: > I am not sure I understand you. Can you elaborate?
> 
> >

@Ana06 if a user is confirmed then we don't want to send an invitation for registration in OSEM. 
But if the user is unconfirmed, then we should send an invitation again to him to join the app and become booth responsible.

> User id

@Ana06  we have user id in view only for those who are currently responsibles, but here, we are asking the creators to enter email id of the users.

Gemfile
```diff
@@ -76,6 +76,7 @@ gem 'uglifier', '>= 1.3.0'
 # as the front-end framework
 gem 'autoprefixer-rails'
 gem 'bootstrap-sass', '~> 3.4.0'
+gem 'bootstrap-tagsinput-rails'
```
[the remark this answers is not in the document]
rahul2240: @Ana06 I was not able to reuse the user_selector_input to allow user to enter email id and if it is a valid email then only it was converting to tags(while creating a booth) so that user can differentiate easily between emails.

rahul2240:
@Ana06 I forgot to ask :P, Thanks for the help, For the upcoming work that I mentioned I am creating an invitation controller (for admins only) and model also to handle things related to model. But for booths I thought it would be better if I use invitation in booths controller itself as it is the only feature that user going to have, related to invitation.

Gemfile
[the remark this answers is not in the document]
rahul2240: @Ana06 How can I use user_selector_input when I am not looking for users, I am just typing new unregistered email ids to invite.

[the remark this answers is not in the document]
differentreality: I believe Ana actually meant that you can use selectize - did you search about using it without searcing through existing data? Is it possible?

Gemfile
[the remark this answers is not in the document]
differentreality: Please update your PR as per Ana's feedback, so that we can review it.

differentreality:
@rahul2240 let's discuss a bit what you are trying to accomplish. What's the plan?
I'd expect you to provide the workflow you are trying to implement in the PR description or the issue relevant to this particular feature and PR.

The following is more of a conversation starter than a final plan.

- User submits booth request
- User invites 2 people to become booth responsibles
- Devise invitable creates 2 new user records, and sends invitation
- The new user accounts are already associated with the booth request 
(but in the end we should only associate real users with a Booth, at least upon acceptance)

app/controllers/booths_controller.rb
```diff
@@ -38,6 +56,23 @@ def update
     @url = conference_booth_path(@conference.short_title, @booth.id)
     @booth.update_attributes(booth_params)
 
+    if booth_params[:invited_users]
```
differentreality: There is too much going on here.
The same code is duplicated in 2 different places, and you are using `User.find_by(email: email)` multiple times (it should at least be a variable so that it is easier to read). 

Let's think on how can we make this more clean.

spec/factories/booths.rb
```diff
@@ -12,5 +12,6 @@
 
     submitter { create(:user) }
     responsible_ids { [create(:user).id] }
+    invited_users { 'fake@user.com, invalid_email' }
```
differentreality: Let's use `@example.com`

Gemfile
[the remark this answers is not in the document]
rahul2240: @differentreality I am planning to send small new PRs with more clear explanations and I think commits in this PR are interrelated so I will just close this one.

Gemfile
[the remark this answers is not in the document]
rahul2240: @differentreality I was working on it. Yes it is possible. After Ana suggested it, I was trying to use that. I will update the PR.

app/controllers/booths_controller.rb
rahul2240: @differentreality I will work on improving it.

spec/factories/booths.rb
rahul2240: ok :)

rahul2240:
@differentreality @Ana06 I have removed some commits to make this PR small and related to 1 thing only. I will create other PR for an unsubscribing link in the invitation email and customized invitation message.
Please review this PR again :smile:

rahul2240:
> @rahul2240 let's discuss a bit what you are trying to accomplish. What's the plan?
> I'd expect you to provide the workflow you are trying to implement in the PR description or the issue relevant to this particular feature and PR.
> 
> The following is more of a conversation starter than a final plan.
> 
> * User submits booth request
> * User invites 2 people to become booth responsibles
> * Devise invitable creates 2 new user records, and sends invitation
> * The new user accounts are already associated with the booth request
>   (but in the end we should only associate real users with a Booth, at least upon acceptance)

@differentreality User can invite any number of people, I will work on associating only confirmed users with roles, booths, conferences in separate PR.

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,8 @@ class ApplicationController < ActionController::Base
   # Ensure every controller authorizes resource or skips authorization (skip_authorization_check)
   check_authorization unless: :devise_controller?
 
+  before_action :devise_configure_permitted_parameters, if: :devise_controller?
```
rahul2240: devise_parameter_sanitizer.permit(:accept_invitation, keys: [:username])

used to add username field in the registration page for invited users otherwise only password and password confirmation fields were present.

differentreality:
> 4. Once the users will register to the app then they will be able to edit the booth for which they were invited.

How is this going to happen exactly?

In your steps you mention that you will create an unconfirmed user. How is this user going to be related to a booth?

rahul2240:
@differentreality unconfirmed user's id will be used to associate that user with the booth using

```
BoothRequest.create(booth_id: @booth.id, user_id: created_user.id,
                              role: 'responsible')
```

If a user registers then there is no problem.
If he doesn't, then invited user will not be visible to the booth owner in the responsibles field in edit booth page. This way booth owner will know that he needs other booth responsibles or reinvite the same user.

app/helpers/invite_helper.rb
```diff
@@ -0,0 +1,19 @@
+module InviteHelper
+  def invite_booth
+    if booth_params[:invited_users]
+      emails_array = booth_params[:invited_users].split(',')
+
+      emails_array.each do |email|
+        new_user = User.find_by(email: email)
+        if new_user.nil? || !new_user.confirmed?
```
differentreality: `!new_user.confirmed?` that's not bulletproof. It's possible that we have unconfirmed users (when using) but the users are actually able to log in and don't require invitation.

app/helpers/invite_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+          User.invite!({ email: email }, current_user)
+        end
+        created_user = User.find_by(email: email)
+        if created_user && @booth.responsible_ids.exclude?(created_user.id)
+          BoothRequest.create(booth_id: @booth.id, user_id: created_user.id,
```
differentreality: Are you created a new booth request, just to add a booth responsible?

rahul2240: @differentreality Whenever we add users to booth responsibles (not during invitation), a new booth request is created with same booth id but with different user ids. 
So I did the same for invited users also.

app/helpers/invite_helper.rb
rahul2240: @differentreality we can add user's "last_sign_in_at", if it is nil then we can send invitations again, is that okay?

app/helpers/invite_helper.rb
rahul2240: @differentreality Is this a wrong approach? I have confirmed that adding booth responsible creates new booth request for those users.

Gemfile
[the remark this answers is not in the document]
Ana06: I am confused. Should we review this PR @rahul2240 ?

app/helpers/invite_helper.rb
```diff
@@ -0,0 +1,19 @@
+module InviteHelper
```
Ana06: this is a module, it shouldn't be called helper

app/controllers/admin/booths_controller.rb
Ana06: > @Ana06 if a user is confirmed then we don't want to send an invitation for registration in OSEM.
But if the user is unconfirmed, then we should send an invitation again to him to join the app and become booth responsible.

a user could have created the account himself but not being confirmed, right?

Gemfile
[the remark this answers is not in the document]
Ana06: I think so

app/controllers/admin/booths_controller.rb
rahul2240: @Ana06 I have updated the condition after having a discussion on it. Please see https://github.com/openSUSE/osem/pull/2535#discussion_r295793443

Now only those users will be invited who haven't logged in to OSEM before.

Gemfile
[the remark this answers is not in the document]
rahul2240: @Ana06 Yes, please. I am resolving the conversations that are outdated.

rahul2240:
@Ana06 Thank you :smile:  I will update the code with the suggested changes.

app/helpers/invite_helper.rb
rahul2240: Ok @Ana06 updating it.

rahul2240: @Ana06 created a concern, is it ok now?

app/views/devise/invitations/edit.html.haml
```diff
@@ -0,0 +1,20 @@
+.container
```
differentreality: Where do we use this?

spec/controllers/admin/booths_controller_spec.rb
```diff
@@ -77,6 +77,14 @@
         it 'shows success message' do
           expect(flash[:notice]).to match('Booth successfully created.')
         end
+
+        it 'creates a new user on inviting booth responsible' do
+          expect(User.last.email).to eq('user@example.com')
+        end
+
+        it 'does not create a user with invalid email on inviting booth responsible' do
+          expect(User.last.email).not_to eq('example')
```
differentreality: You are using User.last in both tests, it's confusing. 
What you want to test is that User A (user@example.com) is part of booth responsibles, and User B (example) is not. Your test is **not** testing for that :/

app/helpers/invite_helper.rb
differentreality: If there is already a user record, why do we need to check anything else?

I mean there are a few ways we can go with this approach and make the feature more useful and meaningful, but what is currently being implemented is pretty basic, so I don't think we should add any more functionality than what is absolutely necessary, unless you plan on documenting (and adding tests) for the more advanced scenarios.

differentreality:
> Only confirmed users are visible in the responsibles field for new or edit booth.

Where do you see that?

rahul2240:
> > Only confirmed users are visible in the responsibles field for new or edit booth.
> 
> Where do you see that?

while adding booth responsibles in the booth, only confirmed users are available in the list due to the condition
[https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb#L135](url)

spec/controllers/admin/booths_controller_spec.rb
rahul2240: I will add a test for booth responsible also.

app/views/devise/invitations/edit.html.haml
rahul2240: When the user clicks on the invitation link in the email message then we are redirected here for registration.

differentreality:
@rahul2240 please rebase

The tests are not very thorough here :/ We should revisit them soon.

app/views/booths/_form.html.haml
```diff
@@ -12,6 +12,8 @@
           hint: 'e.g. employee, comunity manager, etc'
         = f.input :website_url
         = responsibles_selector_input f
+        = f.input :invite_responsible,
+          hint: 'This field is to invite unregistered users using email to become booth responsible.'
```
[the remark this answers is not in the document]
rahul2240: shouldn't it be stand organizer?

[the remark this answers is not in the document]
differentreality: Sure. But what's your approach in general for addressing it?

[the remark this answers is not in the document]
rahul2240: I am planning to show stand organizers instead of booth responsible

rahul2240:
@differentreality sure 

> The tests are not very thorough here :/ We should revisit them soon.

Sure.

app/views/booths/_form.html.haml
```diff
@@ -12,6 +12,8 @@
           hint: 'e.g. employee, comunity manager, etc'
         = f.input :website_url
         = responsibles_selector_input f
+        = f.input :invite_responsible,
+          hint: 'This field is to invite unregistered users using their email to become stand organizers.'
```
differentreality: I don't understand how this is going to be helpful though :/ 

People seeing this interface know about Booths, now here you mention the word stand. Isn't that confusing?

rahul2240: @differentreality true. Should I add booth responsible in brackets at the end?

differentreality: No! We need to come up with an approach for this in general. Not just this little hint here.

rahul2240: If this is for fosdem only, then can't we just rename booth to stand at every place?

differentreality: This, as in the GSoC19 branch? No, this is not necessarily just for FOSDEM. But it will definitely be if we introduce changes that make it impossible to merge it back to master!

rahul2240: @differentreality what about creating locales for booth? We can create a "naming_convention.yml" in "config/locales" and allow users to change to booth or stand.

differentreality: That sounds like a plan! 

So here we should keep, for now, something that makes sense to the rest of the wording being used, ie Booths. 

And a new PR should take care of the Stands wording wherever needed.

rahul2240: right.

differentreality:
> > > Only confirmed users are visible in the responsibles field for new or edit booth.
> > 
> > Where do you see that?
> 
> I was talking about active users 
> while adding booth responsibles in the booth, only active users are available in the list due to the condition
> [https://github.com/openSUSE/osem/blob/master/app/helpers/application_helper.rb#L135](url)

This isn't right though! Active users used to be something completely different, and now they are the ones who logged in during the last 3 months. 
This is the commit that changed the definition https://github.com/openSUSE/osem/commit/516e53d9df10a1231cde8100b8fe74a8f6e42358

We need to actually get all active (as in not disabled) users for our selection field. Do you want to submit a PR for that in master, or should I?

differentreality:
OK let's get this in as a starting point, and work on it a bit more later on to add more possibilities
- see who has been invited
- trigger re-invitation
- maybe even undo a pending invitation (or actually remove the user as booth responsible?)

Issue #2540, opened 2019-06-13
Feature tests are broken due to Chrome error


Ana06:
Since a few days we get the following error in the feature tests:

```
Selenium::WebDriver::Error::UnknownError:
  unknown error: Chrome failed to start: exited abnormally
     (unknown error: DevToolsActivePort file doesn't exist)
     (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

Check for example https://travis-ci.org/openSUSE/osem/jobs/544820037

It seems to be a credentials error:

> ERROR -- omniauth: (google) Authentication failure! invalid_credentials encountered.

This is an important issue as we can not merge new PR without ensuring the tests  are working.

@hennevogel do you know something about this?

Ana06:
@hennevogel @differentreality @bear454 anybody has any idea how to solve this?

hennevogel:
might be https://github.com/openSUSE/open-build-service/pull/7794

Ana06:
@hennevogel 

> might be openSUSE/open-build-service#7794

I don't think it is the same one. In the one you send it is realated to W3C:

> unknown command: Cannot call non W3C standard command while in W3C mode

while in OSEM we have an authentication problem:

> ERROR -- omniauth: (google) Authentication failure! invalid_credentials encountered.

@hennevogel any idea where this problem is coming from?

Ana06:
Fixed in https://github.com/openSUSE/osem/pull/2548

Issue #2528, opened 2019-06-03
Collecting additional information during submissions 


rishabhptr:
As a part of the GSoC project. We want to collect additional details such a contact, social media, additional requests for.

- [ ]  Add social media links
- [ ]  Add contact fields 
- [ ]  Add a field for entering additional requirements.
- [ ]  Change the respective form and views.

For the Social media links, I have proposed to use the contact model which already exists for the social media links and create a polymorphic association with the track and booth models.
One query regarding this is should I leave the existing association with conference model as it is or change that too?

For contact fields I think we should add a website_url for tracks just like booths. Other options such as email / social _tag are covered in contact model.

Ana06:
Where do you want to add this information? I guess you speak about the submission the tracks and booths, right?

> For the Social media links, I have proposed to use the contact model which already exists for the social media links and create a polymorphic association with the track and booth models.

This sounds good to me.

> One query regarding this is should I leave the existing association with conference model as it is or change that too?

What do you exactly mean?

> For contact fields I think we should add a website_url for tracks just like booths

:+1:

rishabhptr:
> Where do you want to add this information? I guess you speak about the submission the tracks and booths, right?

Yeah, this is related to Tracks/Booths.

> One query regarding this is should I leave the existing association with conference model as it is or change that too?

What I mean to say is that using polymorphic association we would have ` contactable_type ` and ` contactable_id ` field for contact, so is it good practice to convert the `has_one` association between conference and contact to polymorphic association too.

Ana06:
@rishabhptr what about you send a PR with a suggestion? It is much easier to discuss about code when seeing the code and we want to see the work you are doing :wink:

rishabhptr:
@Ana06 Yes, you're right and sorry for not updating earlier. I have created a PR explaining my approach and the concerns that I am facing. Thanks!

differentreality:
I am wondering if we want to allow people to submit custom things, in addition to the predefined fields, so that they can determine both the description as well as submit the equivalent value (eg. a link to project's repo).
Then the table can look like this:

ID | description | text
1  | Project license | X license
2  | Project repo | link-to-repo

rishabhptr:
It's a nice idea and I don't think it would affect the contact/social-media field and the other fields that I have mentioned in the PR.

If this is a requirement we can definitely work on it and the discuss the approach here. I am assuming that we would need a different table to implement this.

AndrewKvalheim:
[...]

PR #2541, opened 2019-06-13
Added contactable to tracks

rishabhptr:
Related to #2528 

I have tried couple of different ways to use the contact model for social media fields. I have chosem this approach so as to enable for the user to add all fields at the same time in a single form while creating the track using nested attributes.

Also, as `contact` model has a `has_one` relationship with `conference` model I need to pass the `conference_id` as it is mandatory but I believe this will break the previous mechanism where a conference had one contact which could be fetched to show on the splashpage. _**So we would need to modify this relationship as well.**_

At this point I need to decide if I should use this model or just add the social fields in the main Track/Booth model itself.

image

Apart from this other fields like :code_of_conduct_accepted, :additional_requirements etc are pretty straightforward and after this I can work on the different views related to these details.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_06_03_143107) do
+ActiveRecord::Schema.define(version: 2019_06_12_161235) do
 
   create_table "answers", force: :cascade do |t|
     t.string "title"
@@ -126,6 +126,8 @@ ActiveRecord::Schema.define(version: 2019_06_03_143107) do
     t.string "mastodon"
     t.string "youtube"
     t.string "blog"
+    t.string "contactable_type"
+    t.integer "contactable_id"
   end
 
   create_table "delayed_jobs", force: :cascade do |t|
--- a/app/controllers/tracks_controller.rb
+++ b/app/controllers/tracks_controller.rb
@@ -13,12 +13,15 @@ class TracksController < ApplicationController
 
   def new
     @track = @program.tracks.new(color: @conference.next_color_for_collection(:tracks))
+    @track.build_contact
   end
 
   def edit; end
 
   def create
     @track = @program.tracks.new(track_params)
+    @contact = @track.build_contact(track_params[:contact_attributes])
+    @contact.conference = @conference
     @track.submitter = current_user
     @track.cfp_active = false
     if @track.save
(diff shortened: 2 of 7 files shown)


differentreality:
Please do correct me if I am wrong, but I don't think we are providing an option for people to submit information such as the project's source code location or the project's license.

rishabhptr:
@differentreality The Social Media fields covers the different platforms and yes there is no specific fields for the code repository. I intended to have a **website_url**, **email** , **github/repo** fields in the track model itself and use the contact model for capture the media platforms.

But I liked the idea for project license, but should we add a special field for this?

The other information apart from these in the track submission that I propose are :
**:code_of_conduct_accepted**   `boolean`
**:additional_requirements**         `string`
**size**                                          `integer`

**:track_type**   - which can be selected from a drop down with choices much like event_types which I will cover in my next PR.

These are the details that I am planning to include in the submission from the users.

db/schema.rb
```diff
@@ -10,100 +10,103 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20181113195810) do
+ActiveRecord::Schema.define(version: 20190612161235) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
Ana06: Why is this here? In general, why does this file have that many changes?

rishabhptr: I think some unwanted changes were introduced during db migrations . I will check it again in my new PR against the GSoC19 branch.

differentreality:
@rishabhptr so what are you suggesting for the other links we want to show (email, license, etc)? Add fields in Track model? And then in Booth model?

rishabhptr:
@differentreality As of now yes, any specific field like email, website/license which we can identify would go in the track/booth model itself.

Ana06:
@differentreality can we get this merged or are there still concerns?

@rishabh92 can you please rebase? there are some commits that shouldn't be here.

We should fix https://github.com/openSUSE/osem/issues/2540 because is blocking the PRs as well.

differentreality:
@rishabhptr please keep an eye out on Travis and fix the linters failures, we cannot merge ortherwise!

rishabhptr:
@differentreality I have fixed the linters, but the feature test were failing due to the webdriver issue which is now resolved. Maybe we should bring in the latest changes to the GSoC19 branch?

differentreality:
> @differentreality can we get this merged or are there still concerns?

@Ana06 not concerns per se, but we will probably need to use the same fields for Booth as well.

differentreality:
Screenshots would be helpful here too :wink:

differentreality:
@rishabhptr please rebase

rishabhptr:
@differentreality Yes, the form is quite longer now. Maybe we could ask for the other fields at a later stage.

Right now when the track is accepted and then confirmed by the submitter, the user is able to manage his track but he cannot edit the details as it make sense to not change them after submission.

However, we can include the option to add fields such as social media in a separate tab in the show page. So that the user can update these details there.

Screenshot from 2019-08-09 18-39-40
Screenshot from 2019-08-09 18-38-46

PR #2546, opened 2019-06-22
[WIP] Added track type to admin dashboard

rishabhptr:
This PR includes the changes to add the ability to create the track_types in the admin dashboard as an conference can have different types of tracks or essential tracks like devrooms. This would be later included in submission form.

Taking the example for FOSDEM, these types can be whether you are requesting a devroom for 1day/2day or half day. But this wont be same for other conferences and they might have their own requirements, this way it helps to keep this feature reusable which can be configured as per need.

This PR only covers the admin dashboard changes to create the track types, I will add these as a dropdown in the submission form along with other new fields in a separate PR.

Screenshot from 2019-07-06 14-56-58

Screenshot from 2019-07-06 14-57-10

### DIFF
--- a/app/controllers/admin/track_types_controller.rb
+++ b/app/controllers/admin/track_types_controller.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module Admin
+  class TrackTypesController < Admin::BaseController
+    load_and_authorize_resource :conference, find_by: :short_title
+    load_and_authorize_resource :program, through: :conference, singleton: true
+    load_and_authorize_resource :track_type, through: :program
+
+    def index; end
+
+    def edit; end
+
+    def new
+      @track_type = @conference.program.track_types.new
+    end
+
+    def create
+      @track_type = @conference.program.track_types.new(track_type_params)
+      if @track_type.save
+        redirect_to admin_conference_program_track_types_path(conference_id: @conference),
+                    notice: 'Track type successfully created.'
+      else
+        flash.now[:error] = "Creating track type failed: #{@track_type.errors.full_messages.join('. ')}."
+        render :new
+      end
+    end
+
+    def update
+      if @track_type.update_attributes(track_type_params)
+        redirect_to admin_conference_program_track_types_path(conference_id: @conference),
+                    notice: 'Track type successfully updated.'
+      else
+        flash.now[:error] = "Update track type failed: #{@track_type.errors.full_messages.join('. ')}."
+        render :edit
+      end
+    end
+
+    def destroy
+      if @track_type.destroy
+        redirect_to admin_conference_program_track_types_path(conference_id: @conference),
+                    notice: 'Track type successfully deleted.'
+      else
+        redirect_to admin_conference_program_track_types_path(conference_id: @conference),
+                    error: 'Destroying track type failed! '\
+                    "#{@track_type.errors.full_messages.join('. ')}."
+      end
+    end
+
+    private
+
+    def track_type_params
+      params.require(:track_type).permit(:title, :description)
+    end
+  end
+end
--- a/app/models/track_type.rb
+++ b/app/models/track_type.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class TrackType < ApplicationRecord
+  belongs_to :program, touch: true
+
+  has_paper_trail meta:   { conference_id: :conference_id },
+                  ignore: %i[updated_at]
+
+  validates :title, presence: true
+
+  private
+
+  def conference_id
+    program.conference_id
+  end
+end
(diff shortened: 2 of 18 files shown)


rishabhptr:
For conferences such as FOSDEM we have special essential tracks called devrooms which are self-organised and managed, and then we have main track like hardware, software etc.

- Call for Participation would also differ in such cases as after a devroom is accepted, it's up to them to organise their own CfP and make decisions. 

Therefore having such a field in submission would ensure if any group wants to submit their own devroom or the submission is for a new entry as a main track for which the conference should look out for proposals.

Having the admin add the these types and not hardcoding gives flexibity as user can decide types based on own use case depending on factors such as scale/ resources / general ideas of the conference.

app/controllers/admin/track_types_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+                    notice: 'Track type successfully created.'
+      else
+        flash.now[:error] = "Creating track type failed: #{@track_type.errors.full_messages.join('. ')}."
+        render :new
+      end
+    end
+
+    def update
+      if @track_type.update_attributes(track_type_params)
+        redirect_to admin_conference_program_track_types_path(conference_id: @conference.short_title),
```
Ana06: [nitpick] As `@conference` has a `to_param` method the `.short_title` shouldn't be needed.

app/views/admin/track_types/_form.html.haml
```diff
@@ -0,0 +1,15 @@
...
+  .col-md-12
+    .page-header
+      %h1
+        - if @track_type.new_record?
+          New
+        Track Type
+        = @track_type.title
+.row
+  .col-md-12
+    = semantic_form_for(@track_type, url: (@track_type.new_record? ? admin_conference_program_track_types_path : admin_conference_program_track_type_path(@conference.short_title, @track_type))) do |f|
```
Ana06: [nitpick] same here with `.short_title`

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -78,6 +78,9 @@
           - if can? :update, @conference.program.event_types.build
             %li{class: active_nav_li(admin_conference_program_event_types_path(@conference.short_title))}
               = link_to 'Event Types', admin_conference_program_event_types_path(@conference.short_title)
+          - if can? :update, @conference.program.event_types.build
```
Ana06: this is the same `if` as few lines up

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -78,6 +78,9 @@
...
+            %li{class: active_nav_li(admin_conference_program_track_types_path(@conference.short_title))}
```
Ana06: [nitpick] same here with `.short_title`

app/views/layouts/_admin_sidebar.html.haml
rishabhptr: Sorry, this required a new entry in admin_ability.rb , I must have forgot to update it :sweat_smile:. Thanks! :smile:

app/views/admin/track_types/_form.html.haml
rishabhptr: Sure, Thanks!

rishabhptr:
Not sure what would be the best approach to solve this cop issue :

> app/models/admin_ability.rb:105:3: C: Metrics/AbcSize: Assignment Branch Condition size for signed_in_with_organizer_role is too high. [70.52/70]
>   def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = []) ...

app/models/track_type.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class TrackType < ApplicationRecord
+  belongs_to :program, touch: true
+
+  has_paper_trail meta:   { conference_id: :conference_id },
+                  ignore: %i[updated_at]
+
+  validates :title, presence: true
+  validates :description, presence: true
```
differentreality: I wonder if this is truly necessary.

differentreality:
We need some extra attributes (categories) for tracks and booths. 
The example in the screenshot is quite misleading. We don't need to define the category/type of 'Main tracks', we need the user to choose some additional information, eg
- If a devroom (track) can be of half day, full day, or 2 days
- If a stand is going to need 1 or 2 tables

Obviously these are some examples specifically for FOSDEM while other conferences may need different categories there.

Looks like this approach is creating track_types, and afterwards booth_types.
Seems to me we are duplicating quite a bit :/

app/models/track_type.rb
rishabhptr: Yes, in certain cases the when the type is straight forward then the title itself will be sufficient.

Issue #2086, opened 2018-07-05
Ticket currencies


MysterAitch:
**I'm submitting a ..**
- Bug Report

**Current behaviour:**
Only one currency is permitted per conference. When creating/editing tickets, the option to select alternative currencies is still visible/available.

**Expected correct behaviour:**
If the single currency per conference is to be enforced, then it should be configured on a per-conference rather than per-ticket basis. If currencies are shown on the ticket configuration page, then it should be a read-only/non-editable form field.

Alternatively, support for multiple currency tickets per conference should be available. 

**Steps to reproduce:**
Attempt to create a ticket with a currency that is not the same as existing tickets.

vishaldeepak:
Interested in giving this a shot. Expected Correct behaviour has two ways, which one is  the preferred solution?

differentreality:
This is marked as a good first issue for the option of disabling the ticket currency selection.

Enabling multiple currencies per conference (possibly with conversion) is a bigger issue.

So for the time being, I would suggest you merely disable the currency selection when the currency is already decided.

ed1001:
Hiya, this would be my first ever contribution :) May I take this on?

differentreality:
Sure @ed1001 go ahead :+1:

Ana06:
I do not think we should support several currencies for a conference. OSEM is already complex enough, so I would prefer keeping things simple.

Regarding the single currency per conference, this shouldn't be managed in the tickets. As all the tickets should have the same currency, this should be an attribute of conference. So, the correct approach is in my opinion:

1. Add a currency attribute to conference
2. Allow to set this attribute in the Basics page of the conference (admin page)
3. create a data rake task to set this new attribute to the tickets currency (it should be the same in all the ticket but I would double check this in the rake task - also don't forget to document this rake task for the updates)
4. remove the currency attribute from the ticket

And this may not be a `good first issue`.

PR #2547, opened 2019-06-29
Ed1001 ticket currency select patch

ed1001:
**Checklist**

- [ ] The tests pass locally with my changes. (I'm not sure about this as even the untouched project has failures, so I assume 100% pass is not expected? I may need guidance on this one)

- Ticket currencies - #2086

- Ticket price currency select will behave such that if there is already a ticket created, any new tickets will no longer have the option to select a different currency. If there is only one ticket created I have left the option to edit the currency for that ticket (not sure if this is desired, perhaps could disable if sold tickets > 0). I feel like this part of the code could be 'DRYer' but I couldn't seem to make it so, apologies but I have no experience with haml and couldn't find something suitable by googling, if there's a better way I'd be glad to know. Thanks

### DIFF
--- a/app/views/admin/tickets/_form.html.haml
+++ b/app/views/admin/tickets/_form.html.haml
@@ -12,7 +12,10 @@
       = f.input :title, input_html: { autofocus: true }
       = f.input :description, input_html: { rows: 5, data: { provide: 'markdown-editable' } }
       = f.input :price
-      = f.input :price_currency, as: :select, class: 'form-control', collection: ['USD', 'EUR', 'GBP', 'INR', 'CNY', 'CHF'], include_blank: false
+      - if (@conference.tickets.count >= 1 && @ticket.new_record?) || @conference.tickets.count > 1
+        = f.input :price_currency, as: :select, class: 'form-control', collection: [@conference.tickets.first.price_currency], include_blank: false
+      - else
+        = f.input :price_currency, as: :select, class: 'form-control', collection: ['USD', 'EUR', 'GBP', 'INR', 'CNY', 'CHF'], include_blank: false
       = f.input :registration_ticket, hint: 'A registration ticket is with which user register for the conference.'
       %p.text-right
         = f.action :submit, as: :button, button_html: { class: 'btn btn-primary' }
--- a/docker-compose.override.yml.example
+++ b/docker-compose.override.yml.example
@@ -3,4 +3,4 @@ services:
   osem:
     build:
       args:
-        CONTAINER_USERID: 501
+        CONTAINER_USERID: 13042
--- a/docker-compose.override.yml.example
+++ b/docker-compose.override.yml.example
@@ -3,4 +3,4 @@ services:
   osem:
     build:
       args:
-        CONTAINER_USERID: 13042
+        CONTAINER_USERID: 501


app/views/admin/tickets/_form.html.haml
```diff
@@ -12,7 +12,10 @@
       = f.input :title, input_html: { autofocus: true }
       = f.input :description, input_html: { rows: 5, data: { provide: 'markdown-editable' } }
       = f.input :price
-      = f.input :price_currency, as: :select, class: 'form-control', collection: ['USD', 'EUR', 'GBP', 'INR', 'CNY', 'CHF'], include_blank: false
+      - if (@conference.tickets.count >= 1 && @ticket.new_record?) || @conference.tickets.count > 1
+        = f.input :price_currency, as: :select, class: 'form-control', collection: [@conference.tickets.first.price_currency], include_blank: false
```
Ana06: this shouldn't be a select with a collection, as you only have one. It should be just shown.

app/views/admin/tickets/_form.html.haml
```diff
@@ -12,7 +12,10 @@
       = f.input :title, input_html: { autofocus: true }
       = f.input :description, input_html: { rows: 5, data: { provide: 'markdown-editable' } }
       = f.input :price
-      = f.input :price_currency, as: :select, class: 'form-control', collection: ['USD', 'EUR', 'GBP', 'INR', 'CNY', 'CHF'], include_blank: false
+      - if (@conference.tickets.count >= 1 && @ticket.new_record?) || @conference.tickets.count > 1
```
Ana06: I am not sure if I follow this condition, why isn't it enough with `@conference.tickets.count > 1` `count` raises a query to the database so new records doesn't show up.

ed1001: With only @conference.tickets.count > 1 the second added ticket would still be able to change the currency. @conference.tickets.count >= 1 && @ticket.new_record? makes sure that you can still change currency if you're editing the first ticket (I went on the assumption that was desired, but now I see it isn't).

app/views/admin/tickets/_form.html.haml
ed1001: I had trouble with form submission without keeping the input. Thinking about it now I guess there's probably an easy way to add a read only input.

ed1001:
Hi @Ana06 ! 

Thanks a lot for the feedback :)

I'm certainly up for the more extensive changes you proposed. As this is a slightly larger task I'd like to run a couple of things by you if you don't mind.

 - Remove line 11 (monetize) from ticket.rb and add to conference.rb
 - Add price_currency field to conferences and remove from tickets table
 - change all refs to currency from 'ticket.price_currency' => 'ticket.conference.price_currency'
 - change in views (conference and ticket forms)

With regards to your 3rd suggestion:

>  3. create a data rake task to set this new attribute to the tickets currency (it should be the same in all the ticket but I would double check this in the rake task - also don't forget to document this rake task for the updates)

I don't understand why this is necessary, or how even possible if the price_currency is removed from ticket. If the price_currency is always referenced through conference then the conference's tickets must always share that price_currency. Apologies for my naivety, I'm still finding my feet with Rails :)

ed1001:
Actually maybe best to keep the price_currency in tickets table and have the conference's currency attribute dictate the tickets' currency at creation. Then can keep things like `humanized_money_with_symbol ticket.price`
in the views

ed1001:
Hi @Ana06 ! 

Thanks a lot for the feedback :)

I'm certainly up for the more extensive changes you proposed. As this is a slightly larger task I'd like to run a couple of things by you if you don't mind. 

 - Just to make sure, by `4. remove the currency attribute from the ticket` I guess you mean remove it from the view, not remove the attribute from the ticket model, is that correct?
 - Is the data rake task primarily for checking? I assume the ticket's currency will be set at creation via tickets_controller.
 - As for documenting the rake task, would the details go in README_FOR_APP?

Ana06:
@ed1001 

> Just to make sure, by 4. remove the currency attribute from the ticket I guess you mean remove it from the view, not remove the attribute from the ticket model, is that correct?

No, I mean removing it from the model, as it is already in the conference. It should be still rendered in the view (but you can do this as you have it in the conference).

> Is the data rake task primarily for checking? I assume the ticket's currency will be set at creation via tickets_controller.

The rake task is to set the new attribute in conference for old conferences, as it is empty for new conferences. But as we have it in the first ticket, we can just copy over this information to conference, before getting rid of the the info in the ticket.

> As for documenting the rake task, would the details go in README_FOR_APP?

In https://github.com/openSUSE/osem/blob/master/CHANGES.md as this is only for updating old instances. New instances don't need this. :wink: 

As this is a bigger feature, I recommend you made independent commits for the steps I proposed (step 4 could be done in a different PR as well, so that it doesn't get too big). And of course, If you have doubts/help, do not hesitate to ask :wink:

ed1001:
Thanks for the clarification @Ana06, I will get on it!

Issue #2555, opened 2019-07-18
Streamline Conference public p


esparta:
**I'm submitting a ..**
- [ ] Bug Report
- [ ] Feature Request
- [x] Work in Progress tracker

**Current behavior:**

The root page of OSEM usually have heavy traffic when conferences are announced, this issue is to keep track about that can be done in order to improve the system in general, and the Conference controller, model and views in particular[1]:

- [x] Remove not needed query on not signed users (Fixed on #2551)
- [ ] Use Conference's `:upcoming` & `:past` scopes on `Conference#index` (DRY)
- [ ] Promote New installation to be their own page instead of partial for `Conference#index`
- [ ] Split antiqued and current conferences views
- [ ] Create partials for signed and not signed users (avoid  multiple` if current_user & current_user.feature_x`)
- [ ] Use delegation on Conference -> Venue (avoid `conference.venue.city`)
- [ ] Use delegation on Conference -> Program -> Events
- [ ] Use delegation on Conference -> SplashPage
- [ ] Refactor `Conference#show`, using less ruby and more ActiveRecord.

**Other information:**

If maintainer consider necessary, it can be converted to a Project

PR #2556, opened 2019-07-18
Use New Install within their own template

esparta:
It used to be a partial inside #index, but having their own page
yields some extra benefits:

- Performance: Move logic from the view to the controller
  - Theres' no need to go extra steps inside the :index view
- UI: Remove the extra 'Upcoming Conferences' when the system needs
  to start the flow of the creation of the first user

**Checklist**

- [X] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).

Part of the issue #2555 "Streamline Conference public page"

If accepted, the UI will change from this:

20190718_osem_before

To this:

20190718_osem_after

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -8,6 +8,9 @@ class ConferencesController < ApplicationController
   def index
     @current    = Conference.where('end_date >= ?', Date.current).reorder(start_date: :asc)
     @antiquated = Conference.where('end_date < ?', Date.current)
+    if @antiquated.empty? && @current.empty? && User.empty?
+      render :new_install
+    end
   end
 
   def show
--- a/app/views/conferences/index.html.haml
+++ b/app/views/conferences/index.html.haml
@@ -18,8 +18,6 @@
     #antiquated.collapse
       - @antiquated.each do |conference|
         = render '/conferences/conference_details', conference: conference
-  - if @antiquated.empty? && @current.empty? && User.empty?
-    = render partial: 'new_install'
 
 -content_for :script_body do
   :javascript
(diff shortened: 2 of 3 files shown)


esparta:
I just realize the new test it is not being executed because has a new route: `spec/requests`
I can move/rename the new test into the controller folder or modify travis.sh to add the new folder. I'd prefer the last because it would be interesting to migrate the 'soft-deprecated' controller test to request test.

hennevogel:
>  I'd prefer the last

Me too

esparta:
Ok, I'll push another commit adding the request test to the CI suite.

esparta:
I think is ready to be reviewd, @hennevogel, so whenever you had time it's ok (no rush, at all)
More PR a coming :)

travis_script.sh
```diff
@@ -22,13 +22,17 @@ case $TEST_SUITE in
   features)
     bundle exec rspec --format documentation spec/features
     ;;
+  requests)
```
hennevogel: That's a bit much setup just for one test. Please also run this in the 'controllers' section below. Like

`bundle exec rspec --format documentation spec/controllers spec/requests`

esparta: little late, but done :)
thanks!

esparta:
Made a rebase because my branch was too old and triggered Hakiri warnings
Is again ready to be reviewed, @hennevogel 

Thanks!

hennevogel (approved):

esparta:
Ohhh thanks @hennevogel !

hennevogel:
better late than never :cowboy_hat_face:

PR #2563, opened 2019-07-22
Add alias for booth 

rahul2240:
"Booth" can be replaced with "Stand" which is used by FOSDEM.

### DIFF
--- a/app/views/admin/booths/index.html.haml
+++ b/app/views/admin/booths/index.html.haml
@@ -2,7 +2,7 @@
   .col-md-12
     .page-header
       %h1
-        Booths
+        #{(t'booth').pluralize.capitalize }
         .pull-right
           - if can? :read, Booth
             .btn-group
@@ -11,8 +11,8 @@
                   Export PDF
                   %span.caret
                 %ul.dropdown-menu{ role: 'menu' }
-                  %li= link_to 'All Booths', admin_conference_booths_path(@conference.short_title, format: :pdf, booth_export_option: 'all')
-                  %li= link_to 'Confirmed Booths', admin_conference_booths_path(@conference.short_title, format: :pdf, booth_export_option: 'confirmed')
+                  %li= link_to "All #{(t'booth').pluralize.capitalize }", admin_conference_booths_path(@conference.short_title, format: :pdf, booth_export_option: 'all')
+                  %li= link_to "Confirmed #{(t'booth').pluralize.capitalize }", admin_conference_booths_path(@conference.short_title, format: :pdf, booth_export_option: 'confirmed')
               .btn-group
                 %button.btn.btn-default.dropdown-toggle{ 'data-toggle' => 'dropdown', type: 'button', class: 'btn btn-success' }
                   Export CSV
@@ -29,7 +29,7 @@
                   %li= link_to 'Confirmed', admin_conference_booths_path(@conference.short_title, format: :xlsx, booth_export_option: 'confirmed')
         = "(#{@booths.length})" if @booths.any?
       %p.text-muted
-        All the booth requests
+        All the #{t'booth'} requests
 
 .row
   .col-md-12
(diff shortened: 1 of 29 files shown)


sbernhard:
Nice addition. I guess this feature would be nice to have for other strings, too. Like "conference" -> its maybe not a conference but a training, workshop or just "event".

differentreality:
@hennevogel @Ana06 @bear454 or anyone else, would you like to take a look at this for use in our master?

Ana06:
@sbernhard 

> Nice addition. I guess this feature would be nice to have for other strings, too. Like "conference" -> its maybe not a conference but a training, workshop or just "event".

In this PR, this is configured at the instance level and only for booths. I think for your comment that we you mean is not same. Feel free to open an issue with a feature request is that is the case. :wink:

Ana06:
@differentreality can we merge this for the gsoc branch already?

app/views/admin/booths/index.html.haml
```diff
@@ -2,7 +2,7 @@
   .col-md-12
     .page-header
       %h1
-        Booths
+        #{t'Booth'}s
```
differentreality: Let's introduce the plural in the translation file please

app/views/layouts/_admin_sidebar.html.haml
```diff
@@ -120,7 +120,7 @@
     %li
       = link_to  admin_conference_booths_path(@conference.short_title) do
         %span.fa.fa-shopping-bag
-        Booths
+        = t'Booth'
```
differentreality: User plural translation key

app/views/admin/booths/index.html.haml
rahul2240: alright!

Ana06: I18n supports plurals, please check: https://guides.rubyonrails.org/i18n.html

rahul2240:
@differentreality @Ana06 I have updated the PR so that pluralization can be used easily.
Like this: 
`t'booth', count: 2` -> for "booths"
`t'booth', count: 1 `-> for "booth"
I have used `t 'booth.one' ` and `t 'booth.other'` because it was obvious in most of the places to use "booth" or "booths".

sbernhard:
There is a easier way to handle singular, plural: https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize

rahul2240:
@sbernhard ty :) but I wanted to keep everything related to locale variable in locales itself. I followed the standard [https://guides.rubyonrails.org/i18n.html#pluralization](url) to make it more easily adjusting for different grammars rules.

config/locales/alias.en.yml
```diff
@@ -0,0 +1,10 @@
+en:
+  booth:
+    one: 'stand'
+    other: 'stands'
+  Booth:
```
Ana06: this is the same capitalized. I would use in the ruby code [capitalize](https://ruby-doc.org/core-2.6/String.html#method-i-capitalize) instead, so that when modifying this file there are no repeated entries.

rahul2240: @Ana06 makes sense :sweat_smile:, I will use capitalize.

config/locales/alias.en.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  booth:
+    one: 'stand'
+    other: 'stands'
```
hennevogel: Check out the Ruby on Rails string method [pluralize](https://api.rubyonrails.org/classes/String.html#method-i-pluralize). No need to have 2 translations just for pluralized strings.

differentreality:
@rahul2240 ping

config/locales/alias.en.yml
rahul2240: @hennevogel Alright! I got confused with rails pluralize and i18n pluralize. But I think using rails pluralize makes sense as we don't need a count to apply different translation for "stand".

rahul2240:
@differentreality Yes Sorry. On it :+1:

rahul2240:
@Ana06 on it.

PR #2569, opened 2019-07-22
Add comment functionality to tracks.

rishabhptr:
The PR adds the comment functionality to the track model.

We are already using `act_as_commentable_with_threading`  in our project for events. So there were some minor changes required where we need to check the commentable type as now event would'nt be the only type.

Likewise some changes were required in the mailer for different templates for the two cases.

Screenshot from 2019-07-23 11-11-56
Screenshot from 2019-07-23 11-12-39
Screenshot from 2019-07-23 11-12-46


db/migrate/20190720062321_add_comments_count_to_tracks.rb
```diff
@@ -0,0 +1,10 @@
+class AddCommentsCountToTracks < ActiveRecord::Migration[5.1]
+  def change
+    add_column :tracks, :comments_count, :integer, default: 0, null: false
+
+    Track.find_each do |track|
+      comments_count = track.comment_threads.count
```
differentreality: Ehm, will any track have comment_threads? Why? :thinking:

hennevogel:
No respone, closing.

PR #2573, opened 2019-08-04
show invited users in invite responsible field

rahul2240:
This PR adds the following features -
1. Invited users who have not joined the app will be shown in invite responsible field.
2. The submitter will be able to remove the invited user from the edit page.

**Showing invited users in invite responsible field**
showinvited

### DIFF
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -50,6 +50,7 @@ module Admin
 
     def edit
       @url = admin_conference_booth_path(@conference.short_title, @booth.id)
+      @invitation_pending_emails = @booth.invited_responsibles.pluck(:email).join(',')
     end
 
     def update
--- a/app/controllers/booths_controller.rb
+++ b/app/controllers/booths_controller.rb
@@ -34,6 +34,7 @@ class BoothsController < ApplicationController
 
   def edit
     @url = conference_booth_path(@conference.short_title, @booth.id)
+    @invitation_pending_emails = @booth.invited_responsibles.pluck(:email).join(',')
   end
 
   def update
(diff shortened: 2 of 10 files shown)


differentreality:
Hmmm I just realized it is easy to think you have added an email to invite, when you actually haven't. Wonder if we could take care of that. 

On a different note, booths#show could also show the invited users?

rahul2240:
I will think about that and yes we can show invited users in booths#show.

differentreality:
What does the admin see about such a booth?

rahul2240:
Sorry I didn't get you.

differentreality:
I meant, please add a screenshot for the organizers point of view.
Where do the invited users show up for the organizer (in admin side of the application)?

rahul2240:
Sorry for replying late. Right now! we are showing responsibles in booths#show and booths#edit. For the admins, I will create a new page that will render all invited responsibles.

app/controllers/admin/booths_controller.rb
```diff
@@ -50,6 +50,7 @@ def create
 
     def edit
       @url = admin_conference_booth_path(@conference.short_title, @booth.id)
+      @invitation_pending_emails = @booth.responsibles.where(last_sign_in_at: nil).pluck(:email).join(',')
```
differentreality: Should we extract this to a booth method `invited_responsibles` or something like that?

differentreality:
Yeah it would be useful for organizer to know what's going on as well. 
Other than that, not sure why this is still a WIP?

app/controllers/admin/booths_controller.rb
rahul2240: Yes, we can do that.

rahul2240:
@differentreality I have made the changes. Please review.

PR #2577, opened 2019-08-10
Add vote functionality to tracks

rishabhptr:
To help better assess track submission while working as a team it is important to have a means by which anyone involved in this process can indicate his or her preference regarding any submission.

Most often than not, there will be more submission than what can be accepted and allowing this feature for Tracks/Booths along with events helps to prioritize these submissions which would make decision making easier.

- Changed the association between `Vote` and `Event` to polymorphic and added an association between `Vote` and `Track`. 
- Added the required partials for voting and reused the event_helpers for rating stars and tooltip

Screenshot from 2019-08-10 18-56-38
Screenshot from 2019-08-10 18-56-56

.

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -47,8 +47,8 @@ module Admin
       @versions = @event.versions |
                   PaperTrail::Version.where(item_type: 'Commercial').where('object LIKE ?', "%commercialable_id: #{@event.id}\ncommercialable_type: Event%") |
                   PaperTrail::Version.where(item_type: 'Commercial').where('object_changes LIKE ?', "%commercialable_id:\n- \n- #{@event.id}\ncommercialable_type:\n- \n- Event%") |
-                  PaperTrail::Version.where(item_type: 'Vote').where('object_changes LIKE ?', "%\nevent_id:\n- \n- #{@event.id}\n%") |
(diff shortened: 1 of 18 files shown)


app/controllers/admin/events_controller.rb
```diff
@@ -141,7 +141,7 @@ def restart
     def vote
       @votes = @event.votes.includes(:user)
 
-      if (votes = current_user.votes.find_by_event_id(params[:id]))
+      if (votes = current_user.votes.find_by(votable_type: 'Event', votable_id: params[:id]))
```
differentreality: just `votable: @event` should also work, here and in other places where you query in a similar way. Give it a shot and let me know :)

differentreality:
Thank! Feel free to squash your commits and rebase, to prepare this for merging :tada:  
We should bring this to Booths as well!

app/controllers/admin/events_controller.rb
rishabhptr: Yes it does, thanks :smile:

rishabhptr:
@differentreality done rebase :+1:

app/models/track.rb
```diff
@@ -100,6 +102,25 @@ def to_param
     short_name
   end
 
+  def user_rating(user)
```
differentreality: These methods are used in both Track and Event models. Should we extract them to avoid duplication?

hennevogel:

PR #2581, opened 2019-08-16
Add additional fields to booth submission.

rishabhptr:
This PR adds some additional fields in Booth/Stand submission form:
`email` : This is a public email for the stand.

`special_requirements` : This is a optional fields in case the submitter want to indicate any additional requirements that he may have such as a printer etc.

`accepted_code_of_conduct` : This is a boolean field, incase the organisation has a code of conduct it is necessary for the submitter to read and agree to it before make his submission.

Also the social media fields( perhaps also the source location/ license) would also be included in the form which the user will be able to fill upon acceptance.

image

image

### DIFF
--- a/app/models/booth.rb
+++ b/app/models/booth.rb
@@ -25,9 +25,14 @@ class Booth < ApplicationRecord
             :state,
             :responsibles,
             :conference_id,
+            :email,
             :website_url,
             :submitter_relationship,
             presence: true
+
+  validates :accepted_code_of_conduct, acceptance: {
+    if: -> { conference.try(:code_of_conduct).present? }
+  }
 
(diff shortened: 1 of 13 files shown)


app/models/booth.rb
```diff
@@ -25,10 +25,15 @@ class Booth < ApplicationRecord
             :state,
             :responsibles,
             :conference_id,
+            :email,
```
differentreality: Maybe we do not want to make this mandatory? Let me know your thoughts!

rishabhptr: @differentreality Yes, it makes sense maybe the booth organizers would prefer the communication to happen through a public channel only like twitter etc..

hennevogel:

PR #2583, opened 2019-08-17
Add booth types to booths

rishabhptr:
In certain cases we need a type field for stands which could be dropdown from which the user can choose.
In case of FOSDEM it could be 'one table' or 'two table', these may not be the same for all conferences and may even change from year to year therefore it should be customisable.

Screenshots:
Screenshot from 2019-08-17 18-21-04
Screenshot from 2019-08-17 18-21-11
Screenshot from 2019-08-17 18-21-31
Screenshot from 2019-08-17 18-23-46

### DIFF
--- a/app/views/admin/booths/new.html.haml
+++ b/app/views/admin/booths/new.html.haml
@@ -1,3 +1,3 @@
 %h1 New #{t'booth'}
 
-= render 'booths/form'
+= render 'booths/form', conference: @conference
--- a/app/views/booths/_form.html.haml
+++ b/app/views/booths/_form.html.haml
@@ -13,7 +13,7 @@
 
         = f.input :booth_type_id, as: :select,
           collection: conference.program.booth_types.map {|type| ["#{type.title}", type.id]},
-          include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' } if conference.program.booth_types.any?
+          include_blank: 'Please select a type ...', label: 'Type', input_html: { class: 'select-help-toggle' } if conference.program.booth_types.any?
 
         - conference.program.booth_types.each do |booth_type|
           %span{ class: 'help-block select-help-text booth_booth_type_id collapse', id: "#{booth_type.id}-help" }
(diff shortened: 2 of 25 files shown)


app/views/booths/_form.html.haml
```diff
@@ -10,6 +10,14 @@
         = f.input :submitter_relationship, input_html: { rows: 5, data: { provide: 'markdown-editable' } }, required: true,
           label: 'Submitter\'s relation',
           hint: 'e.g. employee, comunity manager, etc'
+
+        = f.input :booth_type_id, as: :select,
+          collection: conference.program.booth_types.map {|type| ["#{type.title}", type.id]},
+          include_blank: false, label: 'Type', input_html: { class: 'select-help-toggle' } if conference.program.booth_types.any?
```
differentreality: Perhaps it doesn't always make sense to enforce the selection of the type. I'd suggest we do include the blank option with a relevant message (Please select, or sth similar to our other forms)

app/views/admin/booths/new.html.haml
```diff
@@ -1,3 +1,3 @@
 %h1 New #{t'booth'}
 
-= render 'booths/form'
+= render 'booths/form', conference: @conference
```
differentreality: Why do we need to pass the conference variable and not use @conference inside the form?

app/views/booths/_form.html.haml
```diff
@@ -10,6 +10,14 @@
...
+
+        - conference.program.booth_types.each do |booth_type|
```
differentreality: Can you show me what you expect to show with this? Perhaps it makes sense for the type description to be a hint under the input field?

app/views/admin/booths/new.html.haml
rishabhptr: @differentreality Actually [haml-lint](https://github.com/sds/haml-lint) keeps complaining about using instance variable in partials. This has been issue in my other pr's too, maybe I should go ahead and generate a to-do file for all these changes?

app/views/booths/_form.html.haml
rishabhptr: @differentreality sure :+1:

app/views/booths/_form.html.haml
rishabhptr: That is what happens currently, whatever type is chosen its description appears as a hint below the field .
image

differentreality:
Looking at this again with fresher eyes, I am wondering if it makes more sense for this feature to be integrated as a survey. 

We don't currently have an option to show a survey while creating a booth though.

PR #2584, opened 2019-08-18
Combine and evaluate booth submissions

### LINKED ISSUE #2582: Combine/link and evaluate Stands submissions
In certain cases we want to combine two or more stand/booth requests together so that we can accept them for a single table/stand.

The user would submit independent requests and we may be able to link them together to a single table.

The Approach could be as:
1. Create a new model for combined booth request which would have a has_many relationship with booth model.
2. Admin/Organizer would visit booth/index.
3. The will be two tables, the first one will be the default and the second one for the combined requests.
4. User will be able to add entries in second table by clicking on combine request button and selecting the required booths and entering any other field.
5. These combined booths now will have a single state machine, and when accepted will only consume 1 entry from the booth_limit.

rishabhptr:
Fixes #2582

### DIFF
--- a/app/controllers/admin/booth_groups_controller.rb
+++ b/app/controllers/admin/booth_groups_controller.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+module Admin
+  class BoothGroupsController < Admin::BaseController
+    load_and_authorize_resource :conference, find_by: :short_title
+    load_and_authorize_resource :program, through: :conference, singleton: true
+    load_and_authorize_resource :booth_group, through: :program
+
+    def index; end
+
+    def edit; end
+
+    def new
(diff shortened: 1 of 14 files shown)


differentreality:
A note: When publicly showing the available booths of the conference, combined booths should show the group name instead of the individual booths.

rishabhptr:
@differentreality Sure, I will make the change such that after being accepted only the group name of all the combined booth are shown in the splashpage instead of individual entries.

differentreality:
@rishabhptr could you please confirm if this is still WIP?

rishabhptr:
@differentreality This PR currently does the job of introducing booth groups, what we need now is how the state of all included booths will change together( maybe a separate state machine for the booth_group) and also how does this go with the bulk acceptance feature that we plan on having.
But maybe this can be handled in a separate PR and we can discuss the approach in the issue.

hennevogel:

PR #2575, opened 2019-08-06
Booth states can be changed in bulk

rahul2240:
closes #2574 

**Checklist**

- [x] I have read the [Contribution & Best practices Guide](https://github.com/openSUSE/osem/blob/master/CONTRIBUTING.md).
- [x] My branch is up-to-date with the upstream `master` branch.
- [x] The tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works(if appropriate).
- [x] I have added necessary documentation (if appropriate).

**workflow**
1. Admin visit the booths#index page.
2. Select any booth from the select column.
3. Go to top select bar and select the available options. 
4. Click on the option and all the booths with the same state as that of selected booth will be changed to new state.

**Changing state together**

ezgif com-video-to-gif

PR #2586, opened 2019-08-23
add condition for booth rejection mail

rahul2240:
Add a condition to send email only when "send_on_booths_rejection" is true. At present, an email is sent whenever a booth is rejected not depending on whether "send_on_booths_rejection" is true or false.

### DIFF
--- a/app/controllers/admin/booths_controller.rb
+++ b/app/controllers/admin/booths_controller.rb
@@ -92,7 +92,9 @@ module Admin
       @booth.reject!
 
       if @booth.save
-        Mailbot.conference_booths_rejection_mail(@booth).deliver_later
+        if @conference.email_settings.send_on_booths_rejection
+          Mailbot.conference_booths_rejection_mail(@booth).deliver_later
+        end
         redirect_to admin_conference_booths_path(conference_id: @conference.short_title),
                     notice: "#{(t'booth').capitalize} successfully rejected."
       else


differentreality:
@rahul2240 I suggest we make the implementation a bit more helpful to the organizer, because right now the CfP committee member may not be aware that an email is being sent.

The behavior should be similar to what we used to have for events in the past. 
We want the organizer to choose whether or not to reject with an email.

When choosing to reject with an email, we need to check whether or not there is an email template, and if there is one, then send the email.

We should be able to go from state 'reject without email' to 'reject with email' (to trigger the email)

rahul2240:
@differentreality alright! On it :+1:

rahul2240:
@differentreality I think if we add this feature then there is no need to send an email to submitters about their booth acceptance or rejection after the deadline. As we can send the emails to all accepted or rejected booth submitters at last by changing state in bulk ( #2575 )

differentreality:
What you describe sounds right! Not sure what you mean by '...no need to send an email...'

Perhaps you meant that this can be taken care of through the custom emails? As in, when booth is accepted we do *not* send an email. At any given point emails can be sent out to, eg all accepted booths.

rahul2240:
> Not sure what you mean by '...no need to send an email...'

In the GSoC doc, we have this feature to implement

> When the stands are published, there needs to be a customizable section of text that is included in the message sent to the submitters. This needs to be stored in the database so it can be prepared and edited (and agreed internally) in advance of the publication event. (The text will typically provide a basic explanation of factors involved in the decisions.) Additionally, there could be a separate per-stand field (which may be empty) allowing us to include specific comments (feedback) about particular submissions.

so I am confused whether we need the above feature (where we send an email when booths are published as well as when we do accept with email) or the current feature i.e "using bulk actions to change state to accept with email" or "reject with email"  (#2575 ) and also adding a new state that has a transition from "accept" to "accept with email" to send email only once to all.

differentreality:
Merging this because I think the check should exist in our master branch.
I still think we should have proper wording in the view when the user is triggering the rejection.

Issue #2520, opened 2019-05-23
Support talks without speaker


Ana06:
For "talks" like lunch and party (in general any activity you want to add in the schedule but that doesn't have any speaker) it should be possible to have talks without speaker.

snpd25:
@Ana06 , I wanna work on it, will making the speakers optional solve the issue or another option in the speaker list: "Not-particular" or "All present" will do?

Ana06:
@snpd25 

>  I wanna work on it

cool! :wink: Go ahead!

> will making the speakers optional solve the issue or another option in the speaker list: "Not-particular" or "All present" will do?

I think internally we should just save no speaker, but when rendering it to the user it may be a good idea to show something, at least in the admin side. In the user side, I would say that when there is not speaker we should not render any picture or speaker field. But I would say those are details we can discuss in the PR. :smile: 

Take into account we don't want users to send talks for the call for paper without speaker. This should be only for the conference organizers/admins.

snpd25:
Hey @Ana06 , sorry for getting back on it so late. I was just thinking if we can remove the validation
of speakers from the events model. 

> Take into account we don't want users to send talks for the call for paper without speaker. This should be only for the conference organizers/admins.

This will not be a problem as in the proposals controller the speaker is defined as the user who submits the proposal. And I could not find any other way where user defines the speaker. 
Once we get confirmed about what to show on the user side, I will make the final modifications and submit a PR :)

Ana06:
@snpd25 

> I was just thinking if we can remove the validation of speakers from the events model

As we are planning to allow that there is no speaker, the validation has to be removed.

> Once we get confirmed about what to show on the user side, I will make the final modifications and submit a PR :)

IMO it would be enough with which I mentioned in my previos comment:

> I think internally we should just save no speaker, but when rendering it to the user it may be a good idea to show something, at least in the admin side. In the user side, I would say that when there is not speaker we should not render any picture or speaker field. But I would say those are details we can discuss in the PR

Is there something that is not clear or need to be further discussed? If not please go ahead with the PR :muscle:

PR #2587, opened 2019-09-04
Supports talks without speakers

snpd25:
Admins may want to have discussions or any other general activity
that does not have any particular speaker. Thus, speaker validation
is removed.

Closes https://github.com/openSUSE/osem/issues/2520

### DIFF
--- a/app/views/admin/events/show.html.haml
+++ b/app/views/admin/events/show.html.haml
@@ -71,7 +71,7 @@
             %td
               - if @event.speakers.count == 1
                 = link_to 'Fill out speaker\'s biography', edit_admin_user_path(@event.speakers.first)
-              - else
+              - elsif @event.speakers.count > 1
                 Fill out speaker's biography:
                 = speaker_links(@event)
             %td{ 'class' => class_for_todo(progress_status['biographies']) }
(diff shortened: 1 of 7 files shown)


app/views/admin/events/show.html.haml
```diff
@@ -71,7 +71,7 @@
             %td
               - if @event.speakers.count == 1
                 = link_to 'Fill out speaker\'s biography', edit_admin_user_path(@event.speakers.first)
-              - else
+              - elsif @event.speakers.count > 1
```
Ana06: [`count`](https://apidock.com/rails/ActiveRecord/Associations/CollectionProxy/count) raises a query every time it is called. So you shouldn't call it twice. You can use the [case statement](https://www.rubyguides.com/2015/10/ruby-case/) for this. :wink:

Ana06:
@snpd25 have you closed this intentionally? do you need any help to address the changes? Please let me know if there is something you need help with.

snpd25:
I am working on a feature in the master branch. I will just open a PR using another branch.

Ana06:
I am not sure I understand. Why do you need to open another PR? :confused:

PR #2596, opened 2019-10-23
Supports talks without speakers

snpd25:
Admins may want to have discussions or any other general activity
that does not have any particular speaker. Thus, speaker validation
is removed.

Closes https://github.com/openSUSE/osem/issues/2520

### DIFF
--- a/app/views/admin/events/_datatable_row.haml
+++ b/app/views/admin/events/_datatable_row.haml
@@ -24,11 +24,15 @@
         Unknown submitter
 
     %td
-      - event.speakers_ordered.each do |speaker|
-        .speaker
-          = link_to speaker.name, admin_user_path(speaker)
-          - unless speaker.registrations.for_conference(event.conference)
-            %span.label.label-danger Unregistered
+      - if event.speakers.any?
+        - event.speakers_ordered.each do |speaker|
+          .speaker
+            = link_to speaker.name, admin_user_path(speaker)
+            - unless speaker.registrations.for_conference(event.conference)
+              %span.label.label-danger Unregistered
+      - else
+        .small
+          No Speaker
 
     - if @program.languages.present?
       %td
--- a/app/views/admin/events/show.html.haml
+++ b/app/views/admin/events/show.html.haml
@@ -69,9 +69,10 @@
               %span{ 'class' => [icon_for_todo(progress_status['registered']), 'fa-lg'] }
           %tr
             %td
-              - if @event.speakers.count == 1
+              - speakers_count = @event.speakers.count
+              - if speakers_count == 1
                 = link_to 'Fill out speaker\'s biography', edit_admin_user_path(@event.speakers.first)
-              - else
+              - elsif speakers_count > 1
                 Fill out speaker's biography:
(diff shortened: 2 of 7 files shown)


app/views/admin/events/_datatable_row.haml
```diff
@@ -24,11 +24,15 @@
...
-            %span.label.label-danger Unregistered
+      - if event.speakers.any?
+        - event.speakers_ordered.each do |speaker|
+          .speaker
+            = link_to speaker.name, admin_user_path(speaker)
+            - unless speaker.registrations.for_conference(event.conference)
+              %span.label.label-danger Unregistered
+      - else
+        .small
+          No Speaker
```
hennevogel: We don't want to show the users _No speaker_. This is good for the admin (as he may want to know this information), but not for the rest of the users. I think that not showing anything in this case is fine. :wink:

app/views/admin/events/show.html.haml
```diff
@@ -69,9 +69,10 @@
               %span{ 'class' => [icon_for_todo(progress_status['registered']), 'fa-lg'] }
           %tr
             %td
-              - if @event.speakers.count == 1
+              - speakers_count = @event.speakers.count
+              - if speakers_count == 1
                 = link_to 'Fill out speaker\'s biography', edit_admin_user_path(@event.speakers.first)
-              - else
+              - elsif speakers_count > 1
```
hennevogel: You can use the [case statement](https://www.rubyguides.com/2015/10/ruby-case/) for this. :wink:

app/views/admin/events/_datatable_row.haml
snpd25: I thought this view is for admin, since it is in the admin folder.

app/views/admin/events/show.html.haml
snpd25: I will need some help here. I am not sure about fixing the upper-limit for number of speakers (say, `when 1..1000` or `when 1..2000`). Else will have to use some function to compare. Which will be the better option?

Ana06:
This PR is superseding https://github.com/openSUSE/osem/pull/2587 (not sure why a new PR was needed). Some changes not addressed were requested there - such as the one Henne has already pointed out: https://github.com/openSUSE/osem/pull/2596#discussion_r339180012

snpd25:
Sorry for the confusion Ana, actually I was trying to implement a feature on the master branch and it got messed up, and I didn't want this to get delayed, thus this PR. 
I tried addressing the changes, and have used a variable so that the query must not be raised twice. Also, 

> I am not sure about fixing the upper-limit for number of speakers (say, when 1..1000 or when 1..2000). Else will have to use some function to compare. Which will be the better option?

app/views/admin/events/show.html.haml
Ana06: There is a new endless range in Ruby 2.6 that could be used here: https://anamaria.martinezgomez.name/2018/12/21/ruby2_6.html

But we are using Ruby 2.5 :sob: 

I find it good as it is now. :+1: :+1:

PR #2640, opened 2020-03-15
Explicitly configure database authentication in Docker Compose

AndrewKvalheim:
### Checklist

The PostgreSQL Docker image [now requires](https://github.com/docker-library/postgres/pull/658) explicit configuration of the authentication method.

```console
$ docker-compose up database
Creating network "osem_default" with the default driver
Creating osem_database_1 ... done
Attaching to osem_database_1
database_1  | Error: Database is uninitialized and superuser password is not specified.
database_1  |        You must specify POSTGRES_PASSWORD to a non-empty value for the
database_1  |        superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".
database_1  |
database_1  |        You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
database_1  |        connections without a password. This is *not* recommended.
database_1  |
database_1  |        See PostgreSQL documentation about "trust":
database_1  |        https://www.postgresql.org/docs/current/auth-trust.html
osem_database_1 exited with code 1
```

Expect that `OSEM_DB_PASSWORD` is set, and configure PostgreSQL to use that.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,7 +37,7 @@ There is a rudimentary docker-compose configuration for production usage (`docke
 
 
 1. Configure OSEM
-   You have at least to set `SECRET_KEY_BASE`
+   You have at least to set `OSEM_DB_PASSWORD` and `SECRET_KEY_BASE`
    ```
    cp dotenv.example .env.production
    vim .env.production
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     image: postgres
     environment:
       PGDATA: /var/lib/postgresql/data/pgdata
+      POSTGRES_PASSWORD: $OSEM_DB_PASSWORD
   osem:
     build:
       context: .
--- a/docker-compose.yml.production-example
+++ b/docker-compose.yml.production-example
@@ -5,6 +5,7 @@ services:
     image: postgres
     environment:
       PGDATA: /var/lib/postgresql/data/pgdata
+      POSTGRES_PASSWORD: $OSEM_DB_PASSWORD
     volumes:
       - osem_production_database:/var/lib/postgresql/data/pgdata
   production_web:


INSTALL.md
```diff
@@ -37,7 +37,7 @@ There is a rudimentary docker-compose configuration for production usage (`docke
 
 
 1. Configure OSEM
-   You have at least to set `SECRET_KEY_BASE`
+   You have at least to set `OSEM_DB_PASSWORD` and `SECRET_KEY_BASE`
```
differentreality: `OSEM_DB_PASSWORD` may or may not be needed depending on the DB adapter you are using

docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
     image: postgres
     environment:
       PGDATA: /var/lib/postgresql/data/pgdata
+      POSTGRES_PASSWORD: $OSEM_DB_PASSWORD
```
differentreality: Since we are using postgres image, it seems that we need this. Wondering why we didn't have that before...

AndrewKvalheim: It's a recent change that caught many people by surprise; discussion in docker-library/postgres#681.

INSTALL.md
AndrewKvalheim: Note the [context](https://github.com/openSUSE/osem/blob/38b648655060d06109734821095cf32e7a82efb6/INSTALL.md#deploy-via-dockerdocker-compose), though—a section about:

1. using the example Docker Compose configuration, which specifies PostgreSQL
1. production deployment, for which the fallback to a [hard-coded secret](https://github.com/openSUSE/osem/blob/38b648655060d06109734821095cf32e7a82efb6/config/database.yml#L15) is particularly unwise

hennevogel:
Committed slightly different in #2723. Thanks!

AndrewKvalheim:
How about the recommendation of not hard-coding the password? It seems particularly unwise for [production](https://github.com/openSUSE/osem/commit/2e706347a79c7f8bd9fae796b56225924bad87c7#diff-7fafb0484f19aca304cb2ff894479565d42b7ca8e3e3dd7b6d6f0fa68864021fR8).

PR #2646, opened 2020-03-28
Use version range instead of OSEM_RUBY_VERSION.

AndrewKvalheim:
### Checklist

In order to accommodate variation in the version of Ruby across environments, the Gemfile is currently runtime-configurable via `OSEM_RUBY_VERSION`:

https://github.com/openSUSE/osem/blob/e6f3aa559ea4c3850035c2c2b7ae9aea26bb4ed3/Gemfile#L5

This mechanism is incompatible with typical Docker environments since, by design, the runtime configuration has no effect on builds:

```console
$ docker-compose build osem
…
Your Ruby version is 2.5.5, but your Gemfile specified 2.5.0
```

Furthermore, that it's apparently become routine to circumvent the Ruby specification raises the question of why that constraint exists at all. If users are typically expected to set `OSEM_RUBY_VERSION`, what purpose does the Ruby specification serve?

As with the other dependencies, specify a range of compatible versions. I've guessed at a conservative starting point:

```ruby
ruby '~> 2.5.0'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@
 
 source 'https://rubygems.org'
 
-ruby ENV['OSEM_RUBY_VERSION'] || '2.5.0'
+ruby '~> 2.5.0'
 
 # rails-assets requires >= 1.8.4
 if Gem::Version.new(Bundler::VERSION) < Gem::Version.new('1.8.4')
--- a/dotenv.example
+++ b/dotenv.example
@@ -32,9 +32,6 @@
 # OSEM_MEMCACHED_USERNAME='root'
 # OSEM_MEMCACHED_PASSWORD='1234'
 
-# The ruby version to use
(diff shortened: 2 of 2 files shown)


isnuryusuf:
its work for me
solve problem:
```
Your Ruby version is 2.5.7, but your Gemfile specified 2.5.0
ERROR: Service 'production_web' failed to build: The command '/bin/sh -c export NOKOGIRI_USE_SYSTEM_LIBRARIES=1; bundle install --jobs=3 --retry=3 --without test development' returned a non-zero code: 18
```

hennevogel:
We don't really *need* 2.5, so this would limit us too much. Sorry, thanks anyway :bouquet:

AndrewKvalheim:
The point is to document what you do need.

Issue #2639, opened 2020-03-05
oSLO registration T-shirt survey does not return


afaerber:
**I'm submitting a ...**
- [x] Bug Report
- [ ] Feature Request

**Current behavior:**
The oSLO T-Shirt survey leaves the user stranded on that survey page, without indication of whether the survey data was stored successfully nor what the user's next steps should be.

**Expected correct behavior:**
After successfully submitting the survey I would expect either:
* some text message indicating that the data has been saved successfully and to find some link to actively return to the event's registration overview page, or
* returning automatically to the event registration overview page https://events.opensuse.org/conferences/oSLO/register that the survey was opened from.

**Steps to reproduce:**
1. Go to https://events.opensuse.org/.
1. If not yet registered for "openSUSE + LibreOffice Conference October 13 - 16, 2020", click on "Register"(?) and follow through the steps.
If already registered, choose "My Registration".
1. On https://events.opensuse.org/conferences/oSLO/register under "Surveys" click on "What is your T-Shirt Size".
1. If not done yet, select two checkboxes, e.g. "Male T-Shirt" and "Medium".
1. Click "Submit".
1. Notice the mouse cursor changing and the button briefly getting disabled, then returning to previous state (i.e., no additional text, indicator, link or button), still on https://events.opensuse.org/conferences/oSLO/surveys/11.

**Other information:**

PR #2651, opened 2020-03-29
Surveys: Grooming of messages to user

differentreality:
**Checklist**

https://github.com/openSUSE/osem/issues/2639

* Groom error message when survey is closed
* Add flash success and error message for when user replies
* Show **Back** link (for surveys `during_registration` and `after_event`)

### DIFF
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -16,7 +16,7 @@ class SurveysController < ApplicationController
 
   def reply
     unless can? :reply, @survey
-      redirect_to conference_survey_path(@conference, @survey), alert: 'This survey is currently closed'
+      redirect_to conference_survey_path(@conference, @survey), alert: 'This survey is not available'
       return
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/surveys_controller.rb
```diff
@@ -33,13 +33,22 @@ def reply
...
-        user_survey_submission.update_attributes(updated_at: Time.current)
-      else
-        @survey.survey_submissions.create!(user: current_user)
+      begin
+        if user_survey_submission
+          user_survey_submission.update_attributes(updated_at: Time.current)
+        else
+          @survey.survey_submissions.create!(user: current_user)
+        end
+      rescue => error
```
hennevogel, line 42: Hm, how can this happen?

hennevogel (changes requested):

app/controllers/surveys_controller.rb
differentreality: If we cannot create the submission record (when `create!` raises an exception)

hennevogel: Sure but I don't see any condition when this will happen? Only when the user already has submitted. Isn't all of this just some for of  [`find_or_create_by`](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by-21)?

hennevogel:
We want to drop surveys in the near future...

PR #2654, opened 2020-04-03
Fix access to the version history of organization-level roles

AndrewKvalheim:
### Checklist

### Short description of what this resolves, changes proposed

The [tests of the version history of organization-level roles][versions_spec.rb:320] are currently failing. In trying to understand why, I encountered several complications:

1. The tests expect the version history of organization-level roles to be accessible by conference organizers. This surprised me; I expected that organization administrators should have this access, not conference organizers.

    **Proposed change:** Test as an organization administrator.

1. Regardless of that decision, currently neither organization administrators nor conference organizers have access, due to two independent bugs:

    1. No permission has been granted to access versions of organization-level roles.

        **Proposed change:** Grant permission.

    1. In the metadata of versions of organization-level roles, organization IDs are conflated with conference IDs; affiliation with organization *n* is recorded as affiliation with *conference* *n*.

        **Proposed change:** Distinguish between organization IDs and conference IDs.

1. The tests that should have caught this return false negatives under SQLite. In this test configuration, the organization and conference happen to have numerically equal IDs, so their conflation goes undetected.

    **Proposed change:** I don't have any good ideas about how to detect this kind of mistake in the future. Maybe randomize IDs by prepopulating tables with filler rows? Or enforce unique IDs upon creation in factories?

I'm not confident about these changes. I don't fully understand the relationship between the `is_admin` flag and the `organization_admin` role, and I'm unclear about what the `conference_id` version metadata was supposed to be, e.g. should it be a polymorphic association instead?

  [versions_spec.rb:320]: https://github.com/openSUSE/osem/blob/e709d080b46b2cdd7f3b7ec01c9c0d291c029724/spec/features/versions_spec.rb#L320-L342


db/migrate/20200331214534_add_organization_to_versions.rb
```diff
@@ -0,0 +1,35 @@
...
+    PaperTrail::Version.where.not(conference_id: nil).where(item_type: %[Role UsersRole]).each do |version|
+      id = version.conference_id
+
+      if Organization.exists?(id)
+        raise "version #{version.id} conflates organization #{id} with conference #{id}" if Conference.exists?(id)
+
+        version.update_attributes conference_id: nil, organization_id: id
+      end
+    end
+  end
```
AndrewKvalheim, line 34: This is untested and needs discussion about how to handle the situation.

AndrewKvalheim:
>  The Travis CI build failed

Resolved by #2655.

Issue #2568, opened 2019-07-22
Users view doesn't work


sbernhard:
**I'm submitting a ..**
- [x] Bug Report
- [ ] Feature Request

**Current behavior:**
http://demo.osem.io/admin/users doesn't list the users but prints an error

![image](https://user-images.githubusercontent.com/25485845/61662039-d5fe5980-accd-11e9-847e-2e8cc30285e2.png)

**Expected correct behavior:**
List the users

gabrielbaldao:
Can i try fix this bug?

cycomachead:
I was able to get the page working in my fork with the following tweaks to `app/datatables/user_datatable.rb`. To be honest, I'm not 100% certain I haven't messed something else up, but so far so good....?

```rb
   # rubocop:disable Naming/AccessorMethodName
   def get_raw_records
-    User.left_outer_joins(:registrations, :roles)
-      .distinct
-      .select("users.*, COUNT(CASE WHEN registrations.attended = 't' THEN 1 END) AS attended_count")
-      .group('users.id')
+    User.left_outer_joins(:registrations, :roles).select(
+        "DISTINCT users.id, users.name, users.email, users.confirmed_at, COUNT(CASE WHEN registrations.attended = 't' THEN 1 END) AS attended_count"
+      ).group(
+        'users.id, users.name, users.email, users.confirmed_at'
+      )
   end
   # rubocop:enable Naming/AccessorMethodName
 
   def records_total_count
-    fetch_records.unscope(:group).count(:all)
+    fetch_records.unscoped.count(:id)
   end
 
   def records_filtered_count
-    filter_records(fetch_records).unscope(:group).count(:all)
+    filter_records(fetch_records).unscoped.count
   end
```

PR #2656, opened 2020-04-04
Fix bug in UserDatatable query affecting PostgreSQL

AndrewKvalheim:
### Checklist

The current workaround to jbox-web/ajax-datatables-rails#293 results in an invalid query on PostgreSQL:

```console
$ docker-compose run --rm osem bundle exec rspec ./spec/datatables/user_datatable_spec.rb:97
…
     Failure/Error: fetch_records.unscope(:group).count(:all)

     ActiveRecord::StatementInvalid:
       PG::GroupingError: ERROR:  column "users.id" must appear in the GROUP BY clause or be used in an aggregate function
       LINE 1: SELECT COUNT(*) FROM (SELECT DISTINCT users.*, COUNT(CASE WH...
                                                     ^
```

Unscope both the custom `select` and `group` when counting, not just `group`.

### DIFF
--- a/app/datatables/user_datatable.rb
+++ b/app/datatables/user_datatable.rb
@@ -47,10 +47,12 @@ class UserDatatable < AjaxDatatablesRails::Base
   end
   # rubocop:enable Naming/AccessorMethodName
 
+  # Workaround for jbox-web/ajax-datatables-rails#293
   def records_total_count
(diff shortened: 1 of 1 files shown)


AndrewKvalheim:
This is probably the same issue as #2568.

cycomachead:
I like this fix much better than the one I proposed! :)

PR #2680, opened 2020-07-10
Fix docker setup for newer or updated image builds

chocoelho:
**Checklist**

- #2511 

- Add `OSEM_RUBY_VERSION` as build argument to Docker setup

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,9 @@ override example file and in it, set your user id in the variable
 sed "s/13042/`id -u`/" docker-compose.override.yml.example > docker-compose.override.yml
 ```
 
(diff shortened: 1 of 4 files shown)


AndrewKvalheim:
Curious to hear your thoughts on this vs. #2646, which look like two different attempts at solving the same problem.

chocoelho:
> Curious to hear your thoughts on this vs. #2646, which look like two different attempts at solving the same problem.

I just wanted to kept the existent approach as using `OSEM_RUBY_VERSION`. This change would make it possible to configure this variable at build time.

hennevogel:
The variable really is only valuable if you want to use another ruby version. The docker build failed because the underlying OS changed.

PR #2702, opened 2020-11-02
Allow leaving OSEM_SMTP_DOMAIN unset

AndrewKvalheim:
### Checklist

The SMTP setting [`:domain`](https://guides.rubyonrails.org/v5.2/action_mailer_basics.html#action-mailer-configuration) (i.e. [HELO domain](https://ruby-doc.org/stdlib-2.5.8/libdoc/net/smtp/rdoc/Net/SMTP.html#class-Net::SMTP-label-HELO+domain)) is [often unnecessary](https://stackoverflow.com/a/8439535) and falls back to a reasonable [default](https://github.com/mikel/mail/blob/2.7.1/lib/mail/network/delivery_methods/smtp.rb#L82).

Leaving [`OSEM_SMTP_DOMAIN`](https://github.com/openSUSE/osem/blob/ae98ac1430baa690f744d850a985f63e5e9bd3cc/config/environments/production.rb#L97) unset, however, currently causes `:domain` to be explicitly set to `nil`, which in our case caused mail delivery to fail with `EOFError`.

Omit `nil` values from the Action Mailer configuration.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -97,7 +97,7 @@ Osem::Application.configure do
     domain:               ENV['OSEM_SMTP_DOMAIN'],
     enable_starttls_auto: ENV['OSEM_SMTP_ENABLE_STARTTLS_AUTO'],
     openssl_verify_mode:  ENV['OSEM_SMTP_OPENSSL_VERIFY_MODE']
-  }
+  }.compact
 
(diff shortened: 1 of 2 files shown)


AndrewKvalheim:
CI currently failing at #2703

hennevogel (approved):

PR #2837, opened 2021-08-17
Introduces next-rails

hennevogel:
To aid with rails updates


hennevogel:
This is supposed to fail. Nothing is setup for rails 6. Merging.

PR #2905, opened 2021-11-08
fix service name

eladeyal-intel:
doc change to match the service name in   docker-compose.yml.production-example


hennevogel (approved):

hennevogel:
Thank you so much @eladeyal-intel 💐

eladeyal-intel:
Actually maybe a better fiz is to change service name in compose file to `osem` ?

hennevogel:
I don't think so, that would mean you can't have production and dev-env on the same host.

PR #2913, opened 2022-01-24
Order rooms by name in the schedule

AndrewKvalheim:
### Checklist

On the published schedule, rooms are listed in undefined order—in practice, usually the order of creation. This can make the schedule confusing to read.

An explicit ordering feature would be useful, but for now it’s an improvement to just order by name.

### DIFF
--- a/app/controllers/schedules_controller.rb
+++ b/app/controllers/schedules_controller.rb
@@ -29,7 +29,7 @@ class SchedulesController < ApplicationController
       end
 
       format.html do
-        @rooms = @conference.venue.rooms if @conference.venue
+        @rooms = @conference.venue.rooms.order(:name) if @conference.venue
         @dates = @conference.start_date..@conference.end_date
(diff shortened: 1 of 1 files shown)


spec/features/schedules_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+feature Schedule do
+  let!(:conference) { create(:conference) }
+  let!(:venue) { create(:venue, conference: conference) }
+
+  describe 'with unalphabetized rooms' do
```
hennevogel, line 9: I don't think we need a spec for `.order`. The `ActiveRecord` test suite will make sure that this will never break...

CI time is developer time.

hennevogel (changes requested):

AndrewKvalheim:
Makes sense, thanks.

hennevogel (approved):

PR #2960, opened 2022-03-02
Bugfix some forms

hennevogel:


app/views/devise/registrations/_form_fields.html.haml
```diff
@@ -10,17 +11,17 @@
   .checkbox
     %label
       = f.check_box :email_public
-      Check if you want your email address to appear publicly (e.g. on sessions, registration lists, etc.)
-  .form-group
-  = f.label :name, 'Real Name'
-  = f.text_field :name, required: true, class: 'form-control', placeholder: 'Name'
```
AndrewKvalheim, line 16: Did you mean to remove this? (Following up with #1406)

PR #2969, opened 2022-03-11
Basic feature test of survey creation and response

AndrewKvalheim:
### Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (if appropriate).

I tried using surveys for the first time and ran into a few issues:

1. When I click *Surveys → New* I get a permissions error.
2. When I click *Create Survey* I get an error about a missing view.
3. When I click a date field the date picker doesn’t appear. 
4. When I click *Type of Question* the form control isn’t focused.
5. When I view an undated survey I get an error about a failure to determine permissions.
6. When I respond to a survey nothing appears to happen.

1. Grant conference organizers permission to manage surveys.
2. Add the missing view for survey question creation.
3. Correct the markup of the date pickers.
4. Associate the form label with its control.
5. Correct an assumption that surveys always have start and end dates.
6. Show feedback upon survey response.

### DIFF
--- a/app/models/admin_ability.rb
+++ b/app/models/admin_ability.rb
@@ -142,6 +142,10 @@ class AdminAbility
     can :manage, Room, venue: { conference_id: conf_ids }
     can :manage, Sponsor, conference_id: conf_ids
     can :manage, SponsorshipLevel, conference_id: conf_ids
+    can :manage, Survey, surveyable_type: 'Conference',
+                         surveyable_id:   conf_ids
+    can :manage, SurveyQuestion, survey: { surveyable_type: 'Conference',
+                                           surveyable_id:   conf_ids }
     can :manage, Ticket, conference_id: conf_ids
     can :create, TicketScanning do |ticket_scanning|
       conf_id = ticket_scanning.physical_ticket.ticket_purchase.conference_id
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -426,7 +426,7 @@ Lint/UriRegexp:
 # Offense count: 127
 # Configuration parameters: IgnoredMethods, CountRepeatedAttributes.
 Metrics/AbcSize:
-  Max: 71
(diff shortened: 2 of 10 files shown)


app/models/admin_ability.rb
```diff
@@ -142,6 +142,10 @@ def signed_in_with_organizer_role(user, conf_ids_for_organization_admin = [])
     can :manage, Room, venue: { conference_id: conf_ids }
     can :manage, Sponsor, conference_id: conf_ids
     can :manage, SponsorshipLevel, conference_id: conf_ids
+    can :manage, Survey, surveyable_type: 'Conference',
+                         surveyable_id:   conf_ids
+    can :manage, SurveyQuestion, survey: { surveyable_type: 'Conference',
+                                           surveyable_id:   conf_ids }
```
AndrewKvalheim, line 148: This has now crossed the threshold of `Metrics/AbcSize` but I’m not sure how to address it.

AndrewKvalheim: https://github.com/openSUSE/osem/pull/2546#issuecomment-513542629 has the same question

AndrewKvalheim:
> 5. When I respond to a survey nothing appears to happen.

I just found #2639 which reports the same problem. Here I chose to do just the minimum of displaying a notification but I agree a full solution would be to redirect to the appropriate list of surveys.

app/models/admin_ability.rb
hennevogel: `Metrics/AbcSize` is so far just an undiscussed number in `.rubocop_todo.yml`. Just increase it...

AndrewKvalheim: I see, thanks.

AndrewKvalheim:
This now updates the RuboCop to-do list and includes another bug fix for the survey date pickers.

hennevogel (approved):

PR #3059, opened 2022-06-18
Setting to disable email notifications of comments

AndrewKvalheim:
### Checklist

[SeaGL](https://seagl.org/) uses comments heavily during a specific review process in which email notifications aren’t appropriate. Previously we’d simply disabled notifications by [removing](https://github.com/SeaGL/osem/commit/a09eb20cfa6f5ad75de8ace295672521e6bdf111) that code, but a runtime setting for this would be more convenient.

Add a new setting, defaulting to enabled:

  - *E-Mails → Proposal → Send an email to all organizers and CfP team members when a comment is added?*

### Concerns

- Is *Proposal* the right place to put this, or should there be a *Comments* tab?
- This was the simplest solution to implement, but it might be preferable to instead allow more fine-grained control over which roles get notified, or to expose separate notification settings to each user.

### DIFF
--- a/app/controllers/admin/emails_controller.rb
+++ b/app/controllers/admin/emails_controller.rb
@@ -29,6 +29,7 @@ module Admin
                                              :send_on_accepted, :send_on_rejected, :send_on_confirmed_without_registration,
                                              :send_on_submitted_proposal,
                                              :submitted_proposal_subject, :submitted_proposal_body,
+                                             :send_on_event_comment,
                                              :registration_subject, :accepted_subject, :rejected_subject, :confirmed_without_registration_subject,
                                              :registration_body, :accepted_body, :rejected_body, :confirmed_without_registration_body,
                                              :send_on_conference_dates_updated, :conference_dates_updated_subject, :conference_dates_updated_body,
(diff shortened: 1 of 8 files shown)


hennevogel:
> [SeaGL](https://seagl.org/) uses comments heavily during a specific review process in which email notifications aren’t appropriate

Why? I mean comments are appropriate obviously, notifications are just helping people to not loose track. Can you explain?

AndrewKvalheim:
It’s been tedious to get an explanation but I think the review process is:

1. Individually, each reviewer on their own time:
    1. For each proposal:
        1. Outside of OSEM, read a copy of only the title and abstract.
        1. Outside of OSEM, decide on a vote.
        1. Within OSEM, submit the vote and optionally comment to explain your decision.
1. Collectively, all reviewers in a meeting:
    1. For each proposal:
        1. Within OSEM, read the entire proposal, including its votes and comments.

The reviewers avoid exposure to comments until after they’ve decided on their vote, and active notifications of comments work against that goal.

hennevogel:
> The reviewers avoid exposure to comments until after they’ve decided on their vote, and active notifications of comments work against that goal.

We already have a setting for that "behavior". It's called `Program.blind_voting`. Instead of another configuration we should apply this to comments and sending emails about comments in `Program.voting_period?` too.

PR #3085, opened 2022-09-12
Use CSS-compatible row IDs in datatables

AndrewKvalheim:
### Checklist

The IDs of datatable rows consist only of numbers, e.g. `<tr id="123"`. CSS ID selectors [may not begin with a number](https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier) so using e.g. `#123` in tests results in:

```plain
Selenium::WebDriver::Error::InvalidSelectorError:
  invalid selector: An invalid or illegal selector was specified
```

Namespace the IDs with the class name, e.g. `#user-123`.

### DIFF
--- a/app/datatables/registration_datatable.rb
+++ b/app/datatables/registration_datatable.rb
@@ -3,6 +3,7 @@
 class RegistrationDatatable < AjaxDatatablesRails::ActiveRecord
   extend Forwardable
 
+  def_delegator :@view, :dom_id
   def_delegator :@view, :edit_admin_conference_registration_path
 
   def initialize(params, opts = {})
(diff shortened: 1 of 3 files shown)


app/datatables/registration_datatable.rb
```diff
@@ -41,7 +41,7 @@ def data
         email:                    record.email,
         accepted_code_of_conduct: !!record.accepted_code_of_conduct, # rubocop:disable Style/DoubleNegation
         edit_url:                 edit_admin_conference_registration_path(conference, record),
-        DT_RowId:                 record.id
+        DT_RowId:                 "#{record.class.name.parameterize}-#{record.id}"
```
hennevogel, line 44: [dom_id](https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id) does that for you

hennevogel (changes requested):

app/datatables/registration_datatable.rb
AndrewKvalheim: Perfect, thanks.

hennevogel (approved):

PR #3244, opened 2023-05-31
Resolve inability to schedule events

AndrewKvalheim:
### Checklist

- [x] I have added tests that prove my fix is effective or that my feature works.

When dragging an unscheduled event onto a time slot, the event fails to be scheduled. Rails throws:

```
ActionController::RoutingError: No route matches [PUT] "/admin/conferences/…/event_schedules"
```

#### Details

To schedule an event, the editor must make a different kind of request depending on whether the event is already scheduled. This information is encoded in [`event_schedule_id`](https://github.com/openSUSE/osem/blob/bd937534d35bc63e18e8a0b6bc2d56d1a7cab38d/app/assets/javascripts/osem-schedule.js#L49) as:

| | Type |
| - | - |
| Scheduled | non-empty string |
| Nonscheduled | `""` (empty string) or `undefined` |

In disagreement with this encoding, the editor currently tests for inequality to `null`, a condition that always evaluates to true, so events are always determined to be already scheduled, resulting in the incorrect kind of request being made when scheduling nonscheduled events.

Since the encoding used correlates with truthiness, test for that instead.

### Concerns

I have low confidence in my understanding of this. This bug appears to prevent the scheduling of events, a critical feature of OSEM, yet I don’t see any prior reports about it. How have conferences been scheduling events?


spec/features/schedule_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+def expect_scheduled(event, time)
```
hennevogel, line 5: This sounds like a shared example to me...

https://rspec.info/features/3-12/rspec-core/example-groups/shared-examples/

hennevogel (changes requested):

sgryphon:
> This bug appears to prevent the scheduling of events, a critical feature of OSEM, yet I don’t see any prior reports about it. How have conferences been scheduling events?

I just found OSEM and have been trying to use it, and came across this bug, i.e. inability to schedule.

And your fix is > 10 months, and not merged.

Suggestion: Sometimes when I fix a bug in an open source project, the first thing I do is raise a bug report, then reference it in the fix. That way others can find the issue and the fix. (Of course, after 10 months you may have moved on).

hennevogel:
I've split the fix from the new test so we can work on this...

#3558

PR #3276, opened 2024-03-05
Move word counting javascript to the page

hennevogel:
If there is no `#event_abstract` there is no need to count words in it...


hennevogel:
Codacy seems to have problems right now. Merging...

PR #3277, opened 2024-03-05
Get off off rails-assets

hennevogel:
it's dead and pulls depfu down with it...


hennevogel:
Codacy has problems currently, merging...

PR #3397, opened 2024-06-06
Update Ruby version to 3.3.2

hennevogel:
Also remove support for  OSEM_RUBY_VERSION, we are going to support
one ruby version and one ruby version only now


AndrewKvalheim:
Sure you don’t want to [use a `~>`](https://github.com/openSUSE/osem/pull/2646)? I’m still having to patch that in nearly every time I work on OSEM:

> ```console
> Your Ruby version is 3.3.3, but your Gemfile specified 3.3.2
> ```

hennevogel:
> Sure you don’t want to use a `~>`

Yeah, pretty sure. At least not without a test matrix and more people who can take care of the fallout. 3.3.2 -> 3.3.3 means new bundler, new rexml, new strscan and loads of changed (bugfixed) behavior. We are two people here...

PR #3402, opened 2024-06-06
Drop bootstrap3 datetimepicker rails

hennevogel:


Dockerfile
```diff
@@ -20,7 +20,8 @@ WORKDIR /osem/
 RUN bundle config set --local path 'vendor/bundle'; \
     bundle install --jobs=4 --retry=3
 
-# Install our process manager
-RUN gem install foreman
+# Install our process manager / update chromedriver
+RUN gem install foreman; \
+    bundle exec bin/rails webdrivers:chromedriver:update
```
AndrewKvalheim, line 25: I’m not sure how this is supposed to work; `/osem/bin/rails` doesn’t exist at this point in the build.

AndrewKvalheim: This is how I’ve been handling it: https://github.com/SeaGL/osem/commit/2075d681177ca2653b7b7987cc385d3dfd839c78

hennevogel: Followed what the webdrivers peepz say in their wiki: https://github.com/openSUSE/osem/pull/3419

PR #3803, opened 2026-05-31
Fix: event commercial permissions in non-admin flow

### LINKED ISSUE #3143: Access denied when organizer adds commercial
### I'm submitting a

- [x] Bug Report
- [ ] Feature Request

### Current behavior

> You are not authorized to access this page. 

```
Started POST "/conferences/example/program/proposals/1/commercials"
Processing by CommercialsController#create as HTML
Access denied on create #<Commercial id: nil, commercial_id: nil, commercial_type: nil, commercialable_id: 1, commercialable_type: "Event", created_at: nil, updated_at: nil, url: "https://www.youtube.com/watch?v=YH3c1QZzRK4">
Redirected to http://localhost:3000/
Completed 302 Found
```

### Expected correct behavior

Successful addition of a commercial

### Steps to reproduce

1. As a speaker, submit a proposal.
1. As a conference organizer, add a commercial via *Manage → Events → Event → Edit → Commercials → Create Commercial*.

### Other information

Probably a case of [“wandering around in non-admin views”](https://github.com/openSUSE/osem/blob/master/app/models/ability.rb#L140).

RedZapdos123:
## Description:

Issue #3143 reported that organizers could edit a proposal in the non-admin flow but hit `AccessDenied` when creating a commercial for that proposal.

This PR updates `Ability` so organizers and CFP team members can manage event commercials for conferences where they already have proposal-management access in non-admin views.

It also adds regression tests that cover the relevant ability rules.

Closes #3143.

## Checklist:

- [x] I have added relevant regression test cases for this fix.
- [x] I have run linting checks using `bundle exec rubocop app/models/ability.rb spec/ability/ability_spec.rb`.
- [x] I have run focused tests using `bundle exec rspec spec/ability/ability_spec.rb`.

## Additional validation:

- Reproduced the bug on a clean clone in WSL: organizers could manage the proposal event but could not create an event commercial.
- Verified the fix on the patched clone in WSL: organizers could create the event commercial.
- Ran `bundle exec rake spec:ability` in WSL.

### Before the fix (when signed in as an organizer):

image

### After the fix (when signed in as an organizer):

image


spec/controllers/commercials_controller_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe CommercialsController do
```
hennevogel: No need to test `Ability` twice right? You can drop this...

hennevogel (changes requested):

spec/controllers/commercials_controller_spec.rb
RedZapdos123: Removed the redundant controller spec and have only. the ability regression coverage.

hennevogel (approved):

hennevogel:
What does the last commit have to do with this PR?

hennevogel:
@RedZapdos123 thanks Mridankan :)

RedZapdos123:
> What does the last commit have to do with this PR?

Yes, it was out of scope for this PR, so I had reverted it.

I had added it to address a failing spec/features/proposals_spec.rb check on this branch, where Selenium hit a stale node error during proposal confirmation. I moved that commit's changes for test stabilization to #3808 (based on issue #3807).

